PR #4, opened 2017-02-12
rpi2 test

technicalpyro:
Tested on RaspberyyPi 2


dschaper:
Thanks! Are you able to try and any of the other binaries? (Mostly the `arm-linux-gnuabi` binary?)

technicalpyro:
dammit OCD missed a spacing thing ... two commits turn into three so easily

DL6ER:
No worries. I can smash the three commits to a single one when merging (I will do that now!)

PR #8, opened 2017-02-18
Update README to include command modifiers in telnet

technicalpyro:
updated the README to include the modifiers usable within the telnet.

unfortunately it shows as 5 commits still understanding how this works but might be due to updating my fork today not one hundred percent sure

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ PTR: 12
 SRV: 0
 ```
 
-- `>getallqueries` : get all queries that FTL has in its database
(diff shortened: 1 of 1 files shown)


DL6ER:
Thank you for your contribution, I'd like to postpone that a little bit, since I'm going to revamp the telnet interface slightly so that everything is more consistent.

DL6ER:
I added this information to the README along with making it more consistent in the code itself

PR #14, opened 2017-04-05
Improve IP search; check end character

spacedingo:
In a LAN with clients 192.x.x.1-255, comparing first character (1) will perform **strcmp** 254 times for say 192.x.x.14.  By checking last character, it would do **strcmp** 26 times. In the black-hole, less is more : )
Please feel free to change anything and everything - I'm a newbie.

### DIFF
--- a/parser.c
+++ b/parser.c
@@ -685,10 +685,12 @@ int findDomain(char *domain)
 int findClient(char *client)
 {
 	int i;
+	int client_last_pos = strlen(client) - 1;
 	for(i=0; i < counters.clients; i++)
 	{
-		// Quick test: Does the clients IP start with the same character?
-		if(clients[i].ip[0] != client[0])
+		int clients_last_pos = strlen(clients[i].ip) - 1;
+		// Quick test: Does the clients IP end with the same character?
+		if(clients[i].ip[clients_last_pos] != client[client_last_pos])
 			continue;
 
 		// If so, compare the full domain using strcmp


DL6ER:
I don't think that this is a real improvement. Let me explain why I think that:

The whole idea of comparing the first character was born for the domain comparison. Comparing the very first character can be done extremely fast (direct memory access) and does not require a call to any sophisticated string function. Hence, it can speed up the searching for a suitable domain by up to a factor of 36x.
The situation for which I implemented this was for extreme environments like out testing Pi-hole where the domain struct has > 100,000 entries (wich more than 200 million queries in `FTL`'s memory). With this huge amount of domain entries, searching through the whole list with calling `strcmp()` on each entry is quite slow. Comparing the very first character is very fast.
This was never thought to be added to the clients list which should always be anyhow *much* smaller than this number. I only added this for the client search as well so that both functions are similar in nature.

I tested your suggestion in this extreme testing environment (for the domains array, because we have only one client there) and interestingly, it seems like it does not change anything in terms of run time. Further investigations seem to suggest that `strcmp()` is not much more expensive than `strlen()` and for domains/clients where *both the first and the last* letter are equal (which can happen like almost always with domain endings like `.com`), actually *both* functions have to be called to identify matches. Accordingly, I saw an increase of run time of nearly 30% for the domain finding algorithm.

Hence, I'm going to reject this pull request, because
- the domain searching gets noticeably slower, and
- the client searching never needed an algorithmic speed up.

I hope you understand this decision and we look forward to further contributions from you. Rest assured that I know how disappointing it can be if the first contribution to some project gets rejected, but this decision is only motivated by the two points listed above. Although your idea was good in its nature, it revealed to have adverse effects on `FTL`'s performance.

PR #40, opened 2017-05-10
Snyc master into development 

DL6ER:
Internal. Albeit not entirely sure why that is even needed, since all code went from `somebranch` --PR--> `development` --PR--> `master`


DL6ER:
Seems like the initially shown changes are now gone...
screenshot at 2017-05-10 19-52-48

PR #75, opened 2017-06-12
Use MessagePack for the Unix socket API

DL6ER:

## 10

---

[Edited by Mcat12]

Change Unix socket interface to return data in `MessagePack` format. It still uses the same API commands as the telnet interface. The MessagePack format allows FTL to efficiently stream data in a very compact form to the upcoming HTTP API.

Also included are these smaller changes:
- Limit the total number of client connections
- Refactor checking the type of the connection into the global `istelnet[]` array of bools.
- Improve the socket-test binary to allow custom commands to be sent.

This PR previously held an implementation of an HTTP API, but that will be moved to a separate program. It will access FTL's data through the Unix socket.


DL6ER:
Note that due to its complexity, this PR will most likely be a work in progress for quite some time and will not be part of the next versions of `FTL` we release

request.c
```diff
@@ -357,30 +357,49 @@ void getOverTime(int *sock, char type)
...
 
 	// Exit before processing any data if requested via config setting
 	if(!config.query_display)
 		return;
 
 	// Match both top-domains and top-ads
-	if(sscanf(client_message, ">%*[^(](%i)", &num) > 0)
+	// SOCKET: >top-domains (15)
+	// API:    /stats/top_domains?limit=15
+	if(sscanf(client_message, "%*[^0123456789H\n]%i", &num) > 0)
```
Mcat12: Will this pick up `/stats/top_domains?alkjsdlkfj=15` as valid?

request.c
```diff
@@ -357,30 +357,49 @@ void getOverTime(int *sock, char type)
...
 
 	// Sort in descending order?
-	if(command(client_message, " desc"))
+	// SOCKET: >top-domains desc
+	// API:    /stats/top_domains?order=desc
+	if(type == SOCKET && command(client_message, " desc"))
 	{
 		desc = true;
 	}
+	else if(type != SOCKET && command(client_message, "desc"))
```
Mcat12: This should check for `order=desc`

request.c
```diff
@@ -485,7 +536,10 @@ void getTopClients(char *client_message, int *sock, char type)
 {
 	int i, temparray[counters.clients][2], count=10, num;
 
-	if(sscanf(client_message, ">%*[^(](%i)", &num) > 0)
+	// Match both top-domains and top-ads
+	// SOCKET: >top-clients (15)
+	// API:    /stats/top_clients?limit=15
+	if(sscanf(client_message, "%*[^0123456789H\n]%i", &num) > 0)
```
Mcat12: See other comment.

request.c
```diff
@@ -1272,3 +1281,38 @@ void getDBstats(int *sock, char type)
...
+			// Skip empty lines
+			if(line[0] == '\n')
+				continue;
+
+			if(!first) ssend(*sock, ",");
+			first = false;
+
+			// Trim off the newline, if it exists
+			size_t len = strlen(line);
+			if(len > 0 && line[len-1] == '\n')
```
DL6ER: what about Win line-ends (CR-LF)?

request.c
```diff
@@ -195,28 +195,14 @@ bool command(char *client_message, const char* cmd)
...
-		// Send only header (length of content is not yet known and will be sent out in smaller packets)
+		// Send header only for full HTTP requests
 		ssend(sock,
 		      "HTTP/1.0 200 OK\nServer: FTL\nCache-Control: no-cache\n"
 		      "Content-Type: application/json\n\n");
 	}
-	else
-	{
-		// Simple request: Don't send header, only payload
-		ssend(sock,"%s",content);
```
Mcat12: Are we not replying to simple requests anymore?

ppadial:
Hi!

i'm doing an java api client for PiHole API, currently im implementing the FTL methods existing on my current version 3.0, so, i will monitor this PR to update my lib when is released :)
https://github.com/ppadial/pihole-api-client-java

api_dns.c
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	cJSON_Delete(input_root);
+}
+
+void removeList(int *sock, char type, char list_type, char *client_message) {
+	char *domain = strrchr(client_message, '/');
+
+	// Remove leading '/'
+	domain++;
```
DL6ER: This can be dangerous as `strrchr(client_message, '/')` could have returned `NULL` if the character `/` was not found.

api_dns.c
```diff
@@ -0,0 +1,253 @@
...
+		// Invalid route
+		free(expected_route);
+		sendAPIResponse(*sock, type, NOT_FOUND);
+		ssend(*sock, "\"status\":\"not_found\"");
+		return;
+	}
+
+	free(expected_route);
+
+	// Validate domain
```
DL6ER: I think we should do this earlier (see my other comment as well)

api.c
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// session will have the client's valid session written to, if it's not unauthorized
+enum Auth authenticate(char *with_headers, char *payload, long *session, int sock) {
+	// First figure out if the client has authenticated before.
+	char *sessionStr;
+	AuthData *auth = NULL;
+
+	// Find the cookie header (will contain a long int value)
+	if(strstr(with_headers, "Cookie: ") != NULL && (sessionStr = strstr(with_headers, "FTL_SESSION=")) != NULL) {
```
0xcaff: This looks dangerous. Doesn't this read `FTL_SESSION=` from anywhere in the http request? I can't think of how this would be an attack vector on its own, but coupled with other things, it probably could be. Maybe.

main.c
```diff
@@ -63,19 +63,31 @@ int main (int argc, char* argv[]) {
...
 	pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
 
+	// Start log analyzing thread
 	pthread_t piholelogthread;
 	if(pthread_create( &piholelogthread, &attr, pihole_log_thread, NULL ) != 0)
 	{
 		logg("Unable to open Pi-hole log processing thread. Exiting...");
 		killed = 1;
 	}
+	sleepms(100);
```
0xcaff: Why is this sleep here? Why 100ms?

socket.c
```diff
@@ -263,7 +305,162 @@ void *socket_listenting_thread(void *args)
...
-			logg("WARNING: Unable to open clients processing thread, error: %s", strerror(errno));
+			logg("WARNING: Unable to open client socket thread, error: %s", strerror(errno));
+		}
+	}
+	return 0;
+}
+
+void *api_connection_handler_thread(void *socket_desc)
+{
+	//Get the socket descriptor
```
0xcaff: Nitpick, inconsistent comment spacing.

socket.c
```diff
@@ -263,7 +305,162 @@ void *socket_listenting_thread(void *args)
...
+		char *message = calloc(strlen(client_message)+1,sizeof(char));
+		strcpy(message, client_message);
+
+		// Clear client message receive buffer
+		memset(client_message, 0, sizeof client_message);
+
+		if(debug)
+			logg("Received API request: \n%s", message);
+
+		if(strncmp(message, "GET ", 4) == 0 || strncmp(message, "POST ", 5) == 0 || strncmp(message, "DELETE ", 7) == 0)
```
0xcaff: Is there any reason a HTTP library isn't used?

FTL.h
```diff
@@ -23,6 +23,7 @@
 #include <time.h>
 #include <sys/time.h>
 #include <sys/socket.h>
+#include <regex.h>
```
DL6ER: Do we still use any subroutine from the regex.h header?

FTL.h
```diff
@@ -207,7 +211,8 @@ typedef struct {
 enum { DATABASE_WRITE_TIMER, EXIT_TIMER };
 
 enum { QUERIES, FORWARDED, CLIENTS, DOMAINS, OVERTIME, WILDCARD };
-enum { SOCKET };
+enum { TELNET, SOCKET };
+enum { WHITELIST, BLACKLIST, WILDLIST };
```
DL6ER: This might not be needed as well.

FTL.h
```diff
@@ -252,3 +257,5 @@ long int lastdbindex;
 bool travis;
 bool DBdeleteoldqueries;
 bool rereadgravity;
+char * clientip[MAXCONNS];
```
DL6ER: We might even want to get rid of all of this as we aren't implementing an authentication directly in `FTL` now.

main.c
DL6ER: I don't think this delay is needed at all.

main.c
```diff
@@ -63,36 +63,42 @@ int main (int argc, char* argv[]) {
...
 	}
 
+	// Start SOCKET thread
 	pthread_t socket_listenthread;
 	if(pthread_create( &socket_listenthread, &attr, socket_listening_thread, NULL ) != 0)
 	{
 		logg("Unable to open Unix socket listening thread. Exiting...");
 		killed = 1;
 	}
+	sleepms(100);
```
DL6ER: This one can be removed as well, I think.

msgpack.c
```diff
@@ -0,0 +1,112 @@
...
+	memcpy(&bigEValue, &value, sizeof(bigEValue));
+	bigEValue = leToBe64(bigEValue);
+	pack_basic(sock, 0xd3, &bigEValue, sizeof(bigEValue));
+}
+
+void pack_float(int sock, float value) {
+	// Need to use memcpy to do a direct copy without reinterpreting the bytes. It should get optimized away.
+	uint32_t bigEValue;
+    memcpy(&bigEValue, &value, sizeof(bigEValue));
+    bigEValue = htonl(bigEValue);
```
DL6ER: Indentation is off

msgpack.c
```diff
@@ -0,0 +1,112 @@
...
+	swrite(sock, &bigELength, sizeof(bigELength));
+	swrite(sock, string, length);
+}
+
+void pack_map16_start(int sock, uint16_t length) {
+	uint8_t format = 0xde;
+	swrite(sock, &format, sizeof(format));
+	uint16_t bigELength = htons(length);
+	swrite(sock, &bigELength, sizeof(bigELength));
+}
```
DL6ER: I went through the code and verified that everything seems to be correct. However, I **did not** compare the individual subroutines against the msgpack definition.

msgpack.c
```diff
@@ -0,0 +1,112 @@
...
+	pack_basic(sock, 0xca, &bigEValue, sizeof(bigEValue));
+}
+
+void pack_fixstr(int sock, char *string) {
+	// Make sure that the length is less than 32
+	size_t length = strlen(string);
+
+	if(length >= 32) {
+		logg("Tried to send a fixstr longer than 31 bytes!");
+		exit(EXIT_FAILURE);
```
DL6ER: Is it necessary to terminate `FTL` in this case? We should rather have non-`void` subroutines that return `false` in this case and check in the main code for the return value. Abruptly closing the socket on such a thing seems fine, but killing the whole daemon seems to be too much.

msgpack.c
```diff
@@ -0,0 +1,112 @@
...
+	swrite(sock, string, length);
+}
+
+void pack_str32(int sock, char *string) {
+	// Make sure that the length is less than 4294967296
+	size_t length = strlen(string);
+
+	if(length >= 4294967296) {
+		logg("Tried to send a str32 longer than 4294967295 bytes!");
+		exit(EXIT_FAILURE);
```
DL6ER: See comment on line msgpack.c:83

request.c
```diff
@@ -9,37 +9,19 @@
...
-void getDBstats(int *sock);
-void getClientsOverTime(int *sock);
-void getClientNames(int *sock);
-void getUnknownQueries(int *sock);
+bool command(char *client_message, const char* cmd) {
+	return strstr(client_message, cmd) != NULL;
+}
 
-void process_request(char *client_message, int *sock)
+void process_request(char *client_message, int *sock, char type)
```
DL6ER: Do we still need `type`?

request.c
```diff
@@ -139,6 +121,13 @@ void process_request(char *client_message, int *sock)
 		if(debugclients)
 			logg("Client wants to disconnect, ID: %i",*sock);
 
+		// Free memory containing IP address of this client
```
DL6ER: May want to remove this, see comment in FTL.h:260

routines.h
```diff
@@ -40,19 +40,19 @@ void memory_check(int which);
...
 void *telnet_listening_thread_IPv6(void *args);
-void seom(int sock);
-void ssend(int sock, const char *format, ...);
 
 void *socket_listening_thread(void *args);
 bool ipv6_available(void);
 bool bind_sockets(void);
 
-void process_request(char *client_message, int *sock);
+void process_request(char *client_message, int *sock, char type);
```
DL6ER: Might not need `char type`.

socket.c
```diff
@@ -197,9 +199,12 @@ void removeport(void)
 	fclose(f);
 }
 
-void seom(int sock)
+void seom(int sock, char type)
```
DL6ER: We can use `istelnet[sock]` also here and get rid of `char type`

socket.c
```diff
@@ -217,33 +222,62 @@ void ssend(int sock, const char *format, ...)
...
 
+void swrite(int sock, void *value, size_t size) {
+	if(write(sock, value, size) == -1)
+		logg("WARNING: Socket write returned error code %i", errno);
+}
+
+int checkClientLimit(int socket, char *ipAddr) {
+	if(socket < MAXCONNS)
+	{
+		clientip[socket] = strdup(ipAddr);
```
DL6ER: Can maybe remove this, see comment on FTL.h:260

socket.c
```diff
@@ -217,33 +222,62 @@ void ssend(int sock, const char *format, ...)
 	}
 }
 
+void swrite(int sock, void *value, size_t size) {
+	if(write(sock, value, size) == -1)
+		logg("WARNING: Socket write returned error code %i", errno);
+}
+
+int checkClientLimit(int socket, char *ipAddr) {
```
DL6ER: We can remove the `char * ipAddr` argument (see below)

socket.c
```diff
@@ -217,33 +222,62 @@ void ssend(int sock, const char *format, ...)
...
+			return accept(sockfd, (struct sockaddr *) &un_addr, &socklen);
 
 		case 4: // Internet socket (IPv4)
-		memset(&in4_addr, 0, sizeof(in4_addr));
-		socklen = sizeof(un_addr);
-		return accept(sockfd, (struct sockaddr *) &in4_addr, &socklen);
+			memset(&in4_addr, 0, sizeof(in4_addr));
+			socklen = sizeof(un_addr);
+			socket = accept(sockfd, (struct sockaddr *) &in4_addr, &socklen);
+			return checkClientLimit(socket, inet_ntoa(in4_addr.sin_addr));
```
DL6ER: Change to
```
socket = accept(sockfd, (struct sockaddr *) &in4_addr, &socklen);
return checkClientLimit(socket);
```

socket.c
```diff
@@ -217,33 +222,62 @@ void ssend(int sock, const char *format, ...)
...
 		case 6: // Internet socket (IPv6)
-		memset(&in6_addr, 0, sizeof(in6_addr));
-		socklen = sizeof(un_addr);
-		return accept(sockfd, (struct sockaddr *) &in6_addr, &socklen);
+			memset(&in6_addr, 0, sizeof(in6_addr));
+			socklen = sizeof(un_addr);
+			char str[INET6_ADDRSTRLEN];
+			socket = accept(sockfd, (struct sockaddr *) &in6_addr, &socklen);
+			inet_ntop(AF_INET6, &in6_addr.sin6_addr, str, INET6_ADDRSTRLEN);
+			return checkClientLimit(socket, str);
```
DL6ER: Change to
```
socket = accept(sockfd, (struct sockaddr *) &in6_addr, &socklen);
return checkClientLimit(socket);
```

socket.c
```diff
@@ -217,33 +222,62 @@ void ssend(int sock, const char *format, ...)
...
+			socket = accept(sockfd, (struct sockaddr *) &in4_addr, &socklen);
+			return checkClientLimit(socket, inet_ntoa(in4_addr.sin_addr));
 
 		case 6: // Internet socket (IPv6)
-		memset(&in6_addr, 0, sizeof(in6_addr));
-		socklen = sizeof(un_addr);
-		return accept(sockfd, (struct sockaddr *) &in6_addr, &socklen);
+			memset(&in6_addr, 0, sizeof(in6_addr));
+			socklen = sizeof(un_addr);
+			char str[INET6_ADDRSTRLEN];
```
DL6ER: Can be removed

socket.c
```diff
@@ -293,7 +330,7 @@ void *telnet_connection_handler_thread(void *socket_desc)
 			// Requests should not be processed/answered when data is about to change
 			enable_thread_lock(threadname);
 
-			process_request(message, &sock);
+			process_request(message, &sock, TELNET);
```
DL6ER: `TELNET` is not needed here, as we have `istelnet[sock]`

socket.c
```diff
@@ -319,6 +356,11 @@ void *telnet_connection_handler_thread(void *socket_desc)
 		close(sock);
 	free(socket_desc);
 
+	if(clientip[sock] != NULL) {
```
DL6ER: Can be removed as well.

socket.c
```diff
@@ -351,7 +395,7 @@ void *socket_connection_handler_thread(void *socket_desc)
 			// Requests should not be processed/answered when data is about to change
 			enable_thread_lock(threadname);
 
-			process_request(message, &sock);
+			process_request(message, &sock, SOCKET);
```
DL6ER: Also here, `SOCKET` can be removed.

socket.c
```diff
@@ -376,6 +420,11 @@ void *socket_connection_handler_thread(void *socket_desc)
 		close(sock);
 	free(socket_desc);
 
+	if(clientip[sock] != NULL) {
```
DL6ER: Can remove this as well.

msgpack.c
```diff
@@ -0,0 +1,112 @@
...
+	uint8_t format = (uint8_t) (0xA0 | length);
+	swrite(sock, &format, sizeof(format));
+	swrite(sock, string, length);
+}
+
+void pack_str32(int sock, char *string) {
+	// Make sure that the length is less than 4294967296
+	size_t length = strlen(string);
+
+	if(length >= 4294967296) {
```
DL6ER:
```
msgpack.c: In function ‘pack_str32’:
msgpack.c:95:12: warning: comparison is always false due to limited range of data type [-Wtype-limits]
  if(length >= 4294967296) {
```

Mcat12: When compiling on a 64 bit system, this warning does not appear. Do you think we should check for this?

DL6ER: 4294967296 is slightly more than 4 GB, so we really expect strings to become that large?

If you want to keep the check (which is probably, a good idea, test for the length of `size_t`), maybe like this:
```c
if((sizeof(size_t) == 8) && length >= 4294967296)
{ ... }
```

There shouldn't be any performance penalty here, as with any (non-zero) level of optimization the compiled binary won't actually perform the test as the compiler will already eliminate whichever branch is dead at compile time.

Mcat12: That code still gives the compiler warning. How about reducing it to:
```c
if(length >= 2147483648) { ... }
```
The compiler does not yell about this, but it is a slightly sloppy way to avoid it.

api.c
```diff
@@ -0,0 +1,1216 @@
...
+
+	if(debugclients)
+		logg("Sent stats data to client, ID: %i", *sock);
+}
+
+void getOverTime(int *sock)
+{
+	int i, j = 9999999;
+
+	// Get first time slot with total or blocked greater than zero (the array will go down over time due to the rolling window)
```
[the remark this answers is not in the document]
Mcat12: I think you wrote this comment ;)

api.c
```diff
@@ -0,0 +1,1216 @@
...
+	char *clientname = NULL;
+	bool filterclientname = false;
+
+	// Time filtering?
+	if(command(client_message, ">getallqueries-time")) {
+		sscanf(client_message, ">getallqueries-time %i %i",&from, &until);
+	}
+	// Domain filtering?
+	if(command(client_message, ">getallqueries-domain")) {
+		sscanf(client_message, ">getallqueries-domain %ms", &domainname);
```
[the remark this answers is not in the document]
Mcat12: Here is how development currently handles it:
https://github.com/pi-hole/FTL/blob/master/request.c#L680-L685

This version was implemented in this commit:
https://github.com/pi-hole/FTL/commit/48d355bab0e48b09029c62a0ccd199eb046d90ed

api.c
```diff
@@ -0,0 +1,1216 @@
...
+		strncpy(hash, commit, 7); hash[7] = 0;
+
+		if(istelnet[*sock])
+			ssend(*sock, "version vDev-%s\ntag %s\nbranch %s\ndate %s\n", hash, tag, GIT_BRANCH, GIT_DATE);
+		else {
+			char *hashVersion = calloc(5 + strlen(hash), sizeof(char));
+			sprintf(hashVersion, "vDev-%s", hash);
+
+			pack_str32(*sock, hashVersion);
+			pack_str32(*sock, (char *) tag);
```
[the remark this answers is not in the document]
Mcat12: There is a compiler warning otherwise:
```
api.c: In function 'getVersion':
api.c:1007:22: warning: passing argument 2 of 'pack_str32' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
    pack_str32(*sock, tag);
                      ^~~
In file included from api.c:12:0:
api.h:41:6: note: expected 'char *' but argument is of type 'const char *'
 void pack_str32(int sock, char *string);
      ^~~~~~~~~~
```

api.c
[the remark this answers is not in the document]
DL6ER: > This version was implemented in this commit:
48d355b

Is this the nice way of blaming me for it? ;-) I'm not always and only having great ideas.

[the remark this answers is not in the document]
Mcat12: Do you want me to revert it back to the development way?

msgpack.c
DL6ER: I cannot test this at the moment but there should be a preprocessor constant for the maximum allowed size of `size_t`. Maybe something like
```c
#if __SIZE_MAX__ > 4294967295
if(length >= 4294967296)
    { ... }
#endif
```

api.c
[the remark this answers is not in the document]
DL6ER: Yes, please. Maybe increase it to 256 if you find it necessary.

DL6ER:
**Approved** :tada:

PR #119, opened 2017-07-29
Ignore invalid domains (containing ")

Mcat12:

---

Rarely, a query with just the domain `"` will show up. As it's an invalid domain, ignore it.

### DIFF
--- a/parser.c
+++ b/parser.c
@@ -203,6 +203,12 @@ void process_pihole_log(int file)
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


DL6ER:
How about `'` ? Could that interfere with the future JSON output as well?

Mcat12:
`'` is a valid character in JSON, so we don't need to escape it. JSON strings are always double quotes, and so the only major characters we might need to escape are `\` and `"`, although I don't think you can make a DNS query with the former.

PR #124, opened 2017-08-13
Some tweaks for stats and top-clients

DL6ER:

## 10

---

Currently, `unique_clients` (in the output of `>stats`) returns the number of clients that `FTL` has *ever* seen in its runtime. However, this is inconsistent with the data returned by `>top-clients` which only contains clients that are active, i.e. have made *at least* one query in the most recent 24 hours.

This PR makes the behavior more consistent:

- `>stats` will return `unique_clients` that is consistent with the `>top-lists` output (count only active clients). We add a new property `clients_ever_seen` to preserve the information of how many unique clients FTL has seen in its runtime (whether or not they have been active recently) to not loose the availability of this data.

- We add `withzero` as an option to `>top-clients` (e.g. `>top-clients withzero (15)`). With this flag enabled, the top clients result will contain *all* clients that are know to `FTL`, i.e. also those which haven't been active recently.

- Update `README.md`

### DIFF
--- a/request.c
+++ b/request.c
@@ -234,10 +234,13 @@ void getStats(int *sock)
 	sprintf(server_message,"unique_domains %i\nqueries_forwarded %i\nqueries_cached %i\n", \
 	        counters.domains,counters.forwardedqueries,counters.cached);
 	swrite(server_message, *sock);
+
+	// clients_ever_seen: all clients ever seen by FTL
 	sprintf(server_message,"clients_ever_seen %i\n", \
 	        counters.clients);
 	swrite(server_message, *sock);
 
+	// unique_clients: count only clients that have been active within the most recent 24 hours
 	int i, activeclients = 0;
 	for(i=0; i < counters.clients; i++)
 	{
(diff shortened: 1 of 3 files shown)


request.c
```diff
@@ -441,7 +460,7 @@ void getTopClients(char *client_message, int *sock)
 			}
 		}
 
-		if(clients[j].count > 0)
+		if(includezeroclients || clients[j].count > 0)
```
DL6ER: Return this client if either

1. `withzero` option is set, and/or
2. the client made at least one query within the most recent 24 hours

request.c
```diff
@@ -404,6 +416,13 @@ void getTopClients(char *client_message, int *sock)
 		count = num;
 	}
 
+	// Apply Audit Log filtering?
+	bool includezeroclients = false;
+	if(command(client_message, " withzero"))
```
DL6ER: This option can be combined with existing options, i.e. both `>top-clients withzero` and `>top-clients withzero (123)` are valid

DL6ER:
Note that these changes will propagate through the PHP API so there is no need for changes on the web repo.

PR #125, opened 2017-09-06
Re-resolve client host names in regular intervals

DL6ER:

## 10

---

With this modification, `FTL` tries to re-resolve the client's host names after each GC run (each full hour).
This tackles two issues:

- Host names that couldn't be (for whatever reason) resolved when `FTL` first encountered a new client (see https://github.com/pi-hole/pi-hole/issues/1507 )
- Host names that changed during the run time of `FTL`

### DIFF
--- a/parser.c
+++ b/parser.c
@@ -1032,6 +1032,7 @@ void reresolveHostnames(void)
 			// Delete possibly already existing hostname pointer before storing new data
 			if(clients[clientID].name != NULL)
 			{
+				memory.clientnames -= (strlen(clients[clientID].name) + 1) * sizeof(char);
 				free(clients[clientID].name);
 				clients[clientID].name = NULL;
 			}
--- a/parser.c
+++ b/parser.c
@@ -1015,3 +1015,31 @@ void validate_access_oTfd(int timeidx, int pos, int line, const char * function,
 		logg("             found in %s() (line %i) in %s", function, line, file);
 	}
 }
+
+void reresolveHostnames(void)
+{
+	int clientID;
+	for(clientID = 0; clientID < counters.clients_MAX; clientID++)
+	{
+		// Process this client only if it has at least one active query in the log
+		if(clients[clientID].count < 1)
+			continue;
+
+		// Get client hostname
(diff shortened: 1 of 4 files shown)


parser.c
```diff
@@ -1015,3 +1015,31 @@ void validate_access_oTfd(int timeidx, int pos, int line, const char * function,
...
+
+		// Get client hostname
+		char *hostname = resolveHostname(clients[clientID].ip);
+		if(strlen(hostname) > 0)
+		{
+			// Delete possibly already existing hostname pointer before storing new data
+			if(clients[clientID].name != NULL)
+			{
+				free(clients[clientID].name);
+				clients[clientID].name = NULL;
```
Mcat12: Setting it to null isn't necessary as you set it to the new hostname immediately after this.

parser.c
```diff
@@ -1015,3 +1015,31 @@ void validate_access_oTfd(int timeidx, int pos, int line, const char * function,
...
+			}
+
+			// Store client hostname
+			clients[clientID].name = strdup(hostname);
+			memory.clientnames += (strlen(hostname) + 1) * sizeof(char);
```
Mcat12: Shouldn't you also decrement the memory usage when you free the old hostname?

parser.c
DL6ER: It is good practice to set a pointer to `NULL` after `free`ing it and I'd like to keep it even though we may be setting it to a new value afterwards. I guess the compiler is going to optimize it away anyway in the compiled binaries.

> On a big, long-lived project, there are good reasons to set pointers to NULL
https://stackoverflow.com/a/1879678/2087442

parser.c
DL6ER: Yes, I'll add that.

parser.c
Mcat12: Although I think that post is more focused on freed variables that won't be reassigned, there are some valid arguments there.

parser.c
DL6ER: Added in 6445020

PR #130, opened 2017-09-20
Bugfix: Compute properly normalized percentages

DL6ER:

## 10

---

Properly normalize the percentages computed from the raw statistics. This is a bug fix PR.
Very briefly summarized what went wrong: When `dnsmasq` receives a query that cannot be answered from cache, it may be forwarded to more than one forward destination. Int his case we would count e.g. four forwarded queries although we only made one request to `dnsmasq`. Hence, the count of forwarded requests has to be normalized by the number of queries that required a forwarding. The wrong computation is fixed by this PR. Most notable is that the `local` destination should be more prominent for most users as the previous situation was quite "adverse" for this "destination".

Previously
screenshot at 2017-09-20 17-13-00

Corrected
screenshot at 2017-09-20 17-12-24

This PR changes also the output of the query types from absolute (counts) to relative (percentage) outputs for consistency. However, the displayed statistics do not actually change as no normalization is required here.

### DIFF
--- a/request.c
+++ b/request.c
@@ -853,6 +853,8 @@ void getForwardDestinationsOverTime(int *sock)
 {
 	char server_message[SOCKETBUFFERLEN];
 	int i, sendit = -1;
+	double percentage;
+
 	for(i = 0; i < counters.overTime; i++)
 	{
 		validate_access("overTime", i, true, __LINE__, __FUNCTION__, __FILE__);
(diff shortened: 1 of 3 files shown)


request.c
```diff
@@ -526,22 +535,42 @@ void getForwardDestinations(char *client_message, int *sock)
...
+			//
+			// The fraction a describes now how much share an individual forward destination
+			// has on the total sum of sent requests.
+			// We also know the share of forwarded queries on the total number of queries
+			//   b = counters.forwardedqueries/counters.queries
+			//
+			// To get the total percentage of a specific query on the total number of queries,
+			// we simply have to scale b by a which is what we do in the following.
+			if(forwardedsum > 0 && counters.queries > 0)
+				percentage = 1e2*forwarded[j].count/forwardedsum*counters.forwardedqueries/counters.queries;
```
Mcat12: It would be more readable to include spaces between the mathematical operators.

Mcat12: Same goes for other instances of long lines of math.

request.c
```diff
@@ -854,23 +870,57 @@ void getForwardDestinationsOverTime(int *sock)
...
+				{
+					// This forward destination does not yet exist at this timestamp
+					// -> use zero as number of requests sent to this destination
 					k = 0;
+				}
+
+				// Avoid floating point exceptions
+				if(forwardedsum > 0 && overTime[i].total > 0)
+				{
+					percentage = 1e2*k/forwardedsum*(overTime[i].total - (overTime[i].cached + overTime[i].blocked))/overTime[i].total;
```
Mcat12: Can you clear up some of this math with some more parentheses? (particularly with relation to the division) And/or add a comment explaining the math.

DL6ER: Parentheses are not needed here as `a/b*c/d == (a*c)/(b*d) == a*c/b/d`. However, I will add a description on how this works.

Mcat12: It might be difficult to read the code without the parentheses because those rules might not be known, or aren't very intuitive.

request.c
```diff
@@ -916,7 +930,8 @@ void getForwardDestinationsOverTime(int *sock)
 
 			// Avoid floating point exceptions
 			if(overTime[i].total > 0)
-				percentage = 1e2*(overTime[i].cached + overTime[i].blocked)/overTime[i].total;
+				// Forward count for destinatio "local" is cached + blocked normalized by total:
```
Mcat12: misspelling of "destination"

PR #133, opened 2017-09-22
Ensure percentages in forward destinations do always sum up to one

DL6ER:

## 10

---

After some time, the garbage collection algorithm will invalidate queries such that they are not shown anymore in the data. All counters *except* `counters.queries` are corrected by the amount of removed queries. The total counter is, however, *not* corrected due to bookkeeping reasons. Hence, using it for normalization is an inadequate choice when garbage collection is used. While the ratios are still reported correctly, the too high value in `counters.queries` leads to the phenomenon that the individual percentages do not sum up to 100% any more.

This PR fixes this problem by computing the correct number to be used for the normalization at the time it is needed.

### DIFF
--- a/request.c
+++ b/request.c
@@ -483,7 +483,7 @@ void getForwardDestinations(char *client_message, int *sock)
 {
 	char server_message[SOCKETBUFFERLEN];
 	bool allocated = false, sort = true;
-	int i, temparray[counters.forwarded+1][2], forwardedsum = 0;
+	int i, temparray[counters.forwarded+1][2], forwardedsum = 0, totalqueries = 0;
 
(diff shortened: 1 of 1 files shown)


request.c
```diff
@@ -513,6 +513,8 @@ void getForwardDestinations(char *client_message, int *sock)
 		qsort(temparray, counters.forwarded+1, sizeof(int[2]), cmpdesc);
 	}
 
+	totalqueries = counters.forwardedqueries + counters.cached + counters.blocked;
```
DL6ER: I compute this outside of the loop to avoid re-computing the very same number over and over again

PR #134, opened 2017-09-22
Add clients over time data structure

DL6ER:

## 10

---

See https://github.com/pi-hole/AdminLTE/pull/583

### DIFF
--- a/FTL.h
+++ b/FTL.h
@@ -168,6 +168,8 @@ typedef struct {
 	int forwardnum;
 	int *forwarddata;
 	int *querytypedata;
+	int clientnum;
+	int *clientdata;
 } overTimeDataStruct;
 
 typedef struct {
@@ -178,6 +180,7 @@ typedef struct {
 	int forwardedips;
 	int forwardednames;
 	int forwarddata;
+	int clientdata;
 	int querytypedata;
 } memoryStruct;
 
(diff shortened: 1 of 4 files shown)


DL6ER:
Apply `API_EXCLUDE_CLIENTS` filtering as suggested by @WaLLy3K in private communication.

No filtering:
screenshot at 2017-09-29 22-33-55

Assume I want to filter:

- `localhost` (bright blue) as it is constantly active
- `desktop` (darker blue starting around 20:00) just to demonstrate it works.

I apply the following filter:

screenshot at 2017-09-29 22-33-43
where I intentionally used the host name for `localhost` and the IP address for `desktop` to show that both can be used at liberty to identify individual clients.

The result with the above filter is:

screenshot at 2017-09-29 22-33-31

`localhost` and `desktop` are missing from the data. Note that the colors between the two screenshots have changed as the chart generator isn't even aware of the existence of the other clients. This seems to be useful in many situations, e.g. it allows us to monitor newly added clients like smartphones where we hide the activity of the already known clients.

Mcat12:
The web interface should change its description from just top clients to include this feature.

WaLLy3K:
I believe this is a "source data" issue rather than a "data handling" issue, so I'm assuming this should be under FTL's scope:

untitled

The menu gets cut off when there's more than a handful of clients, and there's an entry with a blank name - perhaps because there's two devices on my network which haven't had hostnames manually assigned to them via `/etc/hosts`

PR #135, opened 2017-09-24
Always send EOM at end of transmission

DL6ER:

## 10

---

Always send `EOM` at end of transmission (if connection is still alive).

I found that when we implement new features on the web interface (e.g. the Clients over time graph) but don't have the according functionality already in the background (i.e. `FTL` running on a matching branch), then the `PHP API` will hang. The reason is that it continues to wait for additional data as long as it didn't see the end of transmission (`EOM`) indicator. However - so far - the return for an unknown command
```
unknown command: abcd
```
did not send an `EOM` telegram so the API was always assuming that additional information may be on its way and git stuck in an infinity loop there.

Proposed solution: Always send `EOM` as it doesn't hurt and actually makes the telnet output more consistent.
```
unknown command: abcd
---EOM---
```

Note that this specific "bug" is very unlikely to ever affect users that are using it in production machines as they will always have a version of `FTL` that is in sync with the other branches (as `FTL` is either released at the same time or even a bit before the other branches). Nevertheless, this change eases development as we don't have to care about if the branches are in the correct causal connection.

### DIFF
--- a/request.c
+++ b/request.c
@@ -134,7 +134,7 @@ void process_request(char *client_message, int *sock)
 
 	if(!processed)
 	{
-		sprintf(server_message,"unknown command: %s\n",client_message);
+		sprintf(server_message,"unknown command: %s",client_message);
 		swrite(server_message, *sock);
 	}
 
--- a/request.c
+++ b/request.c
@@ -114,13 +114,6 @@ void process_request(char *client_message, int *sock)
 		getDBstats(sock);
 	}
 
-	// End of queryable commands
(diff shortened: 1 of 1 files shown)


request.c
```diff
@@ -144,6 +137,13 @@ void process_request(char *client_message, int *sock)
 		sprintf(server_message,"unknown command: %s\n",client_message);
```
Mcat12: You can remove the extra newline here, as the EOM now handles that.

PR #142, opened 2017-10-03
only compare the ends of two domains instead of entire domain string

jftuga:

**How familiar are you with the codebase?:** 3

---

**Problem**
On the Top Domains portion of the dashboard, multiple host names appear for the same domain name which can make one domain name dominate the results.  For example, three of the top 10 domains for my system include something1.nflxso.net, something2.nflxso.net, and something3.nflxso.net.  

**Proposal**
Instead of having to individually add each host name to the Settings -> Top Lists -> Excludes, only add the domain (in this case  nflxso.net).  The code for insetupVarsArray() currently checks for an identical strings via strcmp().  I have written a patch that only returns true when the end of the two strings are identical.  Thus, and entry of nflxso.net will return true for anything ending in nflxso.net.  By doing this, a more varied list of Top Domains will be displayed on the dashboard.

### DIFF
--- a/setupVars.c
+++ b/setupVars.c
@@ -183,7 +183,16 @@ bool insetupVarsArray(char * str)
 		}
 		else
 		{
-			if(strcmp(setupVarsArray[i], str) == 0)
+			// Return true only when the ends of the two domains match
+			// This allows an exclusion of 'nflxso.net' to not display:
+			// occ-2-990-987.1.nflxso.net, occ-0-990-987.1.nflxso.net, etc.
+			size_t candidate = strlen(str);
+			size_t excluded = strlen(setupVarsArray[i]);
+			if (excluded > candidate)
+				continue;
+
+			size_t end_pos = candidate - excluded;
+			if(strcmp(str + end_pos, setupVarsArray[i]) == 0)
 				return true;
 		}
 


WaLLy3K:
This feature already exists within FTL, however, you'll need to run `pihole checkout web devel` to be able to exclude `*nflxso.net` from the Web Interface :smile:

DL6ER:
We already support this! The code for it is only a few lines above where you coded your fix. Please explain where you think the difference is.

DL6ER:
> This feature already exists within FTL, however, you'll need to run pihole checkout web devel to be able to exclude *nflxso.net from the Web Interface 

You could use it like it is already if you manually add the domain `*nflxso.net` into your `setupVars.conf`

jftuga:
In the master branch, when I try to add *.example.com, I get this error message:

    Top Domains/Ads entry *.example.com is invalid (use only domains)!
    The settings have been reset to their previous values

I guess I should have tried the development branch first. I will try it out today.
Sorry guys.

DL6ER:
No worries, it is something that is supported in `FTL` for quite some time, but we forgot to actually allow `*` in the domain validator on the web interface, so this feature was kind of hidden, unfortunately.

jftuga:
If I wanted to started working on the domain validator web interface code, would I start here?

https://github.com/pi-hole/AdminLTE/blob/devel/scripts/pi-hole/php/savesettings.php#L46

    scripts/pi-hole/php/savesettings.php 
    validDomainWildcard()

Using the devel branch?

DL6ER:
It has already been fixed in `devel`. See #581

PR #150, opened 2017-10-22
Bump sqlite to 3.20.1

JamieMagee:

3

---

Bump sqlite from `3.20.0`, to `3.20.1`. Release notes can be found [here](https://www.sqlite.org/releaselog/3_20_1.html).

### DIFF
--- a/sqlite3.h
+++ b/sqlite3.h
@@ -121,9 +121,9 @@ extern "C" {
 ** [sqlite3_libversion_number()], [sqlite3_sourceid()],
 ** [sqlite_version()] and [sqlite_source_id()].
 */
-#define SQLITE_VERSION        "3.20.0"
-#define SQLITE_VERSION_NUMBER 3020000
-#define SQLITE_SOURCE_ID      "2017-08-01 13:24:15 9501e22dfeebdcefa783575e47c60b514d7c2e0cad73b2a496c0bc4b680900a8"
+#define SQLITE_VERSION        "3.20.1"
+#define SQLITE_VERSION_NUMBER 3020001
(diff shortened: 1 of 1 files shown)


DL6ER:
Thank you for your submission. We haven't considered this update so far, as the only change,

> Fix a potential memory leak in the new sqlite3_result_pointer() interface. Ticket 7486aa54b968e9b5.

is within a routine we are not even using. We run an extensive set of (internal, undocumented) tests to ensure that an update of SQLite doesn't have negative effects on existing databases. I know that the maintainers run tests themselves but we want to be sure. We haven't undertaken this effort for this update yet, as we haven't seen any value in it. I really don't know if we will do it or not, but I'd say it is rather unlikely.

JamieMagee:
Regarding your internal unit/integration tests, is it possible to make those public? I think that more tests could only help new contributors.

As for the change, I believe that it is a small, low risk change, but in the end it's your call.

DL6ER:
I see your point but the reason for why the procedure is not public is mainly that it is not standardized. What we try to do is to break the database implementation "using all means". See also [Fuzz testing](https://en.wikipedia.org/wiki/Fuzzing), although we do not do it in quite an elegant (automated) fashion.

SQLite 3.21 is (slowly) on its way and should be release in the near future (http://www.sqlite.org/src/timeline?n=100&r=branch-3.21). I am inclined to wait until we can make a bigger update as the currently proposed update does not seem to bring any advantage for our users.

JamieMagee:
I would argue that it is entirely unnecessary to perform your own fuzz testing on a dependency that you don't own.  Also because [SQLite performs fuzz testing](https://www.sqlite.org/testing.html#sql_fuzz_using_the_american_fuzzy_lop_fuzzer) as part of their standard test procedure, I imagine that they already cover a larger surface area than any custom fuzz testing.

PromoFaux:
Closing as we will be waiting for at least 3.21 before updating sqlite

PR #151, opened 2017-10-25
Re-parse gravity data when receiving SIGHUP

DL6ER:

## 10

---

Commit https://github.com/pi-hole/pi-hole/commit/a2825be81961cee4ddbb37f9e984fe6a27faa99a in the core code base changed the standard behavior of `pihole -g` (and some others) to not restart `dnsmasq` any longer, but only instruct it to re-read its config files. Unfortunately, there is no reliable way for `FTL` to detect the latter.
Hence, https://github.com/pi-hole/pi-hole/pull/1751 will make the corresponding code send the `SIGHUP` not only to `dnsmasq` but also to `FTL`. This PR implements the code that is needed to react accordingly to this new signal.

### DIFF
--- a/parser.c
+++ b/parser.c
@@ -704,16 +704,6 @@ void process_pihole_log(int file)
 			validate_access_oTfd(timeidx, forwardID, __LINE__, __FUNCTION__, __FILE__);
 			overTime[timeidx].forwarddata[forwardID]++;
 		}
-		else if((strstr(readbuffer,"IPv6") != NULL) &&
-		        (strstr(readbuffer,"DBus") != NULL) &&
-		        (strstr(readbuffer,"i18n") != NULL) &&
-		        (strstr(readbuffer,"DHCP") != NULL) &&
-		         !initialscan)
-		{
-			// dnsmasq restartet
-			logg("dnsmasq process restarted");
-			read_gravity_files();
-		}
 
 		// Save file pointer position, because we might have to repeat
 		// reading the next line if dnsmasq hasn't finished writing it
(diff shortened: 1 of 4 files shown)


parser.c
```diff
@@ -704,16 +704,6 @@ void process_pihole_log(int file)
 			validate_access_oTfd(timeidx, forwardID, __LINE__, __FUNCTION__, __FILE__);
 			overTime[timeidx].forwarddata[forwardID]++;
 		}
-		else if((strstr(readbuffer,"IPv6") != NULL) &&
```
DL6ER: I remove this code as it was always a dirty hack to detect whether `dnsmasq` was restarted. We cannot use it to detect a `reload` condition. As gravity will now always send a `SIGHUP`, we can safely remove this code here.

Mcat12: How about the case where the user manually restarts dnsmasq such as through `sudo service dnsmasq restart`?

DL6ER: `FTL` doesn't need to know if/when `dnsmasq` is restarted, strictly speaking. We used the fact that `dnsmasq` is restarted in the past to decide that the user may have changed something (updated gravity, added/removed blacklist entries, etc.) that needs our attention. This PR, however, implements a direct communication channel between our core code and `FTL` through which is will be "informed" when it should re-parse the gravity files. If the user just does a manual `restart`, we don't actually have to re-parse the lists as it is highly unlikely that the analyzed lists have changed. Hence, it is okay if we miss these restarts as they contain no information that would be of value to `FTL`.

PR #153, opened 2017-10-26
Musl ARM build

diginc:
- [yup] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [yup] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [yup] I have considered, and confirmed that this submission will be valuable to others.
- [yup] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [yup] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** somewhere between 3 and 3.2 but this isn't really code :D

---

Someone recently pull requested a dockerfile for alpine arm (and musl arm64 but turns out [that's not happening](https://wiki.musl-libc.org/porting/aarch64.html)) in my docker repo.  This gets the pihole-FTL built for the musl libc + arm architecture combo correctly I believe.  I couldn't find a CC='opt' to get musl arm to build on travis's standard x86 build systems but the emulated arm docker debian jessie has a musl-tools which seems to build it no problem:

>The command "$DOCKER make pihole-FTL CC="${BUILD_CC}" CFLAGS="${CFLAGS}"" exited with 0.
> 1.00s$ $DOCKER file pihole-FTL
> pihole-FTL: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-armhf.so.1, not stripped

https://travis-ci.org/diginc/FTL/jobs/292975023

**Pros:** +1 cross compile.  It doesn't break existing arch builds since I prefixed null $DOCKER variables that only the musl-arm env build fill in.
**Cons:** Kinda slow build due to on the fly docker (It should maybe be build separately, pushed, and then we could just pull it much faster than building).  Installing docker-compose takes up some time too but it's just so convenient.  I'll try optimizing this a little tomorrow if desired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ matrix:
     - BUILD_CC="musl-gcc"
     - BIN_NAME="pihole-FTL-musl-armhf"
     - DOCKER_INIT="docker run --rm --privileged multiarch/qemu-user-static:register --reset"
-    - DOCKER='docker run --rm -it -v "$TRAVIS_BUILD_DIR:/build" diginc/multiarch-deb-debootstrap:armhf-musl-jessie'
+    - DOCKER="docker run --rm -it -v ${TRAVIS_BUILD_DIR}:/build diginc/multiarch-deb-debootstrap:armhf-musl-jessie"
     sudo: true
 before_script:
 - mkdir -p "${DEST_DIR}"


DL6ER:
I see in your build that it took `8 min 16 sec` to run the build. Although this is not terribly bad, we'd like to see a faster build if possible.

I think we'll have to add an exception for this new architecture in `pihole checkout ftl ...` or modify the building machine such that it can generate the binaries as well.

> It should maybe be build separately, pushed, and then we could just pull it much faster than building

How would you do that?

diginc:
> How would you do that?

`docker build`ing an image with lots of download, extract, & installs of packages is typically a lot slower than `docker pull`ing an an identical image from hub.docker.com.  If I setup an automated docker hub build (builds + uploads docker images) of the Dockerfile `dockerfiles/armhf` and modified the docker-compose.yaml to say `image: diginc/deb_bootstrap:armhf_musl` instead of the `build` block it would go as fast as travic CI could download `diginc/deb_bootstrap:armhf_musl` rather than reinstalling build-essentials and musl-tools every time.

diginc:
Looking at your overall travis-ci build times the musl ARM build actually doesn't hit the total time (all architectures) too much since it doesn't seem to actually run fully paralleled - i.e. not all architectures start at the same time.  

As long as the musl ARM build isn't ran completely last it seem total matrix build time stays between 10-14m even prior to adding the 6m musl arm build.

DL6ER:
What's suspending this PR is that the build procedure for the non-`master` has to be customized before and I don't know when anyone of us will find the time for doing this...

DL6ER:
We are waiting on #160 to get merged before this PR. However, I just realized that we now also need a criterion on which we would determine that we would have to download *this* instead of another binary. See https://github.com/pi-hole/pi-hole/blob/development/automated%20install/basic-install.sh#L1782 for more detail.

diginc:
Reverted?

I don't think there's any official alpine support, just my docker, so updating the installer shouldn't be necessary.  https://github.com/pi-hole/pi-hole/search?q=alpine&type=Issues&utf8=%E2%9C%93

I'm currently patching the installer to grab the right amd64 musl FTL binary here: https://github.com/diginc/docker-pi-hole/blob/master/install.sh#L29

frenck:
Second vote here... I would like to see official builds based on musl as well.

@diginc Imho you should stop trying to fix the installer and take a different approach here, you might want to take a look at the Pi-hole add-on of Home Assistant that is based on Alpine as well.

PR #158, opened 2017-11-04
Implement "invalid option" and "help" in argument checker

DL6ER:

## 10

---

Entering an invalid option will prevent `pihole-FTL` from starting at all, e.g.
```
$ ./pihole-FTL status
pihole-FTL: invalid option -- 'status'
Try './pihole-FTL --help' for more information
```

I also added a new `help` option, returning:
```
$ ./pihole-FTL -h
pihole-FTL - The Pi-hole FTL engine

Usage:    sudo service pihole-FTL <action>
where '<action>' is one of start / stop / restart

Available arguments:
        d,  debug         More verbose logging,
                          don't go into daemon mode
            test          Don't start pihole-FTL but
                          instead quit immediately
        -v, version       Return version
        -t, tag           Return git tag
        -b, branch        Return git branch
            running       Test if another pihole-FTL
                          process is running and exit
                          even if not (instead of
                          starting a new one)
        -f, no-daemon     Don't go into daemon mode
        -h, help          Display this help and exit


Online help: https://github.com/pi-hole/FTL
```

### DIFF
--- a/args.c
+++ b/args.c
@@ -136,14 +136,14 @@ void parse_args(int argc, char* argv[])
 			printf("\t                  starting a new one)\n");
 			printf("\t-f, no-daemon     Don't go into daemon mode\n");
 			printf("\t-h, help          Display this help and exit\n");
-			printf("\n\nOnline help: https://github.com/pi-hole/FTL\n\n");
+			printf("\n\nOnline help: https://github.com/pi-hole/FTL\n");
 			exit(EXIT_SUCCESS);
 		}
 
 		// Complain if invalid options have been found
 		if(!ok)
 		{
-			printf("pihole-FTL: invalid option -- '%s'\nTry '%s --help' for more information\n\n", argv[i], argv[0]);
+			printf("pihole-FTL: invalid option -- '%s'\nTry '%s --help' for more information\n", argv[i], argv[0]);
 			exit(EXIT_FAILURE);
 		}
 	}
--- a/args.c
+++ b/args.c
@@ -22,38 +22,50 @@ bool travis = false;
 void parse_args(int argc, char* argv[])
 {
 	int i;
-	for(i=0; i < argc; i++) {
-		if((strcmp(argv[i], "d") == 0) || (strcmp(argv[i], "debug") == 0))
+	// start from 1, as argv[0] is the executable name "pihole-FTL"
+	for(i=1; i < argc; i++)
+	{
+		bool ok = false;
+		if(strcmp(argv[i], "d") == 0 ||
+		   strcmp(argv[i], "debug") == 0)
+		{
 			debug = true;
+			ok = true;
+		}
 
 		if(strcmp(argv[i], "debugthreads") == 0)
 		{
 			debug = true;
 			debugthreads = true;
+			ok = true;
 		}
 
 		if(strcmp(argv[i], "debugclients") == 0)
 		{
 			debug = true;
 			debugclients = true;
+			ok = true;
 		}
 
 		if(strcmp(argv[i], "debugGC") == 0)
 		{
 			debug = true;
 			debugGC = true;
+			ok = true;
 		}
 
 		if(strcmp(argv[i], "debugDB") == 0)
 		{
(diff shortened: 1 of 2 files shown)


args.c
```diff
@@ -94,8 +115,39 @@ void parse_args(int argc, char* argv[])
...
+			printf("\t-h, help          Display this help and exit\n");
+			printf("\n\nOnline help: https://github.com/pi-hole/FTL\n");
+			exit(EXIT_SUCCESS);
 		}
 
-		// Other arguments are ignored
+		// Complain if invalid options have been found
+		if(!ok)
+		{
+			printf("pihole-FTL: invalid option -- '%s'\nTry '%s --help' for more information\n", argv[i], argv[0]);
```
DL6ER: Note that this will use `argv[0]` for the `help` command suggestion, i.e. if it is compiled locally in some subfolder and started via `some/folder/pihole-FTL abc` it will also know that and return:
```
$ some/folder/pihole-FTL abc
pihole-FTL: invalid option -- 'abc'
Try 'some/folder/pihole-FTL help' for more information
```

args.c
```diff
@@ -94,8 +115,39 @@ void parse_args(int argc, char* argv[])
...
+
+		// List of implemented arguments
+		if(strcmp(argv[i], "-h") == 0 ||
+		   strcmp(argv[i], "help") == 0)
+		{
+			printf("pihole-FTL - The Pi-hole FTL engine\n\n");
+			printf("Usage:    sudo service pihole-FTL <action>\n");
+			printf("where '<action>' is one of start / stop / restart\n\n");
+			printf("Available arguments:\n");
+			printf("\td,  debug         More verbose logging,\n");
```
Mcat12: Should the short form of `debug` have a hyphen before it, like the others do? `-d`

EvgenijM86:
This breaks lxc-net script, as it expects to be able to give "--" parameters.

DL6ER:
What is `lxc-net` script and why should it expect to be okay to pass `--` to any program?

EvgenijM86:
lxc-net is a script that is used inside LXC virtualization to initialize network. And the problem looks like this:

```
    dnsmasq $LXC_DHCP_CONFILE_ARG $LXC_DOMAIN_ARG $LXC_DHCP_PING_ARG -u ${DNSMASQ_USER} \
            --strict-order --bind-interfaces --pid-file="${varrun}"/dnsmasq.pid \
            --listen-address ${LXC_ADDR} --dhcp-range ${LXC_DHCP_RANGE} \
            --dhcp-lease-max=${LXC_DHCP_MAX} --dhcp-no-override \
            --except-interface=lo --interface=${LXC_BRIDGE} \
            --dhcp-leasefile="${varlib}"/misc/dnsmasq.${LXC_BRIDGE}.leases \
            --dhcp-authoritative $LXC_IPV6_ARG || cleanup
```

If I replace dnsmasq with pihole-FTL it does not work (complains about invalid options). And on Arch Linux dnsmasq is uninstalled because pihole-FTL is supposed to fully replace all of the functionality of dnsmasq. So, if you install pihole-ftl you loose the ability to use software that relies on virtualization, like gitian-builder.

EvgenijM86:
If I understand correctly any script that used dnsmasq with advanced options will not work with pihole-FTL, if you just tell it to use pihole-FTL binary instead of dnsmasq binary.

DL6ER:
> If I replace dnsmasq with pihole-FTL (in that script) it does not work (complains about invalid options). 

Yes, this is correct. Those options are not `pihole-FTL` options.

You can specify `dnsmasq` options after `--`. For example
```
dnsmasq --strict-order --bind-interfaces --pid-file="${varrun}"/dnsmasq.pid
```
could be
```
pihole-FTL -- --strict-order --bind-interfaces --pid-file="${varrun}"/dnsmasq.pid
```

EvgenijM86:
Thanks, it looks that is what I was looking for. You should probably add that to --help description. In my version (v4.3.1) there is no mention of that.

PR #174, opened 2017-12-21
Handle dnsmasq "extra" logging

DL6ER:

## 10

---

Complete re-write of `FTL` log file parser to use `dnsmasq`'s UUIDs present in the log if option `log-queries=extra` is set. This will ensure 100% correctness of the interpreted log files.

One particular difficulty with `dnsmasq`'s `extra` logging is that it resets the UUID counter to 0 when being restarted. This poisons the whole idea of having UUIDs. A proper failsafe condition (using the idea of UUID generations) has been implemented in the new parser.

Note that this is a **breaking change** as the "old" log format cannot be read anymore and as such will bump `FTL`'s version to `v3.0`.

### DIFF
--- a/database.c
+++ b/database.c
@@ -312,12 +312,9 @@ void save_to_DB(void)
 	for(i = lastdbindex; i < counters.queries; i++)
 	{
 		validate_access("queries", i, true, __LINE__, __FUNCTION__, __FILE__);
-		if(queries[i].timestamp <= lasttimestamp || queries[i].db == true)
-		{
-			// Already in database
-			// logg("Skipping %i",i);
+		if(queries[i].timestamp <= lasttimestamp || queries[i].db || !queries[i].complete)
+			// Already in database or not yet complete
 			continue;
-		}
 
 		// Memory checks
 		validate_access("queries", i, true, __LINE__, __FUNCTION__, __FILE__);
--- a/database.c
+++ b/database.c
@@ -309,12 +309,20 @@ void save_to_DB(void)
 		return;
 	}
 
+	int currenttimestamp = time(NULL);
 	for(i = lastdbindex; i < counters.queries; i++)
 	{
 		validate_access("queries", i, true, __LINE__, __FUNCTION__, __FILE__);
-		if(queries[i].timestamp <= lasttimestamp || queries[i].db || !queries[i].complete)
+		if(queries[i].timestamp <= lasttimestamp || queries[i].db)
 			// Already in database or not yet complete
 			continue;
+
+		if(!queries[i].complete && queries[i].timestamp > currenttimestamp-2)
+		{
+			// Break if a brand new query (age < 2 seconds) is not yet completed
+			// giving it a chance to be stored next time
+			break;
+		}
 
 		// Memory checks
 		validate_access("queries", i, true, __LINE__, __FUNCTION__, __FILE__);
--- a/parser.c
+++ b/parser.c
@@ -823,6 +823,60 @@ void process_pihole_log(int file)
 			}
 		}
 
+		// is this a "black.list" line?
+		else if(strstr(readbuffer,"/black.list ") != NULL && strstr(readbuffer," is ") != NULL)
+		{
+			// Check if this domain names contains only printable characters
+			// if not: skip analysis of this log line
+			if(strstr(readbuffer,"<name unprintable>") != NULL)
+			{
+				if(debug) logg("Ignoring <name unprintable> domain (cached)");
+				continue;
+			}
+
+			// Check if this is a PTR query
+			// if so: skip analysis of this log line
+			if(strstr(readbuffer,"in-addr.arpa") != NULL)
+				continue;
+
+			// Get dnsmasq's ID for this transaction
+			int dnsmasqID = getID(readbuffer);
+
+			// Save forwardID in corresponding query indentified by dnsmasq's ID
+			bool found = false;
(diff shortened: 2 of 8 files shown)


DL6ER:
Still working fine for me with no queries that cannot be determined (see `Unknown DNS queries`).
```
[2017-12-27 13:14:45.999]  -> Total DNS queries: 6689
[2017-12-27 13:14:45.999]  -> Cached DNS queries: 2348
[2017-12-27 13:14:45.999]  -> Forwarded DNS queries: 4
[2017-12-27 13:14:45.999]  -> Exactly blocked DNS queries: 61
[2017-12-27 13:14:45.999]  -> Wildcard blocked DNS queries: 2
[2017-12-27 13:14:45.999]  -> Unknown DNS queries: 0
[2017-12-27 13:14:45.999]  -> Unique domains: 880
[2017-12-27 13:14:45.999]  -> Unique clients: 6
```
100% accuracy is verified via visual inspection in a few very busy regions.

This PR can be merged at any point to `development` as the release target `v2.13.1` is already branched out. As soon as this PR is merged, I can open additional PRs, e.g. actually utilizing the new data we collect from the `reply` lines as well as advanced DNSSEC support (which is so far completely ignored by `FTL`) in smaller, much easier to review pull requests.

database.c
```diff
@@ -312,12 +312,9 @@ void save_to_DB(void)
 	for(i = lastdbindex; i < counters.queries; i++)
 	{
 		validate_access("queries", i, true, __LINE__, __FUNCTION__, __FILE__);
-		if(queries[i].timestamp <= lasttimestamp || queries[i].db == true)
-		{
-			// Already in database
-			// logg("Skipping %i",i);
+		if(queries[i].timestamp <= lasttimestamp || queries[i].db || !queries[i].complete)
+			// Already in database or not yet complete
 			continue;
```
Mcat12: If we skip a query while it's still not yet complete and the one after is complete, we would lose the first query. Perhaps change this to a `break` so that we never add any queries after one which is not complete.

request.c
```diff
@@ -1184,3 +1190,45 @@ void getClientNames(int *sock)
...
 		clearSetupVarsArray();
 }
+
+void getUnknownQueries(int *sock)
+{
+	char server_message[SOCKETBUFFERLEN];
+
+	if(!debug)
+	{
+		sprintf(server_message,"For safety reasons, the command is only available in debugging mode!\n");
```
Mcat12: Why is this?

parser.c
```diff
@@ -642,63 +575,101 @@ void process_pihole_log(int file)
...
+			if(!checkQuery(readbuffer, "gravity.list"))
+				continue;
+
+			// Get dnsmasq's ID for this transaction
+			int dnsmasqID = getID(readbuffer);
+			// Skip invalid lines
+			if(dnsmasqID < 0)
+				continue;
+
+			// Save forwardID in corresponding query indentified by dnsmasq's ID
```
Mcat12: Is this comment supposed to reference `forwardID` or `status`?

parser.c
```diff
@@ -642,63 +575,101 @@ void process_pihole_log(int file)
...
+			{
+				if(debug) logg("Skipping malformated forwarded line");
 				continue;
+			}
 
-			// Get timestamp
-			int querytimestamp, overTimetimestamp, timeidx = -1, i;
-			extracttimestamp(readbuffer, &querytimestamp, &overTimetimestamp);
-
+			// Save forwardID in corresponding query indentified by dnsmasq's ID
```
Mcat12: I think this is where that comment came from (see above).

parser.c
```diff
@@ -719,6 +691,252 @@ void process_pihole_log(int file)
...
+				int timeidx = getTimeIndex(readbuffer);
+				validate_access("overTime", timeidx, true, __LINE__, __FUNCTION__, __FILE__);
+				overTime[timeidx].cached++;
+			}
+		}
+		// is this a "wildcard" line?
+		else if(strstr(readbuffer," config ") != NULL && strstr(readbuffer," is ") != NULL)
+		{
+			// Check query for invalid characters
+			if(!checkQuery(readbuffer, "gravity.list"))
```
Mcat12: Is this the correct type for wildcards?

database.c
DL6ER: Hmm, not sure if that is a good solution - although it never happend to me so far, this could be a potential dead-lock assuming that `dnsmasq` didn't write out the result for one query so it would never be completed. What do you think?

parser.c
DL6ER: True, will fix this

parser.c
DL6ER: I agree :-)

request.c
DL6ER: Initially this command was supposed to do something different, I will remove this statement here.

database.c
Mcat12: Maybe bring back the 200 line check or convert it to a timestamp check. If the query hasn't completed after 2 seconds, assume it's a goner and continue.

DL6ER: Okay, I will in addition compare the time stamp of the query with the current time and `break` if the query was from within the most recent two seconds. It will have a chance to get saved (or eventually ignored) on the next DB dumping time.

pralor:

https://discourse.pi-hole.net/t/cache-server-statistics/6092/1

pralor:

https://discourse.pi-hole.net/t/tracking-request-status-unambiguously/6091/1

PR #178, opened 2017-12-23
Wildcard auditing

DL6ER:

## 10

---

Accept lines like `*.spot.com` as wildcard audits for domains.

One note about this:
- `*spot.com` will audit `spot.com`, `some.spot.com`, etc. but also `spotspot.com` or `totalldifferentspot.com`
- `*.spot.com` will audit `some.spot.com`, etc. but neither `spot.com` nor `spotspot.com` nor `totalldifferentspot.com`

### DIFF
--- a/grep.c
+++ b/grep.c
@@ -189,8 +189,23 @@ int countlineswith(const char* str, const char* fname)
 	// Search through file
 	// getline reads a string from the specified file up to either a newline character or EOF
 	while(getline(&buffer, &size, fp) != -1)
-		if((strstr(buffer, str)) != NULL)
+	{
+		// Strip potential newline character at the end of line we just read
+		if(buffer[strlen(buffer)-1] == '\n')
+			buffer[strlen(buffer)-1] = '\0';
+
+		printf("Searching \"%s\" in \"%s\"\n",buffer,str);
+		// Search for exact match
+		if(strcmp(buffer, str) == 0)
 			found++;
+
+		// If line starts with *, search for partial match
+		if(buffer[0] == '*')
+		{
+			if(strstr(str, buffer+1) != NULL)
+				found++;
+		}
+	}
 
 	// Free allocated memory
 	if(buffer != NULL)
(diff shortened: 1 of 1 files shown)


grep.c
```diff
@@ -189,9 +189,22 @@ int countlineswith(const char* str, const char* fname)
 	// Search through file
 	// getline reads a string from the specified file up to either a newline character or EOF
 	while(getline(&buffer, &size, fp) != -1)
-		if((strstr(buffer, str)) != NULL)
+	{
+		// Strip potential newline character at the end of line we just read
+		if(buffer[strlen(buffer)-1] == '\n')
+			buffer[strlen(buffer)-1] = '\0';
+
+		printf("Searching \"%s\" in \"%s\"\n",buffer,str);
```
Mcat12: Was this for debugging? Also, the two variables might be switched.

grep.c
```diff
@@ -189,9 +189,22 @@ int countlineswith(const char* str, const char* fname)
...
+		// Search for exact match
+		if(strcmp(buffer, str) == 0)
 			found++;
```
Mcat12: If there was an exact match, we can probably `continue` here and skip the next `if`.

grep.c
```diff
@@ -189,9 +189,22 @@ int countlineswith(const char* str, const char* fname)
...
 
+		// If line starts with *, search for partial match of
+		// needle "buffer+1" in haystack "str"
+		if(buffer[0] == '*' && strstr(str, buffer+1) != NULL)
```
Mcat12: This will match more than just `xxxxxxexample.com` given `*example.com`. `example.com.xxxxx` will also match.

grep.c
DL6ER: It is `char *strstr(const char *haystack, const char *needle);`, so the variable ordering are correct.

tuurie72:
After running the checkout to FTL (thanks DL6ER), I can confirm this is working as designed.
Love this feature and it saves a huge amount of time if you have a Sonos, Playstation, Spotify connection at your home!

PR #180, opened 2017-12-26
Add dual-stack socket support for FTL

DL6ER:

## 10

---

This pull request adds full dual-stack support for `FTL`'s `telnet`-like socket. Previously, we bound only to an IPv4 port, now we bind to both the IPv4 and IPv6 port spaces.

In case no IPv6 interface is available (e.g. in `docker`), `FTL` will bind only to the IPv4 port - just as it did before.

### DIFF
--- a/socket.c
+++ b/socket.c
@@ -151,7 +151,7 @@ void bind_to_telnet_port(char type, int *socketdescriptor)
 		exit(EXIT_FAILURE);
 	}
 
-	logg("Listening on port %i for incoming %s connections", port, dualstack == true ? "IPv4 + IPv6" : "IPv4");
+	logg("Listening on port %i for incoming %s connections", port, dualstack ? "IPv4 + IPv6" : "IPv4");
 }
 
 
--- a/socket.c
+++ b/socket.c
@@ -86,7 +86,8 @@ void bind_to_port(char type, int *socketdescriptor)
 			break;
 	}
 
-	bool bound = false;
+	bool bound = false, dualstack = false;
+	// Try dual-stack socket
 	for(port = port_init; port <= (port_init + 20); port++)
 	{
 		serv_addr.sin6_port = htons(port);
@@ -97,13 +98,45 @@ void bind_to_port(char type, int *socketdescriptor)
 		else
 		{
 			bound = true;
+			dualstack = true;
 			break;
+		}
+	}
+
+	// Try IPv4 only socket, much of the code seen above is duplicated here for IPv4,
+	// see the comments further up for details
+	if(!bound)
+	{
+		logg("Error listening on any IPv4 + IPv6 port, trying IPv4-only binding");
+		*socketdescriptor = socket(AF_INET, SOCK_STREAM, 0);
+
+		struct sockaddr_in serv_addr4;
+		memset(&serv_addr4, 0, sizeof(serv_addr4));
+		serv_addr4.sin_family = AF_INET;
+
+		if(config.socket_listenlocal && type == SOCKET)
+			serv_addr4.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+		else
+			serv_addr4.sin_addr.s_addr = INADDR_ANY;
+
+		for(port = port_init; port <= (port_init + 20); port++)
+		{
+			serv_addr4.sin_port = htons(port);
+			if(bind(*socketdescriptor, (struct sockaddr *) &serv_addr4, sizeof(serv_addr4)) < 0)
+			{
+				logg("Error on binding on IPv4 port %i", port);
+			}
+			else
+			{
+				bound = true;
+				break;
+			}
 		}
 	}
 
 	if(!bound)
 	{
-		logg("Error listening on any port");
+		logg("Error listening on any IPv4 port");
 		exit(EXIT_FAILURE);
 	}
 
@@ -117,7 +150,7 @@ void bind_to_port(char type, int *socketdescriptor)
 		exit(EXIT_FAILURE);
 	}
 
-	logg("Listening on port %i for incoming connections", port);
+	logg("Listening on port %i for incoming %s connections", port, dualstack == true ? "IPv4 + IPv6" : "IPv4");
 }
 
 // Called from main() at graceful shutdown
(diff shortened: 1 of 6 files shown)


DL6ER:
Before this pull request:
```
$ echo ">quit" | nc -v 127.0.0.1 4711
Connection to 127.0.0.1 4711 port [tcp/*] succeeded!

$ echo ">quit" | nc -v ::1 4711
nc: connect to ::1 port 4711 (tcp) failed: Connection refused
```
With the proposed changes:
```
$ echo ">quit" | nc -v 127.0.0.1 4711
Connection to 127.0.0.1 4711 port [tcp/*] succeeded!

$ echo ">quit" | nc -v ::1 4711
Connection to ::1 4711 port [tcp/*] succeeded!
```

DL6ER:
Resolved merge conflicts, what you you think? @Mcat12 @dschaper

socket.c
```diff
@@ -117,7 +151,7 @@ void bind_to_telnet_port(char type, int *socketdescriptor)
 		exit(EXIT_FAILURE);
 	}
 
-	logg("Listening on port %i for incoming connections", port);
+	logg("Listening on port %i for incoming %s connections", port, dualstack == true ? "IPv4 + IPv6" : "IPv4");
```
Mcat12: The `== true` can be removed since `dualstack` is a boolean already.

socket.c
```diff
@@ -87,23 +88,56 @@ void bind_to_telnet_port(char type, int *socketdescriptor)
...
 		{
 			logg("Error on binding on port %i", port);
 		}
 		else
 		{
 			bound = true;
+			dualstack = true;
 			break;
 		}
 	}
```
Mcat12: On a non-IPv6 system this would produce 20 lines of errors saying it can't bind to a port. Is there a quicker way to figure out if we have IPv6 access?

socket.c
DL6ER: I'm not sure how to detect if the system has a working `ipv6` interface. If we cannot find out how to do it, we could also change the error reporting (don't report on the Individual failures, but only once at the end when we couldn't find at least one port).

socket.c
DL6ER: I'm not sure how to detect if the system has a working `ipv6` interface. If we cannot find out how to do it, we could also change the error reporting (don't report on the Individual failures, but only once at the end when we couldn't find at least one port).

DL6ER:
Marking as `[WIP]`

While `in6addr_any` accepts IPv4 and IPv6 connections simultaneously, `in6addr_loopback` only accepts connections to `::1`. Have to further investigate on this.

socket.c
```diff
@@ -33,18 +34,20 @@ void saveport(int port)
...
+		fprintf(f, "%i", config.port);
 		fclose(f);
 	}
 }
 
-void bind_to_telnet_port(char type, int *socketdescriptor)
+char bind_to_telnet_port_IPv4(char type, int *socketdescriptor)
 {
+	// Try IPv4 only socket
+	// see the comments further up for details
```
Mcat12: This comment is no longer valid

socket.c
```diff
@@ -140,23 +170,22 @@ void bind_to_unix_socket(int *socketdescriptor)
 
 	// Bild to Unix socket handle
 	errno = 0;
-	if(bind(*socketdescriptor, (struct sockaddr *) &address, sizeof (address)) != 0) {
-		logg("Error on binding on unix socket %s", FTLfiles.socketfile);
-		logg("Reason: %s (%i)", strerror(errno), errno);
+	if(bind(*socketdescriptor, (struct sockaddr *) &address, sizeof (address)) != 0)
+	{
+		logg("Error on binding on unix socket %s: %s (%i)", FTLfiles.socketfile, strerror(errno), errno);
```
Mcat12: Unix is capitalized elsewhere.

DL6ER:
Completely re-wrote this feature.

In case you are interested , here is the result of my research:
An IPv4 client can connect to `INADDR_LOOPBACK` (a.k.a. `127.0.0.1`) on an dual-stack listening socket *if the listening socket is bound to `IN6ADDR_ANY` (a.k.a. `::`). This is very convenient and allows to write more simple code that can live in both, the IPv4 and the IPv6 world at the same time.

However, an essential security feature of `FTL` (that can be disabled trough the config file) is that `FTL` does, by default, only bind to the `lo` interface such that data is only available locally and not from anywhere else in the network.
However, when binding a dual-stack socket to `IN6ADDR_LOOPBACK` (a.k.a. `::1`), this will prevent IPv4 clients from being able to connect. Although he reason is simple it not very obvious: First, we have to understand that the socket lives in the IPv6 world, i.e. when an IPv4 client connects, it will be seen by the socket as connecting from the corresponding IPv4-mapped IPv6 address. Hence, when connecting  to/from `127.0.0.1` the dual-stack socket will see the connection going to the IPv4-mapped IPv6 address `::FFFF:7F00:0001` (the equivalent of `127.0.0.1`). However, as the dual-stack socket is explicitly bound to `::1` (and not `::`) the IPv4 client cannot connect.

socket.c
```diff
@@ -335,16 +383,69 @@ void *telnet_listening_thread(void *args)
 	pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
 
 	// Set thread name
-	prctl(PR_SET_NAME,"telnet listener",0,0,0);
+	prctl(PR_SET_NAME,"telnet-v4",0,0,0);
```
Mcat12: Should use `IPv4` to distinguish this from appearing to be `telnet version 4 API`.

socket.c
```diff
@@ -335,16 +383,69 @@ void *telnet_listening_thread(void *args)
...
+	// We will use the attributes object later to start all threads in detached mode
+	pthread_attr_t attr;
+	// Initialize thread attributes object with default attribute values
+	pthread_attr_init(&attr);
+	// When a detached thread terminates, its resources are automatically released back to
+	// the system without the need for another thread to join with the terminated thread
+	pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
+
+	// Set thread name
+	prctl(PR_SET_NAME,"telnet-v6",0,0,0);
```
Mcat12: See above.

PR #187, opened 2017-12-31
Add unix socket

DL6ER:

## 10

---

Add Unix socket to `FTL`. It can be used with the same language as the already known Telnet interface.

Add also a small test program to demonstrate how to connect to `FTL` using the Unix socket (`socket_client.c`)

We need to have a directory `/var/run/pihole` that is owned user `pihole`. Run
```
sudo mkdir /var/run/pihole
sudo chown pihole:pihole /var/run/pihole
```
to prepare this.

### DIFF
--- a/routines.h
+++ b/routines.h
@@ -42,6 +42,7 @@ void close_telnet_socket(char type);
 void seom(char server_message[], int sock);
 void swrite(char server_message[], int sock);
 void *telnet_listenting_thread(void *args);
+void *socket_listenting_thread(void *args);
 
 void process_request(char *client_message, int *sock);
 bool command(char *client_message, const char* cmd);
--- a/routines.h
+++ b/routines.h
@@ -38,10 +38,10 @@ void pihole_log_flushed(bool message);
 
 void memory_check(int which);
 
-void close_socket(char type);
+void close_telnet_socket(char type);
 void seom(char server_message[], int sock);
 void swrite(char server_message[], int sock);
-void *socket_listenting_thread(void *args);
+void *telnet_listenting_thread(void *args);
 
 void process_request(char *client_message, int *sock);
 bool command(char *client_message, const char* cmd);
--- a/socket.c
+++ b/socket.c
@@ -21,7 +21,7 @@
 #define BACKLOG 5
 
 // File descriptors
-int telnetfd;
+int telnetfd, socketfd;
 
 void saveport(int port)
 {
@@ -44,7 +44,7 @@ void bind_to_telnet_port(char type, int *socketdescriptor)
 
 	if(*socketdescriptor < 0)
 	{
-		logg("Error opening socket");
+		logg("Error opening telnet socket");
 		exit(EXIT_FAILURE);
 	}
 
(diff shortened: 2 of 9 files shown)


socket_client.c
```diff
@@ -0,0 +1,77 @@
...
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#define BUF 1024
+#define UDS_FILE "/var/run/pihole/FTL.sock"
```
dschaper: `/var/run` is often a symlink to `/run` which in turn is very often `tmpfs` and volatile. So as long as this survives a reboot then we're okay. Otherwise it will have to be created when the FTL daemon starts. It probably already does that, I'm just reading the `c` code through and leaving this note.

socket_client.c
```diff
@@ -0,0 +1,77 @@
...
+	struct sockaddr_un address;
+	int size, ret;
+
+	socketfd = socket(PF_LOCAL, SOCK_STREAM, 0);
+	if(socketfd <= 0)
+	{
+		printf("Error creating socket!\n");
+		exit(EXIT_FAILURE);
+	}
+	printf ("Socket created\n");
```
dschaper: And yes, it's created (if possible) in `main()`.

socket_client.c
DL6ER: Thanks, I wanted to add this to the service script (and maybe also move the log into a separate folder at the same time if that is okay with you @dschaper )

routines.h
```diff
@@ -38,9 +38,11 @@ void pihole_log_flushed(bool message);
 
 void memory_check(int which);
 
-void close_socket(char type);
+void close_telnet_socket(void);
+void close_unix_socket(void);
 void seom(char server_message[], int sock);
 void swrite(char server_message[], int sock);
+void *telnet_listenting_thread(void *args);
 void *socket_listenting_thread(void *args);
```
Mcat12: Misspelling of listening

socket.c
```diff
@@ -158,22 +195,79 @@ int listener(int sockfd)
...
+
+			process_request(message, &sock);
+			free(message);
+
+			// Release thread lock
+			disable_thread_lock(threadname);
+
+			if(sock == 0)
+			{
+				// Client disconnected by seding EOT or ">quit"
```
Mcat12: Misspelling of `sending`

socket.c
```diff
@@ -210,18 +304,17 @@ void *socket_connection_handler_thread(void *socket_desc)
 
 			if(sock == 0)
 			{
-				// Client disconnected by seding EOT or ">quit"
+				// Socket connection interrupted by seding EOT or ">quit"
```
Mcat12: `sending`

socket.c
```diff
@@ -144,7 +181,7 @@ void swrite(char server_message[SOCKETBUFFERLEN], int sock)
 		logg("WARNING: Socket write returned error code %i", errno);
 }
 
-int listener(int sockfd)
+int telnet_listener(int sockfd)
```
Mcat12: Is this method specific to telnet?

socket_client.c
```diff
@@ -0,0 +1,87 @@
...
+#include <sys/types.h>
+/* Pi-hole: A black hole for Internet advertisements
+*  (c) 2017 Pi-hole, LLC (https://pi-hole.net)
+*  Network-wide ad blocking via your own hardware.
+*
+*  FTL Engine
+*  Unix socket connection test program
+*
+*  This file is copyright under the latest version of the EUPL.
+*  Please see LICENSE file for your rights under this license. */
```
Mcat12: The license header should be at the top.

socket_client.c
```diff
@@ -0,0 +1,87 @@
...
+		// Print received data to stdout
+		printf("%s", buffer);
+
+		// Receive potential new message
+		size = recv(socketfd, buffer, BUF-1, 0);
+
+		// Zero-terminate incoming message
+		if(size > 0)
+			buffer[size] = '\0';
+	}
```
Mcat12: This code is duplicated above and can be converted into either a do-while loop or a simple while loop.

socket_client.c
```diff
@@ -0,0 +1,87 @@
...
+
+	// Print received data to stdout
+	printf("%s", buffer);
+
+	// Close Unix socket connection
+	close(socketfd);
+	return EXIT_SUCCESS;
+	}
```
Mcat12: Ending brace has the wrong indentation.

socket_client.c
DL6ER: Fixed by latest commit

PR #191, opened 2018-01-06
Pi-hole FTL v2.13.2

DL6ER:
Release


DL6ER:
Note: CircleCI is expected to fail as we have no build configuration for it in `master` yet.

PR #192, opened 2018-01-07
Ensure that host names are always lower case

DL6ER:

## 10

---

Ensure that host name returned by `char *resolveHostname(const char *addr)` is always lower case. This is an issue as we currently convert the host name to lower case on the initial reading of the log (when a client is newly added) but we didn't do it again later when we ran `void reresolveHostnames(void)` which is periodically called to recognize updated host names.

### DIFF
--- a/parser.c
+++ b/parser.c
@@ -504,8 +504,6 @@ void process_pihole_log(int file)
 				// Debug output
 				if(strlen(hostname) > 0)
 				{
-					// Convert hostname to lower case
-					strtolower(hostname);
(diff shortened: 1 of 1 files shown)


parser.c
```diff
@@ -504,8 +504,6 @@ void process_pihole_log(int file)
 				// Debug output
 				if(strlen(hostname) > 0)
 				{
-					// Convert hostname to lower case
```
DL6ER: We don't need to do it here, as it is now guaranteed that `char *resolveHostname(const char *addr)` returns a lower case host name.

PR #208, opened 2018-01-20
Import data from long-term database

DL6ER:

## 10

---

This pull requests implements some general improvements for `FTL`.

Changes:
- Instead of completely parsing the log files `pihole.log` and `pihole.log.1` each time `FTL` starts up, we read the already analyzed and categorized data from the long term database. We only read those queries from `pihole.log` that are not already present in the database. This enhances the startup speed of `FTL` as well as provides some functionality we will be using with `FTL` in the future.
- Instead of counting all forward activity, we only count the first server a request is forwarded to. This will change the appearance off the `Forward Destinations (integrated)` plot and is implemented to acknowledge the fact that we store only one forward destination per query in the database.

Minor fixes:
- In case we cached a `<CNAME>` record but still have to ask the upstream servers for the actual IP addresses, we don't count this query as being cached but as being forwarded. Note that this will reduce the `local` forward destination count (it has been too high so far).
- Remove unnecessarily duplicated `validate_access()` calls at some places for a slight performance increase.

### DIFF
--- a/database.c
+++ b/database.c
@@ -14,6 +14,7 @@ sqlite3 *db;
 bool database = false;
 bool DBdeleteoldqueries = false;
 long int lastdbindex = 0;
+long int lastDBimportedtimestamp = 0;
 
 pthread_mutex_t dblock;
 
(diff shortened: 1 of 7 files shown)


database.c
```diff
@@ -480,3 +481,164 @@ void *DB_thread(void *val)
...
+		queries[queryID].timestamp = queryTimeStamp;
+		queries[queryID].type = type;
+		queries[queryID].status = status;
+		queries[queryID].domainID = domainID;
+		queries[queryID].clientID = clientID;
+		queries[queryID].forwardID = forwardID;
+		queries[queryID].timeidx = timeidx;
+		queries[queryID].valid = true; // Mark this as a valid query (false = it has been garbage collected and should be skipped)
+		queries[queryID].db = true; // Mark this as already present in the database
+		queries[queryID].id = 0; // This is dnsmasq's internal ID. We don't store it in the database
```
Mcat12: Just to make sure, the dnsmasq query IDs don't start at 0, right?

parser.c
```diff
@@ -587,27 +578,21 @@ void process_pihole_log(int file)
 				continue;
```
Mcat12: Missing `free(forward)`

parser.c
```diff
@@ -587,27 +578,21 @@ void process_pihole_log(int file)
...
 			}
 
+			// Count only if current query has not been counted so far
+			if(queries[i].complete)
+				continue;
```
Mcat12: Missing `free(forward)`

parser.c
```diff
@@ -275,6 +274,13 @@ void process_pihole_log(int file)
 				break;
 		}
 
+		// Get timestamp
+		int querytimestamp, overTimetimestamp;
+		extracttimestamp(readbuffer, &querytimestamp, &overTimetimestamp);
+
+		// Check if this query has already been imported from the database
+		if(querytimestamp < lastDBimportedtimestamp) continue;
```
Mcat12: There are more of this line below, but since this is run on every line, the others would just be redundant?

database.c
DL6ER: Yes, it starts with `1`.

parser.c
DL6ER: Oh yes, they shouldn't be there, seems like I forgot to remove them at some places.

Issue #71, opened 2017-06-10
Please provide a PPC32 Build


mterron:
**In raising this issue, I confirm the following (please check boxes, eg [X]) Failure to fill the template will close your issue:**

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*
- [x] The issue I am reporting isn't a duplicate 

**How familiar are you with the codebase?:**

_1_

---

**[BUG | ISSUE] Expected Behaviour:**
FTL runs

**[BUG | ISSUE] Actual Behaviour:**
No binary available for CPU arch ppc as per lscpu output
Architecture:          ppc
CPU op-mode(s):        32-bit
Byte Order:            Big Endian
CPU(s):                1
On-line CPU(s) list:   0
Thread(s) per core:    1
Core(s) per socket:    1
Socket(s):             1
Model:                 amcc,apollo3g
BogoMIPS:              1600.00
L1d cache:             32K
L1i cache:             32K
L2d cache:             unknown size
L2i cache:             unknown size

**[BUG | ISSUE] Steps to reproduce:**
Install pi-hole on a PPC based device.

**Device specifics**

Hardware Type: MyBook Live
OS: Linux

technicalpyro:
looking at what i can find regarding your hardware i am unsure it would be supported.  however you can start here: https://github.com/pi-hole/FTL#how-to-compile-ftl-from-source and once DL6ER is awake (he wrote the FTL backend) he may have a better answer

dschaper:
We'll check and see if our builder is capable of producing binaries for that platform. We use Travis to build out our binaries and it may not be possible to get something on that processor, but we will try.

DL6ER:
I haven't had much luck trying to get a ready-to-use cross compiler. Have you tried to compile FTL from source as suggested by @technicalpyro ?

mterron:
I don't think there is a compiler on the wd my book.

mterron:
I was able to cross-compile by doing this (on a fresh Debian Jessie install):
```shell
apt update && apt install curl git
echo 'deb http://emdebian.org/tools/debian/ jessie main' >> /etc/apt/sources.list
curl http://emdebian.org/tools/debian/emdebian-toolchain-archive.key | sudo apt-key add -
dpkg --add-architecture powerpc
apt update
apt install crossbuild-essential-powerpc
[...]

PR #238, opened 2018-02-23
Introducing FTLDNS

DL6ER:

## 10

---

This pull request introduces *FTL*DNS. See [here](https://pi-hole.net/2018/02/22/coming-soon-ftldns-pi-holes-own-dns-dhcp-server/) for more general details and our motivation.

A short (rather technical) change log follows:

- DNS/DHCP server based on `dnsmasq` 2.78
- Show `pi-hole-2.78` as DNS resolver version string
- Add `pihole-FTL --test` to mimic `dnsmasq --test` for testing the configuration files for validity
- Read in all historic details from the database, update database periodically (as before). This makes the log file altogether optional. We may disable it by default in the near future.
- Number of domains on blocking lists: Instead of counting the lines in the files tehmselves, we now use the number of domains `dnsmasq` imported successfully. Note that this count might be less if there are invalid (or empty) lines in the domain lists. In the case of disabled blocking, this means that no domains are on the blocking list.
- Pass all command line options after `--` to the resolver, e.g. `pihole-FTL -a -b -- -c -d` will have `-a -b` handles by FTL and `-c -d` handled by `dnsmasq`.
- Some more general improvements of the code, e.g. don't allocate space for each wildcard entry individually but do it in steps of 100 and skip computationally expensive loops if we can already foresee that an entry will not be found
- Redesign of the database thread: instead of periodically calling it, it says and gets active when needed
- Redesign of the internal memory management: Instead of flushing everything every night (when the log rotation happens), we keep everything and don't periodically reinterpret. The result is an entire rewrite of many data structure routines. This rewrite alone would deserve a full change log, but let's make this brief: We use low-level memory access functions (like `memmove()`) to use allocated memory maximum efficiently. If queries get removed (because of being older than 24 hours), the entire data structure is rearranged in memory to prevent any memory voids from appearing.
- We entirely remove the log parsing functionality as we don't need it any more. This also reduces the memory footprint of the `queries` structure, as many attributes such as `loggeneration` and `valid` are not needed any longer.
- We gather much more detail about individual queries (like their reply type and TTL) and provide some of them for displaying. This will surely be expanded in the future.
screenshot at 2018-02-23
- Add concept of privacy levels, this makes `QUERY_DISPLAY` obsolete. We apply the privacy filters immediately on the raw DNS data such that leaks caused by e.g. a subsequent reduction of the privacy level are not possible. Anonymized domains are replaced by `hidden`, clients by `0.0.0.0`. There are four privacy level at the moment (show all, hide domains, hide domains + clients, hide everything). Depending on the configured privacy level, some features (like Top Lists) may not be available as they get disabled automatically. The privacy level may be changed at any time without needing to restart `pihole-FTL`.
- Reduce verbosity of logging into FTL's log file (no printing of new client or domains)
- Support more query types (A, AAAA, new: ANY, SRV, SOA, PTR, TXT). We compute statistics for them but don't show them on the Query Log page.
- Increase `MAXCLIENTS` because DNS/DHCP connections are now counted therein as well

Changes to `dnsmasq`'s code base:
- Integration of hooks (new queries, received replies, answered from cache, cache cleared, ...)
- Ensure that `dnsmasq` does not close our sockets on start up
- Remove hard-coded cache size of 10,000 (allows users to specify any cache size in their config files)
- Apply patch sent to the dnsmasq mailing list that fixes skipping of file names ending in `~` (in hostsdir and DHCP folders)

Other things:
- We intentionally don't use any IDN libraries due to two reasons: The IDN libraries are use (almost) exclusively when importing the blocking lists at the start of the resolver. This results in a speed up of the initialization / updating of the lists of up to 10x on low-power devices like Raspberry Pis.
- Add extra warning flags we apply only to the FTL part of the code to enforce better (as in warning-free) code

We will compile and provide binaries for the following architectures:
- x86_32
- x86_64
- arm
- armhf
- aarch64


pralor:

https://discourse.pi-hole.net/t/ftldns-pi-holes-own-dns-dhcp-server/7442/1

mterron:
Can you add powerpc to the binaries please? I added support to [pi-hole](https://github.com/pi-hole/pi-hole/pull/1539) a few versions ago and also validated [FTL](https://github.com/pi-hole/FTL/issues/71).

Thanks

Mcat12:
We are unable to support powerpc for a few reasons:
- The upcoming API has core dependencies which do not support it.
- We are unable to test it as we do not have any powerpc hardware.

If in the future we can support it, then we will look into that, but when we checked how many users used the powerpc binaries of FTL version 3, there were only a handful.

mterron:
@Mcat12 can you elaborate on these core dependencies?

Mcat12:
A dependency of the HTTP server it uses has a core crypto dependency which is where the lack of powerpc support lies. Once we are able to make the API source public, you'll be able to see for yourself.

JrCs:
Can you add support for mips so it can be used on embedded router like https://www.ubnt.com/edgemax/edgerouter/ that already use dnsmasq 2.78 ?

PR #260, opened 2018-04-24
Add regex blocking support to FTLDNS

DL6ER:

## 10

---

Add (extended) regular expressions (ERE) support to *FTL*DNS allowing you to define arbitrarily complex filters for domain blocking. All domains matched by this regex filter will be blocked by returning an `NXDOMAIN` response.

One simple example:
```regex
BLOCKINGREGEX=^ab.+\.com$
```
will block all domains that start with "ab" (`^ab`), have at least one further character (`.+`) and end in ".com" (`\.com$`).

Examples for what would be blocked by this rule:
- `abc.com`
- `abtest.com`
- `ab.test.com`
- `abr-------.whatever.com`

Examples for what would **not** be blocked by this rule:
- `testab.com` (the domain doesn't start with "ab")
- `tab.test.com` (the domain doesn't start with "ab")
- `ab.com` (there is no character in between "ab" and ".com")
- `test.com.something` (the domain doesn't end in ".com")


dnsmasq_interface.c
```diff
@@ -909,3 +933,27 @@ int FTL_listsfile(char* filename, unsigned int index, FILE *f, int cache_size, s
...
+{
+	struct all_addr addr4;
+	if(inet_pton(AF_INET, "127.0.0.1", &addr4) <= 0)
+	{
+		logg("inet_pton failed in block_single_domain(%s)!",domain);
+		return;
+	}
+
+	struct crec *cache4;
+	if((cache4 = malloc(sizeof(struct crec) + strlen(domain)+1-SMALLDNAME)))
```
Mcat12: I had to check the definition of `struct crec`, but this does check out even though it seems wrong :wink:.

DL6ER: `crec` has `char *namep` at its end. Hence, when we allocate enough memory for `sizeof(struct crec)` we would have only allocated space for one (!) char in this pointer. Hence, we allocate a bit more to have enough to put the domain name into. The `strlen(domain)+1-SMALLDNAME` magic is a bit more tricky, but that due to that we have a fixed char of length `SMALLDNAME` in `union name` as well.

PR #261, opened 2018-04-25
Merge simple domain format into FTLDNS beta branch

DL6ER:
Internal


pralor:

https://discourse.pi-hole.net/t/problems-after-checking-out-regex-branch/9116/17

PR #266, opened 2018-04-29
Use "blocklist" instead of "blocked" in the forward destions graphic

DL6ER:
Internal


DL6ER:
screenshot at 2018-04-29

PR #277, opened 2018-05-11
Suppress wrong warning

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

Return early if no wildcard domains are defined to avoid warning to be printed although for-loop will anyhow be skipped.

This fixes a bug reported on [Discourse](https://discourse.pi-hole.net/t/fltdns-fatal-errors/9470/).

While I was working on this function, I also replaced the fixed values for the query status by their `enum` equivalents (`QUERY_CACHE == 3`, `QUERY_WILDCARD == 4`) and removed some old commented out code.

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

PR #276, opened 2018-05-10
Remove undocumented, old, and unused debug commands

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

Remove undocumented, old, and unused debug commands. (unused = I haven't used them for > 1 year, they have been helpful in the very early days of FTL < v1.0).

Also remove reading of `API_PRIVACY_MODE` as this was already ignored before. This option is superseded by the new privacy levels.

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

PR #275, opened 2018-05-10
Add blocked per client counter

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:**

## 10

---

This adds a new counter for blocked queries *per client*. This duplicates the functionality we already have for domains (either total or blocked count).

The new info can be queried from the API by specifying the keyword `blocked` in addition to `>top-clients`. This option is in agreement with other available options, e.g. `>top-clients withzero blocked (29)` will work as expected.

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

PR #273, opened 2018-05-05
Clean README of this repo

DL6ER:
Internal

The README got really too long. All removed part are re-added (in extended versions) in the new documentation of *FTL*DNS I'm currently writing.

PR #279, opened 2018-05-11
Update FTLDNS beta testing branch

DL6ER:
This PR updated the beta testing branch. There are no new changes in this PR. It will add the following PRs to the testing:

- Silence wrong FATAL ERROR messages #270
- Allow starting *FTL*DNS as any user #271
- Analyze incoming TCP queries #272
- Clean README of this repo #273 (no functional change for *FTL*DNS)
- Add blocked per client counter #275 

I would like to also get

- Remove undocumented, old, and unused debug commands #276
- Suppress wrong warning #277

in. They should be easy to review.


PromoFaux:
#277 reviewed, though I will leave #276 to @Mcat12 as that is a bit more involved than my current understanding of `FTL` codebase

DL6ER:
We should also merge in https://github.com/pi-hole/FTL/pull/281 to prevent locking of the init.d script

PR #282, opened 2018-05-13
Improve BLOCKINGMODE config setting

DL6ER:

## 10

---

This PR adds two improvements for `BLOCKINGMODE`:

- Re-parse `BLOCKINGMODE` on receipt of `SIGHUP`. This allows changing the parameter without having to completely restart *FTL*DNS.

- Add new `BLOCKINGMODE=NULL` which is similar to `BLOCKINGMODE=IP` but will respond with `0.0.0.0` and `::` for A and AAAA queries, respectively


dnsmasq_interface.c
```diff
@@ -881,14 +909,6 @@ int FTL_listsfile(char* filename, unsigned int index, FILE *f, int cache_size, s
...
 	clearSetupVarsArray(); // will free/invalidate IPv6addr
 
-	// If no IPv4 address was found but user wants us to server NXDOMAIN
-	// we have to mock an IP record (which won't do anything in the end)
-	if(!has_IPv4 && config.blockingmode == MODE_NX)
-	{
-		if(inet_pton(AF_INET, "127.0.0.1", &addr4) > 0)
-			has_IPv4 = true;
-	}
-
```
Mcat12: Why is this section deleted?

config.c
```diff
@@ -302,6 +304,44 @@ void get_privacy_level(FILE *fp)
...
+	// See if we got a file handle, if not we have to open
+	// the config file ourselves
+	bool opened = false;
+	if(fp == NULL)
+	{
+		if((fp = fopen(FTLfiles.conf, "r")) == NULL)
+			// Return silently if there is no config file available
+			return;
+		opened = true;
+	}
```
Mcat12: This seems messy. Why not just always open the file outside of the function and pass it in? Then it is not the function's responsibility to open the file (simpler function, easier to understand and debug).

config.c
```diff
@@ -188,22 +189,17 @@ void read_FTLconf(void)
...
 		case MODE_NX:
 			logg("   BLOCKINGMODE: NXDOMAIN for blocked domains");
 			break;
+		case MODE_NULL:
+			logg("   BLOCKINGMODE: Null IPs for blocked domains");
+			break;
 		default:
-			logg("   BLOCKINGMODE: Pi-hole's IP for blocked domains");
+			logg("   BLOCKINGMODE: Pi-hole's IPs for blocked domains");
 			break;
```
Mcat12: You should probably use a specific switch statement for `MODE_IP` and keep a default for errors.

dnsmasq_interface.c
DL6ER: This was meant for the case when there was no `IPV4_ADDRESS` in `setupVars.conf` as we need an IP for `NXDOMAIN` blocking. However, when now specifying `NXDOMAIN` or `NULL` as option for `BLOCKINGMODE` we will always use the hard-coded IP `0.0.0.0` and hence don't need a fallback solution any more.

config.c
DL6ER: What if opening of the file fails? We would still have to check for if the file was successfully opened in the function to set at least the default value if it is not. Otherwise, if the function would only be called on a successful opening of the config file, it could not change the setting. The purpose here was not to keep the function most simple, but rather to keep the calling most simple (we just need to call `get_blocking_mode(NULL);` and it will take care of getting everything set up correctly). Note that we use a similar technique already in `get_privacy_level(NULL);` immediately above.

config.c
DL6ER: There can be no error. The three possible cases are:
`BLOCKINGMODE=NXDOMAIN`, or `BLOCKINGMODE=NULL` or anything else (even missing option). In the latter case, it will always fall back to the legacy behavior (IP address of Pi-hole). There is no situation that we can consider as failed.

PR #288, opened 2018-05-19
Add support for regex filters from file

DL6ER:

## 10

---

This PR ...

- ... adds support for a regex filter list file (currently hard-coded to `/etc/pihole/regex.list`). List entries are read and compiled at start of *FTL*DNS. FTL will print a log entry like:
   ```
   [2018-05-19 17:27:08.153] Compiled 3 Regex filters in 0.4 msec (0 errors)
   ```
   As before, the resolver needs to be restarted to account for a change in wildcard domains.

- ... ensures that regex filter based blocking respects the user defined blocking mode (previously, *FTL*DNS always served `NXDOMAIN` for Regex blocked domains)

- ... removes support for our former method of wildcard blocking (`dnsmasq` config file) as this method was significantly more costly (in the computational sense) and is superseded by regex blocking in all respects. This enhances performance of *FTL*DNS as we do not have to evaluate for each domain if it's cache entry comes from a wildcard entry or a regular cache entry.

- ... removes the previously used config option `BLOCKINGREGEX` as this is superseded by the ability to define any number of regex filters in the file `/etc/pihole/regex.list`.

**ToDo for the other repos:** Add support for the new file in both the CLI and web interface.


pralor:

https://discourse.pi-hole.net/t/blocking-via-regex-now-available-in-ftldns/9101/13

grep.c
```diff
@@ -57,144 +54,13 @@ int readnumberfromfile(const char* fname)
...
-	if(counters.wildcarddomains > 0)
-	{
-		logg("Wildcard blocking list entries: %i", counters.wildcarddomains);
-	}
-	else
-	{
-		logg("No wildcard blocking list present");
-	}
+	// Counting of other blocking domains is done
+	// while these lists are read
```
Mcat12: Not sure what this comment means. Maybe the function needs to be renamed or inlined?

regex.c
```diff
@@ -82,9 +76,64 @@ bool match_regex(char *input)
...
+	}
+
+	// Allocate memory for regex
+	regex = calloc(num_regex, sizeof(regex_t));
+	regexconfigured = calloc(num_regex, sizeof(bool));
+
+	// Search through file
+	// getline reads a string from the specified file up to either a
+	// newline character or EOF
+	while(getline(&buffer, &size, fp) != -1)
```
Mcat12: This could be a for loop (the pieces of it are scattered in the function).

api.c
```diff
@@ -253,7 +253,7 @@ void getTopDomains(char *client_message, int *sock)
 
 		if(blocked && showblocked && domains[j].blockedcount > 0)
 		{
-			if(audit && domains[j].wildcard)
+			if(audit && domains[j].regexmatch == REGEX_BLOCKED)
 			{
 				if(istelnet[*sock])
 					ssend(*sock, "%i %i %s wildcard\n", n, domains[j].blockedcount, domains[j].domain);
```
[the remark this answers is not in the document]
DL6ER: No, we can leave it like this and don't need to do any changes on the web front end. It is really the same thing so nothing get's changed.

dnsmasq_interface.c
```diff
@@ -904,10 +863,85 @@ int FTL_listsfile(char* filename, unsigned int index, FILE *f, int cache_size, s
...
+		cache4->flags = F_HOSTS | F_IMMORTAL | F_FORWARD | F_REVERSE | F_IPV4;
+		// If we block in NXDOMAIN mode, we add the NXDOMAIN flag and make this host record
+		// also valid for AAAA requests
+		if(config.blockingmode == MODE_NX) cache4->flags |= F_IPV6 | F_NEG | F_NXDOMAIN;
+		cache4->ttd = daemon->local_ttl;
+		add_hosts_entry(cache4, addr4, INADDRSZ, index, rhash, hashsz);
+		name_count++;
+	}
+	// Add IPv6 record only if we respond with an IP address to blocked domains
+	if(has_IPv6 && config.blockingmode != MODE_NX &&
```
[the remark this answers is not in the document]
DL6ER: Oh, you mean if their system doesn't even support IPv4 but only IPv6? I'm pretty sure that `dnsmasq` doesn't support this option at the moment (it will fail when it cannot bind to `0.0.0.0:53`). If they have no `IPV4_ADDRESS` in `setupVars.conf`, it will still work as we then use `0.0.0.0` as IPv4 address (see the first `if` block in `prepare_blocking_mode()`)

regex.c
DL6ER: My personal opinion is that `while(getline(&buffer, &size, fp) != -1)` is better readable than `for(;getline(&buffer, &size, fp) != -1;i++)` and it would only make one line (the `i++`) vanish.

Mcat12: It could be
```c
for(int i = 0; getline(&buffer, &size, fp) != -1; i++)
```

Unless we have too old of a version of C.

DL6ER: Yeah, that was my concern:
```
regex.c: In function ‘read_regex_from_file’:
regex.c:116:2: error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode
  for(int i=0; getline(&buffer, &size, fp) != -1; i++)
```

DL6ER: Okay, let's switch to C11 standard, then we can use this syntax.

DL6ER:
For the sake of clarity: Regex blocked queries have not been blocked after receive of `SIGHUP` because they have not been added to the cache and the resolver has removed them upon receiving `SIGHUP`. This bug is also present in the current beta version of *FTL*DNS (regex blocked queries are unblocked after receive of `SIGHUP`).

This is now fixed as all domains are again validated against the (possibly changed) regex filters after `pihole-FTL` received `SIGHUP`.

PR #295, opened 2018-06-02
Recompile Regex API Command

Mcat12:

---

Add an API command to recompile the blocking regex. This will be used by the API instead of sending SIGHUP because that would require more possibly sudo-level permissions.

API command: `>recompile-regex`

The `done` message was removed from the `>reresolve` command to avoid issues with the msgpack API, and there is already an EOM sent at the end.


Mcat12:
If this merges I will need to add it to the ways you can reload the regex list in the [docs](https://github.com/pi-hole/docs).

PR #296, opened 2018-06-02
Limit Regex Logging

Mcat12:

---

Previously every regex evaluation would be logged even if they took very little time:
```
[2018-06-02 00:59:14.450] Regex evaluation took 0.036 msec
[2018-06-02 00:59:15.553] Regex evaluation took 0.006 msec
[2018-06-02 00:59:16.602] Regex evaluation took 0.007 msec
[2018-06-02 00:59:18.110] Regex evaluation took 0.009 msec
[2018-06-02 00:59:18.589] Regex evaluation took 0.006 msec
[2018-06-02 00:59:19.437] Regex evaluation took 0.007 msec
```

Now they will only be logged if they go over 100ms (up for debate if we want this lower).

### DIFF
--- a/regex.c
+++ b/regex.c
@@ -68,7 +68,13 @@ bool match_regex(char *input)
 			break;
 		}
 	}
-	logg("Regex evaluation took %.3f msec", timer_elapsed_msec(REGEX_TIMER));
+
+	double elapsed = timer_elapsed_msec(REGEX_TIMER);
+
+	// Only log evaluation times if they are longer than normal
+	if(elapsed > 100.0)
+		logg("WARN: Regex evaluation took %.3f msec", elapsed);
+
 	// No match, no error, return false
 	return matched;
 }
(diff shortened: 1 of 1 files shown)


DL6ER:
This message should only come once per regex, but I agree that it should be limited. I'd, however, set the limit to a much lower time (maybe 10 msec) for the reason stated below.

The regex evaluation process is, by construction, a blocker for the resolver. Assume a query for a new domain is received. We now have to identify if this domain is subject to regex blocking and hence have to walk through all compiled regex before we can answer this reply. Due to the nature of `dnsmasq`, we cannot serve other queries in between, so if evaluations would always take 100 msec, this would already lead to significant delays in name resolution (especially when many domains are queried, 100 would already mean hard-blicking by 10 seconds).

The performance metrics for regex evaluation are still very unclear, I  have no real feeling for how many complex regex a Raspberry Pi can evaluate within a millisecond. The currently measured times in the nanoseconds regime may be dominated by other effects so they cannot be linearly scaled upwards with increasing number of regex filters.

Mcat12:
The changes look good :+1:

PR #297, opened 2018-06-03
Only send client names (and IP for socket connection) in getClientNames

Mcat12:

---

Previously the client query count (and an incrementing number) was also sent along, although it made no sense to do so (it basically mirrored the top clients response, and was never used by the receiver). Now it only sends the client name for the telnet API, and sends both the hostname and IP for the socket API.

### DIFF
--- a/api.c
+++ b/api.c
@@ -1019,12 +1019,14 @@ void getClientNames(int *sock)
 			client = clients[i].ip;
 
 		if(istelnet[*sock])
-			ssend(*sock, "%i %i %s\n", i, clients[i].count, client);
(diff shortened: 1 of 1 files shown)


DL6ER:
Thanks for this PR. We should also send the IP through the telnet interface to address the issue where hostname are duplicated (same device, once over IPv6, once over IPv4). Instead of the IP we may simply transmit the mere information if this client is connecting over IPv4 or IPv6. What do you think?

Mcat12:
The only reason I didn't add the IP for telnet was because the PHP API didn't use it, but I can change that.

PR #299, opened 2018-06-04
Add whitelisting support for regex filters

DL6ER:

## 10

---

This PR adds more extensive whitelisting support for *FTL*DNS's RegEx filtering in a novel way: *After* regex filters have been applied, *FTL*DNS will *again* check if a domain is set for explicit whitelisting.

This should finally implement a 100% flexible solution for blocking and un-blocking of stuff.

Example 1:
You want to block all Internet and permit only `kidsafe.com`. This wasn't possible before, however, *FTL*DNS supports this now. You just add `.*` as regex (matches everything) and put `kidsace.com` on the whitelist.

Example 2:
You want to block `adserver.com` but want to be able to access `some.handy.tool.adserver.com`. Fine, just add `adserver\.com$` as regex filter and put `some.handy.tool.adserver.com` into your whitelist - done!

A note on possible performance impact: Domains that aren't matched by a regex filter are not compared against the list of whitelisted domains, so there is no cost involved for them. When a filter match happens, the check against the whitelisted domains in only done. The result is stored in memory and will be used from this point on.
TL;DR: The performance impact is expected to be minimal.

### DIFF
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -184,7 +184,13 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 		// - regex checking is enabled, and
 		// - this domain has not already been validated against the regex.
 		// This effectively prevents multiple evaluations of the same domain
-		if(match_regex(domain))
+		//
+		// If a regex filter matched, we additionally compare the domain
+		// against all known whitelisted domains to possibly prevent blocking
+		// of a specific domain. The logic herein is:
+		// If matched, then compare against whitelist
+		// If in whitelist, negate matched so this function returns: not-to-be-blocked
+		if(match_regex(domain) && !in_whitelist(domain))
 		{
 			// We have to block this domain if not already done
 			if(debug) logg("Blocking %s due to RegEx match", domain);
--- a/regex.c
+++ b/regex.c
@@ -221,6 +221,11 @@ void read_regex_from_file(void)
 	// newline character or EOF
 	for(int i=0; getline(&buffer, &size, fp) != -1; i++)
 	{
+		// Test if file has changed since we counted the lines therein (unlikely
+		// but not impossible). If so, read only as far as we have reserved memory
+		if(i >= whitelist.count)
+			break;
+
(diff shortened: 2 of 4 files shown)


regex.c
```diff
@@ -107,6 +142,61 @@ void free_regex(void)
...
+	free_whitelist_domains();
+}
+
+static void read_whitelist_from_file(void)
+{
+	FILE *fp;
+	char *buffer = NULL;
+	size_t size = 0;
+
+	// Get number of lines in the regex file
```
Mcat12: These comments still need to be adjusted (they say "regex" instead of "whitelist")

regex.c
```diff
@@ -69,6 +96,14 @@ bool match_regex(char *input)
 		}
 	}
 
+	// If a regex filter matched, we additionally compare the domain
+	// against all known whitelisted domains to possibly prevent blocking
+	// of a specific domain. The logic herein is:
+	// If matched, then compare against whitelist
+	// If in whitelist, negate matched so this function returns: not-to-be-blocked
+	if(matched)
+		matched = !in_whitelist(input);
```
Mcat12: Should this get moved to the dnsmasq hook code? It pollutes the purpose of the method a bit. Instead of just matching against the regex, it also handles some blocking logic.

dnsmasq_interface.c
```diff
@@ -184,7 +184,13 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
...
 		// - this domain has not already been validated against the regex.
 		// This effectively prevents multiple evaluations of the same domain
-		if(match_regex(domain))
+		//
+		// If a regex filter matched, we additionally compare the domain
+		// against all known whitelisted domains to possibly prevent blocking
+		// of a specific domain. The logic herein is:
+		// If matched, then compare against whitelist
+		// If in whitelist, negate matched so this function returns: not-to-be-blocked
+		if(match_regex(domain) && !in_whitelist(domain))
```
DL6ER: Using short-circuit evaluation to ensure `in_whitelist()` is only called when `match_regex()` was `true`.

PR #304, opened 2018-06-10
Systemdetails

rrobgill:

---

Added ability for the api to provide information about temperature, load average, number of cores and system host name.

Adding these to the api, should make it easier to extend them into the new web interface, towards feature-parity with adminLTE. It will also give any 3rd party applications api access to the same information.

Two functions are added : hostname and systemstats. 
hostname returns the current host name
systemstats returns the number of active cores, standard time-weighted load average, and the temperature if it is available on the current system.

### DIFF
--- a/api.c
+++ b/api.c
@@ -11,6 +11,7 @@
 #include "FTL.h"
 #include "api.h"
 #include "version.h"
+#include <sys/sysinfo.h>
 
 #define min(a,b) ({ __typeof__ (a) _a = (a); __typeof__ (b) _b = (b); _a < _b ? _a : _b; })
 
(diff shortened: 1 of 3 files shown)


Mcat12:
I would not put these into FTL. FTL is focused on providing DNS information, whereas the new API will handle the rest of the data. We are also still considering if data such as load and temperature should be added to the API and shown in the web interface.

DL6ER:
I agree with @Mcat12. Every API call to FTLDNS will block the DNS server as long as the corresponding request is processed. Furthermore, the used way of getting the temperature is a mere copy from the hacky solution in AdminLTE.

This should really only be added in the API (it is anyhow very close to feature bloat) as it can be added there in a non-blocking way.

DL6ER:
I agree with @Mcat12. Every API call to FTLDNS will block the DNS server as long as the corresponding request is processed. Furthermore, the used way of getting the temperature is a mere copy from the hacky solution in AdminLTE.

This should really only be added in the API (it is anyhow very close to feature bloat) as it can be added there in a non-blocking way.

PR #306, opened 2018-06-12
[WIP] Remove expensive query loops

DL6ER:

## 10

---

Since the very early days, FTL had to be able to parse `dnsmasq`'s log file. This file got rotated at midnight or flushed mid-day and, furthermore, we have `dnsmasq` log asynchronously. Hence, FTL always had to identify which query a forwarded/cache/reply line belonged to by looping over all queries.

Luckily, with *FTL*DNS we can finally overcome this computationally very intense looping over all queries. We can do this now as we have a 1:1 mapping between `dnsmasq`'s UUIDs and `FTL`'s UUIDs.

Although one might argue that we had this since the transition to FTL v3.0, however, there was no guaranteed 1:1 mapping as there was still the possibility that `dnsmasq` could get restarted independently of FTL. With *FTL*DNS, however, the resolver cannot be restarted independently any longer.

This is expected to provide a major reduction in computational costs in *FTL*DNS's hook functions, increasing the overall speed by a notable amount.

I also fixed a few typos I stumbled upon while working on the source files.

### DIFF
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -18,7 +18,28 @@ void save_reply_type(unsigned int flags, int queryID, struct timeval response);
 unsigned long converttimeval(struct timeval time);
 static void block_single_domain(char *domain);
 
+int queryIDoffset = 0;
+
 char flagnames[28][12] = {"F_IMMORTAL ", "F_NAMEP ", "F_REVERSE ", "F_FORWARD ", "F_DHCP ", "F_NEG ", "F_HOSTS ", "F_IPV4 ", "F_IPV6 ", "F_BIGNAME ", "F_NXDOMAIN ", "F_CNAME ", "F_DNSKEY ", "F_CONFIG ", "F_DS ", "F_DNSSECOK ", "F_UPSTREAM ", "F_RRNAME ", "F_SERVER ", "F_QUERY ", "F_NOERR ", "F_AUTH ", "F_DNSSEC ", "F_KEYTAG ", "F_SECSTAT ", "F_NO_RR ", "F_IPSET ", "F_NOEXTRA "};
+
+static inline int getQueryID(int id)
+{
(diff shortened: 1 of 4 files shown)


dnsmasq_interface.c
```diff
@@ -266,27 +288,30 @@ void FTL_forwarded(unsigned int flags, char *name, struct all_addr *addr, int id
...
 
-	if(!queries[i].complete)
+	if(!queries[queryID].complete)
 	{
-		int j = queries[i].timeidx;
+		int j = queries[queryID].timeidx;
 		validate_access("overTime", j, true, __LINE__, __FUNCTION__, __FILE__);
 
-		if(queries[i].status == QUERY_CACHE)
+		if(queries[queryID].status == QUERY_CACHE)
```
Mcat12: If the status is always set to `QUERY_FORWARDED` on line 301, then wouldn't this check always fail?

Mcat12:
Just tested this:
- Started with no query history
- Ran `dig google.com` and `dig doubleclick.com`
- Checked `pihole.log` and saw the queries.
- Checked the web interface and FTL's API for queries, but it doesn't return any.
- Checked `pihole-FTL.log` and saw no errors or relevant lines.

Mcat12:
There are cases where the queryIDoffset is not set, such as when the database is not used or if there was an error importing data.

DL6ER:
> There are cases where the `queryIDoffset` is not set, such as when the database is not used or if there was an error importing data.

This variable is initialized with zero, which is the correct value for when nothing got imported from the database (due to whatever reason):

https://github.com/pi-hole/FTL/blob/68d154fe9112b1aa6d6605cccc22326150276013/dnsmasq_interface.c#L21

DL6ER:
Marking as Work In Progress as there is a problem with non-monotonic IDs within `dnsmasq` when TCP children are forked.

DL6ER:
Closing in favor of #314

PR #310, opened 2018-06-15
Change default blocking mode to "IP-NODATA-AAAA".

DL6ER:

## 10

---

Change default blocking mode to [`IP-NODATA-AAAA`](https://docs.pi-hole.net/ftldns/blockingmode/#pi-holes-ipv6-nodata-blocking) for Pi-hole v4.0.
This blocking mode fully supports the blocking page but should alleviate all issues with non-static IPv6 addresses as all `AAAA` requests to blocked domains will be answered with `NODATA`.

This is currently under discussion and may be closed without merging at any point.


config.c
```diff
@@ -330,7 +330,7 @@ void get_blocking_mode(FILE *fp)
 		else if(strcasecmp(buffer, "NULL") == 0)
 			config.blockingmode = MODE_NULL;
 		else if(strcasecmp(buffer, "IP-NODATA-AAAA") == 0)
-			config.blockingmode = MODE_IP_AAAA_NODATA;
+			config.blockingmode = MODE_IP_NODATA_AAAA;
```
Mcat12: Need to add another check for `"IP"` or there is no way to enable that mode.

Mcat12:
This will also require docs changes

DL6ER:
Merging this now. We're internally discussing if we should even switch to `NULL` blocking as the default to avoid HTTPS loading issues many users are seeing.

If we change to `NULL` blocking as default mode, we'll do this in a separate PR.

PR #314, opened 2018-06-17
Use reverse iterations for searching queryIDs

DL6ER:

## 10

---

Use reverse iterations when searching for `queryID` matches. This greatly increases computational efficiency of several subroutines critical for the performance of *FTL*DNS.

This renders #306 obsolete as queries will most likely be found in the first iteration. However, if this is not the case, then we will do at most 1000 iterations to avoid expensive looping for queries that have not been stored in FTL's data structure.

### DIFF
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -231,7 +231,7 @@ void FTL_forwarded(unsigned int flags, char *name, struct all_addr *addr, int id
 	// Save status and forwardID in corresponding query identified by dnsmasq's ID
 	bool found = false;
 	int i;
-	// Loop through all queries - this is an expensive loop, however, there is no
+	// Loop over all queries - this is an expensive loop, however, there is no
 	// good alternative as we will loose the relation between dnsmasq's id and our
 	// id due to garbage collection, hence, it may be that a query that with an ID
 	// of dnsmasq of 123.456 is our query with ID 567 when the other queries have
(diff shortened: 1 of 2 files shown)


dnsmasq_interface.c
```diff
@@ -231,20 +231,19 @@ void FTL_forwarded(unsigned int flags, char *name, struct all_addr *addr, int id
...
+	// We iterate from the most recent query down to at most MAXITER queries in the past to avoid
+	// iterating through the entire array of queries when queries that have not been recorded
+	// (like PTR queries, etc.) are processed.
+	// MAX(0, a) is used to return 0 in case a is negative (negative array indices are harmful)
 
 	// Validate access only once for the maximum index (all lower will work)
 	validate_access("queries", counters.queries-1, false, __LINE__, __FUNCTION__, __FILE__);
-	for(i=0; i<counters.queries; i++)
+	int until = MAX(0, counters.queries-MAXITER);
+	for(i = counters.queries-1; i > until; i--)
```
Mcat12: Should be `i >= until` so that it checks `MAXITER` number of queries instead of `MAXITER-1` number of queries.
Example:
```
counters.queries = 2
MAXITER = 1
until = counters.queries - MAXITER = 2 - 1 = 1
i = counters.queries-1 = 1
i > until = 1 > 1 = false
In this case, nothing is checked even though 1 query should be checked.
```
This will need to be changed in other places as well.

PR #326, opened 2018-07-16
RegEx debugging mode

DL6ER:

## 10

---

Add debug mode for RegEx. This should allow users to autonomously resolve issues related to user-configured regex filters.
For the sake of convenience, this debug mode can be enabled through the `pihole-FTL.conf` file.

It lead to the resolution of issue https://github.com/pi-hole/pi-hole/issues/2281#issuecomment-405312236 . An exemplary output is given therein.

### DIFF
--- a/config.c
+++ b/config.c
@@ -215,9 +215,9 @@ void read_FTLconf(void)
 		config.regex_debugmode = true;
 
 	if(config.regex_debugmode)
-		logg("   REGEX_DEBUGMODE: RegEx debugging mode. May increase log file size!");
+		logg("   REGEX_DEBUGMODE: Active. May increase log file size!");
 	else
-		logg("   REGEX_DEBUGMODE: Normal RegEx mode");
+		logg("   REGEX_DEBUGMODE: Inactive");
 
 	logg("Finished config file parsing");
 
--- a/config.c
+++ b/config.c
@@ -206,6 +206,19 @@ void read_FTLconf(void)
 			break;
 	}
 
+	// REGEX_DEBUGMODE
+	// defaults to: No
+	config.regex_debugmode = false;
+	buffer = parse_FTLconf(fp, "REGEX_DEBUGMODE");
+
+	if(buffer != NULL && strcasecmp(buffer, "true") == 0)
+		config.regex_debugmode = true;
+
+	if(config.regex_debugmode)
+		logg("   REGEX_DEBUGMODE: RegEx debugging mode. May increase log file size!");
+	else
+		logg("   REGEX_DEBUGMODE: Normal RegEx mode");
+
 	logg("Finished config file parsing");
 
 	// Release memory
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -193,7 +193,7 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 		if(match_regex(domain) && !in_whitelist(domain))
 		{
 			// We have to block this domain if not already done
-			if(debug) logg("Blocking %s due to RegEx match", domain);
+			if(debug) logg("Blocking %s due to Regex match", domain);
 			block_single_domain(domain);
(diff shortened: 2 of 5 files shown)


DL6ER:
Exemplary output in Regex debugging mode:
```
[2018-07-17 17:40:51.304] DEBUG: Regex in line 2 ("((^)|(\.))twitter\.") matches "whatever.twitter.com"
```

config.c
```diff
@@ -206,6 +206,19 @@ void read_FTLconf(void)
...
+	// REGEX_DEBUGMODE
+	// defaults to: No
+	config.regex_debugmode = false;
+	buffer = parse_FTLconf(fp, "REGEX_DEBUGMODE");
+
+	if(buffer != NULL && strcasecmp(buffer, "true") == 0)
+		config.regex_debugmode = true;
+
+	if(config.regex_debugmode)
+		logg("   REGEX_DEBUGMODE: RegEx debugging mode. May increase log file size!");
```
Mcat12: Change description to "Active. May increase log file size!"

dnsmasq_interface.c
```diff
@@ -193,7 +193,7 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 		if(match_regex(domain) && !in_whitelist(domain))
 		{
 			// We have to block this domain if not already done
-			if(debug) logg("Blocking %s due to RegEx match", domain);
+			if(debug) logg("Blocking %s due to Regex match", domain);
```
Mcat12: Does this config option supersede this line?

regex.c
```diff
@@ -38,6 +39,14 @@ static bool init_regex(const char *regexin, int index)
 		log_regex_error("compiling", errcode, index);
 		return false;
 	}
+
+	// Store compiled regex string in buffer if in regex debug mode
+	if(config.regex_debugmode)
+	{
+		regexbuffer[index] = strdup(regexin);
+		if(regexbuffer[index] == NULL)
+			logg("WARN: init_regex() failed to allocate memory for regexbuffer[index]");
```
Mcat12: We already have these allocation and free cases handled in `memory.c`, and they provide more detail.

regex.c
```diff
@@ -86,6 +95,10 @@ bool match_regex(char *input)
 		{
 			// Match, return true
 			matched = true;
+
+			// Print match message when in regex debug mode
+			if(config.regex_debugmode)
+				logg("DEBUG: Regex in line %i (\"%s\") matches \"%s\"", index+1, regexbuffer[index], input);
```
Mcat12: It might make it clearer if we only wrap the regex in parentheses, and not parentheses and quotes.

dnsmasq_interface.c
DL6ER: Yes, there is no reason to show it in FTL's standard debug mode any more as it has proven to work exactly as expected even for large lists.

memory.c
```diff
@@ -268,4 +268,15 @@ void FTLfree(void *ptr, const char * file, const char * function, int line)
...
+	// Set freed pointer explicitly to NULL to avoid carrying around a dangling pointer
+	// Dangling pointers are pointers that do not point to a valid object of the
+	// appropriate type. They represent a special case of memory safety violation.
+	// Dangling pointers arise during object destruction, when an object is deallocated,
+	// without modifying the value of the pointer, so that the pointer still points to
+	// the memory location of the deallocated memory.
+	// If the program writes to memory referenced by a dangling pointer, a silent
+	// corruption of unrelated data may result, leading to subtle bugs that can be
+	// extremely difficult to find.
+	ptr = NULL;
```
Mcat12: This line will not affect anything, because it is on the stack and will be immediately dropped. Sadly, it is not a replacement for setting the pointers you are freeing via this method to NULL.

dnsmasq_interface.c
Mcat12: Should this line then be removed?

PR #332, opened 2018-07-30
Store IP addresses in natural form

DL6ER:

## 10

---

Instead of storing IP addresses as individually allocated strings, we store them in their natural byte form. Both IPv4 and IPv6 addresses fit into the 16 bytes variable we use for this (IPv4 addresses will only use the first four bytes, though).

### DIFF
--- a/datastructure.c
+++ b/datastructure.c
@@ -136,7 +136,6 @@ int findForwardID(const char * forward, bool count)
 	else
 		forwarded[forwardID].count = 0;
 	// Save forward destination IP address
-	// Store client IP
 	saveForwardIP(i, forward);
 	forwarded[forwardID].failed = 0;
 	// Initialize forward hostname
@@ -264,9 +263,10 @@ void saveClientIP(int i, const char *ipaddr)
 
 char* getClientIP(int i)
 {
-	char *buffer = calloc(INET6_ADDRSTRLEN, sizeof(char));
 	int proto = clients[i].IPv4 ? AF_INET : AF_INET6;
-	inet_ntop(proto, clients[i].addr, buffer, INET6_ADDRSTRLEN);
+	size_t size = clients[i].IPv4 ? INET_ADDRSTRLEN : INET6_ADDRSTRLEN;
+	char *buffer = calloc(size, sizeof(char));
+	inet_ntop(proto, clients[i].addr, buffer, size);
 	return buffer;
 }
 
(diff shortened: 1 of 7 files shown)


datastructure.c
```diff
@@ -83,16 +92,31 @@ int findOverTimeID(int overTimetimestamp)
 int findForwardID(const char * forward, bool count)
 {
 	int i, forwardID = -1;
+	int ret, proto = !(strstr(forward,":") != NULL) ? AF_INET : AF_INET6;
```
Mcat12: Avoid a double negative (here and below)

datastructure.c
```diff
@@ -112,7 +136,8 @@ int findForwardID(const char * forward, bool count)
 	else
 		forwarded[forwardID].count = 0;
 	// Save forward destination IP address
-	forwarded[forwardID].ip = strdup(forward);
+	// Store client IP
```
Mcat12: Is this saving the client IP? Maybe delete this comment and use the above comment.

datastructure.c
```diff
@@ -217,14 +254,34 @@ int findClientID(const char *client)
...
+	int ret, proto = clients[i].IPv4 ? AF_INET : AF_INET6;
+	if((ret = inet_pton(proto, ipaddr, clients[i].addr)) != 1)
+		logg("ERROR: inet_pton(%i, %s, %p) failed with %i", proto, ipaddr, clients[i].addr, ret);
+}
+
+char* getClientIP(int i)
+{
+	char *buffer = calloc(INET6_ADDRSTRLEN, sizeof(char));
+	int proto = clients[i].IPv4 ? AF_INET : AF_INET6;
+	inet_ntop(proto, clients[i].addr, buffer, INET6_ADDRSTRLEN);
```
Mcat12: These get IP functions could check for IPv4 vs IPv6 before allocating.

dnsmasq_interface.c
```diff
@@ -87,7 +87,7 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 	if(config.privacylevel >= PRIVACY_HIDE_DOMAINS_CLIENTS)
 	{
 		free(client);
-		client = strdup("0.0.0.0");
+		client = strdup("::");
```
Mcat12: Is this intended?

DL6ER: Yes, `0.0.0.0` is not really well-defined whereas `::` is officially the "unspecified address".

DL6ER:
Closing this for now as sometimes spurious (wrong!) IPs appear. I will probably re-implement this in another form.

PR #337, opened 2018-08-05
Fix incorrect log_regex_error() subroutine

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:**

## 10

---

In `log_regex_error()`, the `logg()` routing was given the arguments in the wrong order. Further, we should not call `free_regex()` when a single regex compilation fails as this will `free` all regex filters (incl. the correct ones).

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

PR #338, opened 2018-08-05
Create regex checklist

DL6ER:

 ## 10

---

This PR is milestoned for Pi-hole v4.1

It adds a new file (`/etc/pihole/regex.checklist`) containing lines with either `OK` or a regex compilation error to be shown on the Pi-hole dash board.

### DIFF
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -57,6 +57,9 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 		return;
 	}
 
+	// Store plain text domain in buffer for regex validation
+	char *domainbuffer = strdup(domain);
+
 	// Check and apply possible privacy level rules
 	// We do this immediately on the raw data to avoid any possible leaking
 	get_privacy_level(NULL);
@@ -77,6 +80,7 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 	   (strcmp(client, "127.0.0.1") == 0 || strcmp(client, "::1") == 0))
 	{
 		free(domain);
+		free(domainbuffer);
 		free(client);
 		disable_thread_lock();
 		return;
@@ -115,6 +119,7 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 		// Return early to avoid accessing querytypedata out of bounds
 		if(debug) logg("Notice: Skipping unknown query type: %s (%i)", types, id);
 		free(domain);
+		free(domainbuffer);
 		free(client);
 		disable_thread_lock();
 		return;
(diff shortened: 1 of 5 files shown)


DL6ER:
Note that this PR shows a few extra commits as it is based on `release/v4.0` rather than `development` itself. These commits will disappear from this PR's `diff` as soon as `release/v4.0` is merged into `development`.

Mcat12:
Can this data be pulled from the FTL API instead of put in another file? It seems cleaner to do it that way.

DL6ER:
Hmm, it's actually much easier and in some sense also logical (FTL creates a new list with the same ordering as the read `regex.list` file). Doing this through the API is much more involved and requires a lot more code changes.

dnsmasq_interface.c
```diff
@@ -57,6 +57,9 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 		return;
 	}
 
+	// Store plain text domain in buffer for regex validation
+	char *domainbuffer = strdup(domain);
```
TiboDevC: a large comment on the code: why still using C rather than C++?
For example this could be an array and I think it would be easier to program and to read

regex.c
```diff
@@ -17,16 +17,19 @@ static bool *regexconfigured = NULL;
...
 {
 	// Regex failed for some reason (probably user syntax error)
 	// Get error string and log it
 	size_t length = regerror(errcode, &regex[index], NULL, 0);
 	char *buffer = calloc(length,sizeof(char));
 	(void) regerror (errcode, &regex[index], buffer, length);
-	logg("ERROR %s regex %i: %s (%i)", index, where, buffer, errcode);
+	logg("ERROR %s regex on line %i: %s (%i)", where, index+1, buffer, errcode);
+	if(checklist)
+		fprintf(checklist, "%s\n", buffer);
```
TiboDevC: where is it going to be printed then?

TiboDevC: (I mean, in the interface, in the dashboard)

DL6ER: If you look closely at this pull request, you'll find there reference to the connected dashboard pull request: 
screenshot at 2018-08-20 13-39-05

https://github.com/pi-hole/AdminLTE/pull/804

dnsmasq_interface.c
DL6ER: C++ is another language. Although it is surely somehow originating from C, I don't see it as its successor any anything like that. I agree that there are many convenience improvements in C++ (such as higher-level processing of strings, vectors, ...) but with C we have more fine grained control over performance as we do everything a bit more low-level. The embedded DNS/DHCP/TFTP server `dnsmasq` is written entirely in C and I want to avoid having mixed language program for no immediate benefits (despite, admittedly, maybe a bit more readability here and there).

DL6ER:
> Why not passing everything into Regex? Regex can match the normal blacklist as well. It would lite the code and make it faster.

No, it would make it much slower. The blocklist/blacklist is stored in a hashed cache table. Lookups even with hundreds of thousands of entries are very fast (less than a few milliseconds) as we need to look at only a tiny subsection of the table. However, with regex, they are so powerful that potentially each filter could match anything (trivial example `dfg` would also match `ab.cdfgh.t.com`). Hence, we need to iterate over *each and every* regex filter to see if one of them matches. This is **much** slower than the simple lookup table we use for the black-/blocklists.

This is maybe something similar to the C++ vs. C question: Yes, we may make it a bit more complicated but (by doing it how we do it) we gain orders of magnitude of performance compared to the most simple / straightforward / easy to read way of doing it.

DL6ER:
Superseded by https://github.com/pi-hole/FTL/pull/337. We log the error message to FTL's log file.

PR #357, opened 2018-08-19
Add PRIVACY_NOSTATS mode that skips all analysis

DL6ER:

## 10
---

Add `PRIVACY_NOSTATS` mode that skips all (!) analysis. We furthermore don't even try to import or save anything to the database.

This will not only be helpful for those asking for even increased privacy over what we already offer (counters and overtime statistics with fully anonymized query data) but also for those running huge Pi-hole deploys where the initial reading in of the historic data (with up to millions of queries) causes a massive startup delay of several minutes (see, e.g., https://github.com/pi-hole/pi-hole/issues/2374).


Mcat12:
When adding the corresponding documentation, it should be pointed out that users will have to disable logging as well (`pihole logging off`) so the dnsmasq log gets disabled.

PR #382, opened 2018-09-18
Use negative query IDs for queries loaded from the database

Mcat12:

---

This is targeted at the `new/shmem` branch.

Use negative query IDs for queries loaded from the database. This is necessary for all queries to have unique IDs. Previously, all queries loaded from the database had an ID of zero.

The API uses the query ID as a unique identifier for cursor pagination, and so queries loaded from the database cause an issue with this. There is no other unique identifier for queries. The index of the query in the queries array can't be used because it could change after garbage collection.

### DIFF
--- a/database.c
+++ b/database.c
@@ -617,12 +617,6 @@ void read_data_from_DB(void)
 	// Loop through returned database rows
 	while((rc = sqlite3_step(stmt)) == SQLITE_ROW)
 	{
-		// Ensure we have enough space in the queries struct
-		memory_check(QUERIES);
-
-		// Set ID for this query
-		int queryID = counters->queries;
-
 		int queryTimeStamp = sqlite3_column_int(stmt, 1);
 		// 1483228800 = 01/01/2017 @ 12:00am (UTC)
 		if(queryTimeStamp < 1483228800)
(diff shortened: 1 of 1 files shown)


database.c
```diff
@@ -699,21 +693,31 @@ void read_data_from_DB(void)
...
 
+		// Ensure we have enough space in the queries struct
+		memory_check(QUERIES);
+
+		// Set index for this query
+		int queryIndex = counters->queries;
+
+		// Set the ID for this query. Queries loaded from the database use negative IDs.
+		// 1 is added to the counter before flipping the sign so the IDs start at -1 instead of 0.
+		int queryID = -1 * (counters->queries + 1);
```
DL6ER: Is the logic of this correct?

The first loaded database entry will be -1, the second -2, etc. leading to the situation:
```
queries[0].id = -1 // loaded from database
queries[1].id = -2 // loaded from database
queries[2].id = -3 // loaded from database
queries[3].id = -4 // loaded from database
queries[4].id = -5 // loaded from database
queries[5].id = -6 // loaded from database
queries[6].id = -7 // loaded from database
queries[7].id = -8 // loaded from database
queries[8].id = 0 // first incoming live query
queries[9].id = 1 // second incoming live query
```

I'd rather assume that there should be no sharp transition between no. 7 and 8

Mcat12: I originally thought to do it the second way, but that would require either a complex SQL query to find all the queries that are valid that will be imported, or a second O(n) for loop to go through and set the IDs in that order. For the purposes of being unique, it isn't that much of an issue, but if you want we can add a loop to set them in that order.

PR #389, opened 2018-09-26
Allocate individual client overTime objects guaranteed pagesize-aligned

DL6ER:
See title. Internal review pull request.

### DIFF
--- a/shmem.c
+++ b/shmem.c
@@ -87,7 +87,7 @@ void newOverTimeClient() {
 
 	// Create the shared memory with enough space for the current overTime slots
 	shm_unlink(name);
-	SharedMemory shm = create_shm(name, sizeof(int) * counters->overTime);
+	SharedMemory shm = create_shm(name, (counters->overTime/pagesize + 1)*pagesize*sizeof(int));
 	if(shm.ptr == NULL) {
 		free(shm.name);
 		logg("Failed to initialize new overTime client %d", overTimeClientCount);
@@ -105,18 +105,18 @@ void newOverTimeClient() {
 }
 
 void addOverTimeClientSlot() {
-	// For each client slot, add an overTime slot
-	for(int i = 0; i < overTimeClientCount; i++) {
-		// Get the name of the shared memory
-		char *name = clientShmName(i);
+	// For each client slot, add pagesize overTime slots
+	for(int i = 0; i < overTimeClientCount; i++)
+	{
+		// Only increase size of the shm object is needed
+		if(shm_overTimeClients[i].size > (size_t)counters->overTime)
+			continue;
 
 		// Reallocate with one more slot
-		realloc_shm(&shm_overTimeClients[i], shm_overTimeClients[i].size + sizeof(int));
+		realloc_shm(&shm_overTimeClients[i], (counters->overTime/pagesize + 1)*pagesize*sizeof(int));
 
 		// Update overTimeClientData
 		overTimeClientData[i] = shm_overTimeClients[i].ptr;
-
-		free(name);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


shmem.c
```diff
@@ -87,7 +87,7 @@ void newOverTimeClient() {
 
 	// Create the shared memory with enough space for the current overTime slots
 	shm_unlink(name);
-	SharedMemory shm = create_shm(name, sizeof(int) * counters->overTime);
+	SharedMemory shm = create_shm(name, (counters->overTime/pagesize + 1)*pagesize*sizeof(int));
```
DL6ER: The first part of this computation,
```c
(counters->overTime/pagesize + 1)
```
deliberately uses integer computation (i.e., `15/16 == 0`, `16/16 == 1`) to determine the number of pages to allocate for having enough space for `counters->overTime` entries.

shmem.c
```diff
@@ -105,18 +105,18 @@ void newOverTimeClient() {
...
 void addOverTimeClientSlot() {
-	// For each client slot, add an overTime slot
-	for(int i = 0; i < overTimeClientCount; i++) {
-		// Get the name of the shared memory
-		char *name = clientShmName(i);
+	// For each client slot, add pagesize overTime slots
+	for(int i = 0; i < overTimeClientCount; i++)
+	{
+		// Only increase size of the shm object is needed
+		if(shm_overTimeClients[i].size > (size_t)counters->overTime)
```
DL6ER: Cast to `size_t` is done because `counters->overTime` is a signed integer whereas `shm_overTimeClients[i].size` is of type `size_t` (unsigned type)

shmem.c
```diff
@@ -105,18 +105,18 @@ void newOverTimeClient() {
 }
 
 void addOverTimeClientSlot() {
-	// For each client slot, add an overTime slot
-	for(int i = 0; i < overTimeClientCount; i++) {
-		// Get the name of the shared memory
-		char *name = clientShmName(i);
```
DL6ER: There seems to be no need for the `name` buffer here, so I removed this code. Please verify this @Mcat12

shmem.c
```diff
@@ -105,18 +105,18 @@ void newOverTimeClient() {
...
+	// For each client slot, add pagesize overTime slots
+	for(int i = 0; i < overTimeClientCount; i++)
+	{
+		// Only increase size of the shm object is needed
+		if(shm_overTimeClients[i].size > (size_t)counters->overTime)
+			continue;
 
 		// Reallocate with one more slot
-		realloc_shm(&shm_overTimeClients[i], shm_overTimeClients[i].size + sizeof(int));
+		realloc_shm(&shm_overTimeClients[i], (counters->overTime/pagesize + 1)*pagesize*sizeof(int));
```
DL6ER: Uses the same math as above. This line is skipped if the object doesn't need to be resized.

shmem.c
Mcat12: Correct

shmem.c
```diff
@@ -105,18 +105,18 @@ void newOverTimeClient() {
...
 
 void addOverTimeClientSlot() {
-	// For each client slot, add an overTime slot
-	for(int i = 0; i < overTimeClientCount; i++) {
-		// Get the name of the shared memory
-		char *name = clientShmName(i);
+	// For each client slot, add pagesize overTime slots
+	for(int i = 0; i < overTimeClientCount; i++)
+	{
+		// Only increase size of the shm object is needed
```
Mcat12: > Only increase size of the shm object is needed

Only increase _the_ size of the shm object _if_ needed

shmem.c
Mcat12: `shm_overTimeClients[i].size` stores the size of the memory in bytes whereas `counters->overTime` (effectively) stores the number of slots each overTime client should have, so the memory size should be divided by `sizeof(int)` to get the number of overTime slots stored in shared memory.

shmem.c
Mcat12: It might be clearer to use this expression instead: `shm_overTimeClients[i].size + pagesize * sizeof(int)`

PR #396, opened 2018-10-05
Add shared memory lock

Mcat12:

---

This replaces the previous mutex managed by threads.c. The lock is stored in shared memory with the name "/FTL-lock", and any clients connecting to the shared memory are expected to use the lock when accessing memory.


DL6ER:
The function names are a bit odd.

```
shm_write_lock();
shm_unlock_lock();
```

I'd suggest using similar names, like
```
shm_lock();
shm_unlock();
```
or maybe
```
shm_write_lock();
shm_remove_lock();
```
(I'd prefer the former).

I expect to be able to add more review comments / test your changes by tomorrow afternoon.

Mcat12:
The simpler names are a bit too simple since there are both a read and write lock. The change to the "unlock" function is fine though.

socket.c
```diff
@@ -321,13 +322,13 @@ void *telnet_connection_handler_thread(void *socket_desc)
 
 			// Lock FTL data structure, since it is likely that it will be changed here
 			// Requests should not be processed/answered when data is about to change
-			enable_thread_lock();
+			shm_write_lock();
```
DL6ER: Shouldn't we use `shm_read_lock();` here? The comment may be outdated (or was actually never correct...).

socket.c
```diff
@@ -378,13 +379,13 @@ void *socket_connection_handler_thread(void *socket_desc)
...
```
DL6ER: See above comment about using `shm_read_lock();` in the socket routines.

shmem.c
```diff
@@ -120,11 +123,58 @@ void addOverTimeClientSlot() {
...
+	// Allow the lock to be used by other processes
+	pthread_rwlockattr_setpshared(&lock_attr, PTHREAD_PROCESS_SHARED);
+
+	// Initialize the lock
+	pthread_rwlock_init(&lock, &lock_attr);
+
+	return lock;
+}
+
+void shm_read_lock() {
```
DL6ER: We should a comment like: If a writer holds the lock, the calling thread cannot acquire the read lock. It blocks (that is, it does not return from `shm_read_lock()`) until the read lock is acquired. The benefit is that **multiple readers can access the data at the same time**.

shmem.c
```diff
@@ -120,11 +123,58 @@ void addOverTimeClientSlot() {
...
+	int result = pthread_rwlock_rdlock(sharedMemoryLock);
+
+	if(result != 0)
+		logg("Failed to obtain SHM read lock: %s", strerror(result));
+}
+
+void shm_write_lock() {
+	int result = pthread_rwlock_wrlock(sharedMemoryLock);
```
DL6ER: We should a comment like: The calling thread acquires the write lock if no other thread (regardless if reader or writer) holds the lock. Otherwise, the thread blocks (that is, does not return from `shm_write_lock()`) until it can acquire the lock. Only one thread can be a writer (at the same time) and hold the lock. Any reader can block any writer while readers do not block themselves.

Mcat12: See the header for the documentation.

Mcat12:
The issue of starving FTL has been fixed by using a robust mutex. The API also makes sure to not starve FTL by not making any new locks if FTL signals that it needs the lock (it waits until the current locks are dead).

PR #398, opened 2018-10-07
Store database ID in queries struct

DL6ER:
This DB ID will be available through shmem to the API to support transparent overlap with the SQlite3 database. It is marked `[WIP]` as it needs to be further tested (e.g. when no database is present on the system).

### DIFF
--- a/FTL.h
+++ b/FTL.h
@@ -160,7 +160,7 @@ typedef struct {
 	int domainID;
 	int clientID;
 	int forwardID;
-	bool db;
+	sqlite3_int64 db;
 	int id; // the ID is a (signed) int in dnsmasq, so no need for a long int here
 	bool complete;
 	bool private;
--- a/database.c
+++ b/database.c
@@ -382,6 +382,9 @@ void save_to_DB(void)
 	long int i;
 	sqlite3_stmt* stmt;
 
+	// Get last ID stored in the database
+	sqlite3_int64 lastID = last_ID_in_DB();
+
 	bool ret = dbquery("BEGIN TRANSACTION");
 	if(!ret)
 	{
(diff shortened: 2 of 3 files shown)


database.c
```diff
@@ -487,12 +515,21 @@ void save_to_DB(void)
 		return;
 	}
 
+	// Update individual queryIDs in the queries struct
+	sqlite3_int64 lastID = last_ID_in_DB();
+	for(i=0; i < total; i++)
+	{
+		// Subtract i from coutners-queries-1 as the database
+		// loop goes only until i < counters->queries
+		queries[(counters->queries-1)-i].db = lastID-i;
```
Mcat12: This does not consider private queries,  queries too new (see lines 414-419), or save errors (if less than 3).

Maybe it would be better to get the last ID in the database before going into the main loop, and then setting the ID as each query is entered into the database?

FTL.h
```diff
@@ -160,7 +160,7 @@ typedef struct {
 	int domainID;
 	int clientID;
 	int forwardID;
-	bool db;
+	sqlite3_int64 db;
```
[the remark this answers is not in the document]
DL6ER: The SQlite3 definition itself is a bit more involved than simply using something like `long long int` because there are architectures (like the RPi 1 B+ IIRC) that do not really support any 64 byte integer architecture-wise.

See also https://www.sqlite.org/c3ref/int64.html

database.c
```diff
@@ -712,7 +743,7 @@ void read_data_from_DB(void)
 		queries[queryIndex].clientID = clientID;
 		queries[queryIndex].forwardID = forwardID;
 		queries[queryIndex].timeidx = timeidx;
-		queries[queryIndex].db = true; // Mark this as already present in the database
+		queries[queryIndex].db = dbid;
 		queries[queryIndex].id = queryID;
```
Mcat12: Do we now want to revert this to just 0, like how `development` handles it? The database ID is sufficient for the API to paginate on (for queries already in the database, otherwise it needs the dnsmasq ID).

PR #401, opened 2018-10-23
Upgrade embedded dnsmasq to v2.80

### LINKED ISSUE #383: Can't dig +trace

1

---

**[BUG] Expected Behaviour:**

```
dig +trace www.example.com

; <<>> DiG 9.10.6 <<>> +trace www.example.com
;; global options: +cmd
.			518400	IN	NS	a.root-servers.net.
.			518400	IN	NS	b.root-servers.net.
.			518400	IN	NS	c.root-servers.net.
.			518400	IN	NS	d.root-servers.net.
.			518400	IN	NS	e.root-servers.net.
```

dig +trace performs a query for root zone NS records, it should start by returning a list of root servers. As above.

```
dig +trace www.example.com

; <<>> DiG 9.10.6 <<>> +trace www.example.com
;; global options: +cmd
;; Received 17 bytes from 172.28.28.25#53(172.28.28.25) in 12 ms
```
With packet capture, pihole-FTL returns 'Flags: 0x80a2 Standard query response, Server failure', and dig +trace fails.

This is occurring on 2 different pihole installs, both pihole-FTL v4.0

If I stop the pihole-FTL service and start dnsmasq service then correct behaviour returns.

```
systemctl stop pihole-FTL && systemctl start dnsmasq
```
dig +trace is now working correctly.

```
systemctl stop dnsmasq && systemctl start pihole-FTL
```
dig +trace is now failing again as above example.

OS:

I have tested this on 2 different machines on up to date Ubuntu 18 installs.

pihole was installed via install script : curl -sSL https://install.pi-hole.net | bash

I updated pihole by running pihole -up on both servers and the issue persists.

DL6ER:

## 10

---

**Official changelog:**
- Add support for RFC 4039 DHCP rapid commit. Thanks to Ashram Method for the initial patch and motivation.
- Alter the default for `dnssec-check-unsigned`. Versions of dnsmasq prior to 2.80 defaulted to not checking unsigned replies, and used `--dnssec-check-unsigned` to switch this on. Such configurations will continue to work as before, but those which used the default of no checking will need to be altered to explicitly select no checking. The new default is because switching off checking for unsigned replies is inherently dangerous. Not only does it open the possiblity of forged replies, but it allows everything to appear to be working even when the upstream namesevers do not support DNSSEC, and in this case no DNSSEC validation at all is occuring.
- Fix DHCP broken-ness when `--no-ping` AND `--dhcp-sequential-ip` are set. Thanks to Daniel Miess for help with this.
- Add a facilty to store DNS packets sent/recieved in a pcap-format file for later debugging. The file location is given by the `--dumpfile` - option, and a bitmap controlling which packets should be dumped is given by the `--dumpmask` option.
- Handle the case of both standard and constructed `dhcp-range`s on the same interface better. We don't now construct a `dhcp-range` if there's - already one specified. This allows the specified interface to have different parameters and avoids advertising the same prefix twice. Thanks - to Luis Marsano for spotting this case.
- Allow zone transfer in authoritative mode if `auth-peer` is specified, even if `auth-sec-servers` is not. Thanks to Raphael Halimi for the suggestion.
- Fix bug which sometimes caused dnsmasq to wrongly return answers without DNSSEC RRs to queries with the do-bit set, but only when DNSSEC - validation was not enabled. Thanks to Petr MenÅ¡Ã­k for spotting this.
- Fix missing fatal errors with some malformed options (`server`, `local`, `address`, `rebind-domain-ok`, `ipset`, `alias`). Thanks to Eugene Lozovoy for - spotting the problem.
- Fix crash on startup with a `--synth-domain` which has no prefix. Introduced in 2.79. Thanks to Andreas Engel for the bug report.
- Fix missing EDNS0 section in some replies generated by local DNS configuration which confused `systemd-resolvd`. Thanks to Steve Dodd for characterising the problem.
- Add `--dhcp-name-match config` option. 
- Add `--caa-record config` option.
- Implement `--address=/example.com/#` as (more efficient) syntactic sugar for `--address=/example.com/0.0.0.0` and `--address=/example.com/::` - Returning null addresses is a useful technique for ad-blocking. Thanks to Peter Russell for the suggestion.
- Change anti cache-snooping behaviour with queries with the recursion-desired bit unset. Instead to returning `SERVFAIL`, we now always - forward, and never answer from the cache. This allows "dig +trace" command to work. (*fixes #383*)
- Include in the example config file a formulation which stops DHCP clients from claiming the DNS name "`wpad`". This is a fix for the CERT Vulnerability VU#598349.

**Note:** This PR is Work-In-Progress and only meant for testing. As it implements a still rather fresh release of `dnsmasq`, it **should not be merged to `development` until Pi-hole v4.1 has been released.** Consider this branch as experimental as there is a certain likeliness that more internals of `pihole-FTL` may need to be adapted due to changes deeper inside the `dnsmasq` code.

The PR statistics are ` +1,460 -283` when ignoring whitespace changes.


DL6ER:
Ready to be merged into `development`. The release candidate (`release/4.1`) to be merged to `master` has already been branched out from `development`.

pralor:

https://discourse.pi-hole.net/t/pi-hole-v4-2-available-with-shared-memory-new-blocking-modes-and-more/16763/1

PR #405, opened 2018-10-25
Move resolution of client and upstream host names to a dedicated thread

### LINKED ISSUE #404: Upstream Names Not Resolved If Database Is Disabled
**In raising this issue, I confirm the following:**

10

---

Upstream names should be resolved even if not using the database, since that feature does not depend on the database functionality.

Upstream names are only resolved in the database thread, which does not run if the database is not enabled:
https://github.com/pi-hole/FTL/blob/059693a32ace9e98b9c3df065b5419f16a18e9b9/database.c#L553

The client name resolution is duplicated in `reresolveHostnames`:
https://github.com/pi-hole/FTL/blob/059693a32ace9e98b9c3df065b5419f16a18e9b9/resolve.c#L67
The upstream name resolution should be moved to this function and `resolveNewClients` should be deleted, since `reresolveHostnames` is called from the GC thread and will run even if the database is disabled. The functionality of resolving only new clients/upstreams should be kept as a flag to the function.
https://github.com/pi-hole/FTL/blob/059693a32ace9e98b9c3df065b5419f16a18e9b9/gc.c#L174

Precondition: FTL is set to use an upstream with a resolvable name, and there exists at least one client with a resolvable name.
- Stop FTL
- Disable the database by adding `DBFILE=` to the config file at `/etc/pihole/pihole-FTL.conf`
- Start FTL
- Make queries against FTL from at least one client with a resolvable name.
- After waiting for at least one hour (GC runs every hour), the upstream names should still remain unresolved but client names should be resolved. This can be verified by checking the web interface.

N/A

**Device specifics**

Hardware Type: Tinkerboard
OS: Armbian 5.60 (Ubuntu 16.04.5 LTS)

DL6ER:

## 10

---

Move resolution of client and upstream host names to a dedicated thread. This PR fixes #404 

### DIFF
--- a/FTL.h
+++ b/FTL.h
@@ -279,3 +279,4 @@ extern pthread_t telnet_listenthreadv6;
 extern pthread_t socket_listenthread;
 extern pthread_t DBthread;
 extern pthread_t GCthread;
+extern pthread_t resolverthread;
--- a/FTL.h
+++ b/FTL.h
@@ -279,4 +279,4 @@ extern pthread_t telnet_listenthreadv6;
 extern pthread_t socket_listenthread;
 extern pthread_t DBthread;
 extern pthread_t GCthread;
-extern pthread_t resolverthread;
+extern pthread_t DNSclientthread;
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -907,6 +907,13 @@ void FTL_fork_and_bind_sockets(struct passwd *ent_pw)
 		exit(EXIT_FAILURE);
 	}
 
+	// Start thread that will stay in the background until host names needs to be resolved
+	if(pthread_create( &resolverthread, &attr, resolver_thread, NULL ) != 0)
+	{
+		logg("Unable to open resolver thread. Exiting...");
+		exit(EXIT_FAILURE);
+	}
+
 	// Chown files if FTL started as user root but a dnsmasq config option
 	// states to run as a different user/group (e.g. "nobody")
 	if(ent_pw != NULL && getuid() == 0)
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -853,6 +853,7 @@ pthread_t telnet_listenthreadv6;
 pthread_t socket_listenthread;
 pthread_t DBthread;
 pthread_t GCthread;
+pthread_t DNSclientthread;
 
 void FTL_fork_and_bind_sockets(struct passwd *ent_pw)
 {
@@ -908,9 +909,9 @@ void FTL_fork_and_bind_sockets(struct passwd *ent_pw)
 	}
 
 	// Start thread that will stay in the background until host names needs to be resolved
-	if(pthread_create( &resolverthread, &attr, resolver_thread, NULL ) != 0)
+	if(pthread_create( &DNSclientthread, &attr, DNSclient_thread, NULL ) != 0)
 	{
-		logg("Unable to open resolver thread. Exiting...");
+		logg("Unable to open DNS client thread. Exiting...");
 		exit(EXIT_FAILURE);
 	}
 
--- a/resolve.c
+++ b/resolve.c
@@ -99,7 +99,7 @@ void resolveClients(bool onlynew)
 }
 
 // Resolve upstream destination host names
-static void resolveForwardDestinations(bool onlynew)
+void resolveForwardDestinations(bool onlynew)
 {
 	int i;
 	for(i = 0; i < counters.forwarded; i++)
--- a/resolve.c
+++ b/resolve.c
@@ -126,8 +126,7 @@ void resolveForwardDestinations(bool onlynew)
 	}
 }
 
-pthread_t resolverthread;
-void *resolver_thread(void *val)
+void *DNSclient_thread(void *val)
 {
 	// Set thread name
 	prctl(PR_SET_NAME, "DNS client", 0, 0, 0);
--- a/resolve.c
+++ b/resolve.c
@@ -149,6 +149,8 @@ void *DNSclient_thread(void *val)
 		{
 			// Try to resolve all client host names (onlynew=false)
 			resolveClients(false);
+			// Try to resolve all upstream destination host names (onlynew=false)
+			resolveForwardDestinations(false);
 			// Prevent immediate re-run of this routine
 			sleepms(500);
 		}
--- a/resolve.c
+++ b/resolve.c
@@ -10,6 +10,14 @@
 
 #include "FTL.h"
 
+// Resolve new client and upstream server host names
+// once every minute
+#define RESOLVE_INTERVAL 60
+
+// Re-resolve client names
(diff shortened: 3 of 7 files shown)


resolve.c
```diff
@@ -62,68 +70,93 @@ char *resolveHostname(const char *addr)
...
 		}
-		free(hostname);
 	}
 }
 
-// This routine is run *before* saving to the database (default interval is once per minute)
-// to account for new clients (and forward destinations)
-void resolveNewClients(void)
+// Resolve upstream destination host names
+static void resolveForwardDestinations(bool onlynew)
```
Mcat12: We have the capability to resolve all forward destinations instead of just new ones, but we don't use it currently. When all clients are re-resolved, we should also re-resolve forward destination names.

resolve.c
```diff
@@ -62,68 +70,93 @@ char *resolveHostname(const char *addr)
...
 		}
 	}
-	for(i = 0; i < counters.forwarded; i++)
+}
+
+pthread_t resolverthread;
+void *resolver_thread(void *val)
+{
+	// Set thread name
+	prctl(PR_SET_NAME,"resolver",0,0,0);
```
Mcat12: This may get confused with DNS resolver.

DL6ER: It's only the thread name that is displayed, e.g., by `htop`. I don't think that anything can really get confused here:

screenshot from 2018-10-26 20-12-47

resolve.c
DL6ER: I wonder if this is really meaningful, as we don't expect the forward destination host names to ever change, do we? In contrast, users may decide to rename their local devices, e.g., from `android-a479b97ef09880b890d233a` to `paul-iphone`

resolve.c
Mcat12: To me, I see in that list the jobs of FTL listed out. It resolves domains on the resolver thread, does database stuff on the database thread, listens to sockets and telnet on their respective threads, and does some housekeeping on the housekeeper thread. I don't see any other thread that looks like "DNS" or "dnsmasq", so the resolver thread seems to be doing that job. A better name might be "internal resolver".

resolve.c
Mcat12: Perhaps if the upstream is localhost, and the hostname changes? It may not be worth implementing, but on the other hand, it is a very small cost (usually only 1 to 3 upstreams).

resolve.c
Mcat12: DNS client sounds fine.

resolve.c
```diff
@@ -62,68 +70,93 @@ char *resolveHostname(const char *addr)
...
+			forwarded[i].name = hostname;
+			forwarded[i].new = false;
+
+			disable_thread_lock();
 		}
 	}
-	for(i = 0; i < counters.forwarded; i++)
+}
+
+pthread_t resolverthread;
```
Mcat12: Can this be moved to where the other threads are defined (main)?

dnsmasq_interface.c
```diff
@@ -907,6 +907,13 @@ void FTL_fork_and_bind_sockets(struct passwd *ent_pw)
 		exit(EXIT_FAILURE);
 	}
 
+	// Start thread that will stay in the background until host names needs to be resolved
+	if(pthread_create( &resolverthread, &attr, resolver_thread, NULL ) != 0)
+	{
+		logg("Unable to open resolver thread. Exiting...");
```
Mcat12: As the thread was renamed to DNS client, rename this reference.

FTL.h
```diff
@@ -279,3 +279,4 @@ extern pthread_t telnet_listenthreadv6;
 extern pthread_t socket_listenthread;
 extern pthread_t DBthread;
 extern pthread_t GCthread;
+extern pthread_t resolverthread;
```
Mcat12: As the thread was renamed to DNS client, rename this variable.

resolve.c
```diff
@@ -62,68 +70,93 @@ char *resolveHostname(const char *addr)
...
+void *resolver_thread(void *val)
```
Mcat12: As the thread was renamed to DNS client, rename this function.

resolve.c
DL6ER: They are all defined in `dnsmasq_interface.c` (close to where the threads are launched). I'll move the definition there.

Mcat12: Ah ya, sorry, thought that was main for a second (we used to launch the threads from there).

resolve.c
```diff
@@ -62,68 +70,92 @@ char *resolveHostname(const char *addr)
...
 
-		// Only try to resolve new forward destinations
-		if(forwarded[i].new)
+		// Run every hour to update possibly changed client host names
+		if(time(NULL) % RERESOLVE_INTERVAL == 0)
 		{
-			forwarded[i].name = resolveHostname(forwarded[i].ip);
-			forwarded[i].new = false;
+			// Try to resolve all client host names (onlynew=false)
+			resolveClients(false);
```
Mcat12: Only clients are re-resolved here, not upstreams like in `>reresolve`.

resolve.c
```diff
@@ -62,68 +70,92 @@ char *resolveHostname(const char *addr)
...
-		if(clients[clientID].count < 1)
-			continue;
+		validate_access("clients", i, true, __LINE__, __FUNCTION__, __FILE__);
 
-		// Get client hostname
-		char *hostname = resolveHostname(clients[clientID].ip);
-		if(strlen(hostname) > 0)
+		// If onlynew flag is set, we will only resolve new clients
+		// If not, we will try to re-resolve all known clients
+		if(onlynew && clients[i].new)
```
Mcat12: If `onlynew` is false, no resolving will occur at all.

resolve.c
```diff
@@ -62,68 +70,92 @@ char *resolveHostname(const char *addr)
...
-		validate_access("clients", i, true, __LINE__, __FUNCTION__, __FILE__);
+		validate_access("forwarded", i, true, __LINE__, __FUNCTION__, __FILE__);
 
-		// Only try to resolve new clients
-		// Note that it can happen that we are not able to find hostnames but we don't
-		// want to try to resolve them every minute in this case.
-		if(clients[i].new)
+		// If onlynew flag is set, we will only resolve new upstream destinations
+		// If not, we will try to re-resolve all known upstream destinations
+		if(onlynew && forwarded[i].new)
```
Mcat12: See above comment.

resolve.c
DL6ER: I'm still not a big fan of resolving the upstream names frequently, but I agree that it should be consistent, so I'll add it.

Mcat12: I'm not totally for or against it, so if you think it is fine without it, then it can be removed.

DL6ER: We can leave it in, but should keep in mind that this changes how FTL did it before (although there will probably no effect at all) and that there will be a few more PTR queries.

PR #416, opened 2018-11-07
Adjustable file locations

DL6ER:

## 10

---

This PR increases the flexibility concerning the files FTL will use. It adds a number of additional config options, listed below with their default value:

```
LOGFILE=/var/log/pihole-FTL.log
PIDFILE=/var/run/pihole-FTL.pid
PORTFILE=/var/run/pihole-FTL.port
SOCKETFILE=/var/run/pihole/FTL.sock
WHITELISTFILE=/etc/pihole/whitelist.txt
BLACKLISTFILE=/etc/pihole/blacklist.txt
GRAVITYFILE=/etc/pihole/gravity.list
REGEXLISTFILE=/etc/pihole/regex.list
SETUPVARSFILE=/etc/pihole/setupVars.conf
AUDITLISTFILE=/etc/pihole/auditlog.list
```

Furthermore, we will now check in three different places for a FTL config file before we go ahead and use the default values (as listed above):

1. `/etc/pihole/pihole-FTL.conf`
2. `/var/snap/pihole/common/etc/pihole/pihole-FTL.conf`
3. a file `pihole-FTL.conf` in the current directory (`$PWD`)

**Although this PR is made to release/4.1, it should not be understood as a blocker for the release.**

### DIFF
--- a/FTL.h
+++ b/FTL.h
@@ -94,7 +94,7 @@ enum { BLOCKING_DISABLED, BLOCKING_ENABLED, BLOCKING_UNKNOWN };
 typedef struct {
 	const char* conf;
 	const char* conf2;
-	const char* log;
+	char* log;
 	const char* pid;
 	const char* port;
 	char* db;
--- a/FTL.h
+++ b/FTL.h
@@ -93,7 +93,7 @@ enum { BLOCKING_DISABLED, BLOCKING_ENABLED, BLOCKING_UNKNOWN };
 // Static structs
 typedef struct {
 	const char* conf;
-	const char* conf2;
+	const char* snapConf;
 	char* log;
 	char* pid;
 	char* port;
--- a/FTL.h
+++ b/FTL.h
@@ -93,6 +93,7 @@ enum { BLOCKING_DISABLED, BLOCKING_ENABLED, BLOCKING_UNKNOWN };
 // Static structs
 typedef struct {
 	const char* conf;
+	const char* conf2;
 	const char* log;
 	const char* pid;
 	const char* port;
--- a/FTL.h
+++ b/FTL.h
@@ -98,7 +98,7 @@ typedef struct {
 	char* pid;
 	char* port;
 	char* db;
-	const char* socketfile;
+	char* socketfile;
 } FTLFileNamesStruct;
 
 typedef struct {
--- a/FTL.h
+++ b/FTL.h
@@ -102,7 +102,7 @@ typedef struct {
 } FTLFileNamesStruct;
 
 typedef struct {
-	const char* whitelist;
+	char* whitelist;
 	const char* blacklist;
 	const char* gravity;
 	const char* regexlist;
--- a/FTL.h
+++ b/FTL.h
@@ -96,7 +96,7 @@ typedef struct {
 	const char* conf2;
 	char* log;
 	char* pid;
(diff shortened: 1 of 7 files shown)


FTL.h
```diff
@@ -93,23 +93,21 @@ enum { BLOCKING_DISABLED, BLOCKING_ENABLED, BLOCKING_UNKNOWN };
 // Static structs
 typedef struct {
 	const char* conf;
-	const char* log;
-	const char* pid;
-	const char* port;
+	const char* conf2;
```
Mcat12: This should be given a more descriptive name, such as `snapConf`

config.c
```diff
@@ -253,6 +325,35 @@ void read_FTLconf(void)
 		fclose(fp);
 }
 
+void getpath(FILE* fp, const char *option, const char *defaultloc, char **pointer)
+{
+	char *buffer = parse_FTLconf(fp, "PIDFILE");
+
+	errno = 0;
+	// Use sscanf() to obtain filename from config file parameter only if buffer != NULL
+	if(!(buffer != NULL && sscanf(buffer, "%127ms", pointer)))
```
Mcat12: This logic can be simplified (De Morgan's Law) to use a single or expression.

config.c
```diff
@@ -253,6 +325,35 @@ void read_FTLconf(void)
 		fclose(fp);
 }
 
+void getpath(FILE* fp, const char *option, const char *defaultloc, char **pointer)
```
Mcat12: Could this function get some documentation (what it does, what the parameters mean, etc)?

config.c
```diff
@@ -13,15 +13,60 @@
...
+		printf("Notice: Found no readable FTL config file");
+	}
+
+	// Read LOGFILE value if available
+	// defaults to: "/var/log/pihole-FTL.log"
+	buffer = parse_FTLconf(fp, "LOGFILE");
+
+	errno = 0;
+	// Use sscanf() to obtain filename from config file parameter only if buffer != NULL
+	if(!(buffer != NULL && sscanf(buffer, "%127ms", &FTLfiles.log)))
```
Mcat12: Logic can be simplified (see lower comment)

config.c
```diff
@@ -13,15 +13,60 @@
...
+		printf("Notice: Found no readable FTL config file");
+	}
+
+	// Read LOGFILE value if available
+	// defaults to: "/var/log/pihole-FTL.log"
+	buffer = parse_FTLconf(fp, "LOGFILE");
+
+	errno = 0;
+	// Use sscanf() to obtain filename from config file parameter only if buffer != NULL
+	if(buffer == NULL || sscanf(buffer, "%127ms", &FTLfiles.log) == 0)
```
Mcat12: `sscanf` returns `EOF` on error, which is -1 (but the macro should be used instead). This may need to be used instead of zero.

DL6ER: If the option is, e.g. `LOGFILE=`, we read an empty string. From the `man` page it is not entirely clear if `sscanf()` would return `0` or `EOF` is returned if the string passed to the function is of length zero. You may be right. Shall we test for `EOF` or shall we be more general an do something like ` < 1` ?

Mcat12: Testing for `< 1` will probably do fine (we could also do `!= 1`).

PR #417, opened 2018-11-09
Fix/privacy levels regex filtering

### LINKED ISSUE #414: Broken: Fix for regex filters + privacy level > 0
Looks like this fix does not work anymore:
https://github.com/pi-hole/FTL/pull/331

I can reproduce this behavior with the latest version of Pi Hole on different machines.

DL6ER:

## 10

---

Fix regex filtering when privacy levels is > 0. This fixes #414 

When a new query comes in, we check if we know this domain already.

If so: We check against the regex filters. If one of the filters match, we inject the domain as blocked into our DNS cache.
If not: We continue.

When the same domains is queried another time, we don't evaluate the regex filters any more as we rely on that the query will be quickly replied to from cache.

This usually works fine, however, for a non-zero privacy level, all domains are only stored as `hidden` in FTL's memory. Accordingly, any subsequent query (after the first )is not validated against the regex filters any more as they all get the same `domainID`.

This has been fixed.
**As a consequence, you cannot lower the privacy level without restarting `pihole-FTL`**!

### DIFF
--- a/api.c
+++ b/api.c
@@ -815,7 +815,7 @@ void getAllQueries(char *client_message, int *sock)
 		   queries[i].privacylevel < PRIVACY_HIDE_DOMAINS_CLIENTS)
 			client = clients[queries[i].clientID].name;
 		else
-			client = getClientString(i);
+			client = getClientIPString(i);
 
 		unsigned long delay = queries[i].response;
 		// Check if received (delay should be smaller than 30min)
--- a/api.c
+++ b/api.c
@@ -265,7 +265,7 @@ void getTopDomains(char *client_message, int *sock)
 			continue;
 
 		// Hidden domain, probably due to privacy level. Skip this in the top lists
-		if(strcmp(domains[j].domain, "hidden") == 0)
+		if(strcmp(domains[j].domain, HIDDEN_DOMAIN) == 0)
 			continue;
 
 		if(blocked && showblocked && domains[j].blockedcount > 0)
@@ -401,6 +401,10 @@ void getTopClients(char *client_message, int *sock)
 		// Skip this client if there is a filter on it
 		if(excludeclients != NULL &&
 			(insetupVarsArray(clients[j].ip) || insetupVarsArray(clients[j].name)))
+			continue;
+
+		// Hidden client, probably due to privacy level. Skip this in the top lists
+		if(strcmp(clients[j].ip, HIDDEN_CLIENT) == 0)
 			continue;
 
 		// Only return name if available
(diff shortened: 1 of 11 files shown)


dnsmasq_interface.c
```diff
@@ -167,7 +150,12 @@ void FTL_new_query(unsigned int flags, char *name, struct all_addr *addr, char *
 	queries[queryID].db = false;
 	queries[queryID].id = id;
 	queries[queryID].complete = false;
-	queries[queryID].private = (config.privacylevel == PRIVACY_MAXIMUM);
+
+	// Check and apply possible privacy level rules
+	// We do this immediately on the raw data to avoid any possible leaking
```
Mcat12: The second line of this comment may need to be tweaked, since no data is modified by getting the privacy level. Instead, the current privacy level is remembered for that query.

datastructure.c
```diff
@@ -234,3 +234,25 @@ bool isValidIPv6(const char *addr)
...
+	if(queries[queryID].privacylevel < PRIVACY_HIDE_DOMAINS)
+	{
+		validate_access("domains", queries[queryID].domainID, true, __LINE__, __FUNCTION__, __FILE__);
+		return domains[queries[queryID].domainID].domain;
+	}
+	else
+		return HIDDEN_DOMAIN;
+}
+
+char *getClientString(int queryID)
```
Mcat12: The function name should indicate that it returns the client IP, not the name

api.c
```diff
@@ -402,10 +403,6 @@ void getTopClients(char *client_message, int *sock)
 			(insetupVarsArray(clients[j].ip) || insetupVarsArray(clients[j].name)))
 			continue;
 
-		// Hidden client, probably due to privacy level. Skip this in the top lists
-		if(strcmp(clients[j].ip, "0.0.0.0") == 0)
-			continue;
```
Mcat12: I don't see a call to `getClientString` to replace this code.

DL6ER: At the very beginning of this subroutine, we just terminate if the privacy level is high enough that client IPs are hidden. As the privacy level cannot be lowered on-the-fly any longer, we can never get into a situation where the loop would run over any queries with a sufficient privacy level due to this early returning.

Mcat12: If FTL is restarted with a lower privacy level, then it could loop over queries which were saved in the database with hidden client IPs. Would we still want to hide those?

DL6ER: Hmm, I agree that having the Top Items being dominated by `hidden` resp. `0.0.0.0` doesn't make much sense. I will reimplement the code to skip such entries in the Top Item replies.

PR #423, opened 2018-11-24
Fix query status if forwarded query was partially replied to from cache

DL6ER:

## 10

---

Fix query status if forwarded query was partially replied to from cache. This may happen if we were able to answer a CNAME from cache but need to ask the upstream server for details concering one of its components. This PR ensures that the query status is set to forwarded only *after* the check if it has been replied to from cache as the latter is a null-op otherwise.

### DIFF
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -321,7 +321,7 @@ void FTL_forwarded(unsigned int flags, char *name, struct all_addr *addr, int id
 	}
 	else
 	{
-		// Normal cache reply
+		// Normal forwarded query (status is set below)
 		// Query is no longer unknown
 		counters.unknown--;
 		// Hereby, this query is now fully determined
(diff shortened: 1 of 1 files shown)


dnsmasq_interface.c
```diff
@@ -290,54 +287,58 @@ void FTL_forwarded(unsigned int flags, char *name, struct all_addr *addr, int id
...
+		gettimeofday(&response, 0);
+		// Reset timer, shift slightly into the past to acknowledge the time
+		// FTLDNS needed to look up the CNAME in its cache
+		queries[i].response = converttimeval(response) - queries[i].response;
+	}
+	else
 	{
-		int j = queries[i].timeidx;
-		validate_access("overTime", j, true, __LINE__, __FUNCTION__, __FILE__);
+		// Normal cache reply
```
Mcat12: This happens when it's a non-cache reply

PR #429, opened 2018-12-11
Use shared memory objects for data

DL6ER:

## 10

---

This PR is a combined effort of @Mcat12 and @DL6ER. It changes the way FTL stores its data internally. Previously, everything was held in private memory. However, after merging this PR, the data will instead be held in POSIX Shared Memory objects, allowing a straightforward sharing with the future API we are currently working on (see [here](https://github.com/pi-hole/api/pull/60)).

Although shared memory objects could be understood as being slightly less secure compared to private memory, however this is not the case. The data is still protected as it is protected with permissions `600`, i.e., only the owner (the user `pihole`) can read and write on the data. This serves the same level of privacy as we do already now.


Mcat12:
Closing and reopening to try and get Codacy working...

dschaper:
Flipping status for CodeFactor check.

pralor:

https://discourse.pi-hole.net/t/pi-hole-v4-2-available-with-shared-memory-new-blocking-modes-and-more/16763/1

PR #430, opened 2018-12-12
Fix no-daemon mode forking dnsmasq and not logging

Mcat12:

---

no-daemon mode operates just like debug mode, but without the debug output, and dnsmasq is not in debug mode.

Note that there is an issue in no-daemon mode where Ctrl-C and other such signals do not stop FTL like in debug mode. However, I could only fix this by putting dnsmasq in debug mode...


DL6ER:
I don't understand the logic of this PR. It doesn't seem to change the behavior of `-f` aka `no-daemon`, does it?

PR #439, opened 2018-12-17
Zero initialize all_addr structs

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X])**:

## 10

---

Make sure to zero initialize the `all_addr` structs we are using for the IPv4 and IPv6 target addresses.

This should resolve a bug that was reported on [Discourse](https://discourse.pi-hole.net/t/very-strange-ipv6-blocking-addresses/15197/9).

Furthermore, this zero initialization allows a great simplification of `prepare_blocking_mode()` as `inet_pton("0.0.0.0")` resp. `inet_pton("::")` do nothing else than filling the corresponding structs with zeroes.

### DIFF
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -1069,7 +1069,7 @@ static int add_blocked_domain_cache(struct all_addr *addr4, struct all_addr *add
 // Add a single domain to resolver's cache. This respects the configured blocking mode
 static void block_single_domain(char *domain)
 {
-	struct all_addr addr4 = { 0 }, addr6 = { 0 };
+	struct all_addr addr4 = {{{ 0 }}}, addr6 = {{{ 0 }}};
 	bool has_IPv4 = false, has_IPv6 = false;
 
 	// Get IPv4/v6 addresses for blocking depending on user configures blocking mode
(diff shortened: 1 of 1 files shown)


dnsmasq_interface.c
```diff
@@ -1078,7 +1069,7 @@ static int add_blocked_domain_cache(struct all_addr *addr4, struct all_addr *add
 // Add a single domain to resolver's cache. This respects the configured blocking mode
 static void block_single_domain(char *domain)
 {
-	struct all_addr addr4, addr6;
+	struct all_addr addr4 = { 0 }, addr6 = { 0 };
```
DL6ER: For the sake of documentation: If there are fewer initializers in a brace-enclosed list than there are elements or members of an aggregate, or fewer characters in a string literal used to initialize an array of known size than there are elements in the array (this is the case here), the remainder of the aggregate shall be initialized implicitly the same as objects that have static storage duration according to the C99 Standard.

DL6ER:
Building on arm failed. There seems to be a regression with gcc 4.9 used in the arm build. [This link](https://github.com/OP-TEE/optee_os/issues/1998#issuecomment-350997292) suggests to use `{{{0}}}` instead of `{0}` for the initialization.

DL6ER:
arm build passed

dnsmasq_interface.c
```diff
@@ -987,47 +987,38 @@ unsigned long converttimeval(struct timeval time)
...
-		IPv6addr = "::";
+		// Blocking mode will use zero-initialized all_addr struct
+		*has_IPv4 = true;
 	}
-	if(IPv6addr != NULL)
+
+	if(config.blockingmode == MODE_IP)
 	{
+		// Read IPv6 address for host entries from setupVars.conf
+		char* const IPv6addr = read_setupVarsconf("IPV6_ADDRESS");
```
Mcat12: The IPv6 address may not exist in the settings (if IPv6 is not used).

PR #449, opened 2018-12-25
Fix: Deadlock situation in resolving subroutines

DL6ER:

## 10

---

This bug does only exist in `development` and is not contained in any release code.

This PR want to merge three improvements/bugfixes concerning out internal locks:
- **Name resolution is broken in `development`.** This PR ensures that the lock is released when resolving host names. This is necessary as we otherwise block the main thread which is supposed to do the resolving.
- Add debugging output for locks. Although it increases the verbosity of the debug mode, this information can actually be quite valuable and we should keep if until we are sure there are no further obstacles with the now lock mechanism. It will greatly improve the remote support possibilities as we can look more into *where* things are currently happening in the code (at least concerning locks).
- Move obtaining/releasing locks to the place where the functions that need it are called. This fixes a bug where `socket.c` tried to release a lock that was already previously released in `request.c` (resulting in the error message `Failed to unlock SHM lock: Operation not permitted`). Furthermore, this prevents obtaining a lock for both, API calls that do not need a lock and also for unknown API calls.

### DIFF
--- a/resolve.c
+++ b/resolve.c
@@ -85,11 +85,19 @@ void resolveClients(bool onlynew)
 		if(onlynew && !clients[clientID].new)
 			continue;
 
+		// Lock data when obtaining IP of this client
 		lock_shm();
+		const char* ipaddr = getstr(clients[clientID].ippos);
+		unlock_shm();
 
-		clients[clientID].namepos = addstr(resolveHostname(getstr(clients[clientID].ippos)));
+		// Important: Don't hold a lock while resolving as the main thread
+		// (dnsmasq) needs to be operable during the call to resolveHostname()
+		const char* hostname = resolveHostname(ipaddr);
+
+		// Finally, lock data when storing obtained hostname
+		lock_shm();
+		clients[clientID].namepos = addstr(hostname);
 		clients[clientID].new = false;
-
 		unlock_shm();
 	}
 }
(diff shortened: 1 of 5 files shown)


resolve.c
```diff
@@ -85,11 +85,19 @@ void resolveClients(bool onlynew)
 		if(onlynew && !clients[clientID].new)
 			continue;
 
+		// Lock data when obtaining IP of this client
 		lock_shm();
+		const char* ipaddr = getstr(clients[clientID].ippos);
+		unlock_shm();
 
-		clients[clientID].namepos = addstr(resolveHostname(getstr(clients[clientID].ippos)));
```
DL6ER: Doing all this in one line was the origin of the observed issues with name resolution as `resolveHostname()` cannot run when the lock is still obtained. Hence, I have split this one-liner into three blocks where only the first and the third are run locked. This is fine as the function `resolveHostname()` does neither access not modify FTL's datastructure and, hence, does not need to run locked (actually, it *cannot* even run locked).

resolve.c
```diff
@@ -108,11 +116,20 @@ void resolveForwardDestinations(bool onlynew)
 		if(onlynew && !forwarded[forwardID].new)
 			continue;
 
+		// Lock data when obtaining IP of this forward destination
 		lock_shm();
+		const char* ipaddr = getstr(forwarded[forwardID].ippos);
+		unlock_shm();
 
-		forwarded[forwardID].namepos = addstr(resolveHostname(getstr(forwarded[forwardID].ippos)));
```
DL6ER: See above comment for the client host name resolution. This one-lines is affected by the same problem.

shmem.c
```diff
@@ -155,12 +155,16 @@ pthread_mutex_t create_mutex() {
 	return lock;
 }
 
-void lock_shm() {
+void _lock_shm(const char* function, const int line, const char * file) {
```
DL6ER: Provided by a macro in `shmem.h`. Note the `_` at the beginning of the name to symbolize this.

shmem.c
```diff
@@ -175,9 +179,11 @@ void lock_shm() {
 		logg("Failed to obtain SHM lock: %s", strerror(result));
 }
 
-void unlock_shm() {
+void _unlock_shm(const char* function, const int line, const char * file) {
```
DL6ER: Provided by a macro in `shmem.h`. Note the `_` at the beginning of the name to symbolize this.

PR #452, opened 2019-01-01
Bugfix for new, more efficient NULL blocking mode

DL6ER:

## 10

---

Pass larger memory size to add_hosts_entry() if we create an entry that is valid for both, `A` and `AAAA` requests. This prevents uninitialized bytes that caused inconsistent behavior as reported on [Discourse](https://discourse.pi-hole.net/t/very-strange-ipv6-blocking-addresses/15197).

The changes are less than indicated by `git diff` as the code was also reindented to increase readability. The only real change is the introduction of the new variable `memorysize` that indicates if the address to be copied over is only valid for IPv4 or even for IPv6. This is strictly not necessary here, as we ensure that the structs are properly initialized, i.e., even always copying `IN6ADDRSZ` would be okay, however, we honor how `dnsmasq` does it internally and hence pass the appropriate memory size as integer.

Marked as `ON HOLD` as we're waiting on confirmation that the proposed change fixes the observed behavior.

### DIFF
--- a/dnsmasq_interface.c
+++ b/dnsmasq_interface.c
@@ -1070,16 +1070,28 @@ static int add_blocked_domain_cache(struct all_addr *addr4, struct all_addr *add
 	{
 		strcpy(cache4->name.sname, domain);
 		cache4->flags = F_HOSTS | F_IMMORTAL | F_FORWARD | F_REVERSE | F_IPV4;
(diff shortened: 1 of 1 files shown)


pralor:

https://discourse.pi-hole.net/t/very-strange-ipv6-blocking-addresses/15197/49

DL6ER:
`ON HOLD` removed

PR #454, opened 2019-01-04
SHM tweaks

DL6ER:

## 10

---

This PR adds two changes:

- Add error reporting for `shm_unlink()`.
- Move `shm_unlink()` into `create_shm()` to reduce code duplication.

Assume a scenario where `pihole-FTL` was started as `root` and terminated without a proper shutdown (killed, crashed, etc.). This scenario would leave the shm object behind. They would be owned by `root`. A subsequent `sudo service pihole-FTL (re)start` would fail in this case as the `root`-owned shm objects cannot get deleted. This is a fatal error as no memory can be obtained. Previously, there was no checking for such scenarios, this PR adds proper error reporting for the `shm_unlink()` operation.

Exemplary additional output that would be generated in the scenario described above:
```
[2019-01-04 17:46:03.213] create_shm(): shm_unlink("/FTL-lock") failed: Permission denied (13)
```

### DIFF
--- a/shmem.c
+++ b/shmem.c
@@ -194,7 +194,6 @@ bool init_shmem(void)
 	pagesize = getpagesize();
 
 	/****************************** shared memory lock ******************************/
-	shm_unlink(SHARED_LOCK_NAME);
 	// Try to create shared memory object
 	shm_lock = create_shm(SHARED_LOCK_NAME, sizeof(ShmLock));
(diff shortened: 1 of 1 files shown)


DL6ER:
Does not need to be mentioned in the release notes as shared memory is a new feature added by v4.2

PR #490, opened 2019-02-01
Further improvements for external blocking detection

DL6ER:

## 10

---

Change detection of the "externally blocked" criterion after further private communication with Quad9. It seems the intended detection method
```
AD == 1 && RA == 0 && RCODE == NXDOMAIN
```
is not reliable. They suggest to use the new criterion
```
RA == 0 && RCODE == NXDOMAIN
```
which seems to produce the expected results in all tests.

Secondly, this PR cleans up some code that is never called and removes the `queries.AD` boolean which is now obsolete. Furthermore, this PR improves a bit on the debugging messages to reduce confusion in the future.


api.c
```diff
@@ -801,7 +801,7 @@ void getAllQueries(char *client_message, int *sock)
 
 		if(istelnet[*sock])
 		{
-			ssend(*sock,"%i %s %s %s %i %i %i %lu\n",queries[i].timestamp,qtype,domain,client,queries[i].status,queries[i].dnssec,queries[i].reply,delay);
+			ssend(*sock,"%i %s %s %s %i %i %i %lu %i\n",queries[i].timestamp,qtype,domain,client,queries[i].status,queries[i].dnssec,queries[i].reply,delay,i);
```
Mcat12: Is this intended?

DL6ER: Yes. It is harmless for the PHP API as this number will just be appended to the individual JSON arrays. However, as this is a pure debugging property, I will wrap it in an `if(debug)` statement.

FTL.h
```diff
@@ -70,7 +70,7 @@
 enum { DATABASE_WRITE_TIMER, EXIT_TIMER, GC_TIMER, LISTS_TIMER, REGEX_TIMER, ARP_TIMER, LAST_TIMER };
 enum { QUERIES, FORWARDED, CLIENTS, DOMAINS, OVERTIME, WILDCARD };
 enum { DNSSEC_UNSPECIFIED, DNSSEC_SECURE, DNSSEC_INSECURE, DNSSEC_BOGUS, DNSSEC_ABANDONED, DNSSEC_UNKNOWN };
-enum { QUERY_UNKNOWN, QUERY_GRAVITY, QUERY_FORWARDED, QUERY_CACHE, QUERY_WILDCARD, QUERY_BLACKLIST, QUERY_EXTERNAL_BLOCKED };
+enum { QUERY_UNKNOWN, QUERY_GRAVITY, QUERY_FORWARDED, QUERY_CACHE, QUERY_WILDCARD, QUERY_BLACKLIST, QUERY_EXTERNAL_BLOCKED_IP, QUERY_EXTERNAL_BLOCKED_NULL, QUERY_EXTERNAL_BLOCKED_NXRA };
```
Mcat12: It would be useful to document the difference between the external blocked query types.

shmem.c
```diff
@@ -12,7 +12,7 @@
 #include "shmem.h"
 
 /// The version of shared memory used
-#define SHARED_MEMORY_VERSION 2
+#define SHARED_MEMORY_VERSION 3
```
Mcat12: This will now need to be version 4, due to version 3 being taken by the overTime rewrite.

DL6ER: Due to the version jump that will be included in FTL v4.2.3, this now needs to be `5`.

FTL.h
DL6ER: Where should the documentation happen? It is hidden in the telnet API which will be replaced by the new API in the future. The difference between them is:

- `QUERY_EXTERNAL_BLOCKED_IP` = Detected an IP address that is known to correspond to a blocking page of an upstream provider
- `QUERY_EXTERNAL_BLOCKED_NULL` = Upstream server returned `0.0.0.0` as address which suggests they blocked it
- `QUERY_EXTERNAL_BLOCKED_NXRA` = Upstream server returned `NXDOMAIN` with `RA` bit set

Mcat12: The documentation should be located next to the enum definition.

Mcat12: Actually, it would be best documented on the documentation site:
https://docs.pi-hole.net/ftldns/database/#supported-status-types

PR #492, opened 2019-02-03
Read blocking domains from a database

DL6ER:

## 10

---

Read blocking domains from a new database generated by `pihole -g` (see https://github.com/pi-hole/pi-hole/pull/2611)


dnsmasq_interface.c
```diff
@@ -1200,77 +1200,72 @@ static void block_single_domain_regex(char *domain)
...
+	fseek(f, 0, SEEK_END);
+
+	if(blockingstatus == BLOCKING_DISABLED)
+	{
+		logg("Skipping %s because blocking is disabled", filename);
 		return cache_size;
+	}
+	else
+	{
+		logg("Importing content of template %s from database", filename);
```
Mcat12: What does "template" mean in this context? If it is just a placeholder "flag" to tell FTL to import from the database, then this message seems to suggest that we're reading from the template instead of the database.

dnsmasq_interface.c
```diff
@@ -1200,77 +1200,72 @@ static void block_single_domain_regex(char *domain)
...
-		   (strstr(domain, " ") != NULL || strstr(domain, "\t") != NULL))
-		{
-			// Reset file pointer back to beginning of the list
-			rewind(f);
-			logg("File %s is in HOSTS format, please run pihole -g!", filename);
-			return name_count;
-		}
-		firstline = false;
-
 		// Skip empty lines
```
Mcat12: This comment should be adjusted to "Skip empty domains", since there is no concept of "lines" in the database.

dnsmasq_interface.h
```diff
@@ -38,3 +38,5 @@ void _FTL_upstream_error(unsigned int rcode, int id, const char* file, const int
 void FTL_dnsmasq_reload(void);
 void FTL_fork_and_bind_sockets(struct passwd *ent_pw);
 int FTL_listsfile(char* filename, unsigned int index, FILE *f, int cache_size, struct crec **rhash, int hashsz);
+
+void FTL_read_gravity(void);
```
Mcat12: This function does not seem to exist.

regex.c
```diff
@@ -164,79 +164,58 @@ void free_regex(void)
...
-	FILE *fp;
-	char *buffer = NULL;
-	size_t size = 0;
-
-	// Get number of lines in the whitelist file
-	whitelist.count = countlines(files.whitelist);
+	// Get number of lines in the whitelist table
+	whitelist.count = gravityDB_count(WHITE_LIST);
 
 	if(whitelist.count < 0)
```
Mcat12: This seems like it should check for equality as well, since it checks if there are any whitelist entries.

regex.c
```diff
@@ -164,79 +164,58 @@ void free_regex(void)
...
 
 	// Start timer for regex compilation analysis
 	timer_start(REGEX_TIMER);
 
-	// Get number of lines in the regex file
-	num_regex = countlines(files.regexlist);
+	// Get number of lines in the regex table
+	num_regex = gravityDB_count(REGEX_LIST);
 
 	if(num_regex < 0)
```
Mcat12: Same issue here as found in whitelist logic.

regex.c
```diff
@@ -248,58 +227,55 @@ void read_regex_from_file(void)
...
 			break;
 
-		// Strip potential newline character at the end of line we just read
-		if(buffer[strlen(buffer)-1] == '\n')
-			buffer[strlen(buffer)-1] = '\0';
-
 		// Skip this entry if empty: an empty regex filter would match
 		// anything anywhere and hence match (and block) all incoming domains.
 		// A user can still achieve this with a filter such as ".*", however
 		// empty lines in regex.list are probably not expected to have such an
```
Mcat12: Comment should be changed as regexes are loaded from the database instead of a file now.

regex.c
```diff
@@ -248,58 +227,55 @@ void read_regex_from_file(void)
...
+			logg("Skipping empty regex filter");
 			skipped++;
 			continue;
 		}
 
-		// Skip this entry if it is commented out
-		if(buffer[0] == '#')
+		// Skip this entry if it is commented out using the legacy syntax
+		// (should use the disabled property of the table now)
+		if(domain[0] == '#')
```
Mcat12: It might be better to convert the `#`-disabled entries into the proper form when we convert from `regex.list` to the `regex` table. Then we do not have to have this legacy code.

regex.c
```diff
@@ -248,58 +227,55 @@ void read_regex_from_file(void)
...
-		// Compile this regex
-		regexconfigured[i] = init_regex(buffer, i);
-	}
-
-	// Free allocated memory
-	if(buffer != NULL)
-	{
-		free(buffer);
-		buffer = NULL;
+		// Copy this whitelisted domain into memory
```
Mcat12: Copy-paste error

regex.c
```diff
@@ -248,58 +227,55 @@ void read_regex_from_file(void)
...
-	fclose(fp);
+	// Finalize statement and close gravity database handle
+	gravityDB_finalizeTable();
 
-	// Read whitelisted domains from file
-	read_whitelist_from_file();
+	// Read whitelisted domains from database
+	read_whitelist_from_database();
 
 	logg("Compiled %i Regex filters and %i whitelisted domains in %.1f msec (%i errors)", (num_regex-skipped), whitelist.count > 0 ? whitelist.count : 0, timer_elapsed_msec(REGEX_TIMER), errors);
```
Mcat12: This method should only handle compiling regex. The regex-loading code (everything but the last two lines) should be moved into a separate function (and rename this function), such that the final function looks like this:
```cpp
read_regex_from_database();
read_whitelist_from_database();

logg(...);
```

routines.h
```diff
@@ -138,7 +138,14 @@ void addOverTimeClientSlot();
 // capabilities.c
 bool check_capabilities(void);
 
+
```
Mcat12: Extra newline?

gravity.c
```diff
@@ -0,0 +1,185 @@
...
+// Prototypes from functions in dnsmasq's source
+void rehash(int size);
+
+static bool gravityDB_open(void)
+{
+	struct stat st;
+	if(stat(FTLfiles.gravitydb, &st) != 0)
+	{
+		// File does not exist
+		if(config.debug & DEBUG_DATABASE) logg("readGravity(): %s does not exist", FTLfiles.gravitydb);
```
Mcat12: Should this be regular-level output? It seems too important to hide in debug.

gravity.c
```diff
@@ -0,0 +1,185 @@
...
+
+bool gravityDB_getTable(unsigned char list)
+{
+	// Read gravity domains
+	gravityDB_open();
+	char *querystr = NULL;
+	switch(list)
+	{
+		case GRAVITY_LIST:
+			querystr = "SELECT * FROM vw_gravity;";
```
Mcat12: If we're just reading domains, it may be better to only select the `domain` column.

gravity.c
```diff
@@ -0,0 +1,185 @@
...
+*  This file is copyright under the latest version of the EUPL.
+*  Please see LICENSE file for your rights under this license. */
+
+#include "FTL.h"
+#include "shmem.h"
+#include <sqlite3.h>
+
+// Private variables
+static sqlite3 *gravitydb = NULL;
+static FILE *gravityfile = NULL;
```
Mcat12: This variable is never assigned.

dnsmasq_interface.c
DL6ER: We need the domain mass-injection machinery that is prepared by `dnsmasq` for when it reads in HOSTS files. Here, we hijack this method and decide that, if the filenames are `/etc/pihole/gravity.list` or `/etc/pihole/black.list` (or whatever the user might have set manually in `pihole-FTL.conf), we ignore the real content of these files and instead read from the database.

This allows us to have the required tools available while still reading efficiently from the database. My fist implementation tried to manually prepare a suitable artificial mass-injection machinery but was very slow as I'd have had to alter too many cache routines in `dnsmasq`. This way of doing it prevents us from having to make *any* changes to the `dnsmasq` code.

Having this in mind, please suggest an alternative message.

regex.c
DL6ER: The response could also have been `DB_FAILED` (which is defined as `-2`), but I agree that this should be separated out into another, more informative, message

dnsmasq_interface.c
Mcat12: My question was just on the word "template" and the message (everything else seems fine). I can't think of a better name now, but I just wanted to understand the reasoning behind the name.

DL6ER:
Setting milestone v5.0

Since we drastically change the way we store data and open up a variety of possibilities in the future, this warrants a major version jump.

regex.c
```diff
@@ -247,35 +254,23 @@ void read_regex_from_database(void)
...
 		// A user can still achieve this with a filter such as ".*", however
-		// empty lines in regex.list are probably not expected to have such an
-		// effect and would immediately lead to "blocking the entire Internet"
+		// empty filters in the regex table are probably not expected to have such
+		// an effect and would immediately lead to "blocking the entire Internet"
 		if(strlen(domain) < 1)
-		{
-			regexconfigured[i] = false;
-			logg("Skipping empty regex filter");
-			skipped++;
```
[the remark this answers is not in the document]
DL6ER: Yes. Empty lines in the file were the reason for this code. There should be no empty lines in the table.

request.c
```diff
@@ -164,8 +164,17 @@ void process_request(char *client_message, int *sock)
...
+		// Free regex list
 		free_regex();
+		// Start timer for regex compilation analysis
+		timer_start(REGEX_TIMER);
+		// Read and compile possible regex filters
 		read_regex_from_database();
+		// Read whitelisted domains from database
+		read_whitelist_from_database();
+		// Log result
+		log_regex_whitelist(timer_elapsed_msec(REGEX_TIMER));
```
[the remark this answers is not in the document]
DL6ER: I removed it from `main.c`

pralor:

https://discourse.pi-hole.net/t/white-blacklist-add-reason-field/1439/11

gravity.c
```diff
@@ -0,0 +1,180 @@
...
+			break;
+		case WHITE_LIST:
+			querystr = "SELECT COUNT(*) FROM vw_whitelist;";
+			break;
+		case REGEX_LIST:
+			querystr = "SELECT COUNT(*) FROM vw_regex;";
+			break;
+		default:
+			logg("gravityDB_count(%i): Requested list is not known!", list);
+			return false;
```
Mcat12: This returns a boolean, but the return type is int.

config.c
```diff
@@ -306,6 +306,9 @@ void read_FTLconf(void)
 	// AUDITLISTFILE
 	getpath(fp, "AUDITLISTFILE", "/etc/pihole/auditlog.list", &files.auditlist);
 
+	// GRAVITYFILE
+	getpath(fp, "GRAVITYFILE", "/etc/pihole/gravity.db", &FTLfiles.gravitydb);
```
Mcat12: GRAVITYFILE is used above for `/etc/pihole/gravity.list`. A new setting must be used, such as `GRAVITYDB`.

dnsmasq_interface.c
```diff
@@ -1196,77 +1204,72 @@ static void block_single_domain_regex(char *domain)
...
+
+	if(blockingstatus == BLOCKING_DISABLED)
+	{
+		logg("Skipping %s because blocking is disabled", filename);
 		return cache_size;
+	}
+	else
+	{
+		logg("Importing content of template %s from database", filename);
+
```
Mcat12: Extra line

gravity.c
```diff
@@ -0,0 +1,180 @@
...
+*  Network-wide ad blocking via your own hardware.
+*
+*  FTL Engine
+*  Gravity database routines
+*
+*  This file is copyright under the latest version of the EUPL.
+*  Please see LICENSE file for your rights under this license. */
+
+#include "FTL.h"
+#include "shmem.h"
```
Mcat12: Why does the database code need to use shared memory?

regex.c
```diff
@@ -164,79 +164,65 @@ void free_regex(void)
 	free_whitelist_domains();
```
Mcat12: This function should be called separately, like what was done for `read_whitelist_from_database`.

Mcat12:
We should branch off a release branch for v4.3 before merging this PR, as this is part of v5.0.

DL6ER:
Yes, I'm missing an option to transform a PR to a draft PR after having opened it as a regular one. I will put `[WIP]` on this PR just to mark it as not being meant to be merged just yet.

config.c
```diff
@@ -288,18 +288,6 @@ void read_FTLconf(void)
...
-	getpath(fp, "WHITELISTFILE", "/etc/pihole/whitelist.txt", &files.whitelist);
-
-	// BLACKLISTFILE
-	getpath(fp, "BLACKLISTFILE", "/etc/pihole/black.list", &files.blacklist);
-
-	// GRAVITYFILE
-	getpath(fp, "GRAVITYFILE", "/etc/pihole/gravity.list", &files.gravity);
-
-	// REGEXLISTFILE
-	getpath(fp, "REGEXLISTFILE", "/etc/pihole/regex.list", &files.regexlist);
```
Mcat12: We will need to remember to update the docs with these changes.

dnsmasq/dnsmasq.h
```diff
@@ -490,13 +490,16 @@ struct crec {
 #define SRC_INTERFACE UID_NONE
 #define SRC_CONFIG    1
 #define SRC_HOSTS     2
-/*----- Pi-hole modification -----*/
-// ID 3 will be used for the regex list file name
-// ID 4 will be used as starting index for any Additional Hosts (AH) files
+/*------------------------- Pi-hole modification -------------------------*/
+// ID 3 will be used for the regex table
+// ID 3 will be used for the gravity table
+// ID 3 will be used for the blacklist table
```
Mcat12: Copy/paste error, ID 3 is written three times.

dnsmasq_interface.c
```diff
@@ -1323,66 +1323,28 @@ static void block_single_domain_regex(const char *domain)
...
 	// Get IPv4/v6 addresses for blocking depending on user configures blocking mode
 	prepare_blocking_mode(&addr4, &addr6, &has_IPv4, &has_IPv6);
-	regexlistname = files.regexlist;
 	add_blocked_domain(&addr4, &addr6, has_IPv4, has_IPv6, domain, strlen(domain), NULL, 0, SRC_REGEX);
 
 	if(config.debug & DEBUG_QUERIES) logg("Added %s to cache", domain);
 }
 
-int FTL_listsfile(const char* filename, unsigned int index, FILE *f, int cache_size, struct crec **rhash, int hashsz)
+static int FTL_table_import(const char *tablename, const unsigned char list, const unsigned int index,
```
Mcat12: Can you add some documentation for this function? (generally a good idea for any function)

dnsmasq_interface.c
```diff
@@ -1415,7 +1377,44 @@ int FTL_listsfile(const char* filename, unsigned int index, FILE *f, int cache_s
...
+		return cache_size;
+	}
+
+	// Import gravity and blacklist domains
+	int added;
+	added  = FTL_table_import("gravity", GRAVITY_LIST, SRC_GRAVITY, addr4, addr6, has_IPv4, has_IPv6, cache_size, rhash, hashsz);
+	added += FTL_table_import("blacklist", BLACK_LIST, SRC_BLACK, addr4, addr6, has_IPv4, has_IPv6, cache_size, rhash, hashsz);
+
+	// Update counter of blocked domains
 	counters->gravity += added;
```
Mcat12: Shouldn't this be an assignment, because we are re-importing gravity? Otherwise, every reload would inflate the size of the counter.

gravity.c
```diff
@@ -55,10 +55,14 @@ static void gravityDB_close(void)
 	sqlite3_close(gravitydb);
 }
 
-bool gravityDB_getTable(unsigned char list)
+bool gravityDB_getTable(const unsigned char list)
 {
-	// Read gravity domains
-	gravityDB_open();
+	// Open gravity database, fail is not possible
+	// Possible if the database has not yet been created by gravity
```
Mcat12: Comments seem contradictory. "fail is not possible" and "Possible if ..."

gravity.c
```diff
@@ -90,18 +94,21 @@ bool gravityDB_getTable(unsigned char list)
 	return true;
 }
 
-const char* gravityDB_getDomain(void)
+inline const char* gravityDB_getDomain(void)
```
Mcat12: Is there a reason for inlining?

dnsmasq_interface.c
DL6ER: We can make this an assignment without changing the effect. Before this function is run, we zero the counter here:

https://github.com/pi-hole/FTL/blob/076c4ef7d5f7de81388f6d79e2ab33477a5149ab/dnsmasq_interface.c#L372

gravity.c
DL6ER: This routine is basically a wrapper for `sqlite3_step()` and `sqlite3_column_text()` (alongside some error checking). It is called extremely often - once for each blocking domain, so it might be called even millions of times if users have huge blocking lists - so overhead should be avoided here. The compiler itself might decide to `inline` this function. I use this keyword here to support this decision and remove the function calling overhead.

dnsmasq_interface.c
```diff
@@ -1328,6 +1328,10 @@ static void block_single_domain_regex(const char *domain)
 	if(config.debug & DEBUG_QUERIES) logg("Added %s to cache", domain);
 }
 
+// Import a specified table form the gravity database and
+// add the read domains to the cache using the currently
+// selected blocking mode. This function is used to import
+// both, the blacklist as well as the gravity blocking domains
```
Mcat12: `form` should be `from`.
The comma does not seem necessary.

gravity.c
```diff
@@ -49,16 +50,20 @@ static bool gravityDB_open(void)
...
-	// Close database handle
 	sqlite3_close(gravitydb);
 }
 
+// Prepare a SQLite3 statement which can be used by
+// gravityDB_getDomain() to get blocking domains from
+// a table which is specified when calling this function
 bool gravityDB_getTable(const unsigned char list)
 {
 	// Open gravity database, fail is not possible
```
Mcat12: Instead of saying "not possible", perhaps say "stop execution if open fails" or nothing at all?

gravity.c
```diff
@@ -0,0 +1,204 @@
...
+			querystr = "SELECT domain FROM vw_gravity;";
+			break;
+		case BLACK_LIST:
+			querystr = "SELECT domain FROM vw_blacklist;";
+			break;
+		case WHITE_LIST:
+			querystr = "SELECT domain FROM vw_whitelist;";
+			break;
+		case REGEX_LIST:
+			querystr = "SELECT filter FROM vw_regex;";
```
Mcat12: The column in the regex table and view is now called `domain`.

gravity.c
```diff
@@ -0,0 +1,204 @@
...
+		sqlite3_close(gravitydb);
+		return DB_FAILED;
+	}
+
+	// Get result when there was no error
+	const int result = sqlite3_column_int(stmt, 0);
+
+	// Finalize statement and close database handle
+	sqlite3_finalize(stmt);
+	gravityDB_close();
```
Mcat12: This is the same as `gravityDB_finalizeTable()`. (general issue)

gravity.c
```diff
@@ -0,0 +1,204 @@
...
+		sqlite3_free(zErrMsg);
+		sqlite3_close(gravitydb);
+		return false;
+	}
+
+	return true;
+}
+
+// Close gravity database handle
+static void gravityDB_close(void)
```
Mcat12: This is only used in two places, and one of those usages can be replaced by the other. Since this only has one statement and (soon to be only) one usage, it should be removed and the contents moved to where it's used.

gravity.c
DL6ER: I seemingly forgot to hit `git push` after my most recent `git commit`. This is already fixed.

gravity.c
DL6ER: I replaced these two calls by a single call to `gravityDB_finalizeTable()`

gravity.c
DL6ER: I agree, I removed this subroutine.

Mcat12:
I'm seeing an issue now where blocked domains are not counted:
```
$ dig doubleclick.com @localhost

; <<>> DiG 9.10.3-P4-Ubuntu <<>> doubleclick.com @localhost
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 15332
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;doubleclick.com.		IN	A

;; ANSWER SECTION:
doubleclick.com.	2	IN	A	0.0.0.0

;; Query time: 0 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Sat Apr 27 22:20:23 UTC 2019
;; MSG SIZE  rcvd: 60
```

image

DL6ER:
@Mcat12 This issue has been fixed in https://github.com/pi-hole/FTL/pull/492/commits/f9feab492f709940ca7784ca10e999acee866b66

config.c
DL6ER: They aren't even mentioned on the `docs` page.

PR #493, opened 2019-02-03
Panic (crash) if we are unable to create or resize shared memory

Mcat12:

---

If we are unable to create or resize shared memory then there is no point in continuing execution. Further interactions with shared memory may result in crashes which are hard to debug. As a result of this change, there is no need to check if the returned shared memory object pointer is null, because if the function succeeded, the operation was successful.

### DIFF
--- a/shmem.c
+++ b/shmem.c
@@ -105,11 +105,6 @@ void newOverTimeClient(int clientID) {
 	// Create the shared memory with enough space for the current overTime slots
 	shm_unlink(name);
 	SharedMemory shm = create_shm(name, (counters->overTime/pagesize + 1)*pagesize*sizeof(int));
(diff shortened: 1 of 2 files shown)


dschaper:
LGTM, but that's just from a visual overview.

DL6ER:
Not removing the source branch as it might be used by others for testing.

pralor:

https://discourse.pi-hole.net/t/crash-after-update-to-4-2/16769/10

pralor:

https://discourse.pi-hole.net/t/crash-after-update-to-4-2/16769/13

Mcat12:
Removing branch as it is outdated, and the binaries will remain on ftl.pi-hole.net if users are still using them.

PR #497, opened 2019-02-05
Pi-hole FTL v4.2.1

DL6ER:
Release fixing two issues:

- Blocking status display for `NODATA` and `NXDOMAIN` blocking modes
- File permission issues when using shared memory on systems that lack Linux capabilities support

It also adds in a new feature:
- Report both the user `pihole-FTL` **started** as and the user `pihole-FTL` **crashed** as in case it crashes and generated a crash ticket in `pihole-FTL.log`

### DIFF
--- a/dnsmasq/config.h
+++ b/dnsmasq/config.h
@@ -43,8 +43,8 @@
 #define HOSTSFILE "/etc/hosts"
 #define ETHERSFILE "/etc/ethers"
 #define DEFLEASE 3600 /* default lease time, 1 hour */
-#define CHUSER "nobody"
-#define CHGRP "dip"
+#define CHUSER "root"
+#define CHGRP "root"
 #define TFTP_MAX_CONNECTIONS 50 /* max simultaneous connections */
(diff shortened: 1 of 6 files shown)


dnsmasq/config.h
```diff
@@ -43,8 +43,8 @@
 #define HOSTSFILE "/etc/hosts"
 #define ETHERSFILE "/etc/ethers"
 #define DEFLEASE 3600 /* default lease time, 1 hour */
-#define CHUSER "nobody"
-#define CHGRP "dip"
+#define CHUSER "root"
+#define CHGRP "root"
```
Mcat12: Should this be noted as a FTL change?

pralor:

https://discourse.pi-hole.net/t/pi-hole-v4-2-available-with-shared-memory-new-blocking-modes-and-more/16763/10

pralor:

https://discourse.pi-hole.net/t/hotfix-release-of-pi-hole-v4-2-2/17472/1

PR #510, opened 2019-02-15
Pi-hole FTL v4.2.2

DL6ER:
This release simplifies the handling of the `overTime` data structures. These structures store the data of the queries over time (permitted / blocked) as well as the client-specific data needed to compute the following two graphics on the dashboard:

screenshot from 2019-02-15 17-46-56

To allow for this simplification, we fixated the length rolling to 24 hours. The option `MAXLOGAGE` can still be loaded to specify how much should be read from the database but can not be used any longer to, e.g., enlarge the displayed window. The change was long overdue and is necessary to keep the code maintainable. Some users have been suffering from a breakdown of the previously used fully dynamic memory handling. This release fixes these errors.


config.c
```diff
@@ -196,13 +196,13 @@ void read_FTLconf(void)
...
-	// defaults to: 24.0
-	config.maxlogage = 24*3600;
+	// defaults to: 24.0 via MAXLOGAGE defined in FTL.h
+	config.maxlogage = MAXLOGAGE*3600;
 	buffer = parse_FTLconf(fp, "MAXLOGAGE");
 
 	fvalue = 0;
 	if(buffer != NULL && sscanf(buffer, "%f", &fvalue))
-		if(fvalue >= 0.0f && value <= 744.0f)
+		if(fvalue >= 0.0f && fvalue <= 1.0f*MAXLOGAGE)
```
Mcat12: The documentation for `MAXLOGAGE` should be updated with the new maximum value.

Mcat12: Will be resolved with pi-hole/docs#91

PR #519, opened 2019-02-22
Pi-hole FTL v4.2.3

### LINKED ISSUE #496: FTL Crash

3

---

FTL should not crash

FTL crashed

- Running pihole after upgrade to v4.2 makes it crash after some time.
- It has crashed 3 times over 2 hour period after upgrade. Every time with crash in the same place.

```[2019-02-04 20:46:00.478] !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
[2019-02-04 20:46:00.478] ---------------------------->  FTL crashed!  <----------------------------
[2019-02-04 20:46:00.478] !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
[2019-02-04 20:46:00.478] Please report a bug at https://github.com/pi-hole/FTL/issues
[2019-02-04 20:46:00.479] and include in your report already the following details:

[2019-02-04 20:46:00.479] FTL has been running for 384 seconds
[2019-02-04 20:46:00.479] FTL branch: master
[2019-02-04 20:46:00.479] FTL version: v4.2
[2019-02-04 20:46:00.479] FTL commit: 178cc52
[2019-02-04 20:46:00.479] FTL date: 2019-02-03 08:42:40 -0800
[2019-02-04 20:46:00.479] FTL user: pihole
[2019-02-04 20:46:00.480] Received signal: Segmentation fault
[2019-02-04 20:46:00.480]      at address: 3065831424
[2019-02-04 20:46:00.480]      with code: SEGV_MAPERR (Address not mapped to object)
[2019-02-04 20:46:00.481] Backtrace:
[2019-02-04 20:46:00.482] B[0000]: /usr/bin/pihole-FTL(+0x1f860) [0x45c860]
[2019-02-04 20:46:00.482] B[0001]: /lib/arm-linux-gnueabihf/libc.so.6(__default_rt_sa_restorer+0) [0xb6d676c0]
[2019-02-04 20:46:00.482] B[0002]: /usr/bin/pihole-FTL(FTL_new_query+0x540) [0x467f00]
[2019-02-04 20:46:00.482] B[0003]: /usr/bin/pihole-FTL(receive_query+0x6c8) [0x47f7f4]
[2019-02-04 20:46:00.482] B[0004]: /usr/bin/pihole-FTL(+0x56dc0) [0x493dc0]
[2019-02-04 20:46:00.482] B[0005]: /usr/bin/pihole-FTL(main_dnsmasq+0xe7c) [0x495820]
[2019-02-04 20:46:00.483] B[0006]: /usr/bin/pihole-FTL(main+0xd4) [0x45a778]
[2019-02-04 20:46:00.483] B[0007]: /lib/arm-linux-gnueabihf/libc.so.6(__libc_start_main+0x114) [0xb6d51678]
[2019-02-04 20:46:00.483] Thank you for helping us to improve our FTL engine!
[2019-02-04 20:46:00.483] FTL terminated!
```

**Device specifics**

Hardware Type: rPi, VPS, etc
OS: Linux raspberrypi 4.14.79+ #1159 Sun Nov 4 17:28:08 GMT 2018 armv6l GNU/Linux

```
root@raspberrypi:/var/log# pihole version
  Pi-hole version is v4.2.1 (Latest: v4.2.1)
  AdminLTE version is v4.2 (Latest: v4.2)
  FTL version is v4.2 (Latest: v4.2)
```
```
Your debug token is: dcpx20ghx8
```

Thanks for great software 👍

DL6ER:
**This release fixes #496**

Issue
---

DNS requests and responses can happen over UDP or TCP. UDP is the most common, but TCP is used often for large responses or by some specific products (we have seen this with Netflix in the past). Handling TCP connections requires more communication with the client and server, so `pihole-FTL` creates TCP helper **processes** by forking.

When dynamically allocated shared memory objects get resized, they get resized, and then re-mapped. The problem is that if our shared memory object is now significantly larger and if this happens on a TCP handler process, the other processes were not notified about the virtual memory change keep using the old shared memory, i.e., they were using dangling pointers. This caused a memory corruption or access error when accessed.

Solution
---
This pull requests ensures that, when the space of virtual addresses changes, the other forks get notified and can re-map the shared memory objects to get possibly updated virtual memory addresses for their pointers.

### DIFF
--- a/shmem.c
+++ b/shmem.c
@@ -67,6 +67,10 @@ unsigned long long addstr(const char *str)
 	   !realloc_shm(&shm_strings, shm_strings.size + pagesize, true))
 		return 0;
 
+	// Store new string buffer size in corresponding counters entry
+	// for re-using when we need to re-map shared memory objects
+	counters->strings_MAX = shm_strings.size;
+
 	// Copy the C string pointed by str into the shared string buffer
 	strncpy(&((char*)shm_strings.ptr)[next_pos], str, len);
 	((char*)shm_strings.ptr)[next_pos + len] = '\0';
@@ -113,7 +117,7 @@ void remap_shm(void)
 	realloc_shm(&shm_domains, counters->domains_MAX*sizeof(domainsDataStruct), false);
 	realloc_shm(&shm_clients, counters->clients_MAX*sizeof(clientsDataStruct), false);
 	realloc_shm(&shm_forwarded, counters->forwarded_MAX*sizeof(forwardedDataStruct), false);
-	realloc_shm(&shm_strings, counters->strings_MAX*sizeof(char), false);
+	realloc_shm(&shm_strings, counters->strings_MAX, false);
 
 	// Update local counter to reflect that we absorbed this change
 	local_shm_counter = shmSettings->global_shm_counter;
@@ -125,6 +129,10 @@ void _lock_shm(const char* function, const int line, const char * file) {
 
 	if(debug) logg("Waiting for lock in %s() (%s:%i)", function, file, line);
 
+	int result = pthread_mutex_lock(&shmLock->lock);
+
+	if(debug) logg("Obtained lock for %s() (%s:%i)", function, file, line);
+
 	// Check if this process needs to remap the shared memory objects
 	if(shmSettings != NULL &&
 	   local_shm_counter != shmSettings->global_shm_counter)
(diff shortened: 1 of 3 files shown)


shmem.c
```diff
@@ -346,8 +406,9 @@ bool realloc_shm(SharedMemory *sharedMemory, size_t size) {
 	}
 
 	// Close shared memory object file descriptor as it is no longer
-	// needed after having called mmap()
-	close(fd);
+	// needed after having called ftruncate()
+	if(resize)
+		close(fd);
```
Mcat12: The file descriptor could be closed at the end of the first if block where we resize the memory, instead of waiting until here to close it. It's only used in a logging statement outside of the if block (and doesn't need to be there).

PR #522, opened 2019-02-24
Align overTime shared memory to struct size

Mcat12:

---

The overTime shared memory size is now a multiple of the struct size, which is required for the API to safely connect to shared memory.

### DIFF
--- a/shmem.c
+++ b/shmem.c
@@ -195,7 +195,7 @@ bool init_shmem(void)
 	counters->queries_MAX = pagesize;
 
 	/****************************** shared overTime struct ******************************/
-	size_t size = ((OVERTIME_SLOTS*sizeof(overTimeDataStruct))/pagesize + 1)*pagesize;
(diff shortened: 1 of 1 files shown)


DL6ER:
Implemented LCM algorithm as discussed. The API seems to have the hard requirement of not allowing any unused memory to be part of the shared memory object. By using the LCM, we can, at at least, reduce the memory requirements to 52K instead of 213KB. This is the minimum number that allows perfect page-alignment.

It will fail hard if the computed memory footprint exceeds the size returned by the LCM computing routine. This can only happen for `pagesize < 1024` which is something we will probably not encounter. However, if you want to add a fallback, feel free to do so.

Note that we might want to reuse this computation also for the `clients` structure later, since we currently allocate always enough memory for 4096 clients which most users will never even come close to. This could be part of a follow-up PR if we think it is worthwhile.

PR #524, opened 2019-02-28
Use LCM to compute optimal allocation step for dynamic shared memory objects

DL6ER:

## 10

---

Due to the requirement of having to be perfectly page-aligned for the API, we used to allocate the shared memory objects in steps of `pagesize`. While this seems meaningful for both, `queries` and `domains`, it doesn't seem useful for `clients` and `forwarded` as rarely any user should see more than 1000 entries in them.

This PR uses the recently added LCM calculator to compute the optimal (minimum) amount of memory that ensures perfect page-alignment.

- `clients`: Allocation step is now 512 instead of 4096. Memory initially needed by this object is now 324KB instead of 2.6MB. The effect is especially notable for `clients` as each client contains its private `overTime` structure which is consuming quite a bit of memory (648 bytes per client).

- `forwarded`: Allocation step is now 512 instead of 4096. Memory initially needed by this object is now 20KB instead of 160KB. The difference is smaller as the `forwarded` data is only 40 bytes each.

### DIFF
--- a/shmem.c
+++ b/shmem.c
@@ -230,9 +230,10 @@ bool init_shmem(void)
 
 	/****************************** shared forwarded struct ******************************/
 	// Try to create shared memory object
-	shm_forwarded = create_shm(SHARED_FORWARDED_NAME, pagesize*sizeof(forwardedDataStruct));
+	size = get_optimal_object_size(sizeof(forwardedDataStruct));
+	shm_forwarded = create_shm(SHARED_FORWARDED_NAME, size*sizeof(forwardedDataStruct));
 	forwarded = (forwardedDataStruct*)shm_forwarded.ptr;
-	counters->forwarded_MAX = pagesize;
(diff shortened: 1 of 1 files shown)


shmem.c
```diff
@@ -476,5 +485,5 @@ static size_t gcd(size_t a, size_t b)
 // in the shared memory object
 static size_t get_optimal_object_size(size_t objsize)
 {
-	return (pagesize*objsize)/gcd(pagesize, objsize);
+	return pagesize / gcd(pagesize, objsize);
```
DL6ER: Note that I removed the multiplication by `objsize` here to get the `number of struct elements` to be allocated instead of the number of bytes (`(number of struct elements)*(size of each element in bytes)`). We multiply this number by `sizeof ...` where it is necessary to get the number in Bytes.

PR #527, opened 2019-03-02
Fix shm crash caused by an insufficient obtained optimal structure size

DL6ER:

## 10

---

Under certain circumstances, it happened that the obtained optimal (= minimum) page-aligned was too small to hold the entire `overTime` array. Although there was a check for this, we only logged that there is an issue but a `exit(1);` was missing so FTL kept running. This inevitably lead to memory corruption.

We extend this by adding an additional parameter with which we can specify how many slots we want to have *at least*.


Mcat12:
Note that this is only a bug in development

PR #529, opened 2019-03-06
Prevent direct access to global objects queries, clients, domains, and forward

DL6ER:

## 10

---

This change removes the global objects:
- `queries`
- `clients`
- `domains`
- `forwarded`

They are now all only be available through corresponding getter functions that will return a pointer to exactly one specific entry. Each cal to a pointer getter is accompanied by corresponding boundary checks. We will have to evaluate if - and if how much - effect this has on FTL's performance.

This change does not affect the shared memory the API uses.

### DIFF
--- a/shmem.c
+++ b/shmem.c
@@ -37,10 +37,10 @@ static SharedMemory shm_overTime = { 0 };
 static SharedMemory shm_settings = { 0 };
 
 // Variable size array structs
-static queriesDataStruct *queries = NULL;
-static clientsDataStruct *clients = NULL;
-static domainsDataStruct *domains = NULL;
-static forwardedDataStruct *forwarded = NULL;
+static queriesData *queries = NULL;
+static clientsData *clients = NULL;
+static domainsData *domains = NULL;
+static forwardedData *forwarded = NULL;
 
 typedef struct {
 	pthread_mutex_t lock;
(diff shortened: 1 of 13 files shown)


shmem.c
```diff
@@ -511,3 +517,139 @@ static size_t get_optimal_object_size(size_t objsize, unsigned int minsize)
...
+		break;
+		default:
+			/* That cannot happen */
+			logg("Fatal error in memory_check(%i)", which);
+			exit(EXIT_FAILURE);
+		break;
+	}
+}
+
+queriesData* _getQuery(int queryID, bool checkMagic, int line, const char * function, const char * file)
```
Mcat12: This code is very repetitive across the `_get*` functions, could it be generalized into a function?

Makefile
```diff
@@ -66,7 +66,7 @@ CCFLAGS=-std=gnu11 -I$(IDIR) $(WARNFLAGS) -D_FILE_OFFSET_BITS=64 $(HARDENING_FLA
 # for dnsmasq we need the nettle crypto library and the gmp maths library
 # We link the two libraries statically. Although this increases the binary file size by about 1 MB, it saves about 5 MB of shared libraries and makes deployment easier
 #LIBS=-pthread -lnettle -lgmp -lhogweed
-LIBS=-pthread -Wl,-Bstatic -L/usr/local/lib -lhogweed -lgmp -lnettle -Wl,-Bdynamic -lrt -lcap
+LIBS=-pthread -lrt -lcap #-Wl,-Bstatic -L/usr/local/lib -lhogweed -lgmp -lnettle -Wl,-Bdynamic -lrt -lcap
```
Mcat12: Are these changes accidental?

PR #532, opened 2019-03-07
Follow-up on #527

DL6ER:

---

This PR is a follow-up on #527 and adds:

- Output result of optimum size calculation when `DEBUG_SHMEM` is enabled
- Ensure `pihole-FTL.conf` and crash handler are set up before we create shared memory objects (so that the `DEBUG_*` values are already known early on during initialization)
- Log which size has been used we used the multiplier

This PR fixes a bug that is only present in branch `development`.

### DIFF
--- a/shmem.c
+++ b/shmem.c
@@ -516,7 +516,7 @@ static unsigned int get_optimal_object_size(unsigned int objsize, unsigned int m
 	{
 		if(config.debug & DEBUG_SHMEM)
 		{
-			logg("DEBUG: LCM(%i, %zu) == %zu >= %zu",
+			logg("DEBUG: LCM(%i, %u) == %u >= %u",
 			     pagesize, objsize,
 			     optsize*objsize,
 			     minsize*objsize);
--- a/shmem.c
+++ b/shmem.c
@@ -46,7 +46,7 @@ static ShmSettings *shmSettings = NULL;
 static int pagesize;
 static unsigned int local_shm_counter = 0;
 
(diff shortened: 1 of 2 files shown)


shmem.c
```diff
@@ -459,7 +459,7 @@ void delete_shm(SharedMemory *sharedMemory)
 // Recursive function to return GCD of a and b
 // Credits: https://www.geeksforgeeks.org/program-to-find-lcm-of-two-numbers/
 // The code in this link has been modified by the Pi-hole developers
-static size_t gcd(size_t a, size_t b)
+static unsigned int gcd(unsigned int a, unsigned int b)
```
Mcat12: Can this be switched back to `size_t`, as we have found that `size_t` is not the issue? Same with the other usages.

DL6ER: I'd actually like to keep it as `unsigned int` as I don't see the need for `size_t` here. The numbers that are involved in these calculations are all small as we only calculate the allocation *step* not the entire object size.

`pagesize` is (implementation-wise) even a `signed int` and the other partner will always be `sizeof(someStruct)` which is expected to be an even smaller number, i.e., the returned value can, at max, be something like `pagesize^2 == 16777216` which is much much smaller than `UINT_MAX`.

Mcat12: To the contrary, I'd like to keep it as `size_t` as I don't see the need for `unsigned int` here. There is no benefit to use `unsigned int` over `size_t` for this use case, and the latter will always be performant (https://stackoverflow.com/a/4295225). There is no performance gain by using a smaller integer size here.

PR #536, opened 2019-03-10
Further general improvements

DL6ER:

## 10

---

This PR adds further general improvements for our code base.

Highlights:
- Added many additional warning flags that are useful and are neither included by `-Wall` and `-Wextra`
- Declare many `char*` which should not be changeable in the corresponding subroutines as `const char*` to have the compiler enforce their immutability.
- Enforce strict prototyping for FTL routines. This ensures that functions have to be *either* made explicitly global (by adding a prototype to the suitable header file), or they have to be made explicitly local (by adding the `static` keyword). Furthermore, strict prototyping requires that functions with zero arguments need to be written explicitly as `function(void)`. Non-strict prototype such as `function()` are permitted by regular `C`, however they in fact allow an arbitrary amount of parameters (they are a function declaration and not a prototype).
- As before, we split the warnings into a `gcc` version 8 and higher and a CI-compatible section which is automatically selected. This will be helpful for users with sufficiently recent `gcc` to get better hints whereas compilation is still possible on the CI.

Additional note: There are changes to `dnsmasq` code in this PR which are necessary to be able to compile *FTL*DNS with the added warning flags. The changes are meaningful and will be submitted to `dnsmasq` separately to be already part of the next release of `dnsmasq`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,12 @@ EXTRAWARN=-Werror -Waddress -Wlogical-op -Wmissing-field-initializers -Woverleng
 # -Wbad-function-cast: Warn when a function call is cast to a non-matching type
 # -Wcast-align=strict: Warn whenever a pointer is cast such that the required alignment of the target is increased. For example, warn if a "char *" is cast to an "int *" regardless of the target machine.
 # -Wwrite-strings: When compiling C, give string constants the type "const char[length]" so that copying the address of one into a non-"const" "char *" pointer produces a warning
-EXTRAWARN2=-Wduplicated-branches -Wduplicated-cond -Wfloat-equal -Wunsafe-loop-optimizations -funsafe-loop-optimizations -Wbad-function-cast -Wcast-align=strict -Wwrite-strings -Wconversion -Wparentheses
+# -Wparentheses: Warn if parentheses are omitted in certain contexts, such as when there is an assignment in a context where a truth value is expected, or when operators are nested whose precedence people often get confused about
+# -Wlogical-op: Warn about suspicious uses of logical operators in expressions
+# -Wlogical-not-parentheses: Warn about logical not used on the left hand side operand of a comparison
+# -Wstrict-prototypes: Warn if a function is declared or defined without specifying the argument types
+# -Wmissing-prototypes: Warn if a global function is defined without a previous prototype declaration
+EXTRAWARN2=-Wduplicated-branches -Wduplicated-cond -Wfloat-equal -Wunsafe-loop-optimizations -funsafe-loop-optimizations -Wbad-function-cast -Wcast-align=strict -Wwrite-strings -Wparentheses -Wlogical-op -Wlogical-not-parentheses -Wstrict-prototypes -Wmissing-prototypes
 # -FILE_OFFSET_BITS=64: used by stat(). Avoids problems with files > 2 GB on 32bit machines
 CCFLAGS=-std=gnu11 -I$(IDIR) $(WARNFLAGS) -D_FILE_OFFSET_BITS=64 $(HARDENING_FLAGS) $(DEBUG_FLAGS) $(CFLAGS) $(SQLITEFLAGS)
 # for FTL we need the pthread library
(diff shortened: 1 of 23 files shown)


grep.c
```diff
@@ -45,25 +45,6 @@ int countlines(const char* fname)
 	return lines;
 }
 
-int readnumberfromfile(const char* fname)
```
DL6ER: This function is not used anywhere

log.c
```diff
@@ -63,7 +63,7 @@ void get_timestr(char *timestring)
 
 void __attribute__ ((format (gnu_printf, 1, 2))) logg(const char *format, ...)
 {
-	char timestring[32] = "";
+	char timestring[84] = "";
```
DL6ER: This change ensures that the string will be large enough to hold anything that might be produced by
https://github.com/pi-hole/FTL/pull/536/files#diff-f8cdfce5cea83a80a9847ba558991869R61

main.c
```diff
@@ -14,9 +14,6 @@ char * username;
 bool needGC = false;
 bool needDBGC = false;
 
-// Prototype
-int main_dnsmasq(int argc, char **argv);
```
DL6ER: Redundant prototype, already contained in `routines.h`

routines.h
```diff
@@ -131,7 +132,7 @@ void newOverTimeClient(int clientID);
  * Add a new overTime slot to each overTime client shared memory block.
  * This also updates `overTimeClientData`.
  */
-void addOverTimeClientSlot();
+void addOverTimeClientSlot(void);
```
DL6ER: Changed function declaration to a full prototype

Mcat12:
> Furthermore, strict prototyping requires that functions with zero arguments need to be written explicitly as function(void). Non-strict prototype such as function() are permitted by regular C, however they in fact allow an arbitrary amount of parameters (they are a function declaration and not a prototype).

Is this not caught by the compiler at least as a warning?

Makefile
```diff
@@ -70,6 +70,28 @@ else
...
+# -Wbad-function-cast: Warn when a function call is cast to a non-matching type
+# -Wwrite-strings: When compiling C, give string constants the type "const char[length]" so that copying the address of one into a non-"const" "char *" pointer produces a warning
+# -Wparentheses: Warn if parentheses are omitted in certain contexts, such as when there is an assignment in a context where a truth value is expected, or when operators are nested whose precedence people often get confused about
+# -Wlogical-op: Warn about suspicious uses of logical operators in expressions
+# -Wstrict-prototypes: Warn if a function is declared or defined without specifying the argument types
+# -Wmissing-prototypes: Warn if a global function is defined without a previous prototype declaration
+# -Wredundant-decls: Warn if anything is declared more than once in the same scope
+# -Winline: Warn if a function that is declared as inline cannot be inlined
+EXTRAWARN2=-Wfloat-equal -Wunsafe-loop-optimizations -funsafe-loop-optimizations -Wbad-function-cast -Wwrite-strings -Wparentheses -Wlogical-op -Wstrict-prototypes -Wmissing-prototypes -Wredundant-decls -Winline $(EXTRAWARNGCC8)
```
Mcat12: Can this be combined with `EXTRAWARN`?

DL6ER:
> Is this not caught by the compiler at least as a warning?

It is now enforced by the addition of `-Wstrict-prototypes -Wmissing-prototypes`. These warnings are neither enabled by default nor included in `-Wall -Wextra` in `gcc`.

args.c
```diff
@@ -24,8 +24,8 @@ void parse_args(int argc, char* argv[])
 	// Regardless of any arguments, we always pass "-k" (nofork) to dnsmasq
 	argc_dnsmasq = 2;
 	argv_dnsmasq = calloc(argc_dnsmasq, sizeof(char*));
-	argv_dnsmasq[0] = "";
-	argv_dnsmasq[1] = "-k";
+	argv_dnsmasq[0] = (char*)"";
+	argv_dnsmasq[1] = (char*)"-k";
```
Mcat12: Why does this cast need to exist? What other type could a string literal be?

DL6ER: `""` is `const char[1]`. However, `argv_dnsmasq` is `char*`.

```c++
argv_dnsmasq[0] = "";
```
results in the compilation error
```
error: assignment discards ‘const’ qualifier from pointer target type
```

Makefile
DL6ER: Yes, the only reason for separating was that `EXTRAWARN` is already quite long...

args.c
Mcat12: Can we change `argv_dnsmasq` to `const char**` then?

Makefile
Mcat12: Makefiles support splitting lines:
https://www.gnu.org/software/make/manual/html_node/Splitting-Lines.html

PR #548, opened 2019-04-05
Improvements for hostname resolving routines

DL6ER:

## 10

---

This PR adds a few micro-optimizations and a bug fix for the host name resolving routines.

Improvement:
- Do not add host names of clients and upstream servers if they haven't changed.

Bug fix:
- Any determined host name was allocated using `strdup()`. However, as a consecutive `free()` was missing, some memory got lost.

### DIFF
--- a/resolve.c
+++ b/resolve.c
@@ -67,8 +67,10 @@ static char *resolveHostname(const char *addr)
 static size_t resolveAndAddHostname(size_t ippos, size_t oldnamepos)
 {
 	// Get IP and host name strings
+	lock_shm();
 	const char* ipaddr = getstr(ippos);
 	const char* oldname = getstr(oldnamepos);
+	unlock_shm();
 
 	// Important: Don't hold a lock while resolving as the main thread
 	// (dnsmasq) needs to be operable during the call to resolveHostname()
@@ -79,7 +81,9 @@ static size_t resolveAndAddHostname(size_t ippos, size_t oldnamepos)
 	{
 		// We do not need to check for name == NULL as name is
 		// always initialized with an empty string at position 0
+		lock_shm();
 		size_t newnamepos = addstr(newname);
+		unlock_shm();
 		if(newname != NULL)
 			free(newname);
 		return newnamepos;
--- a/resolve.c
+++ b/resolve.c
@@ -99,10 +99,10 @@ void resolveClients(bool onlynew)
 		lock_shm();
 
 		// Only store new hostname if it changed
+		// We do not need to check for name == NULL as name is
+		// always initialized with an empty string at position 0
 		if(hostname != NULL && strcmp(name, hostname) != 0)
 		{
-			// We do not need to check for name == NULL as name is
-			// always initialized with an empty string at position 0
 			clients[clientID].namepos = addstr(hostname);
 		}
 		else if(config.debug & DEBUG_SHMEM)
@@ -112,7 +112,8 @@ void resolveClients(bool onlynew)
 		}
 
 		// Release allocated memory
-		free(hostname);
+		if(hostname != NULL)
+			free(hostname);
 
 		// Mark entry as not new
 		clients[clientID].new = false;
@@ -150,10 +151,10 @@ void resolveForwardDestinations(bool onlynew)
 		// Finally, lock data when storing obtained hostname
 		lock_shm();
 		// Only store new hostname if it changed
+		// We do not need to check for name == NULL as name is
+		// always initialized with an empty string at position 0
 		if(hostname != NULL && strcmp(name, hostname) != 0)
 		{
-			// We do not need to check for name == NULL as name is
-			// always initialized with an empty string at position 0
 			forwarded[forwardID].namepos = addstr(hostname);
 		}
 		else if(config.debug & DEBUG_SHMEM)
@@ -163,7 +164,8 @@ void resolveForwardDestinations(bool onlynew)
 		}
 
 		// Release allocated memory
-		free(hostname);
+		if(hostname != NULL)
+			free(hostname);
 
 		// Mark entry as not new
 		forwarded[forwardID].new = false;
--- a/resolve.c
+++ b/resolve.c
@@ -110,12 +110,18 @@ void resolveClients(bool onlynew)
 
 		// If onlynew flag is set, we will only resolve new clients
 		// If not, we will try to re-resolve all known clients
-		if(onlynew && !clients[clientID].new)
+		lock_shm();
+		bool newlock = clients[clientID].new;
+		unlock_shm();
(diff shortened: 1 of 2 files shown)


DL6ER:
Please do not consider this PR a blocker. If 4.3 will be released before this PR has been tested, feel free to rebase this PR onto `development`.

resolve.c
```diff
@@ -88,16 +88,36 @@ void resolveClients(bool onlynew)
...
 
 		// Finally, lock data when storing obtained hostname
 		lock_shm();
-		clients[clientID].namepos = addstr(hostname);
+
+		// Only store new hostname if it changed
+		if(hostname != NULL && strcmp(name, hostname) != 0)
+		{
+			// We do not need to check for name == NULL as name is
+			// always initialized with an empty string at position 0
```
Mcat12: This comment would make better sense on the if statement instead of its body.

resolve.c
```diff
@@ -88,16 +88,36 @@ void resolveClients(bool onlynew)
...
+			clients[clientID].namepos = addstr(hostname);
+		}
+		else if(config.debug & DEBUG_SHMEM)
+		{
+			// Debugging output
+			logg("Not adding \"%s\" to buffer (unchanged)", name);
+		}
+
+		// Release allocated memory
+		free(hostname);
```
Mcat12: The hostname could be null, so this would try to free a null pointer.

See line 67.

resolve.c
```diff
@@ -119,17 +139,36 @@ void resolveForwardDestinations(bool onlynew)
...
+		{
+			// We do not need to check for name == NULL as name is
+			// always initialized with an empty string at position 0
+			forwarded[forwardID].namepos = addstr(hostname);
+		}
+		else if(config.debug & DEBUG_SHMEM)
+		{
+			// Debugging output
+			logg("Not adding \"%s\" to buffer (unchanged)", name);
+		}
```
Mcat12: This code is almost duplicated, maybe it is possible to extract it into a function.

DL6ER: Is this comment meant to only spawn the `logg()` routine or dis you mention this with respect to the whole functions? If it is the latter, then yes, I agree. However, I don't think it is enough code to be worth putting into a common function that would then also need some pointer magic to be able to fill the `namepos` fields of once the `clients` and once the `forwarded` data structures.

resolve.c
DL6ER: Our own implementation of `free()` is robust against trying to free `NULL` and will only log a warning into FTL's log. However, I agree that this should be fixed and did this now.

resolve.c
Mcat12: The code from line 139 to 172 (excluding line 169) can be easily extracted into a function with the following signature (the name is up for debate):
```c
size_t resolveAndAddHostname(size_t ippos, size_t namepos);
```

And would be used in the following way (example given is for forward destinations):
```c
forwarded[forwardID].namepos = resolveAndAddHostname(
    forwarded[forwardID].ippos,
    forwarded[forwardID].namepos
);

forwarded[forwardID].new = false;
```

If the hostname has changed then the output of the function would be the output of `addstr`. If it has not changed, then the output would be the original `namepos`.

resolve.c
```diff
@@ -71,6 +63,37 @@ static char *resolveHostname(const char *addr)
...
 
+// Resolve upstream destination host names
+static size_t resolveAndAddHostname(size_t ippos, size_t oldnamepos)
+{
+	// Get IP and host name strings
+	const char* ipaddr = getstr(ippos);
+	const char* oldname = getstr(oldnamepos);
+
+	// Important: Don't hold a lock while resolving as the main thread
+	// (dnsmasq) needs to be operable during the call to resolveHostname()
```
Mcat12: This function needs to lock shared memory just as it did before it was a function. Memory is read by `getstr` and written to by `addstr`, but should not be locked when calling `resolveHostname`.

DL6ER: The string buffer is only every appended so `getstr()` can be safely called without memory actions taking place during its run. You are right, however, as the string buffer might get extended just when we run `getstr()` which is something we obviously want to avoid.

resolve.c
```diff
@@ -71,6 +63,41 @@ static const char *resolveHostname(const char *addr)
...
+	// Only store new newname if it changed
+	if(newname != NULL && strcmp(oldname, newname) != 0)
+	{
+		// We do not need to check for name == NULL as name is
+		// always initialized with an empty string at position 0
+		lock_shm();
+		size_t newnamepos = addstr(newname);
+		unlock_shm();
+		if(newname != NULL)
+			free(newname);
```
Mcat12: At this point, `newname` will not be null so this check is not needed.

resolve.c
```diff
@@ -71,6 +63,41 @@ static const char *resolveHostname(const char *addr)
...
+
+	// Important: Don't hold a lock while resolving as the main thread
+	// (dnsmasq) needs to be operable during the call to resolveHostname()
+	char* newname = resolveHostname(ipaddr);
+
+	// Only store new newname if it changed
+	if(newname != NULL && strcmp(oldname, newname) != 0)
+	{
+		// We do not need to check for name == NULL as name is
+		// always initialized with an empty string at position 0
```
Mcat12: This comment got moved back into the if statement body.

resolve.c
```diff
@@ -85,20 +113,20 @@ void resolveClients(bool onlynew)
...
+			lock_shm();
+			clients[clientID].namepos = newnamepos;
+			unlock_shm();
+		}
 
-		// Finally, lock data when storing obtained hostname
-		lock_shm();
-		clients[clientID].namepos = addstr(hostname);
+		// Mark entry as not new
 		clients[clientID].new = false;
```
Mcat12: This should be wrapped in a lock (and the forwarded call below).

Sorry for the third review!

resolve.c
```diff
@@ -85,18 +113,20 @@ void resolveClients(bool onlynew)
 		if(onlynew && !clients[clientID].new)
```
Mcat12: I think all of the references to `clients` and `forwarded` might need a lock.

resolve.c
```diff
@@ -71,66 +63,133 @@ static const char *resolveHostname(const char *addr)
...
-		clients[clientID].namepos = addstr(hostname);
-		clients[clientID].new = false;
-		unlock_shm();
+		// Mark entry as not new even when we don't have a new host name
+		if(newflag)
+		{
+			lock_shm();
+			clients[clientID].new = false;
+			unlock_shm();
+		}
```
Mcat12: Try:
```
Lock
If newpos != oldpos
    namepos = newpos
new = false
Unlock
```

Issue #552, opened 2019-04-18
Can’t compile on musl based Alpine Linux 


debiansid:
**In raising this issue, I confirm the following (please check boxes, eg [X]) Failure to fill the template will close your issue:**

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [] The issue I am reporting can be *replicated*
- [x] The issue I am reporting isn't a duplicate 

**How familiar are you with the codebase?:**

_{1}_

---

**[BUG | ISSUE] Expected Behaviour:**
make 

**[BUG | ISSUE] Actual Behaviour:**
lack of backtrace on musl, it failed.

**[BUG | ISSUE] Steps to reproduce:**

-
-
-
-

**Log file output [if available]**

```
usr/lib/gcc/x86_64-alpine-linux-musl/8.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: obj/signals.o: in function `SIGSEGV_handler':
/home/ezhang/FTL-4.2.3/signals.c:45: undefined reference to `backtrace'
/usr/lib/gcc/x86_64-alpine-linux-musl/8.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: /home/user/FTL-4.2.3/signals.c:46: undefined reference to `backtrace_symbols'
collect2: error: ld returned 1 exit status
make: *** [Makefile:101: pihole-FTL] Error 1
```
**Device specifics**

Hardware Type: rPi, VPS, etc
OS:Alpine Linux 

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

debiansid:
https://github.com/zuhd-org/easyloggingpp/issues/540

DL6ER:
There seems to be no easy solution for this.

- We could use `libunwind`, however, this would add a new dependency. The portability demands for `pihole-FTL` are enormous, one can hardly imagine on how many different devices users run Pi-holes. 
- We could statically link `libunwind` to avoid adding an external dependency, however, a quick search brought up just too many issue tickets related to statically linked `libunwind`.
- We could just remove the backtrace symbol derivation for non-`glibc` builds. This would cause symbol names to be missing in self-generated backtraces, however, it should hopefully allow to compile `pihole-FTL` *at all*, so it seems to be the favorable option here. Note that we can still log the return addressed from the stack frames corresponding to the function calls of the process.

[...]

PR #553, opened 2019-04-18
Improve musl support + run tests on musl and 32bit binaries

DL6ER:

## 10

---
According to #552, MUSL does neinter support `backtrace()` nor `backtrace_symbols()`. We only try to obtain a backtrace when `pihole-FTL` is compiled for `glibc` (which is almost always the case) but skip backtrace generation with MUSL binaries. The reason for lacking a backtrace is logged in the same place where the backtrace is shown in "usual" binaries.


DL6ER:
Just for clarification: This PR still awaits the patch @DDoSolitary submitted to `dnsmasq` being merged to `dnsmasq` so we can cherry-pick it into `FTL`.

dschaper:
The `dnsmasq` mailing list was pinged on the patch yesterday to get it some attention.

DL6ER:
Due to the lack of activity on the `dnsmasq` mailing list, I decided to merge the patch of @DDoSolitary and deal with possible merge conflicts when they arise on the next `dnsmasq` release. This PR is now ready for review and merge.

signals.c
```diff
@@ -29,36 +31,38 @@ static void SIGSEGV_handler(int sig, siginfo_t *si, void *unused)
...
-		}
+		logg("B[%04i]: %p, %s", j, buffer[j],
+		     bcktrace != NULL ? bcktrace[j] : "---");
 	}
-	free(bcktrace);
-
+	if(bcktrace != NULL)
+		free(bcktrace);
+#else
+	logg("!!! INFO: pihole-FTL has not been compiled for glibc !!!");
```
Mcat12: This should have something about not being compiled with backtrace support, because on its own this message does not say what would have happened if it had been compiled with glibc. Maybe something more like:
```
!!! INFO: pihole-FTL has not been compiled with backtrace support, not generating one !!!
```

DL6ER:
This seems to generate a working `musl` binary:
```
/ # file FTL/pihole-FTL 
pihole-FTL: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, with debug_info, not stripped

/ # ldd FTL/pihole-FTL 
	/lib/ld-musl-x86_64.so.1 (0x7f0ecf02f000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f0ecf02f000)

/ # FTL/pihole-FTL -b
tweak/musl_support
```

docker/x86_64-musl/Dockerfile
```diff
@@ -1,12 +1,3 @@
-FROM debian:stretch
+FROM alpine:edge
 
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends nettle-dev musl-tools \
-        make file wget netcat-traditional sqlite3 git ca-certificates ssh
-
-# Install ghr for GitHub Releases: https://github.com/tcnksm/ghr
-RUN wget https://github.com/tcnksm/ghr/releases/download/v0.12.0/ghr_v0.12.0_linux_amd64.tar.gz && \
```
Mcat12: Do we want to keep this in the image?

docker/x86_64-musl/Dockerfile
```diff
@@ -1,12 +1,3 @@
...
-    tar -xzf ghr_*_linux_amd64.tar.gz && \
-    mv ghr_*_linux_amd64/ghr /usr/bin/ghr
-
-ENV CC musl-gcc
+RUN apk add --no-cache alpine-sdk linux-headers gmp-dev nettle-dev
```
Mcat12: Will everything work even without `linux-headers`, i.e. on a fresh Alpine install?

Mcat12:
Also, shouldn't the musl libc be statically linked?

docker/x86_64-musl/Dockerfile
DL6ER: No, leaving off `linux-headers` results in
```
dnsmasq/dnsmasq.h:144:10: fatal error: linux/capability.h: No such file or directory
 #include <linux/capability.h>
          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Mcat12: I mean, at run time. Will the binary execute?

DL6ER: It seems to contain only header files so it shouldn't be required at run time.

https://pkgs.alpinelinux.org/contents?repo=main&page=1&arch=x86&branch=edge&name=linux-headers

docker/x86_64-musl/Dockerfile
DL6ER: Are we actually using it?

Mcat12: We are not currently, but at some point we planned on using it. It's in all of the other images, so might as well keep it for consistency. We can later determine if we want to remove it from all of the images.

DL6ER:
Note: The updated `Dockerfile`s also reduce the size of the `ftl-build` containers:
```
old x86_64 = 499MB
new x86_64 = 376MB

old x86_64-musl = 341MB
new x86_64-musl = 234MB
```

test/test_suite.bats
```diff
@@ -92,56 +92,56 @@
...
-  [[ ${lines[2]} == "1 1 version.ftl" ]]
-  [[ ${lines[3]} == "2 1 version.bind" ]]
-  [[ ${lines[4]} == "3 1 whitelisted.com" ]]
-  [[ ${lines[5]} == "4 1 regexa.com" ]]
-  [[ ${lines[6]} == "5 1 ftl.pi-hole.net" ]]
+  [[ "${lines[@]}" == *" 1 version.ftl"* ]]
+  [[ "${lines[@]}" == *" 1 version.bind"* ]]
+  [[ "${lines[@]}" == *" 1 whitelisted.com"* ]]
+  [[ "${lines[@]}" == *" 1 regexa.com"* ]]
+  [[ "${lines[@]}" == *" 1 ftl.pi-hole.net"* ]]
```
DL6ER: Here and below: It is not meaningful to assume a particular order here as the values are sorted before output. It is unpredictable in which order they may come out. While this has always been the same when compiling for `glibc`, the new `musl` build reveals that another library may have a different interpretation here.

test/test_suite.bats
```diff
@@ -179,54 +179,54 @@
...
-  [[ ${lines[2]} == *"TXT version.bind 127.0.0.1 3 0 6"* ]]
-  [[ ${lines[3]} == *"A blacklisted.com 127.0.0.1 5 0 4"* ]]
-  [[ ${lines[4]} == *"A 0427d7.se 127.0.0.1 1 0 4"* ]]
-  [[ ${lines[5]} == *"A whitelisted.com 127.0.0.1 2 0 4"* ]]
-  [[ ${lines[6]} == *"A regex5.com 127.0.0.1 4 0 4"* ]]
-  [[ ${lines[7]} == *"A regexa.com 127.0.0.1 2 0 7"* ]]
-  [[ ${lines[8]} == *"A google.com 127.0.0.1 2 0 4"* ]]
-  [[ ${lines[9]} == *"AAAA google.com 127.0.0.1 2 0 4"* ]]
-  [[ ${lines[10]} == *"A ftl.pi-hole.net 127.0.0.1 2 0 4"* ]]
+  [[ ${lines[1]} == *"TXT version.ftl 127.0.0.1 "*"3 0 6"* ]]
```
DL6ER: Here and below: Acknowledge that there might be a host name after the IP address of the client.

test/test_suite.bats
```diff
@@ -283,7 +283,8 @@
 @test "Ownership and permissions of pihole-FTL.db correct" {
   run bash -c 'ls -l /etc/pihole/pihole-FTL.db'
   printf "%s\n" "${lines[@]}"
-  [[ ${lines[0]} == "-rw-r--r-- 1 pihole pihole"* ]]
+  [[ ${lines[0]} == *"pihole pihole"* || ${lines[0]} == *"pihole   pihole"* ]]
```
DL6ER: `x86_64-musl` is built on `busybox` which has a slightly different variant of `ls` displaying three, instead of one, spaces between the user and group names.

Mcat12:
Can you add the comments you made on GitHub as comments in the tests?

docker/x86_32/Dockerfile
```diff
@@ -3,11 +3,16 @@ FROM debian:stretch
...
 RUN wget https://github.com/tcnksm/ghr/releases/download/v0.12.0/ghr_v0.12.0_linux_amd64.tar.gz && \
-    tar -xzf ghr_*_linux_amd64.tar.gz && \
+    tar -xzf ghr_*_linux_amd64.tar.gz && rm ghr_*_linux_amd64.tar.gz && \
     mv ghr_*_linux_amd64/ghr /usr/bin/ghr
 
 ENV CC "gcc -m32"
+
+#RUN git clone https://github.com/pi-hole/FTL && cd FTL
+#RUN cd FTL && make -j 4
+#RUN cd FTL && test/run.sh
```
Mcat12: These lines should be removed (and in the other dockerfiles)

docker/x86_64-musl/Dockerfile
```diff
@@ -1,12 +1,19 @@
-FROM debian:stretch
+FROM alpine:edge
```
Mcat12: Is it possible to remain on debian?

DL6ER: I have found no working configuration with `musl-gcc` on `debian`.

Mcat12:
```
$ ldd pihole-FTL-musl-linux-x86_64 
	linux-vdso.so.1 (0x00007ffe6efd7000)
	libc.musl-x86_64.so.1 => not found
```

The musl binary is not statically linked to the musl libc.

PR #554, opened 2019-04-19
Use thread-safe alternatives of ctime and localtime

DL6ER:

## 10

---

Use thread-safe alternatives of `ctime()` and `localtime()`. 

`ctime()` and `localtime()` fill data into a `char` array resp. `tm` struct in global memory and then return a pointer to that memory. If the function is called from multiple places in the same program, and especially if it is called from multiple threads in the same program, then the calls will overwrite each other's data which we want to avoid.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -45,7 +45,8 @@ void *GC_thread(void *val)
 			if(config.debug & DEBUG_GC) timer_start(GC_TIMER);
 
 			int removed = 0;
-			if(config.debug & DEBUG_GC) logg("GC starting, mintime: %lu %s", mintime, ctime(&mintime));
+			char ctimebuffer[32]; // need at least 26 bytes according to ctime_r man page
+			if(config.debug & DEBUG_GC) logg("GC starting, mintime: %lu %s", mintime, ctime_r(&mintime, ctimebuffer));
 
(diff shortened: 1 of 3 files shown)


gc.c
```diff
@@ -45,7 +45,8 @@ void *GC_thread(void *val)
 			if(config.debug & DEBUG_GC) timer_start(GC_TIMER);
 
 			int removed = 0;
-			if(config.debug & DEBUG_GC) logg("GC starting, mintime: %lu %s", mintime, ctime(&mintime));
+			char ctimebuffer[32]; // need at least 26 bytes according to ctime_r man page
+			if(config.debug & DEBUG_GC) logg("GC starting, mintime: %lu %s", mintime, ctime_r(&mintime, ctimebuffer));
```
Mcat12: `ctime_r` seems to be marked obsolete:
> POSIX.1-2008 marks asctime(), asctime_r(), ctime(), and ctime_r() as obsolete, recommending the use of strftime(3) instead.

https://linux.die.net/man/3/ctime_r

DL6ER: I replaced `ctime` by our own routine we already use when generating the string representation for the `pihole-FTL.log` file

PR #577, opened 2019-05-26
Identify clients by hardware address not hardware + IP address

### LINKED ISSUE #576: Network table should only use hardware address to uniquely identify clients

## 10

---

**Preface**

Sequential DHCP servers may hand out different addresses to the same devices if they are connected in a different order. Pi-hole's own DHCP server tries to hand out the same addresses every time to the same devices (based on their hardware addresses).

The network table should uniquely identify clients based on their hardware address (MAC).

The network tables uses both the IP and the MAC to identify clients, causing identical devices that happen to have received different IP addresses as separate devices.

- Use sequential DHCP server (`pihole-FTL` can do this when the setting `dhcp-sequential-ip` is used)
- Connect various devices in different chronological order
- Observe that the same devices are added multiple times

**Solution**

Only use MAC address to identify clients

DL6ER:

## 10

---

This PR fixes #576. Further details can be found therein.

If multiple clients (with the same `hwaddr` but different IP addresses) have been added in the past, they will be fused to unique entries.

This PR bumps the database version to 4 without changing the structure. The version update is merely to signal that the unification of clients in the `network` table has been done on the given database.

### DIFF
--- a/database.c
+++ b/database.c
@@ -269,10 +269,10 @@ void db_init(void)
 		dbversion = db_get_FTL_property(DB_VERSION);
 	}
 
-	//Update to version 4 if lower
+	// Update to version 4 if lower
 	if(dbversion < 4)
 	{
-		// Update to version 3: Create network table
+		// Update to version 4: Unify clients in network table
 		logg("Updating long-term database to version 4");
 		unify_hwaddr(db);
 		// Get updated version
--- a/database.c
+++ b/database.c
@@ -22,6 +22,9 @@ static pthread_mutex_t dblock;
 static bool db_set_counter(const unsigned int ID, const int value);
 static int db_get_FTL_property(const unsigned int ID);
 
+// defined in networktable.c
+extern bool unify_hwaddr(sqlite3 *db);
+
 static bool check_database(int rc)
 {
 	// We will retry if the database is busy at the moment
@@ -235,6 +238,7 @@ void db_init(void)
 		database = false;
 		return;
 	}
+
 	// Update to version 2 if lower
 	if(dbversion < 2)
 	{
@@ -249,6 +253,7 @@ void db_init(void)
 		// Get updated version
 		dbversion = db_get_FTL_property(DB_VERSION);
 	}
+
 	// Update to version 3 if lower
 	if(dbversion < 3)
 	{
@@ -260,6 +265,16 @@ void db_init(void)
 			database = false;
 			return;
 		}
+		// Get updated version
+		dbversion = db_get_FTL_property(DB_VERSION);
+	}
+
+	//Update to version 4 if lower
+	if(dbversion < 4)
+	{
+		// Update to version 3: Create network table
+		logg("Updating long-term database to version 4");
+		unify_hwaddr(db);
 		// Get updated version
 		dbversion = db_get_FTL_property(DB_VERSION);
 	}
--- a/networktable.c
+++ b/networktable.c
@@ -99,7 +99,7 @@ void parse_arp_cache(void)
 		// commitment. Read-only access such as this SELECT command will be
 		// executed immediately on the database.
 		char* querystr = NULL;
-		int ret = asprintf(&querystr, "SELECT id FROM network WHERE hwaddr = \'%s\' LIMIT 1;", hwaddr);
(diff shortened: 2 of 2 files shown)


networktable.c
```diff
@@ -86,15 +87,19 @@ void parse_arp_cache(void)
...
 
-		// Get ID of this device in our network database. If it cannot be found, then this is a new device
-		// We match both IP *and* MAC address
-		// Same MAC, two IPs: Non-deterministic DHCP server, treat as two entries
-		// Same IP, two MACs: Either non-deterministic DHCP server or (almost) full DHCP address pool
-		// We can run this SELECT inside the currently active transaction as only the
-		// changed to the database are collected for latter commitment. Read-only access
-		// such as this SELECT command will be executed immediately on the database.
+		// Get ID of this device in our network database. If it cannot be
+		// found, then this is a new device We only use the hardware address
```
Mcat12: Missing period.

networktable.c
```diff
@@ -196,6 +208,93 @@ void parse_arp_cache(void)
...
+
+	// Ensure hwaddr is a unique field
+	// Unfortunately, SQLite's ALTER TABLE does not support adding
+	// constraints to existing tables. However, we can add a unique
+	// index for the table to achieve the same effect.
+	//
+	// See https://www.sqlite.org/lang_createtable.html#constraints:
+	// >>> In most cases, UNIQUE and PRIMARY KEY constraints are
+	// >>> implemented by creating a unique index in the database.
+	dbquery("CREATE UNIQUE INDEX hwaddr_idx ON network(hwaddr)");
```
Mcat12: The index name should be further namespaced in case we want to make a hwaddr index on another table. A better name would be `network_hwaddr_idx`

database.c
```diff
@@ -264,6 +269,16 @@ void db_init(void)
 		dbversion = db_get_FTL_property(DB_VERSION);
 	}
 
+	//Update to version 4 if lower
+	if(dbversion < 4)
+	{
+		// Update to version 3: Create network table
```
Mcat12: This comment is incorrect

networktable.c
```diff
@@ -196,6 +208,93 @@ void parse_arp_cache(void)
...
+	}
+
+	// Finalize statement and free query string
+	sqlite3_finalize(stmt);
+	free(querystr);
+
+	// Ensure hwaddr is a unique field
+	// Unfortunately, SQLite's ALTER TABLE does not support adding
+	// constraints to existing tables. However, we can add a unique
+	// index for the table to achieve the same effect.
```
Mcat12: What do you think of changing the table to use the MAC address as the primary key?

DL6ER: The `network` table has already been created for most. `ALTER TABLE` cannot be used to change the `PRIMARY KEY` and I don't think it would be worth creating a new and importing the old database table.

The current `id` column comes at no costs, as a column type `INTEGER PRIMARY KEY` is a mere alias for the `ROWID` which is always present (we're not using `AUTOINCREMENT` here).

database.c
```diff
@@ -22,6 +22,9 @@ static pthread_mutex_t dblock;
 static bool db_set_counter(const unsigned int ID, const int value);
 static int db_get_FTL_property(const unsigned int ID);
 
+// defined in networktable.c
+extern bool unify_hwaddr(sqlite3 *db);
```
Mcat12: Would this be better in a header?

networktable.c
```diff
@@ -15,6 +15,7 @@
 
 // Private prototypes
 static char* getMACVendor(const char* hwaddr);
+bool unify_hwaddr(sqlite3 *db);
```
Mcat12: If this is not used in this file, does it need to appear here?

database.c
DL6ER: It cannot be written in `routines.h` as the datatype `sqlite3` is not known in this file. I'd like to refrain from including `sqlite3.h` in each file by adding it through `routines.h`.

networktable.c
DL6ER: We enforce prototyping for all functions. Omitting the prototype would cause a compilation error. Typically, the prototype would be included through `routines.h`. Due to the reasons mentioned above (type `sqlite3` is unknown in `routines.h`), I added the prototype here.

PR #580, opened 2019-05-31
Add support for DESTDIR

### LINKED ISSUE #558: Support DESTDIR for "make install"

_1_

---

`make DESTDIR=foo install` installs the executable to the directory `foo`.

`DESTDIR` is ignored.

- `make`
- `make DESTDIR=foo install`

N/A

**Device specifics**

Hardware Type: All
OS: All

DL6ER:

## 10

---

The `DESTDIR` variable is specified by the user on the make command line as an absolute file name. For example:
```
sudo make DESTDIR=/tmp/stage install
```
Where the installation step would normally install `/usr/bin/pihole-FTL`, then an installation invoked as in the example above would install `/tmp/stage/usr/bin/pihole-FTL` instead. The `DESTDIR` path is created, if necessary. Note that `root` capabilities are nonetheless required for the `install` step as only `root` can add the required capabilities to the `pihole-FTL` binary.

This PR fixes #558

Prepending the variable `DESTDIR` to each target in this way provides for staged installs, where the installed files are not placed directly into their expected location but are instead copied into a temporary location (the `DESTDIR`). However, installed files maintain their relative directory structure and any embedded file names will not be modified following https://www.gnu.org/prep/standards/html_node/DESTDIR.html.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,5 +157,6 @@ prefix=/usr
 # other requirements (correct ownership of files, etc.) is managed by
 # the service script on sudo service pihole-FTL (re)start
 install: pihole-FTL
-	install -m 0755 pihole-FTL $(prefix)/bin
-	/sbin/setcap CAP_NET_BIND_SERVICE,CAP_NET_RAW,CAP_NET_ADMIN+eip $(prefix)/bin/pihole-FTL
+	mkdir -p $(DESTDIR)$(prefix)/bin
+	install -m 0755 pihole-FTL $(DESTDIR)$(prefix)/bin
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -157,5 +157,6 @@ prefix=/usr
 # other requirements (correct ownership of files, etc.) is managed by
 # the service script on sudo service pihole-FTL (re)start
 install: pihole-FTL
-	install -m 0755 pihole-FTL $(prefix)/bin
-	/sbin/setcap CAP_NET_BIND_SERVICE,CAP_NET_RAW,CAP_NET_ADMIN+eip $(prefix)/bin/pihole-FTL
+	mkdir -p $(DESTDIR)$(prefix)/bin
```
DL6ER: We previously assumed `/usr/bin` to always exist. With a custom `DESTDIR`, we can not be sure any longer if this is the case and, hence, we make parent directories as needed.

PR #585, opened 2019-06-07
Reinstall CI tests for FTL

DL6ER:

## 10
---

Reinstall automated tests in the build queue, they have been disabled since we migrated frm Travis to Circle.

We create a `gravity.db` file with a predefined set of queries (partially done by the tests themselves) to get a constant number of queries so we can compare the API replies across code modifications.

Some (more meaningless) tests have been replaced by some "active" tests that actually query some domains to actually test the DNS server.

This PR is meant as a starting point to reinstate testing. Further tests may be added at any time after this PR has been merged.


DL6ER:
@Mcat12 Yes, the idea is to silence the warning you mentioned. A fresh Pi-hole installation will provide a (basically empty) template so providing an empty file seems fine in terms of mocking a "real" Pi-hole environment. In contrast to the tests we ran on the Travis infrastructure, the files are put in appropriate places (like `/etc/pihole`, `/var/run/pihole`, etc.).

.circleci/config.yml
```diff
@@ -28,6 +28,10 @@ version: 2
           [[ "$CIRCLE_PR_NUMBER" == "" ]] && ./FTL-client "${FOLDER}" "${BIN_NAME}.sha1" "${FTL_SECRET}"
           rm ./FTL-client
           ls -lah .
+    - run:
+        name: "Test"
```
Mcat12: Can this be put before the upload step?

test/run.sh
```diff
@@ -1,77 +1,49 @@
...
 
-dnsmasq_pre() {
-  echo -n $(date +"%b %e %H:%M:%S")
-  echo -n "dnsmasq[123]:"
-}
+# Only run tests on x86_64 target
+if [[ "${1}" != "pihole-FTL-linux-x86_64" ]]; then
+  echo "Skipping tests (${1})!"
+  exit 0
+fi
```
Mcat12: Can this check be done only if `CI` is set to `true`? Then it is easier to run the tests locally.

test/run.sh
```diff
@@ -1,77 +1,49 @@
...
+# Install necessary additional components for testing
+apt-get -qq install dnsutils -y > /dev/null
 
-cat <<EOT >> pihole-FTL.conf
-DBFILE=pihole-FTL.db
-LOGFILE=pihole-FTL.log
-SOCKETFILE=pihole-FTL.sock
-EOT
+# Create necessary directories
+mkdir -p /etc/pihole /var/run/pihole /var/log
```
Mcat12: Instead of `mkdir`, `mv`, `cp`, etc, can we use `install` with explicit permissions?

test/run.sh
```diff
@@ -1,77 +1,49 @@
...
+
+# Prepare setupVars.conf
+echo "BLOCKING_ENABLED=true" > /etc/pihole/setupVars.conf
+
+# Prepare pihole-FTL.conf
+echo "" > /etc/pihole/pihole-FTL.conf
 
 # Start FTL
-./pihole-FTL travis-ci
+if ! ./pihole-FTL-linux-x86_64; then
```
Mcat12: This should use the given parameter instead of hardcoding the binary name.

test/run.sh
```diff
@@ -1,77 +1,49 @@
...
-cat <<EOT >> pihole-FTL.conf
-DBFILE=pihole-FTL.db
-LOGFILE=pihole-FTL.log
-SOCKETFILE=pihole-FTL.sock
-EOT
+# Create necessary directories
+mkdir -p /etc/pihole /var/run/pihole /var/log
+
+# Download prepared gravity database
+wget https://ftl.pi-hole.net/gravity.db
```
Mcat12: Please change this to a SQL file in this repo which is used to generate the test database. There are a lot of downsides to hosting a test database like this:
- It requires us to remember to keep a database in ftl.pi-hole.net, which is not meant for hosting test data
- It is not version controlled
- The database as it is now is not reproducible.
- If the tests or database code are modified in a feature branch in a way that is incompatible with the hosted test database, the branch will fail the build because of the incompatibility.
- The tests are not reproducible in an offline system.

test/run.sh
```diff
@@ -1,77 +1,49 @@
...
+  exit 1
+fi
 
 # Prepare BATS
 mkdir -p test/libs
-git submodule add https://github.com/sstephenson/bats test/libs/bats
-git submodule add https://github.com/ztombol/bats-support test/libs/bats-support
-# git submodule add https://github.com/ztombol/bats-assert test/libs/bats-assert
+git clone --depth=1 https://github.com/sstephenson/bats test/libs/bats > /dev/null
+git clone --depth=1 https://github.com/ztombol/bats-support test/libs/bats-support > /dev/null
```
Mcat12: What is the reason for changing from submodule to clone? If changing to clone, we should also remove the submodule files.

test/run.sh
```diff
@@ -1,77 +1,49 @@
...
-n=0
-until [ $n -ge 45 ]; do
-  nc -vv -z -w 30 127.0.0.1 4711 && break
-  n=$[$n+1]
-  echo "..."
-  tail -n2 pihole-FTL.log
-  echo "..."
-  sleep 1
-done
+sleep 2
```
Mcat12: What is the reason for using a static sleep instead of the retry mechanism?

test/test_suite.sh
```diff
@@ -1,253 +1,276 @@
...
+  [[ ${lines[4]} == "hash "* ]]
+  [[ ${lines[5]} == "date "* ]]
+  [[ ${lines[6]} == "" ]]
+}
+
+@test "Blacklisted domain is blocked" {
+  run bash -c "dig blacklisted.com @127.0.0.1 +short"
+  printf "%s\n" "${lines[@]}"
+  [[ ${lines[0]} == "0.0.0.0" ]]
+  [[ ${lines[1]} == "" ]]
```
Mcat12: What is the significance of the second line being blank?

test/run.sh
Mcat12: This may not be necessary, so don't take it as a must do.

.circleci/config.yml
DL6ER: The test will only be run for the `x86_64` binary. I intentionally put it after the Upload step so even if the tests fail, we upload the binary. If we put the tests in front of the Upload step, then all binaries would still be uploaded, but the `x86_64` would either be missing or would stay on a previous version.

test/run.sh
DL6ER: I'm not sure if `install` can evn be used when you just want to create directories but don't have files to put into. Note that this is not a real system where permissions are important. The environment is meant to be killed after testing.

test/run.sh
DL6ER: What do you mean by "the submodule files"? I changed from submodule to clone because the submodule command didn't actually populate the directories. I have not investigated why but saw that a shallow clone was much faster than the submodule call and serves the same purpose.

test/run.sh
DL6ER: Code simplification. The startup of `pihole-FTL` is very fast (takes only a few dozen milliseconds) on CircleCI.

test/test_suite.sh
DL6ER: I added this everywhere to exclude that there is *more* output than just this. Example in this case would be if there are several `A` records for this domain. Try, e.g.,
```
$ dig ebay.de +short
66.135.196.249
66.135.195.175
```
We want the test to fail if **more** than only `0.0.0.0` is returned.

test/run.sh
Mcat12: Install can be used to just create directories, using the `-d` option. I'll resolve this though.

test/run.sh
Mcat12: The files in this directory:
https://github.com/pi-hole/FTL/tree/master/test/libs

test/run.sh
Mcat12: :+1: If it causes flaky tests in the future (FTL for some reason stalls and isn't ready), we can add it back then.

test/test_suite.sh
Mcat12: Would something like this work better?
```bash
output=$(dig blacklisted.com @127.0.0.1 +short)
{
    echo "0.0.0.0"
}>expected
[[ "${output}" == "${expected}" ]]
```

Mcat12: This makes sure that there is no more output beyond the last empty line.

test/gravity.db.schema
```diff
@@ -0,0 +1,94 @@
...
+(
+	property TEXT PRIMARY KEY,
+	value TEXT NOT NULL
+);
+INSERT INTO info VALUES('version','1');
+DELETE FROM sqlite_sequence;
+INSERT INTO sqlite_sequence VALUES('adlists',1);
+INSERT INTO sqlite_sequence VALUES('blacklist',1);
+INSERT INTO sqlite_sequence VALUES('whitelist',1);
+INSERT INTO sqlite_sequence VALUES('regex',1);
```
Mcat12: These were probably added by the dump command, if you used that, but are they necessary?

test/gravity.db.schema
```diff
@@ -0,0 +1,94 @@
+PRAGMA foreign_keys=OFF;
```
Mcat12: We probably want this on.

test/test_suite.sh
DL6ER: This would be possible as well. I wanted to use the same syntax everywhere. The current way of checking each line individually has the advantage that `bats` will tell us exactly which line isn't equal lessening the work when comparing the outputs. This is obviously no issue here, but, e.g., for the Top Lists tests.

test/gravity.db.schema
DL6ER: They aren't necessary, however, as you assumed, they have been produced by the `dump` command and I left them in there to reproduce the original `gravity.db` file as accurate as possible. Reading https://www.sqlite.org/fileformat2.html#seqtab it seems like we should keep them as the database would not be fully determined / inconsistent without them.

test/test_suite.sh
Mcat12: Will bats output the data when the test case fails? If so, then it shouldn't be that big of a deal.

test/gravity.db.schema
DL6ER: This also comes 1:1 from the `.dump` command. We are not using `foreign` keys anywhere.

test/test_suite.sh
DL6ER: Yes, `bats` will output the data but we will need to check manually which lines mismatch. Again, this is more a question of if we want to use the same checking mechanism in all tests.

test/gravity.db.schema
Mcat12: If we do start using foreign keys, we can enable this at that time.

test/test_suite.sh
```diff
@@ -1,253 +1,276 @@
...
+  [[ "${lines[@]}" == *"CREATE TABLE network ( id INTEGER PRIMARY KEY NOT NULL, ip TEXT NOT NULL, hwaddr TEXT NOT NULL, interface TEXT NOT NULL, name TEXT, firstSeen INTEGER NOT NULL, lastQuery INTEGER NOT NULL, numQueries INTEGER NOT NULL,macVendor TEXT);"* ]]
   [[ "${lines[@]}" == *"CREATE INDEX idx_queries_timestamps ON queries (timestamp);"* ]]
+  [[ "${lines[@]}" == *"CREATE UNIQUE INDEX network_hwaddr_idx ON network(hwaddr);"* ]]
 }
 
-@test "Arguments check: Invalid option" {
-  run bash -c './pihole-FTL abc'
-  echo "output: ${lines[@]}"
+@test "Fail on invalid argument" {
+  run bash -c './pihole-FTL-linux-x86_64 abc'
```
Mcat12: Can we use some variable instead of hardcoding the name? This makes local testing harder, as the binary is not named this way by default.

test/test_suite.sh
Mcat12: The outputs are all rather short. We should try to keep the mechanism the same (although exceptions can be made), so this is more a question of if this method is better than the other.

DL6ER: The method you suggested wouldn't work for `@test "pihole-FTL.db schema as expected"`. I vote for keeping what we already have.

Mcat12: That one is an exception, which is allowed.

We can keep what we have, but we just need to remember that the tests could miss output after the empty line.

DL6ER: Ah, `bats` doesn't allow empty lines. If the output is
```
A

B




C
```
Then we will have
```
output[0] == "A"
output[1] == "B"
output[2] == "C"
output[3] == ""
```

test/gravity.db.schema
Mcat12: Can this file be renamed to have a `.sql` extension?

test/test_suite.sh
```diff
@@ -1,253 +1,276 @@
 #!./test/libs/bats/bin/bats
```
Mcat12: The bat documentation recommends to use the `bats` extension:
https://github.com/bats-core/bats-core#usage

test/run.sh
```diff
@@ -1,77 +1,48 @@
...
+  echo "pihole-FTL failed to start"
+  exit 1
+fi
 
 # Prepare BATS
 mkdir -p test/libs
-git submodule add https://github.com/sstephenson/bats test/libs/bats
-git submodule add https://github.com/ztombol/bats-support test/libs/bats-support
-# git submodule add https://github.com/ztombol/bats-assert test/libs/bats-assert
+git clone --depth=1 https://github.com/sstephenson/bats test/libs/bats > /dev/null
```
Mcat12: This repo is the current home of bats, so it should be used:
https://github.com/bats-core/bats-core#background

test/run.sh
```diff
@@ -1,77 +1,48 @@
...
+git clone --depth=1 https://github.com/ztombol/bats-support test/libs/bats-support > /dev/null
```
Mcat12: Is this used?

.circleci/config.yml
```diff
@@ -28,6 +28,11 @@ version: 2
           [[ "$CIRCLE_PR_NUMBER" == "" ]] && ./FTL-client "${FOLDER}" "${BIN_NAME}.sha1" "${FTL_SECRET}"
           rm ./FTL-client
           ls -lah .
+    - run:
```
Mcat12: In the upload step, can this be added as the first command?
```
[ -z "$FTL_SECRET" ] && exit 0
```

This would make local testing via `circleci local execute --job x86_64` easier, as the step would not fail when uploading. This is already used in the API's Circle config.

(I commented on this line because GitHub won't let me comment on the upload step)

DL6ER: Would this require the Test step to be run before the Upload step?

Mcat12: No, this command will just skip the upload when running locally, as the secret is missing.

Mcat12: Also, as we rename the binary, the `mv pihole-FTL "${BIN_NAME}"` command will have to be before this command.

PR #588, opened 2019-06-10
Tests: Run as user pihole

DL6ER:

## 10

---

Run tests as user `pihole`. This requires a number of preparations similar to what we do in FTL's `service` file.

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -7,10 +7,21 @@ if [[ ${CI} == "true" && "${CIRCLE_JOB}" != "x86_64" ]]; then
 fi
 
 # Install necessary additional components for testing
-apt-get -qq install dnsutils -y > /dev/null
+apt-get -qq install dnsutils libcap2-bin -y > /dev/null
 
-# Create necessary directories
+# Create pihole user
+useradd -m -s /bin/bash pihole
+
+# Create necessary directories and files
 mkdir -p /etc/pihole /var/run/pihole /var/log
+touch /var/log/pihole-FTL.log
+chown pihole:pihole /etc/pihole /var/run/pihole /var/log/pihole-FTL.log
+
+# Copy binary into a location the new user pihole can access
+cp ./pihole-FTL /home/pihole
+chmod +x /home/pihole/pihole-FTL
+# Note: We cannot add CAP_NET_RAW and CAP_NET_ADMIN at this point
+setcap CAP_NET_BIND_SERVICE+eip /home/pihole/pihole-FTL
 
 # Prepare gravity database
 sqlite3 /etc/pihole/gravity.db < test/gravity.db.sql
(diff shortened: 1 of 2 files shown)


test/run.sh
```diff
@@ -7,10 +7,21 @@ if [[ ${CI} == "true" && "${CIRCLE_JOB}" != "x86_64" ]]; then
 fi
 
 # Install necessary additional components for testing
-apt-get -qq install dnsutils -y > /dev/null
+apt-get -qq install dnsutils libcap2-bin -y > /dev/null
 
-# Create necessary directories
+# Create pihole user
+useradd -m -s /bin/bash pihole
```
Mcat12: This will create the user with a login shell and home directory, but the `pihole` user is supposed to be a nologin user.

test/run.sh
```diff
@@ -7,10 +7,21 @@ if [[ ${CI} == "true" && "${CIRCLE_JOB}" != "x86_64" ]]; then
...
+
+# Create necessary directories and files
 mkdir -p /etc/pihole /var/run/pihole /var/log
+touch /var/log/pihole-FTL.log
+chown pihole:pihole /etc/pihole /var/run/pihole /var/log/pihole-FTL.log
+
+# Copy binary into a location the new user pihole can access
+cp ./pihole-FTL /home/pihole
+chmod +x /home/pihole/pihole-FTL
+# Note: We cannot add CAP_NET_RAW and CAP_NET_ADMIN at this point
```
Mcat12: Why is this?

test/test_suite.bats
```diff
@@ -249,14 +249,14 @@
 }
 
 @test "Fail on invalid argument" {
-  run bash -c './pihole-FTL abc'
+  run bash -c '/home/pihole/pihole-FTL abc'
```
Mcat12: Is there any way to not hard-code the binary path and name? This is not good for local testing.

test/run.sh
DL6ER: I'm not entirely sure, but I observed that, whenever I include one of these two capabilities, the `setcap` command will error with:
```
setcap: Operation not permitted.
```

test/test_suite.bats
DL6ER: I don't know how to transport a variable into the `bats` tests. If you know how to do it, you are more than invited to add this. We could also `cd` into this directory and just run `./pihole-FTL` if that is more convenient fir local tests.

Mcat12: You can probably use an environment variable, but if `cd`-ing is easier (and doesn't introduce more issues with local testing) then that's fine.

Mcat12: For now, we can leave this as is and execute the tests as part of the CI pipeline locally (`circleci local execute --job x86_64`).

PR #591, opened 2019-06-10
Add IPv6 support for network table

DL6ER:

## 10

---

Add IPv6 support for the network table. This updates the database version to 5.

As we expect an interface to have a number of addresses in the IPv6 world, we introduce a new database table `network_addresses` with three columns:

1. `id` corresponds to the `id` of the `network` table and, as such, to a unique hardware address.
2. `ip` contains a given IP address corresponding to the hardware address identified by this particular `id`
3. `lastQuery`: Unix timestamp of last query seen from this particular IP address.

Conflict management has been implemented to only update the `lastQuery` column if a given `(id,ip)` pair already exists in the database. This avoids multiple entries from the same data.

This is a backend implementation. This frontend needs to be modified as well in order to be able to source the new data. This is not a breaking change as the new version can be used with the current dashboard without any issues.


dschaper:
Can we be sure that the `ip` utility is installed in every platform?

DL6ER:
@dschaper Yes, we install it as dependency [here](https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh#L243) or [here](https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh#L286) and use it in the installer [here](https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh#L512), [here](https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh#L528), [here](https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh#L534) and in a few other places.

networktable.c
```diff
@@ -39,23 +38,40 @@ bool create_network_table(void)
 	return true;
 }
 
-// Read kernel's ARP cache using procfs
-void parse_arp_cache(void)
+bool create_network_addresses_table(void)
+{
+	bool ret;
+	// Create network_addresses table in the database
+	ret = dbquery("CREATE TABLE network_addresses ( id INTEGER NOT NULL, " \
```
Mcat12: As this ID is a foreign key and not owned by the table, it should be renamed to `network_id` and have the additional property of `REFERENCING network (id)`.

If we don't have the foreign key pragma turned on, we should do that too.

networktable.c
```diff
@@ -39,23 +38,40 @@ bool create_network_table(void)
...
+
+	// Update database version to 5
+	ret = db_set_FTL_property(DB_VERSION, 5);
+	if(!ret){ dbclose(); return false; }
+
+	return true;
+}
+
+// Parse kernel's neighbor cache
+void parse_neigh_cache(void)
```
Mcat12: For three extra letters, we can use the full word "neighbor" in the function name, and it may be less confusing.

networktable.c
```diff
@@ -76,15 +92,16 @@ void parse_arp_cache(void)
 	// Read ARP cache line by line
 	while(getline(&linebuffer, &linebuffersize, arpfp) != -1)
 	{
-		int num = sscanf(linebuffer, "%99s 0x%x 0x%x %99s %99s %99s\n",
-		                 ip, &type, &flags, hwaddr, mask, iface);
+		int num = sscanf(linebuffer, "%99s dev %99s lladdr %99s %99s",
```
Mcat12: On my machine, there is an extra word between the MAC address and the status: `router`. This may only appear on special neighbors, like the router, but with this code `status` would be set to `router` instead of `REACHABLE`.

DL6ER: I haven't seen such a `router` line before, can you post an (obfuscated) example?

Mcat12:
```
dead:beef:dead:beef:dead:beef dev wlp3s0 lladdr de:ad:be:ef:de:ad router REACHABLE
```

networktable.c
```diff
@@ -42,10 +42,11 @@ bool create_network_addresses_table(void)
 {
 	bool ret;
 	// Create network_addresses table in the database
-	ret = dbquery("CREATE TABLE network_addresses ( id INTEGER NOT NULL, " \
+	ret = dbquery("CREATE TABLE network_addresses ( network_id INTEGER NOT NULL, "\
 	                                               "ip TEXT NOT NULL, "\
-	                                               "lastQuery INTEGER NOT NULL, " \
-	                                               "UNIQUE(id,ip));");
+	                                               "lastQuery INTEGER NOT NULL, "\
+	                                               "UNIQUE(network_id,ip), "\
```
Mcat12: This can actually be the primary key

networktable.c
```diff
@@ -93,17 +93,35 @@ void parse_neighbor_cache(void)
...
+				   strcasecmp(status1, "stale")     != 0)
+				{
+					continue;
+				}
+				break;
+
+			// Example: dead:beef:dead:beef:dead:beef dev wlp3s0 lladdr de:ad:be:ef:de:ad router REACHABLE
+			case 5:
+				if(strcasecmp(status1, "router")     != 0 &&
+				   (strcasecmp(status2, "reachable") != 0 &&
```
Mcat12: The extra parentheses are unnecessary.

Mcat12:
Now that we are storing IPs in the `network_addresses` table, can we remove the `ip` column from the `network` table?

networktable.c
```diff
@@ -39,23 +38,41 @@ bool create_network_table(void)
 	return true;
 }
 
-// Read kernel's ARP cache using procfs
-void parse_arp_cache(void)
+bool create_network_addresses_table(void)
```
Mcat12: As part of this DB upgrade, all entries in the `network` table should have a row in `network_addresses` using the value of the `ip` column from the `network` table. Then this `ip` column can be removed from `network`.

DL6ER:
> Now that we are storing IPs in the network_addresses table, can we remove the ip column from the network table?

The question is: Is it worth it?

Removing a column is only possible by creating a new table, copying all but the `ip` column to it, removing the old table, and renaming the new table to `network`. This might take some time (thinking about those users with large deployments like in schools) and the user should better not interrupt this upgrading process (even if it shouldn't cause harm as the first step will take the majority of time and is non-destructive).

I agree that we can create an entry in `network_addresses` for each device during the upgrade process.

networktable.c
```diff
@@ -41,17 +41,101 @@ bool create_network_table(void)
...
 	                                               "ip TEXT NOT NULL, "\
-	                                               "lastSeen INTEGER NOT NULL, "\
+	                                               "lastSeen INTEGER NOT NULL DEFAULT (cast(strftime('%%s', 'now') as int)), "\
 	                                               "UNIQUE(network_id,ip), "\
 	                                               "FOREIGN KEY(network_id) REFERENCES network(id));");
-	if(!ret){ dbclose(); return false; }
+	if(!ret)
+	{
+		logg("create_network_addresses_table(): CREATE TABLE network_addresses failed!");
+		dbclose();
```
Mcat12: Should we also run a `ROLLBACK` here?

DL6ER: If `CREATE TABLE` failed, it shouldn't have done anything.

Mcat12: Maybe for the later statements then.

Mcat12: It looks like SQLite will close the transaction for you in this case.

networktable.c
```diff
@@ -41,17 +41,101 @@ bool create_network_table(void)
...
+	                     "FROM network;");
+	if(!ret)
+	{
+		logg("create_network_addresses_table(): INSERT INTO network_bck failed!");
+		dbclose();
+		return false;
+	}
+
+	// Step 3: Drop unique index in hwaddr before dropping the network table
+	ret = dbquery("DROP INDEX network_hwaddr_idx;");
```
Mcat12: This should be handled as part of `DROP TABLE network`, right?

networktable.c
```diff
@@ -41,17 +41,101 @@ bool create_network_table(void)
...
+	//         We add the UNIQUE constraint on (ip,hwaddr) already here
+	ret = dbquery("CREATE TABLE network_bck ( id INTEGER PRIMARY KEY NOT NULL, " \
+	                                         "hwaddr TEXT NOT NULL, " \
+	                                         "interface TEXT NOT NULL, " \
+	                                         "name TEXT, " \
+	                                         "firstSeen INTEGER NOT NULL, " \
+	                                         "lastQuery INTEGER NOT NULL, " \
+	                                         "numQueries INTEGER NOT NULL, " \
+	                                         "macVendor TEXT, " \
+	                                         "UNIQUE(id,hwaddr));");
```
Mcat12: If we just have a unique constraint on `hwaddr`, that would be even better than on on `(id, hwaddr)`.

networktable.c
DL6ER: Quoting the SQLite documentation:

> Transactions can be started manually using the BEGIN command. Such transactions usually persist until the next COMMIT or ROLLBACK command. But a transaction will also ROLLBACK if the database is closed or if an error occurs and the ROLLBACK conflict resolution algorithm is specified. See the documentation on the ON CONFLICT clause for additional information about the ROLLBACK conflict resolution algorithm. 

We use the standard which is `ON CONFLICT ABORT`:

> When an applicable constraint violation occurs, the ABORT resolution algorithm aborts the current SQL statement with an SQLITE_CONSTRAINT error and backs out any changes made by the current SQL statement

networktable.c
DL6ER: Why do you think so? I don't think it would make much of a difference, but we can surely change it.

Mcat12: It fits the data better. The important thing is that the MAC address is unique, not necessarily that it is unique with respect to some arbitrary ID. With the previous code, you could have duplicate MAC addresses as long as they have a different ID.

networktable.c
DL6ER: Ah, yeah, you are right.

networktable.c
```diff
@@ -39,23 +38,115 @@ bool create_network_table(void)
...
+	}
+
+	// Create a network_addresses row for each entry in the network table
+	ret = dbquery("INSERT INTO network_addresses (network_id,ip) SELECT id,ip FROM network;");
+	if(!ret)
+	{
+		logg("create_network_addresses_table(): INSERT INTO network_addresses failed!");
+		dbclose();
+		return false;
+	}
```
Mcat12: It looks like this pattern of `dbquery`, `if !ret`, etc can be generalized into a function or macro. What do you think of this?

networktable.c
```diff
@@ -39,23 +38,115 @@ bool create_network_table(void)
...
+
+	// Remove IP column from network table.
+	// As ALTER TABLE is severely limit, we have to do the column deletion manually.
+	// Step 1: We create a new table without the ip column
+	//         We add the UNIQUE constraint on (ip,hwaddr) already here
```
Mcat12: This comment is now out of date

networktable.c
```diff
@@ -76,16 +167,35 @@ void parse_arp_cache(void)
...
+				   strcasecmp(status2, "stale")     != 0)
+				{
+					continue;
+				}
+				break;
+
+			// Skip lines without hardware information
+			default:
+				continue;
+		}
```
Mcat12: Would the logic be simpler if we only checked for the lines with the status that we want?

There is also a `nud` option which can filter by status (see the man page for `ip-neighbour`:
```
nud STATE
                     only list neighbour entries in this state.  NUD_STATE takes values listed below or the special value all which means all states. This option may occur more than once.  If this option
                     is absent, ip lists all entries except for none and noarp.
```

test/test_suite.bats
```diff
@@ -243,9 +243,9 @@
   [[ "${lines[@]}" == *"CREATE TABLE queries ( id INTEGER PRIMARY KEY AUTOINCREMENT, timestamp INTEGER NOT NULL, type INTEGER NOT NULL, status INTEGER NOT NULL, domain TEXT NOT NULL, client TEXT NOT NULL, forward TEXT );"* ]]
   [[ "${lines[@]}" == *"CREATE TABLE ftl ( id INTEGER PRIMARY KEY NOT NULL, value BLOB NOT NULL );"* ]]
   [[ "${lines[@]}" == *"CREATE TABLE counters ( id INTEGER PRIMARY KEY NOT NULL, value INTEGER NOT NULL );"* ]]
-  [[ "${lines[@]}" == *"CREATE TABLE network ( id INTEGER PRIMARY KEY NOT NULL, ip TEXT NOT NULL, hwaddr TEXT NOT NULL, interface TEXT NOT NULL, name TEXT, firstSeen INTEGER NOT NULL, lastQuery INTEGER NOT NULL, numQueries INTEGER NOT NULL,macVendor TEXT);"* ]]
+  [[ "${lines[@]}" == *"CREATE TABLE IF NOT EXISTS \"network\" ( id INTEGER PRIMARY KEY NOT NULL, hwaddr TEXT UNIQUE NOT NULL, interface TEXT NOT NULL, name TEXT, firstSeen INTEGER NOT NULL, lastQuery INTEGER NOT NULL, numQueries INTEGER NOT NULL, macVendor TEXT);"* ]]
```
Mcat12: Why does `IF NOT EXISTS` show up here?

DL6ER: It is automatically added when renaming the table using `ALTER TABLE ... RENAME TO ...`

networktable.c
DL6ER: What do you propose? We need to be able to return `false` in `create_network_addresses_table()` if we detect an error.

Mcat12: Some macro which given `sql` will expand to:
```c
{
    int ret = dbquery(sql);
    if(!ret) {
        logg("create_network_addresses_table(): \"%s\" failed!", sql);
        dbclose();
        return false;
    }
}
```

networktable.c
```diff
@@ -38,86 +38,56 @@ bool create_network_table(void)
 	return true;
 }
 
+// Private macro
+#define SQL(sql) {\
+    if(!dbquery(sql)) {\
+        logg("create_network_addresses_table(): \"%s\" failed!", sql);\
```
Mcat12: Can this be generalized to all methods? Perhaps by replacing `create_network_addresses_table` with `__FUNCTION__`

networktable.c
```diff
@@ -39,23 +38,85 @@ bool create_network_table(void)
...
+	// Step 2: Copy data (except ip column) from network into network_back
+	SQL("INSERT INTO network_bck "\
+	              "SELECT id, hwaddr, interface, name, firstSeen, "\
+	                     "lastQuery, numQueries, macVendor "\
+	                     "FROM network;");
+
+	// Step 3: Drop the network table, the unique index will be automatically dropped
+	SQL("DROP TABLE network;");
+
+
```
Mcat12: Extra newline here

networktable.c
```diff
@@ -16,75 +16,84 @@
...
+		return false;\
+	}\
+}
+#define SQL_void(sql) {\
+	if(!dbquery(sql)) {\
+		logg("%s(): \"%s\" failed!", __FUNCTION__, sql);\
+		dbclose();\
+		return;\
+	}\
+}
```
Mcat12: These can be moved to `routines.h`, as they can be used by anything that interacts with the database.

database.c
```diff
@@ -88,14 +88,7 @@ bool dbopen(void)
 	// Foreign key constraints are disabled by default (for backwards
 	// compatibility), so must be enabled separately for each database
 	// connection (see also https://www.sqlite.org/foreignkeys.html).
-	rc = dbquery("PRAGMA FOREIGN_KEYS=ON;");
-	if( !rc )
-	{
-		logg("dbopen() - FOREIGN_KEYS error (%i): %s", rc, sqlite3_errmsg(db));
-		dbclose();
-		check_database(rc);
```
Mcat12: Do we want to add this `check_database` to the macro?

networktable.c
```diff
@@ -96,24 +94,23 @@ bool create_network_addresses_table(void)
...
 		return;
 	}
 
-	// Open database file
-	if(!dbopen())
-	{
-		logg("parse_arp_cache() - Failed to open DB");
-		pclose(arpfp);
-		return;
-	}
```
Mcat12: Why is this moved to before the `ip` command? If the `ip` command fails, the database will remain open.

DL6ER: The idea is to not run `ip neigh` when the database is not available. I will add a `dbclose()` in the case the `ip` command fails.

Mcat12:
I got this error when it tried to upgrade the database:
```
[2019-06-27 22:50:41.131 14380] SQLite3 message: abort at 24 in [DROP TABLE network;]: FOREIGN KEY constraint failed (787)
[2019-06-27 22:50:41.131 14380] dbquery(DROP TABLE network;) - SQL error (19): FOREIGN KEY constraint failed
[2019-06-27 22:50:41.131 14380] check_database(19): Disabling database connection due to error
[2019-06-27 22:50:41.131 14380] create_network_addresses_table(): "DROP TABLE network;" failed!
[2019-06-27 22:50:41.131 14380] Network-addresses table not initialized, database not available
```

Mcat12:
It looks like since we enable foreign keys by default, we need to disable them as part of step one:
> 1. If foreign key constraints are enabled, disable them using PRAGMA foreign_keys=OFF. 

https://www.sqlite.org/lang_altertable.html#otheralter

networktable.c
```diff
@@ -88,6 +92,9 @@ bool create_network_addresses_table(void)
 	// Finish transaction
 	SQL_bool("COMMIT");
 
+	// Re-enable foreign key enforcement
+	SQL_bool("PRAGMA foreign_keys=OFF");
```
Mcat12: This is not re-enabling support, likely a copy-paste error.

pralor:

https://discourse.pi-hole.net/t/apple-devices-not-in-network-overview/22177/6

pralor:

https://discourse.pi-hole.net/t/network-table-seeing-clients-as-hostname-of-router-rather-than-devices/31597/7

PR #593, opened 2019-06-15
Use SFTP to upload binaries

Mcat12:

---

Change from using a custom uploader client to using SFTP. The custom uploader is sometimes unreliable, and SFTP is more secure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,8 +23,8 @@ version: 2
           mv pihole-FTL "${BIN_NAME}"
           sha1sum pihole-FTL-* > ${BIN_NAME}.sha1
           ssh-keyscan -H $SSH_HOST >> ~/.ssh/known_hosts
-          [[ "$CIRCLE_PR_NUMBER" == "" ]] && sftp $SSH_USER@$SSH_HOST <<<$'mkdir ${CIRCLE_TAG:CIRCLE_BRANCH}'
-          [[ "$CIRCLE_PR_NUMBER" == "" ]] && sftp $SSH_USER@$SSH_HOST <<<$'put ${BIN_NAME}*'
+          [[ "$CIRCLE_PR_NUMBER" == "" ]] && sftp $SSH_USER@$SSH_HOST <<<$"mkdir ${CIRCLE_TAG:CIRCLE_BRANCH}"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -20,16 +20,13 @@ version: 2
...
           mv pihole-FTL "${BIN_NAME}"
           sha1sum pihole-FTL-* > ${BIN_NAME}.sha1
-          wget https://ftl.pi-hole.net:8080/FTL-client
-          chmod +x ./FTL-client
-          [[ "$CIRCLE_PR_NUMBER" == "" ]] && ./FTL-client "${FOLDER}" "${BIN_NAME}" "${FTL_SECRET}"
-          [[ "$CIRCLE_PR_NUMBER" == "" ]] && ./FTL-client "${FOLDER}" "${BIN_NAME}.sha1" "${FTL_SECRET}"
-          rm ./FTL-client
+          ssh-keyscan -H $SSH_HOST >> ~/.ssh/known_hosts
+          [[ "$CIRCLE_PR_NUMBER" == "" ]] && sftp $SSH_USER@$SSH_HOST <<<$"mkdir ${DIR}"
+          [[ "$CIRCLE_PR_NUMBER" == "" ]] && sftp $SSH_USER@$SSH_HOST <<<$"put ${BIN_NAME}* ${DIR}"
```
DL6ER: Can we run both commands in the same connection to avoid two handshakes, etc.?

```suggestion
          if [[ "$CIRCLE_PR_NUMBER" == "" ]]; then
            sftp -b - $SSH_USER@$SSH_HOST << "-mkdir ${DIR}
            put ${BIN_NAME}* ${DIR}
          fi
```

Mcat12: The `mkdir` command may fail if the directory already exists, and there is no good way to prevent that error. Therefore, the commands are separated so the second one will execute even if the first fails.

Mcat12: https://stackoverflow.com/questions/27420116/how-to-create-directory-if-doesnt-exists-in-sftp

DL6ER: > The mkdir command may fail if the directory already exists, and there is no good way to prevent that error.

That's why I suggested `batch` mode for SFTP. In this case, we can prepend `mksir` by `-` to continue batch execution even if the `mkdir` fails.

PR #594, opened 2019-06-18
Generate self-contained musl binary

DL6ER:

## 10

---

Compile x86_64-musl binary as static position independent executable. A static position independent executable is similar to static executable, but can be loaded at any address without a dynamic linker.

As a test, I downloaded the generated `pihole-FTL-musl-linux-x86_64` binary into a Debian-based `x86_64` container and made it executable. Thereafter, ran our test suite* using this binary and all tests succeeded.

---
*) For running the tests, 
- cloned the `FTL` repository, 
- checked out the current `development` branch,
- copied the previously downloaded binary into the cloned directory, and
- ran `test/run.sh`.

I did *not* build the binary in the cloned repository.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,8 +13,7 @@ version: 2
         name: "Build"
         command: |
           BRANCH=$([ -z "$CIRCLE_TAG" ] && echo "$CIRCLE_BRANCH" || echo "master")
-
-          make GIT_BRANCH="${BRANCH}" GIT_TAG="${CIRCLE_TAG}"
+          make GIT_BRANCH="${BRANCH}" GIT_TAG="${CIRCLE_TAG}" CIRCLE_JOB="${CIRCLE_JOB}"
           file pihole-FTL
     - run:
         name: "Upload"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ version: 2
         name: "Build"
         command: |
           BRANCH=$([ -z "$CIRCLE_TAG" ] && echo "$CIRCLE_BRANCH" || echo "master")
-          make GIT_BRANCH="${BRANCH}" GIT_TAG="${CIRCLE_TAG}" CIRCLE_JOB="${CIRCLE_JOB}"
+          make GIT_BRANCH="${BRANCH}" GIT_TAG="${CIRCLE_TAG}" STATIC="${STATIC}"
           file pihole-FTL
     - run:
         name: "Upload"
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ LIBS=-pthread -lrt -Wl,-Bstatic -L/usr/local/lib -lhogweed -lgmp -lnettle
 # Flags for compiling with libidn2: -lidn2
 
 # Do we want to compile a statically linked musl executable?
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -98,11 +97,19 @@ CCFLAGS=-std=gnu11 -I$(IDIR) $(WARNFLAGS) -D_FILE_OFFSET_BITS=64 $(HARDENING_FLA
...
 # for dnsmasq we need the nettle crypto library and the gmp maths library
 # We link the two libraries statically. Although this increases the binary file size by about 1 MB, it saves about 5 MB of shared libraries and makes deployment easier
-#LIBS=-pthread -lnettle -lgmp -lhogweed
-LIBS=-pthread -lrt -Wl,-Bstatic -L/usr/local/lib -lhogweed -lgmp -lnettle -Wl,-Bdynamic
+LIBS=-pthread -lrt -Wl,-Bstatic -L/usr/local/lib -lhogweed -lgmp -lnettle
 # Flags for compiling with libidn : -lidn
 # Flags for compiling with libidn2: -lidn2
 
+# Do we want to compile a statically linked musl executable?
+ifeq "$(CIRCLE_JOB)" "x86_64-musl"
```
Mcat12: Can this be a more general check (not tied to the CI)? Something like checking if `MUSL` equals `true`?

log.c
```diff
@@ -140,7 +140,7 @@ void log_counter_info(void)
 void log_FTL_version(const bool crashreport)
 {
 	logg("FTL branch: %s", GIT_BRANCH);
-	logg("FTL version: %s", GIT_TAG);
+	logg("FTL version: %s", GIT_VERSION);
```
Mcat12: Is this meant for #595?

Makefile
Mcat12: Actually, doesn't need to be tied to musl, the variable could be named `STATIC`.

DL6ER: I had the same idea initially. For the same of simplicity (use the same build job for all architectures), we'd realize this via an environment variable set in the musl docker container. What do you think about this?

log.c
DL6ER: Yes, this got committed by accident, I'll remove it from this PR.

.circleci/config.yml
```diff
@@ -13,8 +13,7 @@ version: 2
         name: "Build"
         command: |
           BRANCH=$([ -z "$CIRCLE_TAG" ] && echo "$CIRCLE_BRANCH" || echo "master")
-
-          make GIT_BRANCH="${BRANCH}" GIT_TAG="${CIRCLE_TAG}"
+          make GIT_BRANCH="${BRANCH}" GIT_TAG="${CIRCLE_TAG}" STATIC="${STATIC}"
```
DL6ER: I added it here to be explicit. As `STATIC` is an environment variable, we can also leave it off as `make` will silently pick it up.

Makefile
Mcat12: To make it easier to change the value, we should instead set the variable in the CircleCI environment variables for that job.

DL6ER: I intentionally didn't choose this path as it again ties us more to the CI. The sole purpose of the `musl` container is to provide a self-contained binary. The container should behave similarly on the CI and when compiling local. This wouldn't be the case if we add in a (hidden) extra environment variable in the CI.

PR #601, opened 2019-06-23
Move ./aux/macvendor.py -> ./macvendor.py

### LINKED ISSUE #599: Unable to clone/DEV via Windows due to repo directory naming using reserved device names
The use of the directory named `aux` in this repo makes it impossible to successfully fully clone and build/develop at all on a Windows environment due to this being one of the several "reserved" names (for device use) implemented back in the mid 1970's.  

I was aware of this so was not surprised when I saw the git error upon cloning on WIN regarding the directory named `AUX` causing the catastrophic git clone failure (my hair is really grey) but for most it is obviously probably not known and/or they are not attempting to clone and develop using a Windows environment.

Here is a bit more info from MS stating as much and some background ("file" also obviously goes for "directory" as they are both just file table/journal entries):

https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file

```
Do not use the following reserved names for the name of a file:

CON, PRN, AUX, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9. Also avoid these names followed immediately by an extension; for example, NUL.txt is not recommended. For more information, see Namespaces.
```

Since many DEVs, despite having access to Linux-based systems, perform their DEV workflows in a Windows environment it would be great to have this fixed so the repo can exist on a windows volume.  TIA!

DL6ER:

## 10

---

Move `./aux/macvendor.py -> ./macvendor.py`. This fixes #599.


Mcat12:
The gitignore should be updated as well

Mcat12:
What do you think of renaming the folder instead of moving them to the root? Moving to the root adds more clutter to that level.

PR #610, opened 2019-07-06
Database error checking

DL6ER:

## 10

---

Improve error checking in the database routines to cover all actions that can fail, e.g., due to readonly database errors.

- Check if the `pihole-FTL.db` database is still available at the beginning of all database routines in `database/common.c`
- Log if we failed to update the database from version 3 -> 4 and return early if this is the case.
- Simplify logging in case of an error.
    Previously:
    ```
    [2019-07-06 17:00:55.526 32339] SQLite3 message: statement aborts at 36: [CREATE UNIQUE INDEX network_hwaddr_idx ON network(hwaddr)] attempt to write a readonly database (8)
    [2019-07-06 17:00:55.526 32339] dbquery(CREATE UNIQUE INDEX network_hwaddr_idx ON network(hwaddr)) - SQL error (8): attempt to write a readonly database
    [2019-07-06 17:00:55.526 32339] check_database(8): Disabling database connection due to error
    [2019-07-06 17:00:55.526 32339] unify_hwaddr(): "CREATE UNIQUE INDEX network_hwaddr_idx ON network(hwaddr)" failed!
    ```
    Now:
    ```
    [2019-07-06 17:00:55.526 32339] SQLite3 message: statement aborts at 36: [CREATE UNIQUE INDEX network_hwaddr_idx ON network(hwaddr)] attempt to write a readonly database (8)
    [2019-07-06 17:00:55.526 32339] check_database(8): Disabling database connection due to error
    [2019-07-06 17:00:55.526 32339] ERROR: unify_hwaddr() failed!
    ```

### DIFF
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -94,6 +94,15 @@ bool dbquery(const char *format, ...)
 		return false;
 	}
 
+	// Log generated SQL string when dbquery() is called
+	// although the database connection is not available
+	if(!database)
+	{
+		logg("dbquery(\"%s\") called but database is not available!", query);
+		sqlite3_free(query);
+		return false;
+	}
+
 	if(config.debug & DEBUG_DATABASE)
 	{
 		logg("dbquery: \"%s\"", query);
@@ -230,6 +239,9 @@ void db_init(void)
 		return;
 	}
 
+	// Database connection is not open
+	database = true;
+
 	// Test FTL_db version and see if we need to upgrade the database file
 	int dbversion = db_get_FTL_property(DB_VERSION);
 	logg("Database version is %i", dbversion);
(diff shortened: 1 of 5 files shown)


src/database/common.c
```diff
@@ -307,11 +320,15 @@ void db_init(void)
...
-	database = true;
 }
 
 int db_get_FTL_property(const unsigned int ID)
 {
+	if(!database)
+	{
+		logg("db_get_FTL_property(%u) called but database is not available!", ID);
+		return DB_FAILED;
+	}
```
Mcat12: This can be made into a macro to avoid duplicated code.

src/database/common.c
```diff
@@ -19,7 +19,10 @@
 #include "files.h"
 
 sqlite3 *FTL_db;
-bool database = false;
+// This boolean is set to false once we hit the
+// first database error to prevent further access
+// to the pihole-FTL.db database
+bool database = true;
```
Mcat12: This check in main should now be moved into `db_init` as the `database` value will remain `true` if the check fails even though the database is not available.
https://github.com/pi-hole/FTL/blob/9f127e9838c172d5557b969c56164f0627953e86/src/main.c#L70-L72

src/database/common.c
DL6ER: Not really. The functions aren't always the same. Some of them return an integer (`DB_FAILED`), some return a boolean, some are `void`. Also, some of them take an argument, some even two, sometimes unsigned, sometimes signed. I don't think this really justifies any kind of macro.

PR #612, opened 2019-07-09
Add whitelist regex support

DL6ER:

## 10

---

Implement regex support for the whitelist. Just as with blacklists, whitelist regex support is intended for few, specialized filters. For mass-whitelisting, the known exact whitelisting should be used*.

Regex whitelist filter will affect the entire blocking Pi-hole is performing:

- Gravity: Domains from the blocklists that a match whitelist regex filter will not be imported.
- Blacklist: Blacklisted domains that match a whitelist regex filter will not be imported.
- Regex blacklist: Domains matching both, blacklist **and** whitelist regex filters will be permitted as the whitelist is given a higher importance in FTL.

Using whitelist regex filters can slow down the startup process of FTL, however it can also result in improved performance when fewer blocking domains have to be imported.

Performance examples on a NanoPi NEO:

- No regex whitelist:
    ```
    Database (gravity): imported 114851 domains (took 3269.0 ms)  
    ```
- One regex whitelist filter (not matching):
    ```
    Database (gravity): imported 114851 domains (took 3707.6 ms)
    ```
- Five regex whitelist filter (none of them matching anything):
    ```
    Database (gravity): imported 114851 domains (took 4374.3 ms) 
    ```
- One whitelist regex (`\.com$`, matching all `.com` addresses):
    ```
    Database (gravity): imported 86736 domains (took 2182.2 ms)
    ```
    Note the reduced number of imported domains and the reduced time it takes to import gravity.

The used not-matching filters are long filters of medium complexity (containing a few character groups and alternations).

This pull request also adds eight new tests around blocking and whitelisting.

---

*) Pi-hole v5.0 ships with a massively improved exact whitelist performance. Up to millions of domains on the whitelist are now possible even on low-end hardware.


DL6ER:
Discourse feature request: [Wildcard and regex support for whitelisting](https://discourse.pi-hole.net/t/wildcard-and-regex-support-for-whitelisting/14538)

pralor:

https://discourse.pi-hole.net/t/filter-blocklist-on-update/21679/8

dschaper:
Is this a WIP?

DL6ER:
No, it is ready for review (and possible discussions based thereon).

src/database/gravity-db.c
```diff
@@ -302,9 +310,13 @@ static bool domain_in_list(const char *domain, sqlite3_stmt* stmt)
...
 
-	// Return result.
+	// Return early in case we already have found an exact match.
 	// SELECT EXISTS(...) either returns 0 (false) or 1 (true).
-	return result == 1;
+	if(result == 1)
+		return true;
+
+	// If not: Walk regex-based whitelist filters in addition
+	return match_regex(domain, REGEX_WHITELIST);
```
Mcat12: This function is executed for both whitelist and audit list, so the whitelist-specific code should be lifted into `in_whitelist`.

src/main.c
```diff
@@ -102,9 +102,6 @@ int main (int argc, char* argv[])
 	close_telnet_socket();
 	close_unix_socket();
 
-	// Free regex list and array of whitelisted domains
-	free_regex();
```
Mcat12: Shouldn't this be kept?

src/regex.c
```diff
@@ -19,69 +19,72 @@
...
 {
 	// Regex failed for some reason (probably user syntax error)
 	// Get error string and log it
-	const size_t length = regerror(errcode, &regex[index], NULL, 0);
+	const size_t length = regerror(errcode, &regex[regexid][index], NULL, 0);
 	char *buffer = calloc(length,sizeof(char));
-	(void) regerror (errcode, &regex[index], buffer, length);
-	logg("ERROR %s regex on line %i: %s (%i)", where, index+1, buffer, errcode);
+	(void) regerror (errcode, &regex[regexid][index], buffer, length);
+	logg("ERROR %s regex %s no. %i: %s (%i)", where, regextype[regexid], index+1, buffer, errcode);
```
Mcat12: Instead of `no.`, `id = ` may be clearer.

src/regex.c
```diff
@@ -107,65 +110,78 @@ void free_regex(void)
...
+		// Reset counter for number of regex
+		num_regex[regexid] = 0;
+	}
 }
 
-void read_regex_from_database(void)
+static void read_regex_table(unsigned char regexid)
 {
+	// Get database ID
+	unsigned char databaseID = (regexid == REGEX_BLACKLIST) ? REGEX_BLACK_LIST : REGEX_WHITE_LIST;
```
Mcat12: The difference between these enums is a little confusing. Perhaps the DB one should have `DB` or `TABLE` added to it?

Mcat12:
@DL6ER As this is ready to review, the core PR should also be made ready for review (un-draft).

test/gravity.db.sql
```diff
@@ -125,16 +142,28 @@ CREATE TRIGGER tr_blacklist_update AFTER UPDATE ON blacklist
...
-    WHERE regex.enabled = 1 AND (regex_by_group.group_id IS NULL OR "group".enabled = 1)
-    ORDER BY regex.id;
+CREATE VIEW vw_regex_blacklist AS SELECT DISTINCT domain
+    FROM regex_blacklist
+    LEFT JOIN regex_blacklist_by_group ON regex_blacklist_by_group.regex_blacklist_id = regex_blacklist.id
+    LEFT JOIN "group" ON "group".id = regex_blacklist_by_group.group_id
+    WHERE regex_blacklist.enabled = 1 AND (regex_blacklist_by_group.group_id IS NULL OR "group".enabled = 1)
+    ORDER BY regex_blacklist.id;
+
+CREATE TRIGGER tr_regex_update AFTER UPDATE ON regex_blacklist
```
Mcat12: This trigger is now named `tr_regex_blacklist_update`

src/regex.c
DL6ER: The regex no. 0 might be not what is at table ID = 0. FTL starts to count drom zero on for the first regex it read from the table. However, there could be a regex with ID=0 which is either disabled or has been removed. Then "ID" would be misleading. We could modify the view to return both the ID and the regex string (instead of currently only the domain, `CREATE VIEW vw_regex_blacklist AS SELECT DISTINCT domain`). I don't think we could get a meaningful ID, otherwise.

The obvious alternative is to remove this as we can print the failed regex string itself.

Mcat12: It would be easier for the user to debug if we print the regex instead of the number or ID.

src/main.c
DL6ER: The scope of the `free_regex()` routine has been reduced to the respective module. At this point, `pihole-FTL` will (unstoppable) terminate. This will free all allocated memory automatically.

src/regex.c
DL6ER: With the latest commit, we will print
```
[2019-07-16 12:24:47.778 4694] ERROR: compiling regex whitelist filter "?": Invalid preceding regular expression (13)
```
when the compilation fails.

However, we will have to store them permanently if we also want to be able to report failures during runtime evaluation. Currently, we store the string representation of the filters only if `DEBUG_REGEX` is enabled.

I will investigate on if it is at all possible to encounter an error during runtime if a filter was compiled successfully.

DL6ER: Updated the log message to be shown as a warning rather than an error (it doesn't really affect the DNS functionality):
```
[2019-07-16 22:31:52.558 10489] Warning: Invalid regex whitelist filter "?": Invalid preceding regular expression (error code 13)
```

src/database/gravity-db.c
```diff
@@ -119,37 +124,37 @@ bool gravityDB_getTable(const unsigned char list)
 {
 	if(!gravity_database_avail)
 	{
-		logg("gravityDB_getTable(%d): Gravity database not available", list);
+		logg("gravityDB_getTable(%u): Gravity database not available", list);
+		return false;
+	}
+
+	// Checking for smaller than GRAVITY_LIST is omitted due to list being unsigned
```
Mcat12: `GRAVITY_LIST` is now `GRAVITY_TABLE`

DL6ER: Since this is only a minor comment inaccuracy, I will change it in another PR.

PR #626, opened 2019-08-04
Fix possible crash if database is locked on startup

### LINKED ISSUE #625: SQLite3 message: database is locked in "SELECT VALUE FROM ftl WHERE id = 0;"

_2_

---

When restarting the DNS service with the following command:

```sh
sudo pihole restartdns
```
the DNS service should properly restart without crashing.

When executing the command (see above), FTL reports the following error messages (entries taken from the log file under `/var/log/pihole-FTL.log`:

```log
[2019-07-30 16:30:20.036 1727] SQLite3 message: database is locked in "SELECT VALUE FROM ftl WHERE id = 0;" (5)
[2019-07-30 16:30:20.036 1727] db_query_int(SELECT VALUE FROM ftl WHERE id = 0;) - SQL error prepare (5): database is locked
[2019-07-30 16:30:20.036 1727] Database version is -2
[2019-07-30 16:30:20.036 1727] Database version incorrect, database not available
```

and FTL crashes. This behaviour does only occur after executing the restart command, not at normal startup.

This might be related to the merged PR #185 

- execute `sudo pihole restartdns`

see above

**Device specifics**

Hardware Type: rPi, VPS, etc
OS:
```
Description:	Raspbian GNU/Linux 10 (buster)
Release:	10

Linux hostname 4.19.58-v7+ #1245 SMP Fri Jul 12 17:25:51 BST 2019 armv7l GNU/Linux

Pi-hole version v4.3.1
AdminLTE version is v4.3
FTL version is v4.3.1
```

DL6ER:

## 10

---

Ensure we always initialize the database lock mutex even if an error causes us to return early in `db_init()`.

This PR fixes #625 (at least I assume this given the limited information available there so far).

### DIFF
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -221,6 +221,14 @@ void db_init(void)
 		return;
 	}
 
+	// Initialize database lock mutex
+	if (pthread_mutex_init(&dblock, NULL) != 0)
+	{
+		logg("FATAL: FTL_db mutex init failed\n");
+		// Return failure
+		exit(EXIT_FAILURE);
+	}
+
 	// Initialize SQLite3 logging callback
 	// This ensures SQLite3 errors and warnings are logged to pihole-FTL.log
 	// We use this to possibly catch even more errors in places we do not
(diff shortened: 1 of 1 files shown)


src/database/common.c
```diff
@@ -221,6 +221,17 @@ void db_init(void)
 		return;
 	}
 
+	// Initialize database lock mutex
+	if (pthread_mutex_init(&dblock, NULL) != 0)
+	{
+		logg("FATAL: FTL_db mutex init failed\n");
```
Mcat12: Is there some error output we can use to know why the mutex failed to initialize?

PR #631, opened 2019-09-01
Update dnsmasq code

### LINKED ISSUE #603: Cannot use libnettle 3.5

## 10

---

`pihole-FTL` should compile when using the most recent version of `libnettle`

Last week, `libnettle` v3.5 has been released. It obsoletes some symbols used by `dnsmasq` such as `nettle_secp_256r1` and `nettle_secp_384r1`.

- Try to compile any branch of `pihole-FTL` using the most recent version of `libnettle`.

A temporary workaround is mentioned here: https://github.com/themiron/dnsmasq/commit/6fd9aba7abe1e084123bc5002959350897774ace

DL6ER:

---

Highlights:

- **Improve cache behavior for TCP connections**. For ease of implementation, we always forked a new process to handle each incoming TCP connection. A side-effect of this is that any DNS queries answered from TCP connections are not cached: when TCP connections were rare, this was not a problem. With the coming of DNSSEC, it's now the case that some DNSSEC queries have answers which spill to TCP, and if, for instance, this applies to the keys for the root then those never get cached, and performance is very bad. This fix passes cache entries back from the TCP child process to the main server process, and fixes the problem.
- Remove the `NO_FORK` compile-time option, and support for uclinux. In an era where everything has an MMU, this looks like an anachronism, and it adds to (Ok, multiplies!) the combinatorial explosion of compile-time options.
- **Fix line-counting when reading `/etc/hosts` and friends**; for correct error messages.
- Fix bug in DNS non-terminal code which could sometimes cause a `NODATA` rather than an `NXDOMAIN` reply.
- **Support `TCP-fastopen`** (RFC-7413) on both incoming and outgoing TCP connections, if supported and enabled in the OS.
- **Improve kernel-capability manipulation code** under Linux. The code now fails early if a required capability is not available, and tries not to request capabilities not required by its configuration.
- **Add `--shared-network` config.** This enables allocation of addresses the DHCP server in subnets where the server (or relay) doesn't have an interface on the network in that subnet.
- **Fix compilation against nettle version 3.5 and later.** This fixes #603 
- **Fix spurious DNSSEC validation failures** when the auth section of a reply proving that a DS record does not exist contains unsigned RRs. Only the NSEC/NSEC3 records needed to prove the non-existence of the DS record must be signed. 
-  **Always force AD bit to zero in authoritative DNS answers. **
-  **Cache SRV records.**
- ** Treat DS and DNSKEY queries being forwarded the same as those locally originated.** The queries will not be forwarded to a server for a domain, unless there's a trust anchor provided for that domain. This allows, especially, suitable proof of non-existance for DS records to come from the parent domain for domains which are not signed.
-  **Fix spurious AD flags** in some DNS replies from local config. 
- Ensure that AD bit is reset on answers from `--address=/<domain>/<address>`.


Mcat12:
Should we wait for a stable release, as we have done previously?

TC1977:
I'm having repeated crashes of FTL 4.3.2 over the last 48 hours, and `pihole -d` seems to be pointing to dnsmasq as the cause of the errors. How can I switch my pihole-FTL to this PR to test it?

DL6ER:
@TC1977 Please open a new issue ticket with more details. I'm not sure of switching to this branch would solve your issues, however, as this branch is based off of `development`, you'll have to check out the `development` branches for this.

DL6ER:
Superseded by https://github.com/pi-hole/FTL/pull/652

PR #635, opened 2019-09-05
Check pointers before accessing content

DL6ER:

## 10

---

 In case requested object is out of range (`ID < 0` or `ID > maximum`) or has an incorrect magic byte, get getter functions in `shmem.c` return `NULL`. We have check against such `NULL`s before accessing the objects to avoid memory issues and segmentation faults.

Unfortunately, there are many getters and, hence, we have to add many guards. In a more high-level language, something like this will probably done by the language/compiler for us, but as C is a rather low-level language (you may want to call it "Comfort-Assembler"), we have to so this checking ourselves. This freedom opens a few possibilities, like we can skip updating only parts of the information if only some pointers are not available.

Finally, I should note that none of the checks added in this PR are expected to be triggered. However, I have once seen one failed magic byte (in several years looking almost daily at the logs) so there is a certain necessity for the checks to ensure the absence of SEGFAULTs in the crucial service a DNS server offers.

---

Footnote: The failing tests will be fixed when #634 is merged into this branch.

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -81,6 +81,9 @@ void getStats(const int *sock)
 	{
 		// Get client pointer
 		const clientsData* client = getClient(clientID, true);
+		if(client == NULL)
+			continue;
+
 		if(client->count > 0)
 			activeclients++;
 	}
@@ -227,6 +230,8 @@ void getTopDomains(const char *client_message, const int *sock)
 	{
 		// Get domain pointer
 		const domainsData* domain = getDomain(domainID, true);
+		if(domain == NULL)
+			continue;
 
 		temparray[domainID][0] = domainID;
 		if(blocked)
@@ -287,6 +292,8 @@ void getTopDomains(const char *client_message, const int *sock)
 		const int domainID = temparray[i][0];
 		// Get domain pointer
 		const domainsData* domain = getDomain(domainID, true);
+		if(domain == NULL)
+			continue;
 
 		// Skip this domain if there is a filter on it
 		if(excludedomains != NULL && insetupVarsArray(getstr(domain->domainpos)))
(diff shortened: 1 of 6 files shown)


Mcat12:
Wouldn't it be better to change it to simply fail (exit) if the memory is corrupt? If the memory is corrupt, there isn't much point in continuing.

DL6ER:
Checking pointers that might come back as `NULL` seems to be the right thing to do.

This PR is motivated by a few (namely three) reports I've been sent by mail over the last year where users saw non-reproducible bugs. After switching on all debugging options, we were able to identify that there seems to be a certain possibility that `dnsmasq` logs a `forwarded ... to ...` without having issued a `new query ...` before. When this happens at the very beginning of the FTL run (and only then), we have not stored a single query and the first call to `getQuery(i, true)` (where `i=0`) will already return a `NULL` (as the magic byte *cannot* match). Our code apparently tries to access this `NULL` [here](https://github.com/pi-hole/FTL/pull/635/files#diff-092b49420239862fdbdcc87fdc72e10cL320) and hits a `SEGFAULT`.

However, in this case -- and also when a bug (not a memory corruption) "just" caused some magic bytes to be incorrect --, we could have safely continued without hard failing by just skipping this one reply.

An alternative to checking against `NULL` is to change the functions to something like `getQuery(ID, true, &okay)` and then check the `okay` boolean, however, this is just a more verbose way of ding the same thing in my eyes.

As accessing a `NULL` pointer is always bad and can cause a `SEGFAULT` anywhere, I added according checking code also to the telnet API and the other places where I found `get{Query,Domain,Forward,Client}(...)`.

Mcat12:
If there are only a few known cases where this can happen, we should add an explicit check at those points instead of loosening our safety guarantees. In the forwarded without query case, first check to make sure we have recorded a query and stop if we have not recorded any queries.

DL6ER:
There is only one known case I know of where a crash happened, actually. Why do you think we shouldn't add a check everywhere a crash would occur otherwise? Just because of if memory corruption occurred, there is no point in letting it run further?

I'm not sure whether it'll be easy to *exactly* predetermine exactly where the crashes are non-fatal and where they are fatal, but I'll check it again tomorrow.

dschaper:
@DL6ER Status? Is this good to check?

DL6ER:
Yes

dschaper:
After this merges we will need to set `ci/circleci: arm-qemu` back to required for `development` branch.

PR #642, opened 2019-09-18
Per-client blocking rules, intermediate CNAME path blocking and some more

DL6ER:

## 10

---

This pull request proposes extensive changes in the way FTL decides how and when to block queries.

One result of our work is that we were able to **reduce overall memory requirements of the `pihole-FTL` process significantly** allowing for the size blocking lists solely being limited by the SD card and not anymore by RAM size.

Another side effect is that our improved implementation now also allows for **client-dependent blocking** settings which became the tag of this pull request.

Per-client blocking will allow you to independently tag
- regex blacklist entries
- exact blacklist entries
- regex whitelist entries
- exact whitelist entries, but also
- entire adlists
to groups and assign groups (also multiple) to individual clients.

This allows you to, e.g., set up a group of blacklisted domains you only want to see blocked on Apple devices whereas you want to set up some whitelisted domains you need for properly working Windows devices. The same with adlists: You can specify that some of your adlists (maybe Windows telemetry blocking) to be used by a specific group of your clients in which you put your Windows computers whereas you don't add these adlists for Linux machines.

Groups can also be used to enable or disable many domains as once, there is no need to remove entries from lists to disable them. In addition, clients can now be associated with these groups but will, obviously, only use enabled lists.
Untagged (= no groups associated to them) clients are the default and use *all* enabled lists. In contrast, groups also provide an elegant way to disable (or reduce) blocking for individual clients. Just assign them to an "empty" group, that is a group not connected to any adlist or blacklist entry.

Besides all the great advantages our new implementation brings, we do not want to hide the downsides. It was always of great importance for us to have Pi-hole being able to run on even the lowest-end devices such as the Raspberry Pi version 1. With our new implementation - and all the new logic that enables us to do much more now - performance restrictions seem to suggest that we slowly start to deviate from the lowest-end devices. Pi-hole will still run on such devices, however, many clients with complex per-client configurations or huge blocking lists or many regex filters (or any combination of these) will cause Pi-hole to maybe require a more beefy device to still meet your latency expectations.
Having said all this, note that if you run Pi-hole in a normal household network (< 10 clients), use standard lists and no or only a few regex filters, there should be nothing to worry about for you.

**Update 1** We were able to resolve the majority of the feared performance penalty with a clever vector implementation memorizing whether a specific client has queried a specific domain already before and what the result was.
**Update 2** The traditional way to reload the information after changing the lists was to send a reload command (`SIGHUP`) to FTL. This had the (necessary) side-effect of clearing the entire DNS cache. ~~With this new version of FTL, the user can send `SIGUSR2` to only clear the domain-blocking cache.~~ The DNS cache itself stays intact and is **not flushed**.
**Update 3** The signal for flushing only the domain-blocking cache (and reload the regex lists) has been changes to the unused `SIGRT0`. Use, e.g., `sudo kill -SIGRTMIN $(pidof pihole-FTL)`.


src/dnsmasq/cache.c
```diff
@@ -1687,14 +1683,6 @@ char *record_source(unsigned int index)
     return "config";
   else if (index == SRC_HOSTS)
     return HOSTSFILE;
-  /*----- Pi-hole modification -----*/
-  else if (index == SRC_REGEX)
-    return SRC_REGEX_NAME;
-  else if (index == SRC_GRAVITY)
-    return SRC_GRAVITY_NAME;
-  else if (index == SRC_BLACK)
-    return SRC_BLACK_NAME;
```
Mcat12: How are blocked requests recorded in dnsmasq's log now?

src/dnsmasq/forward.c
```diff
@@ -1981,14 +2003,28 @@ unsigned char *tcp_request(int confd, time_t now,
...
+	  else
+	  {
+	  /**********************************************/
 	   
 	   /* m > 0 if answered from cache */
 	   m = answer_request(header, ((char *) header) + 65536, (size_t)size, 
 			      dst_addr_4, netmask, now, ad_reqd, do_bit, have_pseudoheader);
 	  
-	  /* Do this by steam now we're not in the select() loop */
-	  check_log_writer(1);
```
Mcat12: Is this deletion intentional?

src/dnsmasq_interface.c
```diff
@@ -212,26 +218,48 @@ void _FTL_new_query(const unsigned int flags, const char *name, const struct all
 	// Get domain pointer
 	domainsData* domain = getDomain(domainID, true);
 
-	// Try blocking regex if configured
-	if(domain->regexmatch == REGEX_UNKNOWN && blockingstatus != BLOCKING_DISABLED)
+	// Only check domains for blocking conditions when global blocking is enabled
+	char blockDomain = 0;
```
Mcat12: This should be a boolean, including the function return type.

src/main.c
```diff
@@ -22,6 +22,7 @@
 #include "regex_r.h"
 #include "shmem.h"
 #include "capabilities.h"
+#include "datastructure.h"
```
Mcat12: What is this used for here?

src/datastructure.h
```diff
@@ -57,6 +60,10 @@ typedef struct {
 	int overTime[OVERTIME_SLOTS];
 	unsigned int numQueriesARP;
 	bool new;
+	sqlite3_stmt* whitelist_stmt;
+	sqlite3_stmt* gravity_stmt;
+	sqlite3_stmt* blacklist_stmt;
```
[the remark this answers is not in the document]
DL6ER: I think the size of these objects should always be known (`sizeof sqlite3_stmt*`), however, I agree that they are something the API will not be able to access anyway, so I'll move is into a separate FTL-only struct.

src/dnsmasq/cache.c
DL6ER: They are currently logged only as a new query without a "result line". I thought about this as being a good realization at in keeping the dnsmasq changes minimal, however, I agree that having a result logging seems a good idea.

src/dnsmasq/forward.c
```diff
@@ -1981,14 +2003,28 @@ unsigned char *tcp_request(int confd, time_t now,
 	   /* RFC 6840 5.7 */
 	   if (header->hb4 & HB4_AD)
 	     ad_reqd = 1;
+	  
+	  /* Do this by steam now we're not in the select() loop */
+	  check_log_writer(1);
```
DL6ER: `check_log_writer(1);` is moved to before our modification block.

src/dnsmasq_interface.c
DL6ER: The type `bool` isn't available in the majority of `dnsmasq`'s codebase. I didn't want to add `#include <stdbool.h>` all over the place for adding this type without any real benefit as the typical realization is a char on all major platforms IIRC.

src/dnsmasq/forward.c
DL6ER: It is not deleted but moved up to line R2008 (see my review comment above).

src/main.c
DL6ER: This was needed during experiments in this file and I forgot to remove it, thanks for catching this!

src/dnsmasq_interface.c
Mcat12: Would it be possible to use `bool` in our code, since it is just a macro?

DL6ER: No, mixing is not possible. If we use `bool` in the `dnsmasq_interface.c` file, we will also need to use `bool` in the corresponding header file or the prototype will not match. However, we include the file `dnsmasq_interface.h` in `dnsmasq`'s code where `bool` is not available.
We guess should not `typedef char bool` on our own, I think this would be considered bad practice.

We *could* probably add
```c
#include <stdbool.h>
```
at the top of `dnsmasq_interface.h` and by this smuggle the header into `dnsmasq`'s code but I'd first have to test whether the code would still compile and also for possible unwanted side-effects (which I don't think would exist, however). If it turns out to work, we can use `bool` anywhere.

I'll investigate on this tomorrow.

DL6ER: Done

src/dnsmasq/cache.c
DL6ER: The most recent commit adds back logging for blocked domains to `pihole.log`. Implementing it 1:1 to what we have now would require changes in dnsmasq's logging routine, however, I think my new way of logging is close enough (1 line addition):
```
Sep 24 11:32:11 dnsmasq[2081]: query[A] gravity-blocked.test.pi-hole.net from 127.0.0.1
Sep 24 11:32:11 dnsmasq[2081]: gravity blocked gravity-blocked.test.pi-hole.net is 0.0.0.0

Sep 24 11:32:13 dnsmasq[2081]: query[A] blacklist-blocked.test.pi-hole.net from 127.0.0.1
Sep 24 11:32:13 dnsmasq[2081]: exactly blacklisted blacklist-blocked.test.pi-hole.net is 0.0.0.0

Sep 24 11:32:17 dnsmasq[2081]: query[A] regex-blocked.test.pi-hole.net from 127.0.0.1
Sep 24 11:32:17 dnsmasq[2081]: regex blacklisted regex-blocked.test.pi-hole.net is 0.0.0.0
```

src/datastructure.h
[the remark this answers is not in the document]
DL6ER: Checking this, it would require a lot of extra housekeeping. As they are just pointers, they should always be 8 byte on 64bit targets and 4 byte on 32bit targets. The new `regex_enabled` is worth two pointers so either 16 or 8 bytes.

*edit*: Confirmed at least for x86_64:
```
(gdb) p sizeof(sqlite3_stmt*)
$1 = 8
(gdb) p sizeof(client->whitelist_stmt)
$2 = 8

(gdb) p sizeof(bool*[2])
$3 = 16
(gdb) p sizeof(client->regex_enabled)
$4 = 16
```

src/main.c
DL6ER: Done

pralor:

https://discourse.pi-hole.net/t/exclude-certain-lan-addresses-from-filtering/2014/28

pralor:

https://discourse.pi-hole.net/t/allow-temporarily-disabling-pi-hole-by-device/21675/7

pralor:

https://discourse.pi-hole.net/t/per-client-white-and-black-lists/9424/3

pralor:

https://discourse.pi-hole.net/t/apply-pi-hole-blocking-to-cnames/25445/58

DL6ER:
Updated PR description.

src/datastructure.h
```diff
@@ -57,14 +62,18 @@ typedef struct {
...
+	bool *regex_enabled[2];
 } clientsData;
 
 typedef struct {
 	unsigned char magic;
-	unsigned char regexmatch;
 	size_t domainpos;
 	int count;
 	int blockedcount;
+	ucharvec *clientstatus;
```
Mcat12: This vector is not safe to be used by consumers of shared memory, as it contains pointers which point to the process's private memory.

Mcat12: This should also trigger an updated shared memory version.

DL6ER: This is intentional. Realizing this in FTL itself instead of in shared memory is a lot easier. Also, this is more of an internal DNS cache for FTL, nothing an external users should really look at. The contained information (whether a specific query is blocked or not and if, why) is still available as always in the shared queries data.

pralor:

https://discourse.pi-hole.net/t/apply-pi-hole-blocking-to-cnames/25445/108

DL6ER:
To the best of my ability, this is ready for testing and review. To be honest, this code came a very very long way with many busy-due-to-work breaks in between. I might have missed something. Proper testing will show.

src/vector.c
```diff
@@ -0,0 +1,292 @@
...
+	free(v->items);
+	// ...and then then vector itself
+	free(v);
+	v = NULL;
+}
+/********************************* type ucharvec *********************************/
+
+
+/*
+vector *vNew(void)
```
Mcat12: Do we need all of this commented-out code?

DL6ER: Oh, no, obviously not.

My first vector implementation was for arbitrary objects (`void **items`), however, time showed that it was more practical to use an actual storage type rather than one additional level of abstraction. This simplifies the interactions with this dynamic vector and avoids possible de-/referencing pitfalls the compiler might not always be able to catch for the user.

I will remove the commented out code here and elsewhere. It will still live on in the `git` history so we can always go back to it when we need this again.

PR #647, opened 2019-10-02
Restructure network table upgrade procedure

DL6ER:

## 10

---

Ensure to reset (=terminate) running statement before calling `DELETE` that affect the transaction. SQLite3 does not accept such a `DELETE` in a running statement resulting in a failed network table upgrade if there are multiple entries for the same hardware address.

### DIFF
--- a/src/database/network-table.c
+++ b/src/database/network-table.c
@@ -303,7 +303,8 @@ bool unify_hwaddr(void)
 	// The grouping is constrained by the HAVING clause which is
 	// evaluated once across all rows of a group to ensure the returned
 	// set represents the most recent entry for a given hwaddr
-	const char* querystr = "SELECT id,hwaddr FROM network GROUP BY hwaddr HAVING MAX(lastQuery);";
+	// Get only duplicated hwaddrs here (HAVING cnt > 1).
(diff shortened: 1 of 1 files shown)


src/database/network-table.c
```diff
@@ -327,7 +328,11 @@ bool unify_hwaddr(void)
 
 		// Obtain id and hwaddr of the most recent entry for this particular client
 		const int id = sqlite3_column_int(stmt, 0);
-		const char *hwaddr = (const char *)sqlite3_column_text(stmt, 1);
+		char *hwaddr = strdup((char*)sqlite3_column_text(stmt, 1));
+
+		// Finalize statement
+		sqlite3_finalize(stmt);
```
Mcat12: Once the statement is finalized, isn't it undefined behavior to call `sqlite3_step` on it?
>  Any use of a prepared statement after it has been finalized can result in undefined and undesirable behavior such as segfaults and heap corruption.

https://www.sqlite.org/c3ref/finalize.html

DL6ER: You are right, thank you for catching this. `sqlite3_reset()` should be sufficient here.

PR #648, opened 2019-10-02
Use network table as secondary source for client host names

DL6ER:

## 10

---

This PR implements a [feature requested on Discourse](https://discourse.pi-hole.net/t/show-name-in-top-clients-lists-rather-than-ipv6-addresses/23893/).

With this PR we add the `network` table as secondary data source for client host names. This ensures that devices with multiple IPs (on the same network interface) can all be identified by the host name of the machine even if a only one IP address <-> host name association is available.

Especially concerning IPv6 addresses, this will enhance the user experience as host names are now "automatically learned" by FTL.

The principle is the following:
1. For each known client with at least one query to `pihole-FTL`, we query the upstream server (possibly configured using conditional forwarding) for a host name. If such a host name is found, we do not only use it for displaying but we also store it **correlated to the network interface's hardware address** in the `network` table.
2. In case we found no host name for a given IP address (this is commonly the case for IPv6 addresses), we query the `network` table for a host name. Through the path `IP address -> hardware address -> host name` we might be able to find the correct host name even if there is (and maybe was never) a host name correlated to the given IP address.

We implement this lookup in the `network` table intentionally as secondary source of truth to allow users to continue assigning different host names to different IP addresses on the same network interface. It is unclear if anyone used this, however as this was possible before, we think it should still be possible now.

The effect of this PR should be immediately visible to users as FTL should be able to use this new strategy to determine host names for clients it didn't find one for before.

### DIFF
--- a/src/database/network-table.c
+++ b/src/database/network-table.c
@@ -496,3 +496,60 @@ void updateMACVendorRecords(void)
 	sqlite3_finalize(stmt);
 	dbclose();
 }
+
+char* getDatabaseHostname(const char* ipaddr)
+{
+	// Open pihole-FTL.db database file
+	if(!dbopen())
+	{
+		logg("getDatabaseHostname(%s) - Failed to open DB", ipaddr);
+		return strdup("");
+	}
+
+	// Prepare SQLite statement
+	sqlite3_stmt* stmt = NULL;
+	const char *querystr = "SELECT name FROM network WHERE id = (SELECT network_id FROM network_addresses WHERE ip = ?);";
+	int rc = sqlite3_prepare_v2(FTL_db, querystr, -1, &stmt, NULL);
+	if( rc != SQLITE_OK ){
+		logg("getDatabaseHostname(%s) - SQL error prepare (%i): %s",
+		     ipaddr, rc, sqlite3_errmsg(FTL_db));
+		return strdup("");
+	}
+
+	// Bind domain to prepared statement
+	// SQLITE_STATIC: Use the string without first duplicating it internally.
+	// We can do this as domain has dynamic scope that exceeds that of the binding.
+	if((rc = sqlite3_bind_text(stmt, 1, ipaddr, -1, SQLITE_STATIC)) != SQLITE_OK)
+	{
+		logg("getDatabaseHostname(%s): Failed to bind domain (error %d) - %s",
+		     ipaddr, rc, sqlite3_errmsg(FTL_db));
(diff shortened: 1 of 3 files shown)


pralor:

https://discourse.pi-hole.net/t/show-name-in-top-clients-lists-rather-than-ipv6-addresses/23893/11

kevintuhumury:
I just started using Pi-Hole and actually ran into the issue described on Discourse. I noticed this has been merged into the `development` branch two months ago. I was wondering whether there is an ETA on when this is released? The last version released was back in May, so is it going to take a while?

DL6ER:
@kevintuhumury A lot of new features went into the next version of Pi-hole which will become v5.0. We're working hard behind the scenes to get the release out, there are so many features that testing is really extensive. However, it should not take that long until it gets released. We consider `development` to be sufficiently stable at the moment, so you can give it a go if you like.

In case you encounter any issues, we're here to help you and - in fact - would appreciate anything about possible issues before we release ;-) However, mind that going back to `master` from `development` will not be possible without data loss which is also why we bump v4.x -> v5.0.

Prophe-T:
Thanks for trying to improve the user experience and sorry if i'm posting to the wrong place.

The behaviour for this feature is, as stated above:
> 2. In case we found no host name for a given IP address (this is commonly the case for IPv6 addresses), we query the `network` table for a host name.

However, my ISP reserves and resolves a (somewhat dynamic) hostname for any given ipv6 assigned to devices in my home network, like:
dynamic-*ipv6 with dashes*.pool.*ISP*.de
So for me the current behaviour still populates or "clogs" the top-client list and dashboard in general with a whole lot of really long hostnames still belonging to mostly the same devices. Especially since (as is my current understanding, but i'm no expert) with SLAAC, my mobile devices get a new ipv6 every time they wake from wifi sleep.

Is it possible to give the user the option to query the network table for a hostname first for an ipv6 address, instead of trying to query the upstream dns server? Like a checkbox "prefer network table hostnames vor ipv6", thereby doin 2. first, then 1.? This should be optional, since as DL6ER said, not every ISP gives ipv6 hostnames.

yubiuser:
Maybe this is something that could fix your issue:
https://github.com/pi-hole/FTL/pull/784

And the discussion around it:
https://discourse.pi-hole.net/t/incorrect-hostnames/33111

Prophe-T:
Yes, it seems like thats the same thing i'm mentioning. Ty

PR #652, opened 2019-10-13
Update embedded dnsmasq to v2.81

### LINKED ISSUE #603: Cannot use libnettle 3.5

DL6ER:

## 10

---

Important or interesting features for us are **highlighted in bold**.

### New features
- **Support `TCP-fastopen`** (RFC-7413)
  on both incoming and outgoing TCP connections, if supported and enabled in the OS.
- **Add `--shared-network` config.**
  This enables allocation of addresses the DHCP server in subnets where the server (or relay) doesn't have an interface on the network in that subnet.
- **Add `dhcp-ignore-clid` configuration option**
  The idea of this option was already discussed years ago on the mailing
  list: https://dnsmasq-discuss.thekelleys.org.narkive.com/ZoFQNaGo/  always-ignore-client-identifier#post4
  In our production environment, we discovered that some devices are using 'client identifier' not unique at all, resulting on IP addresses conflicts between several devices (we saw up to four devices using same IP address). The root cause is probably a buggy operating system/configuration of devices, but this patch adds a configuration workaround on server side when fixing clients is impossible.
- Support DHCP option 150 (TFTP server address, RFC 5859)
- Support prefixed ranges of IPv6 addresses in `dhcp-host`
- Add tag filtering of `dhcp-host` directives
- Add `--script-on-renewal` option
- Add DHCPv6 ntp-server (56) option handling

### Tweaks
- **Improve cache behavior for TCP connections.**
  For ease of implementation, we always forked a new process to handle each incoming TCP connection. A side-effect of this is that any DNS queries answered from TCP connections are not cached: when TCP connections were rare, this was not a problem. With the coming of DNSSEC, it's now the case that some DNSSEC queries have answers which spill to TCP, and if, for instance, this applies to the keys for the root then those never get cached, and performance is very bad. This fix passes cache entries back from the TCP child process to the main server process, and fixes the problem.
- Ensure that AD bit is reset on answers from `--address=/<domain>/<address>`.
- Remove ability to compile without IPv6 support.
  This was the source of a large number of `#ifdefs`, originally included for use with old embedded `libc` versions. I'm sure no-one wants or needs IPv6-free code these days, so this is a move towards more maintainable code.
- Remove the `NO_FORK` compile-time option, and support for uclinux. In an era where everything has an MMU, this looks like an anachronism, and it adds to (Ok, multiplies!) the combinatorial explosion of compile-time options.
- Free config file values on parsing errors.
- Alter DHCP address selection after `DECLINE` in `consec-addr` mode.
  Avoid offering the same address after a receiving a DECLINE message to stop an infinite protocol loop. This has long been done in default address allocation mode: this adds similar behavior when allocating addresses consecutively.
- Severe restructuring of the internally used datastructures (`all_addr` union, log and rcode fields, etc.)
- **Cache SRV records.**
- Change `read_leases()` to skip invalid entries.
  There's no reason to stop reading the existing lease file when dnsmasq is started and an invalid entry is found, it can just be ignored.  This was fallout from an Openstack bug where the file was being written incorrectly with `[]` around IPv6 addresses.
- **DHCP leases: prune lease as soon as expired**
  We detected a performance issue on a dnsmasq running many dhcp sessions (more than 10 000). At the end of the day, the server was only releasing old DHCP leases but was consuming a lot of CPU.
  On a server with very large number of leases and releasing often sessions, that can waste a very big CPU time.
- **Improve kernel-capability manipulation code** under Linux.
  The code now fails early if a required capability is not available, and tries not to request capabilities not required by its configuration.
- Improve UBus support
- Truncate stupidly large cache sizes.
  If the cache size is very large, the `malloc()` call will overflow on 32 bit platforms and `dnsmasq` will crash. Limit to an order of magnitude less.
- Generalise CNAME handling
    Cope with cached and configured CNAMES for all record types we support, including local-config but not cached types such as TXT.
    Also, if we have a locally configured CNAME but no target for the requested type, don't forward the query.
- Don't waste time caching zero-TTL DNS records
- Generalise locally-configured CNAME handling
    It's now possible for the target of a CNAME to be any locally configured RR or even point to a non-existent RR.
- Check for `SERV_NO_REBIND` on unqualified domains
- Tweaks to TFTP:
    - Fail on overlarge files (block numbers are limited to 16 bits)
    - Honour tftp-max setting in single port mode.
    - Tweak timeouts, and fix logic which suppresses errors if the last ACK is missing.
- Remove experimental DHCPv6 prefix-class support (The standard for this never made it beyond an internet-draft which expired in 2012, so it can be considered dead)
- **Enhance `--conf-dir` to load files in a deterministic order**
- Allow empty server spec in `--rev-server`, to match `--server`
- Optimise closing file descriptors
- Ignore routes in non-main tables (can reduce load with large routing tables)
- Allow overriding of `ubus` service name

### Fixes
Basically all fixes immediately apply to FTL.
- Don't forward `*.bind`/`*.server` queries upstream
  Chaos `.bind` and `.server` (RFC4892) zones are local, therefore don't forward queries upstream to avoid mixing with supported locally and `false` replies with `NO_ID` enabled.
- Fix spurious AD flags in some DNS replies from local config.
- Fix build after `y2038` changes in `glibc`.
- Do not rely on dead code elimination, use array instead.
  Make options bits derived from size and count. Use size of option bits and last supported bit in computation. No new change would be required when new options are added. Just change OPT_LAST constant.
- Keep suitable capabilities if we may bind server sockets to interface or port.
- Fix crash when negative SRV response over TCP gets stored in LRU cache entry.
- Fix too small control array in tftp code on BSD and SOLARIS
  This causes tftp to fail on some BSD versions, for sure. It works by chance on others. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=241068
- Fix crash freeing negative SRV cache entries. 
- Check for `not(DS or DNSKEY)` in `is_outdated_cname_pointer()`
-  Fix entries in /etc/hosts disabling static leases.
  It is possible for a config entry to have one address family specified by a `dhcp-host` directive and the other added from `/etc/hosts`. This is especially common on OpenWrt because it uses `odhcpd` for DHCPv6 and IPv6 leases are imported into `dnsmasq` via a hosts file.
  To handle this case there need to be separate *_HOSTS flags for IPv4 and IPv6. Otherwise when the hosts file is reloaded it will clear the CONFIG_ADDR(6) flag which was set by the `dhcp-host` directive.
- Fix line-counting when reading `/etc/hosts` and friends
  For correct error messages.
- Fix bug in DNS non-terminal code which could sometimes cause a `NODATA` rather than an `NXDOMAIN` reply.
- **Fix compilation against nettle version 3.5 and later.**
  This fixes #603 
- Fix memory leak in `helper.c`
- Always force AD bit to zero in authoritative DNS answers.
- Fix too small control array in tftp code on BSD and SOLARIS
- Don't silently discard all-zeroes addresses in `host-record`
- Fix `dhcp-name-match` to function when name supplied in `dhcp-host`
- Fix bug which gave zero-length DHCPv6 packets if `sendto()` is interrupted
- Fix `dhcp-name-match` to always match client-supplied name
- Avoid RA code trampling on DHCPv6 messages
- Don't send RAs on interfaces without a link-local address (@DL6ER: the source address of the RAs must be the link-local address)
- Do not crash for empty domains in `server=//...` configurations (this bug has been reported by a Pi-hole user, @DL6ER submitted a patch upstream)
- Fix RA problems with two interfaces on same IPv6 subnet
- Fix infinite-loop router advert problems
- Fix problem with netlink socket and TCP DNS 
- Restore ability to answer non-recursive requests
- Fix parameters to `setsockopt()` for `TCP_FASTOPEN`
- Update decline address handling in DHCPv6 for new multi-address world
     When dhcp-host options can have many IPv6 addresses, we need to deal with one of them being declined by a client. The other addresses are still valid
- Fix error in IPv6 prefix calculation
- Fix resource leak on `ubus_init` failure
- Fix nameserver list in auth mode
- 

### DNSSEC related changes and fixes
- Implement RFC-4036 para 5.3.3. rules on TTL values.
- **Fix spurious DNSSEC validation failures** when the auth section of a reply proving that a DS record does not exist contains unsigned RRs. Only the NSEC/NSEC3 records needed to prove the non-existence of the DS record must be signed. 
- Treat DS and DNSKEY queries being forwarded the same as those locally originated.
  The queries will not be forwarded to a server for a domain, unless there's a trust anchor provided for that domain. This allows, especially, suitable proof of non-existence for DS records to come from the parent domain for domains which are not signed.
- Unsigned RRs in auth section proving that a DS doesn't exist are OK.
  In a reply proving that a DS doesn't exist, it doesn't matter if RRs in the auth section _other_ than NSEC/NSEC3 are not signed. We can't set the AD flag when returning the query, but it still proves that the DS doesn't exist for internal use. As one of the RRs which may not be signed is the SOA record, use the TTL of the NSEC record to cache the negative result, not one derived from the SOA.
- Unsigned RRs in the auth section are not bogus.
  Even if they are in a signed zone.
- **Fix crash on `REFUSED` answers to DNSSEC queries.**
  Some `REFUSED` answers to DNSSEC-originated queries would bypass the DNSSEC code entirely, and be returned as answers to the original query. In the process, they'd mess up datastructures so that a retry of the original query would crash dnsmasq.
- Add missing `dump_packet()` for DNSSEC query retries.
- Check for `REFUSED` and `SERVFAIL` replies to DNSKEY queries. 
- Make the existing "insecure DS received" warning more informative by reporting the domain name reporting the issue. This may help identify a problem with a specific domain or server configuration.
- Remove DSA signature verification from DNSSEC, as specified in RFC 8624
- Add support for ED448 DNSSEC signature verification (requires `libnettle 3.6`, yet to be released)
- Support ECC-GOST DNSSEC signature algorithm (requires `libnettle 3.6`, yet to be released)
- Fix rare problem allocating frec for DNSSEC
- Tweak to DNSSEC logging


PromoFaux:
@DL6ER, any reason for this to be still draft status, or can it be merged?

DL6ER:
We're waiting for the official release of `dnsmasq v2.81`. I just keep this updates so

- users can try if an issue they're seeing is already resolved in `dnsmasq`
- it keeps the workload lower for me at release time as I don't have to process hundreds of commits at once

pralor:

https://discourse.pi-hole.net/t/pi-hole-covid-19-and-a-special-easter-gift/30437/1

PR #659, opened 2019-11-21
Add HTTP server and JSON API

DL6ER:

## 10

---

This PR adds a fully-fledged HTTP server and JSON API to FTL.
It will replace `lighttpd` and serve both the API and the next-gen web interface.

More detail is to be added here once this feature advances further.


pralor:

https://discourse.pi-hole.net/t/make-some-advance-api/26786/2

pralor:

https://discourse.pi-hole.net/t/black-whitelisting-does-not-work-on-v5-api/33119/1

DL6ER:
We're still following this. I just rebased it onto the latest `development` branch.

pralor:

https://discourse.pi-hole.net/t/php-webscripts/31016/28

pralor:

https://discourse.pi-hole.net/t/php-webscripts/31016/32

DL6ER:
This branch can now host the web interface. Many things will not work as we removed the rusty telnet interface and added a React-based API in this version of FTL. This will have to be adapted for the current API. Eventually, we will be able to get rid of `api.php` and siblings and have all JS scripts directly talking to FTL.

However, pages not directly depending on the interactive API do already work.

For instance:

Screenshot at 2020-05-29 00-14-58
Screenshot at 2020-05-29 00-20-25

pralor:

https://discourse.pi-hole.net/t/php-webscripts/31016/35

pralor:

https://discourse.pi-hole.net/t/how-to-update-dhcp-with-api/34773/1

davidhessler:
Is there a rough date when this is going to merge?

jojost1:
To add to that: will the old API still exist?

jojost1:
> Which API are you referring to? There are a few different things that people call an API.
The current /api.php? summary, lists, topItems, ... endpoints.

dschaper:
> /api.php? summary, lists, topItems, ... endpoints.

The goal is to have as little disruption as possible and to open even more opportunity for API use in a true API. The URL endpoint may change, but there will be plenty of notification and discussion prior to that, if that needs to happen.

DL6ER:
The PHP-based API (which is what `/api.php?summary`, etc. are based on) will be *replaced* by a proper, more uniform and clear JSON interface. Documentation for this is [being written](https://deploy-preview-338--pihole-docs.netlify.app/api/dns/blocking/) whenever we find the time for it. As @dschaper mentioned, this will by no means be a sudden transition. We will have an extended beta-period preceding the release of the embedded system. However, it brings an enormous amount of benefits and this clearly outweighs the disadvantage of breaking a few scripts. Also, because this allows us to have the API be much more flexible in the future to extend new functionality.

SkYNewZ:
Hi @DL6ER, as a Go developer I would like to start writing a Go SDK for this new API.
I succeeded to install Pi-Hole on a development instance with [this branch](https://github.com/pi-hole/FTL/tree/new/http) with the file `/etc/pihole/ftlbranch`.
image

I also have the latest Pi-Hole documentation with the API part your are working on.
But, request the `/api` endpoint always try to redirect me to the "old" `/admin`:
```
root@pi-hole-dev-instance:~# curl -v http://127.0.0.1/api
* Expire in 0 ms for 6 (transfer 0x56408f477f50)
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Expire in 200 ms for 4 (transfer 0x56408f477f50)
* Connected to 127.0.0.1 (127.0.0.1) port 80 (#0)
> GET /api HTTP/1.1
> Host: 127.0.0.1
> User-Agent: curl/7.64.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-type: text/html; charset=UTF-8
< Expires: Tue, 08 Dec 2020 20:42:10 GMT
< Cache-Control: max-age=0
< Content-Length: 645
< Date: Tue, 08 Dec 2020 20:42:10 GMT
< Server: lighttpd/1.4.53
<
```

Can you help me to set up Pi-Hole in order to have this API and try to starting playing with ?

Thanks.

DL6ER:
I should mention that this API is *very much* work-in-progress. Even when we made a plan how the API should look like, we do not prevent it from maturing on the way. If we realize only later that a specific path should be changes because this just feels more natural (or similar), we take the liberty to change it. While we're documenting this by updating the documentation itself, it would be too much work to write change logs so tracing our changes may be some substantial amount of work (read as: may or may not, no guarantees in any direction). As such, I have the feeling that it may be too early for even staring an SDK, but I will also not stop you :-)

The redirection to `/admin` was added only (rather) recently and there is no exception for `/api`. I'm also not sure if we should use `/api` when the entire rest of the system is inside `/admin` so what we currently have is probably even correct. Users may be hosting different stuff at the root-level of the domain and I'm not sure if we should enforce an immutable endpoint on them.

Everything is still open for discussion in this branch. More eyes can often help so any comments are appreciated.

SkYNewZ:
> If we realize only later that a specific path should be changes because this just feels more natural (or similar), we take the liberty to change it

I truly understand that your job is work in progress. Let's say I don't have a personal project at the moment, and it could be one :)

So all of your current work is available on `/admin` instead of `/api` for now ?

jojost1:
Not sure what scope this PR is and where to request new API features so I'll place a comment here. I'd love API functionality for the following:
- Enable/Disable Black/Whitelist entries
- Detect and start an update
- Get instance temp/cpu/mem (same that's displayed on the web interface)
- Manage groups, adlists, clients
- Manage settings (DNS, DHCP)
- Manage local DNS records

Cheers!

yubiuser:
>Enable/Disable Black/Whitelist entries
Get instance temp/cpu/mem (same that's displayed on the web interface)
Manage groups, adlists, clients

This is already implemented

DL6ER:
@jojost1 The will include *everything* about Pi-hole that needs to be addressed from the web interface. Because we have a very tight security concept, a few things will not be possible (like restarting the system or running updates) because the FTL daemon never runs itself as `root` so it cannot do certain things. I do believe that such a thought-through security concept is an essential thing and more other project should follow this. But, as it makes things considerably more difficult in some locations, I don't really expect them to do similar things.

You may say that updating gravity or restarting the system is working now, however, this is only possible because of some dirty privilege tricks which we'll throw out to sandbox the Pi-hole daemon without the overhead of a "real" sandbox.

yubiuser (commented):
Refering to: https://github.com/pi-hole/FTL/pull/659/commits/774bd0338aa90005e77ff510193aebf65bf9ee2b

Isn't this going to increase the database size a lot?

What use case do you see in keeping those information in the database? E.g. `ttl` might be changing rather often and an old regex (`regex_id`) might be already deleted?

DL6ER:
> Isn't this going to increase the database size a lot?

Yes, experiments I ran over the last few days suggest an increase of 30-40% (the majority is storing the client host name).

It is advantageous as it simplifies the user interface because now it doesn't matter if you query on-disk (long-term) or in-memory (fast 24 hours). The database structures are now identical and both loading and storing becomes are easy as

``` sql
INSERT INTO queries SELECT * FROM disk.queries WHERE timestamp >= ?;
```

``` sql
INSERT INTO disk.queries SELECT * FROM queries WHERE id > ? AND timestamp < ?;
```

Furthermore, the entire content is restored after an FTL restart and there are no artifacts like DNSSEC status always being `UNKNOWN` after a re-import. This is also what `regex_id` wants to cover - full restorability.

I did experiment with not storing `TTL` and `REPLY_TIME` but these two are only a few bytes (if stored at all, not all queries have TTL and reply time (e.g. no reply arrived at all)) and the logic for excluding them and handling this properly increases the complicity beyond what the gain is.

I currently plan to reduce the duration of the database we typically keep from one year to maybe one month. One could even argue to reduce it to one week or even 24 hours (I'm not planing to be that drastic right now).
This will give us enough room to possibly add more stuff in the future. Users can always decide that they still want a specific number of days using [`MAXDBDAYS`](https://docs.pi-hole.net/ftldns/configfile/#maxdbdays). We'd only change the default value.

jfb-pihole:
I agree with reducing MAXDBDAYS to less than a year. I would go no lower than 30 days as the default. Consider 60 or 90.

yubiuser:
I consider 90 days a good range.

> Yes, experiments I ran over the last few days suggest an increase of 30-40% (the majority is storing the client host name).

Could this be a place for normalization? DNSSEC and REPLY will be fairly constant, and even client hostname could benefit from a separate table. This could save some space (although space is cheap...).
Or would the trade-off, more complex queries, be to high?

DL6ER:
> Could this be a place for normalization?

> DNSSEC and REPLY will be fairly constant

Yes, we only store an ID (like we do already for type, status, etc.) so it will be one byte per entry. A linked table could possibly need more storage, but not less.

> client hostname

Yes, for this it would work. However, I kept the IP address in `client` for backwards compatibility. Because we replicate the table in multiple places (two times in-memory and once on-disk), the additional complexity introduced by a linking table would actually be large. Even when we put the client linking-table only in the on-disk database, the storing/restoring SQL statements would be much more complex than what I shows here: https://github.com/pi-hole/FTL/pull/659#issuecomment-848672429

I'd say "storage is cheap" is a clear winner here.

adyanth:
Amazing work here! Is the API stable enough to build on top of? I have plans on building an [external-dns](https://github.com/kubernetes-sigs/external-dns) provider for dynamically modifying DNS records and the legacy API does not support it.

DL6ER:
Thanks @adyanth! After spending significant time on the current release branch (v5.x), focus only recently (end of last week) moved back to the development on this branch. We are currently exploring several independent options for the web interface (AdminLTE, Vue-based, etc.) and reverse the right to tweak the individual API endpoint as we see fit. Reason for this is that we designed the REST API implementation by looking at what v5.x has and transforming this to REST best practices. However, it may turn out that it may be more natural to change a few things (like putting objects in an array rather than individual items in an array) as we develop the frontend.

adyanth:
Understood @DL6ER , thanks! This is the place I can follow for updates right? Or is there a discource/any other place where I can keep myself updated on the progress of this?

DL6ER:
We're posting noteworthy updates here: https://discourse.pi-hole.net/t/pi-hole-v6-0-development-is-well-on-its-way/43314

adyanth:
> We're posting noteworthy updates here: https://discourse.pi-hole.net/t/pi-hole-v6-0-development-is-well-on-its-way/43314

Looks like I either do not have access to it or the link is broken. Will wait for v6.0, thanks!

src/api/docs/content/specs/dns.yaml
```diff
@@ -0,0 +1,169 @@
...
+                  oneOf:
+                  - $ref: '#/components/schemas/errors/no_payload'
+                  - $ref: '#/components/schemas/errors/item_missing'
+          '401':
+            description: Unauthorized
+            content:
+              application/json:
+                schema:
+                  $ref: 'common.yaml#/components/errors/unauthorized'
+    cache:
```
yubiuser, line 68: `/dns/cache` 
retrieves more than `.cache_size`, `.cache_inserted` and `cache_evited`

src/api/ftl.c
```diff
@@ -305,21 +305,28 @@ int get_system_obj(struct ftl_conn *api, cJSON *system)
 	JSON_ADD_ITEM_TO_OBJECT(system, "sensors", sensors);
 
 	// Try to obtain device model
-	FILE *f_model = fopen("/sys/firmware/devicetree/base/model", "r");
-	char model[1024] = { 0 };
-	if(f_model && fgets(model, sizeof(model)-1, f_model))
+	if(file_exists("/sys/firmware/devicetree/base/model"))
```
yubiuser: In `PADD` we have a more extensive way of identifying the device model than just looking at `/sys/firmware/devicetree/base/model`.

Do you think this is something to consider for FTL as well?

https://github.com/pi-hole/PADD/blob/a9681264c83529a2bd826bc86dd933d85b9b15bf/padd.sh#L214-L250

Cleaning of unwanted OEM information could still happen at the client side.

DL6ER: Screenshot from 2023-01-28 07-48-21

yubiuser: O.E.M. is strong with you ;-)

DL6ER: It's a cheap yet extremely powerful Ryzen 5 powered barebone device so one might argue that I am the OEM.

DL6ER:
Switching this to ready for review + merge to development. You can ignore the old builder jobs still marked as required here. We can change them when they landed in development

yubiuser (dismissed):

PromoFaux:
development continues in `development-v6` branch

PR #672, opened 2019-12-30
Add shared DNS cache between forks

DL6ER:

## 10

---

When adding per-client blocking, we added two new layers layers for storing which domain should be blocked for a given client. Unfortunately, this causes unexpected side-effects between TCP workers and the main UDP process where the caches can go out-of-sync, leading to `SEGFAULTS` when one of them tries to access a part of the cache that only lives in one of the children.

The only way to resolve this is to redesign FTL's DNS cache from scratch in a shared memory object. This PR creates two new shared memory objects with different structure, one for regular domains (has a domain been analyzed and decided before for a particular client?) and one for regular expressions (is a particular white/black regex enabled for this client?).

The new DNS cache has the additional benefit of being extremely easy to extend by arbitrary elements. Currently, it holds only the information if a particular domain-client-combination has been seen before and what the result of FTL's analysis way (should this query be blocked? if so, why?).

While working on this, I discovered another bug where clients not associated to any group (this can be achieved by deleting all associations of a client to a group, including the assignment to the default group 0) were still considered to be member of group 0, may leading to unexpected blocking behavior.

This re-implementation removes the dynamic vectors added before as they cannot live within shared memory and are no longer used in the code.


dschaper:
Enabled squash merge for this repo.

PR #677, opened 2020-01-20
Fix for blocking modes

DL6ER:

## 10

---

This PR fixes two bugs in FTL. One was the detection of the address which should be used to reply to a request, the other was the correct synthesis of the DNS header response codes and the requirement of the absence of answer section records for `NODATA` and `NXDOMAIN` blocking modes.

Both issues are fixed by the PR and only appeared when using a different blocking mode than `NULL` (which is Pi-hole's default).


pralor:

https://discourse.pi-hole.net/t/blockingmode-nxdomain-is-being-ignored/27167/9

PR #689, opened 2020-02-06
Ensure each regex is imported only once

DL6ER:
**By submitting this pull request, I confirm the following:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

Fixes a bug brought up on [Discourse](https://discourse.pi-hole.net/t/groups-and-regex/27732).

PR #687, opened 2020-02-04
Fix negative blocked queries on v5.0 beta

DL6ER:
**By submitting this pull request, I confirm the following:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

Fixes https://github.com/pi-hole/AdminLTE/issues/1145

PR #690, opened 2020-02-10
Revert "Ensure each regex is imported only once"

dschaper:
Reverts pi-hole/FTL#689

This commit is causing crashes, reported by multiple users.


pralor:

https://discourse.pi-hole.net/t/ftl-crashes-while-using-the-web-interface/28115/6

DL6ER:
Crashes are caused by a typo in https://github.com/pi-hole/FTL/pull/687

A fix is already up and I asked users to test it. Will open a PR once confirmed that this fixes the issue.

PR #696, opened 2020-02-20
Remove UNKNOWN DNSSEC status and import queries with UNSPECIFIED DNSS…

DL6ER:

## 10

---

Remove `UNKNOWN` DNSSEC status and import queries with `UNSPECIFIED` DNSSEC status (this will not cause a text to be shown on the Query Log).
We do not use the `UNKNOWN` status anywhere else, so we can now get rid of it.


pralor:

https://discourse.pi-hole.net/t/queries-being-responded-as-ok-cached-unknown/28469/11

dschaper:
Is UNSPECIFIED an RFC status or one that we use only internally?

DL6ER:
We use it internally. The RFCs do not specify an "not-yet-known" state.

PR #702, opened 2020-02-29
Allow for graceful filtering of AAAA record requests.

hvindin:

---

This is a little unusual to do on most home networks, but on some networks the DNS resolvers can't handle AAAA queries properly (Lookin at you F5 Wide IP Resolvers....) in the case of those of us who are cursed with GTM Wide IP resolvers doing our DNS Resolution, if we send a query then it takes a full 10 seconds for every resolution. F5 have a KB about it here - https://support.f5.com/csp/article/K7308

While I've just been maintaining this for a while now as a patch I apply to FTL releases I was asked by someone if there was any way to block AAAA queries, having recommended that they use pi-hole on their home network, and it occured to me that there may be other use cases where it is desirable for someone to apply this sort of control to their network, I don't know, I don't know everyone's requirements

It does seem to _kind of_ fit within the purview of pi-hole as the entire project is largely about being able to control what happens on _your_ network. And in the case where it's not necessarily _your_ network, but one managed by people who have definitely kept up to date on their networking knowledge and try and fix the 10s problem by disabling IPv6 everywhere 🤦‍♂ because obviously you can query AAAA records over ipv4... it's also kinda handy.

The dnsmasq upstream don't appear to be overly sympathetic to this problem, as they have stated a number of times, but this feels like a useful "provide flexibility on your network" kind of feature that _sort of_ fits in the purview of pi-hole given that it let's you better control what is going in and out of your network.

And given the minuscule size of the change (and that I know that an almost identical patch to dnsmasq is running in production in multiple extremely large organisations across literally tens of thousands of servers) it feels like it's not likely to do any harm stabilitywise so if it's desirable by the maintainers - here you go 🙂

### DIFF
--- a/src/dnsmasq/dnsmasq.h
+++ b/src/dnsmasq/dnsmasq.h
@@ -266,7 +266,8 @@ struct event_desc {
 #define OPT_TFTP_APREF_MAC 56
 #define OPT_RAPID_COMMIT   57
 #define OPT_UBUS           58
-#define OPT_LAST           59
+#define OPT_FILTER_AAAA    59
+#define OPT_LAST           60
 
 /* extra flags for my_syslog, we use a couple of facilities since they are known 
    not to occupy the same bits as priorities, no matter how syslog.h is set up. */
--- a/src/dnsmasq/option.c
+++ b/src/dnsmasq/option.c
@@ -170,6 +170,7 @@ struct myoption {
 #define LOPT_UBUS          354
 #define LOPT_NAME_MATCH    355
 #define LOPT_CAA           356
+#define LOPT_FILTER_AAAA   357
  
 #ifdef HAVE_GETOPT_LONG
 static const struct option opts[] =  
@@ -341,6 +342,7 @@ static const struct myoption opts[] =
     { "dhcp-rapid-commit", 0, 0, LOPT_RAPID_COMMIT },
     { "dumpfile", 1, 0, LOPT_DUMPFILE },
     { "dumpmask", 1, 0, LOPT_DUMPMASK },
+    { "filter-aaaa", 0, 0, LOPT_FILTER_AAAA },
     { NULL, 0, 0, 0 }
   };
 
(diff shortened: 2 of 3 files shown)


dschaper:
The statement was implied that disabling IPv6 is for people that aren't "modern" in their networking skills so I'll happily play that role.

Question: If I have IPv6 on my network but no ability to get an AAAA response, how do I use IPv6 hosts without using just a bare IPv6 address?

hvindin:
It's really more of a "Can I do an AAAA record lookup without IPv6"

Should you? probably not, it's not likely to be useful.

But does most software try and get an A and an AAAA record when performing name resolution? It seems to be the case.

So disabling IPv6 on a network would make those AAAA records _definitely_ useless, but tends to ignore the issue where an AAAA lookup can still happen and if you happen to not want to query for AAAA records there are surprisingly few ways to do it.

If you go and set all your `sysctl` ipv6 config to make ipv6 disabled and then just run a `getent` for some random host (having not specifically used a ahostsv4 lookup, but instead a regular old hosts lookup) you will still see queries for AAAA records in the logs.

In the particular case that I've hit a couple of times on very large internal networks it's almost always a misbehaving F5 GTM that screws you, it waits for 10s to return any response even if it already has an A record to give you.

As I mentioned, this isn't really a "everyone would want to use it all the time" sort of flag, it's a fairly limited amount of networks that actually have this problem, but it's also an exceedingly low risk change given how many production environments I know this patch to dnsmasq is running in and it doesn't actually require any change for things to just keep doing what they are doing today, if you happened to want to avoid AAAA lookups entirely then you could block them from going to your upstream DNS servers locally if you decided you wanted to do that.

dschaper:
The payload of an AAAA record is an IPv6 address and IPv6 transit is disabled on the network, I really don't see the utility in diverging from the upstream code for this. The record type is valid and I don't want to start breaking RFCs with disabling certain record types. 

Have you approached `dnsmasq` and Simon Kelley to see if there's any reception to the idea in the source application?

hvindin:
the upstream response to this functionality has been, quite sensibly, reasonably similar to yours.

I'm not overly attached to this PR being merged in, I've just been maintaining this patch for the fundamentally broken networks that I occassionally need to deal with and thought I would check if there was any sort of appetite to include this as a "apply more controls to the network you are running pihole on" perspective.

I entirely agree that it shouldn't be something that would ever be required on a network that wasn't already broken :laughing:

DL6ER:
> I entirely agree that it shouldn't be something that would ever be required on a network that wasn't already broken laughing

Closing based on this comment. You can obviously always build your own binary based on this patch whenever you need it. It just doesn't seem like we want to include this in the official code. Thanks for your contribution nonetheless!

PR #708, opened 2020-03-04
Implement subnet support for clients

DL6ER:

## 10

---

This PR implements subnet support for clients. Previously, only exact client / group associations have been possible, like `192.168.0.1`. This PR allows for arbitrarily complex subnet configurations using the widely known [Classless Inter-Domain Routing (CIDR) notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_blocks).

This allows users to specify "broad clients" such as
- `192.168.1.0/24` which will match al clients in the range `192.168.1.1` to `192.168.1.255` (256 devices),
- `10.8.0.0/16` will match all clients in the range `10.8.0.1` to `10.8.255.255` (65,536 devices),
- `192.168.100.0/22` represents the 1024 IPv4 addresses from `192.168.100.0` to `192.168.103.255`.

This implementation is IPv6-aware and can be used for IPv6 subnets as well, such as the IPv6 block `2001:db8::/48` representing the block of IPv6 addresses from `2001:db8:0:0:0:0:0:0` to `2001:db8:0:ffff:ffff:ffff:ffff:ffff` (1,208,925,819,614,629,174,706,176 devices).

Our implementation is more generic than what is written on the linked Wikipedia article as you can use *any* CIDR block (not only multiples of 4).


pralor:

https://discourse.pi-hole.net/t/subnetting-doesnt-appear-to-work-for-groups/28813/25

pralor:

https://discourse.pi-hole.net/t/netzwerk-in-clientliste-hinzufugen/34779/2

PR #710, opened 2020-03-07
Internal resolver tweaks

DL6ER:

## 10

---

In order to be able to resolve host names specified in Custom DNS, `/etc/hosts`, etc. we have to ask FTL itself when trying to find out host names to IP addresses by generating PTR requests for this. Previously, this was automatically done, however, with the removal of forcing the local nameserver to `127.0.0.1` we lost the ability to resolve only locally known host names temporarily. We reinstalled the ability by forcing FTL as the *first* resolver on the system. As this effectively overwrote `/etc/resolv.conf` (even if only inside FTL itself), it had adverse effect on docker instances where local name resolution failed for some special devices.

This PR addresses the issue in a (user-confirmed) docker-friendly way by adding FTL as *last* resolver.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -358,6 +358,19 @@ void read_FTLconf(void)
 	else
 		logg("   CNAME_DEEP_INSPECT: Inactive");
 
+	// FORCE_LOCAL_RESOLVER
+	// defaults to: true
+	config.force_local_resolver = true;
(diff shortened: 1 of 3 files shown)


pralor:

https://discourse.pi-hole.net/t/ptr-lookups-for-docker-networks/28753/15

DL6ER:
Updated description.

PR #711, opened 2020-03-07
Add all clients FTL knows about to the network table

DL6ER:

## 10

---

Add all clients FTL knows about to the network table. This requires a yet to be done modification to the web interface to hide the mock hardware addresses we generate for these devices.

This change will import distant clients not visible over ARP/neigh (like VPN connected or VLAN separated devices) in the network table. As there is no one unique identification criterion now (we used the MAC address for this, before), each IP change of a non-deterministic DHCP server will fill the network table with additional entries as there is no way to detect if client `.3` was `.5` yesterday without the MAC address.

Maybe we need a setting for this defaulting to `false` which can then be enabled by power-users running VPNs, VLANs, etc. However, "normal" (as in: no power-) users will likely not have any clients in their network which are not also covered by ARP/neigh, so there may be no issue even with having this defaulting to `true`.

Note that clients found through ARP/neigh will always have higher priority than "remaining" clients known to FTL. FTL's knowledge is only used to "fill the gaps".


pralor:

https://discourse.pi-hole.net/t/request-ability-to-add-clients-to-known-client-list-from-network-page/28601/10

pralor:

https://discourse.pi-hole.net/t/long-term-data-weirdness/29308/17

DL6ER:
This PR also adds host name validation (length & valid characters) to FTL

pralor:

https://discourse.pi-hole.net/t/long-term-data-weirdness/29308/30

pralor:

https://discourse.pi-hole.net/t/include-clients-from-different-subnets-vlans-in-network-overview/27967/17

PR #716, opened 2020-03-24
Add DELAY_STARTUP setting to delay startup of the embedded dnsmasq

DL6ER:

## 10

---

Adding a supported way of delaying the startup of the embedded `dnsmasq`. Even though it would be good to solve this otherwise, it is still better to do this in a supported way instead of hacking `cron` jobs or modify the `init.d` files which makes debugging much harder.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -358,6 +358,19 @@ void read_FTLconf(void)
 	else
 		logg("   CNAME_DEEP_INSPECT: Inactive");
 
+	// DELAY_STARTUP
+	// defaults to: zero (seconds)
+	buffer = parse_FTLconf(fp, "DELAY_STARTUP");
+
+	config.delay_startup = 0;
+	if(buffer != NULL && sscanf(buffer, "%u", &config.delay_startup) &&
+	   (config.delay_startup > 0 && config.delay_startup <= 300))
+	{
+		logg("   DELAY_STARTUP: Requested to wait %u seconds during startup.", config.delay_startup);
+	}
+	else
+		logg("   DELAY_STARTUP: No delay requested.");
+
 	// Read DEBUG_... setting from pihole-FTL.conf
 	read_debuging_settings(fp);
 
--- a/src/config.h
+++ b/src/config.h
@@ -21,6 +21,7 @@ typedef struct {
 	int DBinterval;
 	int port;
 	int maxlogage;
+	unsigned int delay_startup;
 	int16_t debug;
 	unsigned char privacylevel;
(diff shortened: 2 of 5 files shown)


pralor:

https://discourse.pi-hole.net/t/add-option-to-delay-pihole-startup-to-set-up-wireguard-interfaces-before-correctly/29715/2

pralor:

https://discourse.pi-hole.net/t/have-to-pihole-restartdns-after-reboot/28772/25

dschaper:
The log display is only when debugging mode is enabled? I don't think we need to have a countdown in general use.

yubiuser:
With all DEBUG options disabled I still get

```
[2020-03-24 19:45:27.823 3888]    DELAY_STARTUP: Requested to wait 15 seconds during startup.

2020-03-24 19:45:28.039 3888] Sleeping as requested ...
[2020-03-24 19:45:28.039 3888]    15
[2020-03-24 19:45:31.042 3888]    12
[2020-03-24 19:45:34.046 3888]     9
[2020-03-24 19:45:37.049 3888]     6
[2020-03-24 19:45:40.052 3888]     3
[2020-03-24 19:45:43.055 3888] Done!
```

dschaper:
Thanks. The countdown loop may have been useful during testing but I don't see a real need for it. And longer delay values are just going to ramp up the output. We grab the head and tail of the log for debugging and this output will push needed information out of the log capture.

yubiuser:
I get an error switching back from this branch to `new/all_clients_network_table`. Pihole complains about checksum error... :-/

DL6ER:
We only show five steps in between, no matter how long the delay is. If you still find this too much of output, we can remove this. I felt it is a nice way of showing at least a bit that FTL is still there, waiting for you when you put in a large delay like, lets say, one minute.

yubiuser:
If your primary intent is to let the user know that FTL is still there you could change it to terminal output. Then there would be no need to log it.

DL6ER:
FTL is typically invoked as daemon, the terminal output goes nowhere. We even close all terminal connections in daemon mode as a security measure.

DL6ER:
I force-pushed the timer away. Output will now be (snippet):
```plain
[2020-03-25 10:07:18.812 32462] Sleeping for 10 seconds as requested by configuration ...
[2020-03-25 10:07:28.812 32462] Done sleeping, continuing startup of resolver...
```

If `DELAY_STARTUP` is not set (or set to zero), i.e., the default, no output will be shown at all.

PR #722, opened 2020-04-01
Restored asterisk match for auditlog

silibum:

**How familiar are you with the codebase?:** _3_

---

My auditlog includes things like `*.domain.local` and `*.arpa` and as those entries were converted to the databse while testing the v5.0 branch I noticed that my auditlog show'd up with tons of 'wrong' entries and I decided to put some research into the new database based version of the auditlog:

The current stable version has an asterisk matching for the first char of an auditlog.list line: If the line starts with * a substring from the linevalue will be used to check if the current domain contains the entry. This feature wasn't there anymore as the sqlversion strictly uses 'domain = ?' for the where-clause...

To avoid changes in 'domain_in_list' I decided to replace the auditlist_stmt: 

`SELECT EXISTS(SELECT domain, CASE WHEN substr(domain, 1, 1) = '*' THEN '*' || SUBSTR(?, - length(domain) + 1) ELSE SUBSTR(?, - length(domain)) END matcher FROM domain_audit WHERE matcher = domain);`

I'm open for a better solution (or query) - Please feel free to improve my solution.

---

*hope it's all at the right branch/place now as I noticed my earlier request was done against release/v5.0 and also I now used my local branch to commit my change as the github web gui wasn't the best way... - sorry for the trouble - first timer with pull request via github...*

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -72,7 +72,7 @@ bool gravityDB_open(void)
 	}
 
 	// Prepare audit statement
-	rc = sqlite3_prepare_v2(gravity_db, "SELECT EXISTS(SELECT domain from domain_audit WHERE domain = ?);", -1, &auditlist_stmt, NULL);
+	rc = sqlite3_prepare_v2(gravity_db, "SELECT EXISTS(SELECT domain, CASE WHEN substr(domain, 1, 1) = '*' THEN '*' || substr(?, - length(domain) + 1) else  substr(?, - length(domain)) END matcher FROM domain_audit WHERE matcher = domain);", -1, &auditlist_stmt, NULL);
 	if( rc != SQLITE_OK )
 	{
 		logg("gravityDB_open(\"SELECT EXISTS(... domain_audit ...)\") - SQL error prepare (%i): %s", rc, sqlite3_errmsg(gravity_db));


DL6ER:
Thanks for your contribution, against `release/v5.0` would actually be the right target branch at this point to get it still included in the upcoming v5.0 release. Submitting to `development` is always proposing code contributions towards the *next* release, which will be v5.1 as v5.0 has already been branched out.

I looked at your proposed change and wonder if you tested it carefully enough. No blaming at all here, I only ask because, when executing this SQL later during scanning of the audit table, we bind `domain` only to the first `?` as we needed it only once:
```sql
SELECT EXISTS(
  SELECT domain from domain_audit WHERE domain = ?
);
```
With your proposed query, however, `domain` would have to be bound *two* times (two `?`). This will lead to that the second `?` will be `NULL` and exact matching (no wildcard character) will not work with your proposed change.
The solution will be to bind the same string two times or use named SQL parameters here. In this case, I prefer the latter (cleaner approach). ~~See my suggestion below.~~

Furthermore, I think we should add the SQL with some indentations and also add a comment for how it is thought to work. I initially planned to show you in here how to improve your code, however, Github won't let me suggest changes to files you haven't edited, so I will open a competing pull request, preserving your initial commit as good as possible to retain your authorship on this idea!

Thank you for using the Pi-hole and your code contribution which is helping us making Pi-hole a better software for us all!

DL6ER:
Thanks for your contribution, against `release/v5.0` would actually be the right target branch at this point to get it still included in the upcoming v5.0 release. Submitting to `development` is always proposing code contributions towards the *next* release, which will be v5.1 as v5.0 has already been branched out.

I looked at your proposed change and wonder if you tested it carefully enough. No blaming at all here, I only ask because, when executing this SQL later during scanning of the audit table, we bind `domain` only to the first `?` as we needed it only once:
```sql
SELECT EXISTS(
  SELECT domain from domain_audit WHERE domain = ?
);
```
With your proposed query, however, `domain` would have to be bound *two* times (two `?`). This will lead to that the second `?` will be `NULL` and exact matching (no wildcard character) will not work with your proposed change.
The solution will be to bind the same string two times or use named SQL parameters here. In this case, I prefer the latter (cleaner approach). ~~See my suggestion below.~~

Furthermore, I think we should add the SQL with some indentations and also add a comment for how it is thought to work. I initially planned to show you in here how to improve your code, however, Github won't let me suggest changes to files you haven't edited, so I will open a competing pull request, preserving your initial commit as good as possible to retain your authorship on this idea!

Thank you for using the Pi-hole and your code contribution which is helping us making Pi-hole a better software for us all!

DL6ER:
Superseded by #727

PR #725, opened 2020-04-06
Restore fork-safety

DL6ER:

## 10

---

FTL is a complex multi-dimensional application. The main process has a number of threads surrounding it, each of which performing specific duties (like updating the database, resolving host names and parsing the network (ARP) cache, ...). To make things even more complicated, FTL adds yet another level of parallelism, namely it forks the main process in order to maintain individual DNS clients connecting over a persisting TCP connections. The forking is done to ensure a non-blocking DNS resolution even in the presence of persistent connections.

During the investigation of #698, we have been remindet that SQLite3 is [not fork-safe](https://www.sqlite.org/howtocorrupt.html#_carrying_an_open_database_connection_across_a_fork_) even when it is thread-safe. This PR fixes a possible crash as, previously, FTL tried to use the same prepared statements and database connections across forks. As this is not a good idea, we change this with the given PR to individual database connections and prepared statements for different forks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ FTL_DEPS = *.h database/*.h api/*.h version.h
 FTL_DB_OBJ = database/common.o database/query-table.o database/network-table.o database/gravity-db.o database/database-thread.o database/sqlite3-ext.o
 FTL_API_OBJ = api/socket.o api/request.o api/msgpack.o api/api.o
 FTL_OBJ = $(FTL_DB_OBJ) $(FTL_API_OBJ) main.o memory.o log.o daemon.o datastructure.o signals.o files.o setupVars.o args.o gc.o config.o \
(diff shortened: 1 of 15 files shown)


dschaper:
Okay to squash?

DL6ER:
I'd rather like to keep the commits. There are no minor commits, also the seemingly unnecessary commit  7d3e960 should stay in there so we can check with `git blame` *why* we added this exactly. This will avoid similar issues when coding something similar again in the future when we have long forgotten about such musl specifics.
Furthermore, c9eb4a1 removes subfeatures of the dynamic vectors we don't need right now (as we always address vector components absolutely. However, in the future we may want to have relative vectors ("append to the end") and we may also want to delete components (incl. moving the remaining vector to avoid "holes"). If we'd squash the commits, we'd effectively wipe these functions, I already wrote and tested out of Github's memory.

PR #727, opened 2020-04-07
Restore audit wildcard-support

DL6ER:

## 10

---

With Pi-hole v4.x we had wildcard-support in the audit log. When migrating this list into the database, we lost this ability. This PR is the successor of #722 which aimed at restoring this feature. A few changes were necessary that exceeded Github's ability to suggest changes so I moved this into a new PR, preserving the OP's authorship of the first commit.

This PR is milestoned for still being included in the v5.0 release.

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -126,7 +126,7 @@ bool gravityDB_open(void)
 	// Prepare audit statement
 	if(config.debug & DEBUG_DATABASE)
 		logg("gravityDB_open(): Preparing audit query");
-	rc = sqlite3_prepare_v2(gravity_db, "SELECT EXISTS(SELECT domain from domain_audit WHERE domain = ?);", -1, &auditlist_stmt, NULL);
+	rc = sqlite3_prepare_v2(gravity_db, "SELECT EXISTS(SELECT domain, CASE WHEN substr(domain, 1, 1) = '*' THEN '*' || substr(?, - length(domain) + 1) else  substr(?, - length(domain)) END matcher FROM domain_audit WHERE matcher = domain);", -1, &auditlist_stmt, NULL);
 	if( rc != SQLITE_OK )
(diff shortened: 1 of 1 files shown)


DL6ER:
@silibum Please review my proposed changes.

You can even test this yourself when running
```
pihole checkout ftl new/audit_wildcards
```
(note that this does only work when you are already on the beta testing branches).

I changed three things in here:

1. Added indentation and comments for clarity (always a good thing)
2. Changed from numbered to names SQL parameters (this allows us to only have to bind once), I added a comment about how this works further down in the same file, where the domain is actually bound to the prepared statement
3. I fixed the `ELSE` clause from `substr(:input, - length(domain))` to just `:input`.
    Otherwise, the `else` clause would cause wildcard-behavior as well, e.g. `local` in the database would cause all input domains to be truncated from the right, i.e. `"something.local" -> "local"` which would then be a **false positive**.

silibum:
Hey,

looks good. Compiled and running without any issue - I "missed" to test the non prefixed version: I have 122 entries in my table and just noticed that are 0 non prefixed... I don't user the AdminLTE for adding to the list (now I use phpLiteAdmin and before I edited the .list-file) - Wrong testing 👎 

Comments - Yeah not everyone knows what that query does - just thought it was self explaining enough but better to comment that query a bit.

Thanks for the fixing/improving - looks solid. Credits aren't important to me (specially because this wasn't 100% tested before submitting) - I (and I'm sure many others) just like to retain wildcard-support with v5.x ;-)

DL6ER:
Then this is ready for review + approval + merge.

Git can automatically preserve authorship when you know how to do it, see the details of the first commit in this PR:
Screenshot from 2020-04-08 10-32-56

pralor:

https://discourse.pi-hole.net/t/wildcard-auditing/6080/7

PR #733, opened 2020-04-21
Automatically block _esni.* subdomains of blocked domains

DL6ER:

## 10

---

[Encrypted Server Name Indication (ESNI)](https://tools.ietf.org/html/draft-ietf-tls-esni-06) is certainly a good step into the right direction to enhance privacy on the web. It prevents on-path observers, including ISPs, coffee shop owners and firewalls, from intercepting the TLS Server Name Indication (SNI) extension by encrypting it. This prevents the SNI from being used to determine which websites users are visiting.

ESNI will obviously cause issues for `pixelserv-tls` which will be unable to generate matching certificates on-the-fly when it cannot read the SNI. Cloudflare and Firefox are already enabling ESNI.

According to the IEFT draft, we can easily restore `piselserv-tls`'s operation by replying `NXDOMAIN` to `_esni.*` subdomains of blocked domains as this mimics a "not configured for this domain" behavior.

Discourse feature request: https://discourse.pi-hole.net/t/support-dropping-esni-records-for-blocked-domains/13250

Note that this feature may be disabled by setting `BLOCK_ESNI=false` in `/etc/pihole/pihole-FTL.conf`.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -372,6 +372,19 @@ void read_FTLconf(void)
 	else
 		logg("   DELAY_STARTUP: No delay requested.");
 
+	// BLOCK_ESNI
+	// defaults to: true
+	config.block_esni = true;
+	buffer = parse_FTLconf(fp, "BLOCK_ESNI");
+
+	if(buffer != NULL && strcasecmp(buffer, "false") == 0)
+		config.block_esni = false;
+
+	if(config.block_esni)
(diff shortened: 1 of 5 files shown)


pralor:

https://discourse.pi-hole.net/t/support-dropping-esni-records-for-blocked-domains/13250/10

DL6ER:
Confirmed working on Discourse.

PR #780, opened 2020-05-23
Add warning for invalid hostnames to FTL message table

DL6ER:

## 10

---

Add a message for Pi-hole diagnostics when host names with invalid characters have been found during FTL's internal name resolution (typically client host names containing non-ASCII characters).

Example:
```
$ grep /etc/hosts "127.0.0.53
127.0.0.53 Cöfte
```

results in
id|timestamp|type|message|blob1|blob2|blob3|blob4|blob5
--|--|--|--|--|--|--|--|--
98|1590222660|HOSTNAME|127.0.0.53|Cöfte|1|||

where `message (TEXT)` = IP of the host names, `blob1 (UTF8 TEXT)` = the obtained host names, `blob2 (INT)` = position of the invalid character

### DIFF
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -24,8 +24,14 @@ sqlite3 *FTL_db = NULL;
 bool database = true;
 bool DBdeleteoldqueries = false;
 long int lastdbindex = 0;
+static bool db_avail = false;
 
 static pthread_mutex_t dblock;
+
+__attribute__ ((pure)) bool FTL_DB_avail(void)
+{
+	return db_avail;
+}
 
 void dbclose(void)
 {
@@ -36,6 +42,8 @@ void dbclose(void)
 		rc = sqlite3_close(FTL_db);
 		FTL_db = NULL;
 	}
+
+	db_avail = false;
 
 	// Report any error
 	if( rc != SQLITE_OK )
@@ -85,6 +93,8 @@ bool dbopen(void)
 		dbclose();
 		return false;
 	}
+
+	db_avail = true;
 
 	return true;
 }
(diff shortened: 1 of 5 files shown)


yubiuser:
Starting counting for blob 2 with zero might irritated some users as it is more natural starting counting with one in natural languages.
Invalide character is ö and most users will tend to see it on position 2.

DL6ER:
I agree, however, users should not really interpret the message table themselves, see https://github.com/pi-hole/AdminLTE/pull/1356

DL6ER:
Furthermore, as both languages in involved here (`C` and `Javascript`) count from zero and this table is not supposed to be used directly by the user, it seems strange to do an artificial `+1` in `C` followed by a similarly artificial `-1` in JS.

However, thanks for your input!

yubiuser:
> I agree, however, users should not really interpret the message table themselves, see pi-hole/AdminLTE#1356

Ok :-) Haven't imagine such an overkill presentation of the error message

DL6ER:
I will do one more optimization concerning locks, until then, this PR is again in draft mode. Feel free to still merge the AdminLTE PR, they are completely independent and it would anyways be better if the AdminLTE PR is merged before.

yubiuser:
What happens if there is more than one illegal character?

DL6ER:
We only check the string until we find the first invalid character. This is also the one we report to the user. I don't think we need to add multi-error handling. When the fix the first incorrect character (and there are multiple), the next iteration will yield the next invalid character.

PromoFaux:
image

@DL6ER did you forget to sign a commit?

DL6ER:
Re-signed all commits using
```
git rebase --exec 'git commit --amend --no-edit -n -S' -i development
```

DL6ER:
@PromoFaux Kindly asking for a re-approval

pralor:

PR #781, opened 2020-05-24
Add cmake build

fhriley:
Signed-off-by: Frank Riley <fhriley@gmail.com>

**How familiar are you with the codebase?:** 5

---

This implements a full cmake build. I believe it fully replicates the existing Makefile build. A cmake build is great because it makes it much easier for new developers to get started. For example, with this commit I can clone FTL in CLion and within seconds, I can be deploying and debugging on one of my remote targets.

### DIFF
--- a/src/database/CMakeLists.txt
+++ b/src/database/CMakeLists.txt
@@ -15,8 +15,6 @@ set(SQLITE_DEFINES
         )
 
 set(sqlite3_sources
-        sqlite3-ext.c
-        sqlite3-ext.h
         sqlite3.c
         sqlite3.h
         )
@@ -38,6 +36,8 @@ set(database_sources
         network-table.h
         query-table.c
         query-table.h
+        sqlite3-ext.c
+        sqlite3-ext.h
         )
 
 add_library(database OBJECT ${database_sources})
--- a/src/database/CMakeLists.txt
+++ b/src/database/CMakeLists.txt
@@ -1,0 +1,44 @@
+# SQLITE_OMIT_LOAD_EXTENSION: This option omits the entire extension loading mechanism from SQLite, including sqlite3_enable_load_extension() and sqlite3_load_extension() interfaces. (needs -ldl linking option, otherwise)
+# SQLITE_DEFAULT_MEMSTATUS=0: This setting causes the sqlite3_status() interfaces that track memory usage to be disabled. This helps the sqlite3_malloc() routines run much faster, and since SQLite uses sqlite3_malloc() internally, this helps to make the entire library faster.
+# SQLITE_OMIT_DEPRECATED: Omitting deprecated interfaces and features will not help SQLite to run any faster. It will reduce the library footprint, however. And it is the right thing to do.
+# SQLITE_OMIT_PROGRESS_CALLBACK: The progress handler callback counter must be checked in the inner loop of the bytecode engine. By omitting this interface, a single conditional is removed from the inner loop of the bytecode engine, helping SQL statements to run slightly faster.
+# SQLITE_DEFAULT_FOREIGN_KEYS=1: This macro determines whether enforcement of foreign key constraints is enabled or disabled by default for new database connections.
+# SQLITE_DQS=0: This setting disables the double-quoted string literal misfeature.
+set(SQLITE_DEFINES
+        SQLITE_OMIT_LOAD_EXTENSION
+        SQLITE_DEFAULT_MEMSTATUS=0
+        SQLITE_OMIT_DEPRECATED
+        SQLITE_OMIT_PROGRESS_CALLBACK
+        SQLITE_OMIT_MEMORYDB
+        SQLITE_DEFAULT_FOREIGN_KEYS=1
+        SQLITE_DQS=0
+        )
+
+set(sqlite3_sources
+        sqlite3-ext.c
+        sqlite3-ext.h
+        sqlite3.c
+        sqlite3.h
+        )
+
+add_library(sqlite3 OBJECT ${sqlite3_sources})
+target_compile_definitions(sqlite3 PRIVATE ${SQLITE_DEFINES})
+target_compile_options(sqlite3 PRIVATE -Wno-implicit-fallthrough -Wno-cast-function-type)
+
+set(database_sources
+        common.c
+        common.h
+        database-thread.c
+        database-thread.h
+        gravity-db.c
+        gravity-db.h
+        message-table.c
+        message-table.h
+        network-table.c
+        network-table.h
+        query-table.c
+        query-table.h
+        )
+
+add_library(database OBJECT ${database_sources})
+target_compile_options(database PRIVATE "${EXTRAWARN}")
--- a/src/database/CMakeLists.txt
+++ b/src/database/CMakeLists.txt
@@ -1,26 +1,12 @@
-# SQLITE_OMIT_LOAD_EXTENSION: This option omits the entire extension loading mechanism from SQLite, including sqlite3_enable_load_extension() and sqlite3_load_extension() interfaces. (needs -ldl linking option, otherwise)
-# SQLITE_DEFAULT_MEMSTATUS=0: This setting causes the sqlite3_status() interfaces that track memory usage to be disabled. This helps the sqlite3_malloc() routines run much faster, and since SQLite uses sqlite3_malloc() internally, this helps to make the entire library faster.
-# SQLITE_OMIT_DEPRECATED: Omitting deprecated interfaces and features will not help SQLite to run any faster. It will reduce the library footprint, however. And it is the right thing to do.
-# SQLITE_OMIT_PROGRESS_CALLBACK: The progress handler callback counter must be checked in the inner loop of the bytecode engine. By omitting this interface, a single conditional is removed from the inner loop of the bytecode engine, helping SQL statements to run slightly faster.
-# SQLITE_DEFAULT_FOREIGN_KEYS=1: This macro determines whether enforcement of foreign key constraints is enabled or disabled by default for new database connections.
-# SQLITE_DQS=0: This setting disables the double-quoted string literal misfeature.
-set(SQLITE_DEFINES
-        SQLITE_OMIT_LOAD_EXTENSION
-        SQLITE_DEFAULT_MEMSTATUS=0
-        SQLITE_OMIT_DEPRECATED
-        SQLITE_OMIT_PROGRESS_CALLBACK
-        SQLITE_OMIT_MEMORYDB
-        SQLITE_DEFAULT_FOREIGN_KEYS=1
-        SQLITE_DQS=0
-        )
-
+# This file has no dependencies on other code in the repo and takes a long time
+# to build. It is placed in its own target so that it does not include any other
+# headers in the build command and thus does not need to be rebuilt when headers
+# are modified.
 set(sqlite3_sources
         sqlite3.c
-        sqlite3.h
         )
 
 add_library(sqlite3 OBJECT ${sqlite3_sources})
-target_compile_definitions(sqlite3 PRIVATE ${SQLITE_DEFINES})
 target_compile_options(sqlite3 PRIVATE -Wno-implicit-fallthrough -Wno-cast-function-type)
 
 set(database_sources
@@ -36,9 +22,11 @@ set(database_sources
         network-table.h
         query-table.c
         query-table.h
+        sqlite3.h
         sqlite3-ext.c
         sqlite3-ext.h
         )
 
 add_library(database OBJECT ${database_sources})
 target_compile_options(database PRIVATE "${EXTRAWARN}")
+target_include_directories(database PRIVATE ${PROJECT_SOURCE_DIR}/src)
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,8 @@ pihole-FTL
 version*
 
 # CMake files
-/CMakeLists.txt
 /cmake-build-debug/
+/cmake-build-release/
 
 # IDE files
 .idea/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,6 @@
+cmake_minimum_required(VERSION 3.1)
+project(PIHOLE_FTL C)
+
+set(DNSMASQ_VERSION pi-hole-2.81)
+
+add_subdirectory(src)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,0 +1,160 @@
+set(CMAKE_C_STANDARD 11)
+
+# Default to a release with debug info build
+if (NOT EXISTS ${CMAKE_BINARY_DIR}/CMakeCache.txt)
+    if (NOT CMAKE_BUILD_TYPE)
+        set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "" FORCE)
+    endif()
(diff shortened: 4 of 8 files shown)


dschaper:
The fails are on the upload step, it's just a CI issue.

DL6ER:
Further testing results show:
```
[ 73%] Building C object src/CMakeFiles/pihole-FTL.dir/capabilities.c.o
```
while this file should be places outside the `src` directory as `build/capabilities.o`. I don't mind whether it is `.o` or now `.c.o` if it is consistent.

DL6ER:
I just tried to compile the binary in our `arm` container. It works as expected, however, I see that the binary is places in `src/` which is also something we'd need to change. Nothing should be put into the `src` subdirectory.

We can discuss if the binary should be placed into the `build` folder as all the other objects. I think this would be consistent. However, this would also need to be considered in the tests as well so the test suite can find the `pihole-FTL` executable.

fhriley:
> 1. We should not have two concurrent building systems
>    This adds maintenance workload and would likely lead to things working one but not the other way, eventually, hence we should use _either_ `make` _or_ `cmake`. Also because `cmake .` will overwrite the existing `Makefile`.
> 2. Using `cmake` will require changes to the CI commands, located [here](https://github.com/pi-hole/FTL/blob/3d7c095cb10a1b692c80dd28f6051dea91d45baa/.circleci/config.yml#L17) to use `cmake` instead of `make`. Furthermore, we need to add the package `cmake` to our CI docker builders.

Agreed. I see this as a two step process. 1) Get the cmake build vetted and checked in and then 2) Convert the CI over to cmake. Also, I don't have any experience with CircleCI (we use Gitlab) so I didn't want to change that just yet. I think the transition should be pretty simple so I can start looking at it. I probably don't have access to do any testing though?

> Also because `cmake .` will overwrite the existing `Makefile`.

Typical usage of cmake is to do an out of source build in a directory of your choosing so nothing in the source is modified. There is one exception to that, which is the `version.h` file that gets generated.

> 3. Release build should contain debugging symbols.
>    I know this is somewhat uncommon, however, many users aren't able to work with debuggers and, hence, we have a pretty extensive internal backtrace generator. For this to work, we always need debugging symbols added to the binary. We already do this, however, your release target doesn't seem to include then (missing `-g3`).

In cmake, there is a distinction between a release build (`Release`) and a release build with debug symbols (`RelWithDebInfo `). You can see I defaulted the build to `RelWithDebInfo` [here](https://github.com/pi-hole/FTL/blob/c7d6d505f84ceeb066559dacadccb85f7f1fcdc4/src/CMakeLists.txt#L3). Thus both `cmake` and `cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo` will generate a release with debug symbols. The updated CI would probably want to do `cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo` just to be explicit about it.

> 4. Furthermore, the [building documentation](https://docs.pi-hole.net/ftldns/compile/) needs to be changed. I can do this in a follow-up PR.

Agreed. 

> Sorry for all this, however, I support the addition (or transition) we just need to clarify that things do not break. I will prepare `cmake` availability in the containers so you can experiment.

That will be very helpful. Thanks!

fhriley:
> Further testing results show:
> 
> ```
> [ 73%] Building C object src/CMakeFiles/pihole-FTL.dir/capabilities.c.o
> ```
> 
> while this file should be places outside the `src` directory as `build/capabilities.o`. I don't mind whether it is `.o` or now `.c.o` if it is consistent.

I missed this. Where does it need to go and what is it used for?

fhriley:
> I just tried to compile the binary in our `arm` and `armhf` containers (simple cross-platform builders). It works as expected, however, I see that the binary is places in `src/` which is also something we'd need to change. Nothing should be put into the `src` subdirectory.

I forgot to add the install step. `cmake` is like `autoconf` where building is one step and installing is a second step.  So `make` does the build and `make install` installs the built product. The install step can  put  the binary wherever we want. I'll see if I can get it placed the same as the `Makefile` build so existing scripts continue to work.

fhriley:
> **edit** FWIW, this is the `RUN` command I used to test-compile and check the ELF type of the generated binary:
> 
> ```shell
> git clone https://github.com/pi-hole/FTL.git && cd FTL && git checkout development && git checkout -b fhriley-cmake development && git pull https://github.com/fhriley/FTL.git cmake && cmake . && make -j 4 && readelf -A src/pihole-FTL
> ```

You did an in source build. This is how you should be doing a cmake build, which does an out of source build:

```
mkdir build-cmake
cd build-cmake
cmake ..
make -j 4
make install # As noted in my other comment,  I need to add this
```

fhriley:
> I just tried to compile the binary in our `arm` and `armhf` containers (simple cross-platform builders). It works as expected, however, I see that the binary is places in `src/` which is also something we'd need to change. Nothing should be put into the `src` subdirectory.

`make install` now matches the `Makefile` build and places `pihole-FTL` in the root.

DL6ER:
> Also, I don't have any experience with CircleCI (we use Gitlab) so I didn't want to change that just yet. I think the transition should be pretty simple so I can start looking at it. I probably don't have access to do any testing though?

The testing is done automatically for you, just look at the checks of the PR (above the comment block). They all fail, however, we just pushed a change to our `development` branches that should allow successful build reports for foreign pull requests. When you update your branch with changes, the CI will automatically try to do what you want and fail it it does not work.

When you update

https://github.com/pi-hole/FTL/blob/3e258bb0e6d12e616a40a92083ef7e7f98bfe41a/.circleci/config.yml#L38

from `v1.2` to `v1.3`, then you will have containers with `cmake` included. This is already prepared.

---

> Typical usage of cmake is to do an out of source build in a directory of your choosing so nothing in the source is modified. There is one exception to that, which is the version.h file that gets generated.

I see, so the build steps will need additional steps to create a directory and build in there. This does not look too handy, however, we just have to clearly communicate this to our users.

---

> On further thought, I think this was just a misunderstanding of how to do an out of source build with cmake. I don't think there is any action that needs to be done here.

I agree that is a misunderstanding, I'd still think something needs to get changed, my initial comment is still valid, let me try to phrase it differently: Even an out-of-built generates a rather deep directory structure. The way it is right now might be how `cmake` is doing things and how we will do it, it just feels very strange: 
Assume we're compiling `/src/signals.c`. The object file will be put into `/cmake-build/src/CMakeFiles/pihole-FTL.dir/signals.c.o`. I'm wondering if this file should rather go into a place like `/cmake-build/signals.c.o` to have a flatter data hierarchy. The path looks strange.
Similary, I'd see `/cmake-build/src/dnsmasq/CMakeFiles/dnsmasq.dir/slaac.c.o` actually build placed in some place like `/cmake-build/biuld/dnsmasq/slaac.o`.

I do not request a change here. Just pointing this out as I find it odd.

---

>> I just tried to compile the binary in our `arm` and `armhf` containers (simple cross-platform builders). It works as expected, however, I see that the binary is places in `src/` which is also something we'd need to change. Nothing should be put into the `src` subdirectory.

> `make install` now matches the `Makefile` build and places `pihole-FTL` in the root.

I meant that the default `make` target generated the binary in `/cmake-build/src/pihole-FTL`.
This doesn't feel like the proper place to go. I expect the binary to appear in the root directory of the build. This is *not* about the `install` step.

DL6ER:
Another observation of mine:

I switched from your branch `cmake` to `new/NAMES_FROM_NETWORKDB` and `merge`d `cmake` into this branch. I think moved into `/bmake-build` and hit `make`.

Observation 1: target `dnsmasq` was not rebuilt: :green_circle: 
As expected.

Observation 2: target `sqlite3` was rebuilt: :red_square: 
This should not have happened as nothing changed in the two files. We should monitor this. Maybe you already know why this happened.

Observation 3: all other targets were rebuilt: :green_circle: 
They all were rebuilt as `version.h` changed. This is expected and intended.

We should fix observation 2 as recompiling sqlite3 takes a long time (especially when compiling on a Raspberry Pi!) and should not take place when `version.h` changes (branch switching).

dschaper:
You need to pull the `development` branch updates and rebase your PR on that.

fhriley:
> I agree that is a misunderstanding, I'd still think something needs to get changed, my initial comment is still valid, let me try to phrase it differently: Even an out-of-built generates a rather deep directory structure. The way it is right now might be how `cmake` is doing things and how we will do it, it just feels very strange:
> Assume we're compiling `/src/signals.c`. The object file will be put into `/cmake-build/src/CMakeFiles/pihole-FTL.dir/signals.c.o`. I'm wondering if this file should rather go into a place like `/cmake-build/signals.c.o` to have a flatter data hierarchy. The path looks strange.
> Similary, I'd see `/cmake-build/src/dnsmasq/CMakeFiles/dnsmasq.dir/slaac.c.o` actually build placed in some place like `/cmake-build/biuld/dnsmasq/slaac.o`.

Agreed that the structure is confusing, and it takes a bit to get used to it. I don't believe there is any way to change it. `cmake` was developed over many years as a cross platform build generator so they probably had a good reason to do it that way.

> I meant that the default `make` target generated the binary in `/cmake-build/src/pihole-FTL`.
> This doesn't feel like the proper place to go. I expect the binary to appear in the root directory of the build. This is _not_ about the `install` step.

Got  it, I pushed a change to put pihole-FTL at the root of the build dir.

fhriley:
> Observation 2: target `sqlite3` was rebuilt: 🟥
> This should not have happened as nothing changed in the two files. We should monitor this. Maybe you already know why this happened.
> 
> We should fix observation 2 as recompiling sqlite3 takes a long time (especially when compiling on a Raspberry Pi!) and should not take place when `version.h` changes (branch switching).

If the `Makefile` build doesn't behave in the same way, it has a bug. `new/NAMES_FROM_NETWORKDB` modifies `src/config.h`. `config.h` is included in the sqlite3 code [here](https://github.com/fhriley/FTL/blob/3d7c095cb10a1b692c80dd28f6051dea91d45baa/src/database/sqlite3-ext.c#L27), and thus, must be rebuilt.

DL6ER:
> If the Makefile build doesn't behave in the same way, it has a bug. new/NAMES_FROM_NETWORKDB modifies src/config.h. config.h is included in the sqlite3 code here, and thus, must be rebuilt.

Ah, I see, that's where the error is, no big deal. It isn't the Makefile who is wrong, it is the cmake script, however, the fix is simple:

`sqlite3-ext.c` is independent of `sqlite3.c`, in fact it belongs to the `database_sources` as you can see here (last item):

https://github.com/pi-hole/FTL/blob/3d7c095cb10a1b692c80dd28f6051dea91d45baa/Makefile#L20

Fixing this will fix the compilation behavior.

I think we're close to completion, then. We can do the CI integration separately and I can do it as well if you don't want to do it.

fhriley:
> > If the Makefile build doesn't behave in the same way, it has a bug. new/NAMES_FROM_NETWORKDB modifies src/config.h. config.h is included in the sqlite3 code here, and thus, must be rebuilt.
> 
> Ah, I see, that's where the error is, no big deal. It isn't the Makefile who is wrong, it is the cmake script, however, the fix is simple:
> 
> `sqlite3-ext.c` is independent of `sqlite3.c`, in fact it should belong to the `database_sources` instead of `sqlite3_sources` as you can see here (last item):
> 
> https://github.com/pi-hole/FTL/blob/3d7c095cb10a1b692c80dd28f6051dea91d45baa/Makefile#L20
> 
> Fixing this will fix the compilation behavior.
> 
> I think we're close to completion, then. We can do the CI integration separately and I can do it as well if you don't want to do it.

I understand it now. I fixed this issue. Also, I realized that the sqlite defines are used in `sqlite3.h`, which means the defines need to be on all compile commands. I fixed this as well.

src/database/CMakeLists.txt
```diff
@@ -1,26 +1,12 @@
...
-        SQLITE_DQS=0
-        )
-
+# This file has no dependencies on other code in the repo and takes a long time
+# to build. It is placed in its own target so that it does not include any other
+# headers in the build command and thus does not need to be rebuilt when headers
+# are modified.
 set(sqlite3_sources
         sqlite3.c
-        sqlite3.h
```
DL6ER: I'm not sure this is correct. This would allow us to change `sqlite3.h` without this forcing a rebuild of the sqlite3 object? Can we have it in both (so add below but keep here=?

DL6ER:
@fhriley I updated our fork, the CI builds will now succeed.
Please ensure you run `git pull` before investing more work here to ensure everything's in sync.

src/database/CMakeLists.txt
fhriley: I think it is correct as is. `sqlite3.c` does not include `sqlite3.h`. `sqlite3.c` looks like it has `sqlite3.h` embedded in it. Modifying `sqlite3.h` and then rebuilding `sqlite3.c` should end up with the exact same object code. I just tested this by removing the `pMethods` field in `struct sqlite3_file` in `sqlite3.h` and then rebuilt `sqlite3.c`. Indeed, the before and after object files are identical.

DL6ER: Thanks.

fhriley:
I rebased this  off `development` to  keep the history a little cleaner.

DL6ER:
@fhriley Thanks, I just finished testing your `cmake` branch and there are only two things left before I will approve+merge it.

1. Please sign off your most recent commits:
   ```
   git rebase HEAD~6 --signoff
   git push --force-with-lease origin cmake
   ```
   should do it.

2. While my CI tests worked for all architectures, I failed to produce a static build. I assume the commands should be
   ```bash
   mkdir cmake
   cd cmake
   cmake -DSTATIC=true ..
   cmake --build . -v -j $(nproc)
   ```
   but this doesn't seem to do the trick. I added `-v` to the last command to check that the static lines are indeed not added to the linking step.

fhriley:
> 2\. I failed to produce a static build

This should be fixed. Note that I removed `-static-pie`. It doesn't exist on `gcc 7.5`, which is the `gcc` version installed on `Ubuntu 18.04`.

DL6ER:
We should not drop `static-pie` lightly. Mind that the `STATIC` variant is only meant to be used with `gcc-musl` which always seems to supports this. There are good reasons to use this option as position-independent executables prevent possible exploits based on knowing the offset of some code region. Furthermore, it generated binaries that can be loaded at any address without a dynamic linker.

I will test your static build changes now.

fhriley:
It's no problem to add it back. I just can't test it.

DL6ER:
Your `STATIC=true` build triggers:
```
[...] -o ../pihole-FTL -static-libgcc -static -lrt -Wl,-Bstatic -lhogweed -lgmp -lnettle -lidn -Wl,-Bdynamic
```
however, this will lead to a dynamically linked binary:
```
$ readelf -d ./pihole-FTL | grep "Shared library"
0x0000000000000001 (NEEDED)             Shared library: [libc.musl-x86_64.so.1]
```

This last `-Wl,-Bdynamic` needs to be removed to avoid linking `libmusl` dynamically, resulting in:
```
$ readelf -d ./pihole-FTL | grep "Shared library"
(no output)
```

DL6ER:
Don't worry about the testing of `static-pie`, I will test it on the system where it is used on.

fhriley:
Ok, I think the static build should be good to go now.

DL6ER:
I was just going to approve, but how about `static-pie`?...

pralor:

PR #782, opened 2020-05-24
Try CircleCI context for EnvVars

dschaper:
Project variables don't seem to work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,30 +81,37 @@ workflows:
   build:
     jobs:
       - arm:
+          context: FTL-BUILD
           filters:
             tags:
               only: /^v.*/
       - arm-qemu:
+          context: FTL-BUILD
           filters:
             tags:
               only: /^v.*/
       - armhf:
+          context: FTL-BUILD
           filters:
             tags:
               only: /^v.*/
       - aarch64:
+          context: FTL-BUILD
           filters:
             tags:
               only: /^v.*/
       - x86_64:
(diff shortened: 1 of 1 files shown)


dschaper:
Actually, context doesn't matter. The environment variables are stored in secrets that only this repo has access to. Any PR's against this repo will fail on the upload. My view is that we have the option of putting the SSH variables in the build environment without a problem. We would just need to add that to the yaml. 

The actual key to upload is protected and injected at build time. So, do we want PRs to be able to upload binaries? If so, we set the environment. If not then we need to change the upload step to exit 0 if the SSH_HOST is unset.

`[[ "$CIRCLE_PROJECT_REPONAME" == "FTL" && "$CIRCLE_PROJECT_USERNAME" == "pi-hole" && -n "$SSH_HOST"  ]] || exit 0`

dschaper:
Either way this PR is probably useless.

DL6ER:
@dschaper I don't think we want to host binaries build from foreign code on `ftl.pi-hole.net`. It could make potentially dangerous code becoming available through `pihole checkout`. Only imagine a very simple code that scans the entire filesystem and sends everything it can read to some remote server. Admittedly, we would quickly discover this, however, this discovery is a manual step for protection and we should strive to minimal need for manual steps.

I'd prefer #773 which fixes the already existing logic to skip uploading for foreign code.

dschaper:
That makes perfect sense. I'll check that linked PR.

PR #784, opened 2020-05-24
Add NAMES_FROM_NETDB option

DL6ER:

## 10

---

This adds a new option `NAMES_FROM_NETDB` which allows to disable the try to obtain an addresses host name using the network table.

One use case it that there is a IPv6 client without a host names itself.
However, using the network table we can find out that this client is the same device where we also kwon an IPv4 address together with a valid host names for. In this case, we use the host name associated to the other address as this is the same device. This behavior can now be disabled using `NAMES_FROM_NETDB=false`

This feature has been discussed on Discourse: https://discourse.pi-hole.net/t/incorrect-hostnames/33111

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -372,6 +372,25 @@ void read_FTLconf(void)
 	else
 		logg("   BLOCK_ESNI: Disabled");
 
+	// NAMES_FROM_NETDB
+	// Should we use the fallback option to try to obtain client names from
+	// checking the network table? Assume this is an IPv6 client without a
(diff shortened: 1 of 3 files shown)


pralor:

https://discourse.pi-hole.net/t/incorrect-hostnames/33111/21

pralor:

https://discourse.pi-hole.net/t/network-table-seeing-clients-as-hostname-of-router-rather-than-devices/31597/24

pralor:

PR #798, opened 2020-05-28
Be less nice

DL6ER:

## 10

---

Set nice value of pihole-FTL to increase DNS server performance. The new niceness default to `-10` (lower is less nice = getting more CPU cycles for our own). The value can be changed through a config option. Changing the niceness can be avoided by setting `NICE=0` (this will set `0` as niceness, which is the default) or `NICE=-999` (this will skip trying to set the niceness, altogether).

If the user is not allowed to change their niceness, FTL will log:
```
NICE: Cannot change niceness to -10 (permission denied)
```

If the user specified an invalid niceness, it will be clamped to the according extremum (`+/- 20`):
```
NICE: Set process niceness to -20 (asked for -100)
```

If everything is fine, the output is:
```
NICE: Set process niceness to -12
```

TODO (pi-hole/core): Add `CAP_SYS_NICE` to FTL so it is allowed to change its own niceness when when running as user `pihole`. Without this, only `root` can change the niceness of processes.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -13,6 +13,8 @@
 #include "memory.h"
 #include "setupVars.h"
 #include "log.h"
+// nice()
+#include <unistd.h>
 
 ConfigStruct config;
 FTLFileNamesStruct FTLfiles = {
@@ -39,6 +41,7 @@ static size_t size = 0;
 static char *parse_FTLconf(FILE *fp, const char * key);
 static void release_config_memory(void);
 static void getpath(FILE* fp, const char *option, const char *defaultloc, char **pointer);
+static void set_nice(const char *buffer, int fallback);
 
 void getLogFilePath(void)
 {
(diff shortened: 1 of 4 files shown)


src/config.c
```diff
@@ -367,6 +370,20 @@ void read_FTLconf(void)
 	if(buffer != NULL && strcasecmp(buffer, "false") == 0)
 		config.block_esni = false;
 
+	// NICE
+	// Shall we change the nice of the current process?
+	// defaults to: -10 (can be disabled by setting value to -999)
```
dschaper, line 375: Is there a reason for the default to be `-10`? This is a question we will be asked many times.

DL6ER: No and Yes.

Quoting `man sched`:
>       With the advent of the CFS scheduler in kernel 2.6.23, Linux adopted
>       an algorithm that causes relative differences in nice values to have
>       a much stronger effect.  **In the current implementation, each unit of
>       difference in the nice values of two processes results in a factor of
>       1.25 in the degree to which the scheduler favors the higher priority
>       process.**  This causes very low nice values (+19) to truly provide
>       little CPU to a process whenever there is any other higher priority
>       load on the system, and makes high nice values (-20) deliver most of
>       the CPU to applications that require it (e.g., some audio applica‐
>       tions).

This means that a niceness of `-10` will result in roughly 9x more cycles given to `pihole-FTL` compared to other processes also asking for 100% CPU usages.

yubiuser:
Out of curiosity: Will I see DNS server performance increase in my small home network where my CPU is already dying of boredom? Or rephrased: how many clients do I need before I'll profit from this PR?

DL6ER:
@yubiuser It will not affect you *at all* when you run Pi-hole as the only process on your machine. However, when installing Pi-hole next other compute-intense, i.e., CPU-hungry, processes like a database or e-mail server, maybe a VPN service, etc., then this change will ensure that Pi-hole receives a greater share of the available computing power (it will be less equally distributed among the processes, hence, less "nice" to them).

PromoFaux:
Oh. Merge conflicts

DL6ER:
@PromoFaux Merge conflict resolved

pralor:

PR #817, opened 2020-06-24
Remove obsolete AAAA_QUERY_ANALYSIS setting

DL6ER:

## 10

---

When this was [added over 3 years ago](https://github.com/pi-hole/FTL/pull/63), there may have been [reasons to do so](https://github.com/pi-hole/docker-pi-hole/issues/143#issuecomment-305609805). However, with IPv6 progressing further and further, this option (artificially hide AAAA records even when they are still processed) seems a bit ... falling out of story.

We remove this without replacement. I will update the `release/v5.1` documentation when this gets merged.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -115,16 +115,6 @@ void read_FTLconf(void)
 	else
 		logg("   SOCKET_LISTENING: all destinations");
 
-	// AAAA_QUERY_ANALYSIS
-	// defaults to: Yes
-	buffer = parse_FTLconf(fp, "AAAA_QUERY_ANALYSIS");
-	config.analyze_AAAA = read_bool(buffer, true);
-
-	if(config.analyze_AAAA)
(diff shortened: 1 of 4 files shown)


jfb-pihole:
My preference would be to leave this feature intact.  For those not running IPv6 networks, there isn't much value looking at the AAAA entries in the query log.  It's much cleaner to not see these and focus on the DNS queries that are being used.

yubiuser:
We have seen the option and/or the corresponding documentation can cause confusion and misinterpretation.

I personally don't need that feature at all. If pihole is processing AAAA queries I want to know about. I'm not a big fan of hiding things just for the sake of a prettier graphs/table/logs.
But if others see value in this leave it intact - **but** make the documentation even clearer what this feature does and what not.

DL6ER:
https://github.com/pi-hole/docs/commit/8f073c29d5e272c4128c82229a9a29108bdd2cb3

PR #822, opened 2020-06-29
Fix dnsmasq TCP worker buffer overflow

DL6ER:
**By submitting this pull request, I confirm the following:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

This implements the fix by Simon Kelley for the current buffer overflow issue: http://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=commit;h=e2cb6559586d4b0d0087c326eac0129fe8661382

Closes #757 #767 #770 #785 #802 #804 #405 #806 #808 #809

PR #819, opened 2020-06-28
Add support for additional query types

DL6ER:
**By submitting this pull request, I confirm the following:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

Add support for additional query types. This also implements support for blocking for **all** query types instead of the subset we supported blocking for in v5.0

We already supported:
- `A`,
- `AAAA`,
- `ANY`,
- `SRV`,
- `SOA`,
- `PTR`,
- `TXT`,
- `NAPTR`

New:
- `MX`,
- `DS`,
- `RRSIG`,
- `DNSKEY`, and 
- `OTHER` (summing up all other possible DNS types)

PR #813, opened 2020-06-18
[dnsmasq] Fix a possible buffer overflow when handling TCP requests

fhriley:
…tails are on the dnsmasq mailing list.

Signed-off-by: Frank Riley <fhriley@gmail.com>

**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 5

PR #797, opened 2020-05-27
Reacquire client and upstream pointers after a name resolution

DL6ER:
**By submitting this pull request, I confirm the following:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

Reacquire client and upstream pointers after a name resolution. As we're leaving the locked area for the resolve, we cannot control if the shared memory object changed meanwhile. If it did, then the pointers will point into nowhere, leading to a `SEGV_MAPERR`.

This PR fixes #785

PR #827, opened 2020-07-05
Pi-hole FTL v5.1

DL6ER:
## New
- Switch from `make` to `cmake` (use `./build.sh` instead of `make`) #781 #794 (@fhriley / @DL6ER)
- Pi-hole diagnostics system showing warnings on the dashboard #752 #756 #780 (@DL6ER)
- Automatic IDN encoding for international domains in HOSTS files #673 (@DL6ER)
- Add `NAMES_FROM_NETDB` option #784 (@DL6ER)
- Add support for additional query types #819 (@DL6ER)

## Tweaks
- Always use best matching subnet for clients #753 (@DL6ER)
- Improve self-generated backtraces #774 (@DL6ER)
- Allow no/false and yes/true for all config options #796 (@DL6ER)
- Use higher priority for FTL #798 (@DL6ER)
- Terminate main process when a fork fails #818 (@DL6ER)

## Fixes
- Process cached SRV records #769 (@DL6ER)
- Store client group information in shared memory #787 (@DL6ER)
- Fix broken `RESOLVE_IPV{4,6}` setting #792 (@DL6ER)
- Reacquire client and upstream pointers after a name resolution #797 (@DL6ER)
- Do not listen to reload signals in forks #811 (@DL6ER)
- [dnsmasq] Fix a possible buffer overflow when handling TCP requests #813 #822 (@fhriley / @DL6ER)


PromoFaux:
One more approval here, please

PR #828, opened 2020-07-06
Improve FTL tests

DL6ER:

## 10

---

Do not assume a query to google.com returns only one fixed AAAA record. Also, do not assume the IP address of ftl.pi-hole.net to be fixed for all time.

### DIFF
--- a/test/test_suite.bats
+++ b/test/test_suite.bats
@@ -145,7 +145,6 @@
   run bash -c "dig AAAA google.com @127.0.0.1 +short +tcp"
   printf "%s\n" "${lines[@]}"
   [[ ${lines[0]} != "::" ]]
-  [[ ${lines[1]} == "" ]]
 }
 
 @test "Known host is resolved as expected" {


test/test_suite.bats
```diff
@@ -145,14 +145,12 @@
...
-  [[ ${lines[1]} == "" ]]
 }
 
 @test "Known host is resolved as expected" {
-  run bash -c "dig ftl.pi-hole.net @127.0.0.1 +short"
+  run bash -c "dig A ftl.pi-hole.net @127.0.0.1 +short"
   printf "%s\n" "${lines[@]}"
-  [[ ${lines[0]} == "139.59.170.52" ]]
-  [[ ${lines[1]} == "" ]]
+  [[ ${lines[0]} != "0.0.0.0" ]]
```
dschaper, line 153: This will be true for any value that is not 0.0.0.0? Even empty or "`;; connection timed ;`"?

DL6ER: Yes, the test is not meant for successful DNS resolution but for not-blocked DNS resolution. However, we can leave the IP check for FTL as it is.

dschaper: It looks like `dig` failing completely will also end up as a green condition?

DL6ER: Yes

dschaper: If that's intended then I guess it does what it says on the tin.

DL6ER: You are right. I missed that. I will restore the previous test as it seems superior having this in mind.

PR #844, opened 2020-07-19
Fix setcap target path.

DDoSolitary:

**How familiar are you with the codebase?:** _1_

---

- Take the `DESTDIR` variable into consideration.
- Delay expansion of `${CMAKE_INSTALL_PREFIX}`, which can be a different value at install time.


DL6ER:
Sorry for the delay, I'll be back at my full development environment on Monday. I'll then check out this PR.

DL6ER:
@DDoSolitary How would you set/change `DESTDIR` and `CMAKE_INSTALL_PREFIX` to have the desired effect?

DDoSolitary:
@DL6ER 

The `install (CODE ...)` command will add the code to the generated `cmake_install.cmake` file, and one will run `cmake -P cmake_install.cmake`, or `make install`, which indirectly invokes the `cmake -P` command, to install the files.

`DESTDIR`: It is pretty common to invoke `make install` with `DESTDIR` set, like `make DESTDIR=/some/path install`. Then `DESTDIR` will become an environment variable of the `cmake -P` command, so I used `\$ENV{DESTDIR}` to read it.

`CMAKE_INSTALL_PREFIX` can be set when CMake is directly invoked, like `cmake -P cmake_install.cmake -DCMAKE_INSTALL_PREFIX=/some/path`. It is not commonly used by users, but there are tools setting this variable at install time to customize install location. For example, CPack, the official packaging tool from CMake, relies on this to install the files to a temporary location so that it can package the installed files.

PR #845, opened 2020-07-19
Improve TCP worker handling

DL6ER:

## 10

---

We've seen reports where connection termination due to timeout and foreign client disconnecting can happen at the same time (on milliseconds accuracy). The detailed bug description can be found in https://github.com/pi-hole/FTL/issues/816#issuecomment-660013854.

We're performing four changes in this pull request:

1. We skip second termination if there is already a termination event in progress. This ensures that a running termination cannot be interrupted. To ensure non-interruptibility, we use C11 atomic functions ensuring we generate atomic test code.
2. In addition to step 1, we try to ensure that this situation does not even happen by doubling the TCP worker timeout from 105 seconds to 300 seconds
3. Increase limit for concurrently active TCP workers from 20 to 60.
    We've seen reports where 20 wasn't sufficient in user networks. Given that TCP workers do not really consume all that much more memory, this limit may even be increased further in case we recognize that 60 still isn't sufficient.
4. For debugging proposes, we show for which TCP worker FTL created a fork.
    Previously:
    ```
    TCP worker forked
    ```
    Now:
    ```
    TCP worker forked for client 192.168.0.15 on interface enp0s2 (192.168.0.11)
    ```

### DIFF
--- a/src/dnsmasq/config.h
+++ b/src/dnsmasq/config.h
@@ -15,7 +15,7 @@
 */
 
 #define FTABSIZ 150 /* max number of outstanding requests (default) */
-#define MAX_PROCS 20 /* max no children for TCP requests */
+#define MAX_PROCS 60 /* max no children for TCP requests */
 #define CHILD_LIFETIME 300 /* secs 'till terminated (RFC1035 suggests > 120s) */
 #define TCP_MAX_QUERIES 100 /* Maximum number of queries per incoming TCP connection */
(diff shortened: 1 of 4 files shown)


pralor:

https://discourse.pi-hole.net/t/feedback-for-allow-defining-clients-by-their-mac-address-host-name-and-networking-interface/32324/154

pralor:

https://discourse.pi-hole.net/t/multiple-instances-of-pihole-ftl-running/34865/21

PR #857, opened 2020-08-03
Fix rare database importing issue

DL6ER:

## 10

---

Do not try to bind a value when we should actually read one. This glitch was harmless, however, it prevented regex ID from being loaded from the database so it is nevertheless a bug. This bug never existed in `master`, only in `development`.

While we're at it, we increase memory efficiency by not using `asprintf` but binding the integer directly into the statement.


DL6ER:
Confirmed working

https://discourse.pi-hole.net/t/force-update-of-ip-hostname-clientname-relation-ship/36585/22?u=dl6er

PR #872, opened 2020-08-24
Only resolve active clients

DL6ER:

## 10

---

This PR is a feature request from discourse (see link below). Only try to resolve host names of upstream servers which were recently active. The current limit for "recently active" is hard-coded to two hours.

### DIFF
--- a/src/datastructure.c
+++ b/src/datastructure.c
@@ -75,7 +75,11 @@ int findUpstreamID(const char * upstreamString, const bool count)
 
 		if(strcmp(getstr(upstream->ippos), upstreamString) == 0)
 		{
-			if(count) upstream->count++;
+			if(count)
+			{
+				upstream->count++;
+				upstream->lastQuery = time(NULL);
+			}
 			return upstreamID;
 		}
 	}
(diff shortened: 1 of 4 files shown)


dschaper:
The related Discourse discussion auto-closed for lack of action. Is this still a valid PR?

DL6ER:
Yes, or effectively reduces the number of PTR requests FTL does by itself on every full hour. Hence, it is a useful addition.

dschaper:
Is the option to change the time period coming in another PR? (You know there will be requests for it...)

dschaper (approved):

DL6ER:
> Is the option to change the time period coming in another PR?

There is no run-time option for this you'd have to change the value in the source code and recompile:

https://github.com/pi-hole/FTL/blob/dbd4a696afd35efa5e6c4882c74d9102d9d7d6a2/src/FTL.h#L86-L88

Reresolving on the full hour was added two years ago (https://github.com/pi-hole/FTL/pull/405) with the same hardcoded values and I have not seen a request for a run-time value since then. I tend to keep the run-time options to what users actually need to ensure the config file documentation stays readable.

Nevertheless, I have no objections to adding another run-time setting when you find this useful.

PR #878, opened 2020-08-30
Extend domain filtering to also check the CNAME domain for domain-fil

DL6ER:

## 10

---

Extend domain filtering to also check the CNAME domain for domain-filtering (if this is indicated by the query status).
This fixes a subitem of #833 we seemingly have forgotten.

http://pi.hole/admin/queries.php?domain=aax-eu.amazon.de
Screenshot at 2020-08-30 21-34-18

http://pi.hole/admin/queries.php?domain=aax-eu-retail-direct.amazon-adsystem.com
Screenshot at 2020-08-30 21-34-43

Note: Reply type "N/A" is expected as I restarted FTL in between and these are queries re-imported from the database.

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -849,24 +849,28 @@ void getAllQueries(const char *client_message, const int *sock)
 			continue;
 
 		// Skip if domain is not identical with what the user wants to see
-		if(filterdomainname && query->domainID == domainid)
+		if(filterdomainname)
 		{
-			// Get this query
-		}
-		// If the domain of this query did not match, the CNAME
-		// domain may still match - we have to check it in
-		// addition if this query is of CNAME blocked type
-		else if((query->status == QUERY_GRAVITY_CNAME ||
-		         query->status == QUERY_BLACKLIST_CNAME ||
-		         query->status == QUERY_REGEX_CNAME) &&
-		         query->CNAME_domainID == domainid)
-		{
-			// Get this query
-		}
-		else
-		{
-			// Skip this query
-			continue;
+			// Check direct match
+			if(query->domainID == domainid)
+			{
+				// Get this query
+			}
+			// If the domain of this query did not match, the CNAME
+			// domain may still match - we have to check it in
+			// addition if this query is of CNAME blocked type
+			else if((query->status == QUERY_GRAVITY_CNAME ||
+				query->status == QUERY_BLACKLIST_CNAME ||
+				query->status == QUERY_REGEX_CNAME) &&
+				query->CNAME_domainID == domainid)
+			{
+				// Get this query
+			}
+			else
+			{
+				// Skip this query
+				continue;
+			}
 		}
 
 		// Skip if client name and IP are not identical with what the user wants to see
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -849,8 +849,25 @@ void getAllQueries(const char *client_message, const int *sock)
 			continue;
 
 		// Skip if domain is not identical with what the user wants to see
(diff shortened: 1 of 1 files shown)


yubiuser:
Thanks for the fast implementation. But this is the solution to another problem.

The issue I meant is that in the top blocked domains it still shows the "actual blocked domain" (aax-eu-retail-direct.amazon-adsystem.com) instead of the requested CNAME (aax-eu.amazon.de). As you said [here](https://discourse.pi-hole.net/t/display-blocked-deep-cname-domains-in-pi-hole-tail-query/32297/47?u=yubiuser), the intention is to show the requested domain instead of the actual blocked domain.

Using this branch I get both in top blocked domains (links for both work though)
Bildschirmfoto zu 2020-08-30 21-50-16

DL6ER:
Oh, I see, so we have to actually hide `aax-eu-retail-direct.amazon-adsystem.com` because it is already contained in `aax-eu.amazon.de`

I will have to think about how to do this as the domains themselves do not carry the information about if they have been seen in the middle of a CNAME blocking event or not. This is only available at the query level. Well, I will likely have to add a property to the domains. I will check this out another day with fresh eyes. Thanks for the heads up!

src/api/api.c
```diff
@@ -849,8 +849,29 @@ void getAllQueries(const char *client_message, const int *sock)
 			continue;
 
 		// Skip if domain is not identical with what the user wants to see
-		if(filterdomainname && query->domainID != domainid)
-			continue;
+		if(filterdomainname)
+		{
+			// Check direct match
+			if(query->domainID == domainid)
```
PromoFaux, line 855: Just so I know I've got my head around this, does this also read as: (might be slightly syntactically wrong!)

```c
if (! ((query->status == QUERY_GRAVITY_CNAME || query->status == QUERY_BLACKLIST_CNAME || query->status == QUERY_REGEX_CNAME) && query ->CNAME_domainID == domainid)) && ! (query->domainID === domainid))
{
 continue;
}
```

DL6ER: Yes, but I thought it's easier to have it is smaller pieces. I already checked in the assembly output that the compiler optimizer is cleaning up my slightly more verbose syntax by doing some very fancy optimized branching so the more explicit writing style doesn't seem to cause any costs.

yubiuser:
Maybe related to 

>Oh, I see, so we have to actually hide aax-eu-retail-direct.amazon-adsystem.com because it is already contained in aax-eu.amazon.de

But the number on the dashboard is way off to what is displayed when following the link

Bildschirmfoto zu 2020-09-05 20-56-25

Bildschirmfoto zu 2020-09-05 20-56-30

PromoFaux:
> number on the dashboard is way off

I'm unable to reproduce this on a (nearly) fresh dev system

image

image

image

(last image filtered to hide the non-blocked `AAAA` records)

PromoFaux (approved):

yubiuser:
I also tried it again and now the numbers match. Don't know why they didn't the last time...

Please consider hiding the actual blocked domain (and only show CNAME) at the dashboard before this gets merged.

DL6ER:
> Please consider hiding the actual blocked domain (and only show CNAME) at the dashboard before this gets merged.

Where exactly? In the Top Items? What if a domain was partially in a CNAME, partially directly queried? Should we show only the delta in this case?

We can open a new PR once this is clarified.

yubiuser:
Yes, at the top lists. 

I remember you intended to only show directly queried domains. If a domain has been blocked as part of the CNAME chain, also show (or increase the counter) of the actually requested domain.

PR #884, opened 2020-09-07
Fix GCC9 regression for printing the same buffer into itself

DL6ER:

## 10

---

Fixes warnings showing up with the most recent `gcc` compiler. The code still works the same, however, it is a bit safer, so this is a good change. The fix is to not print the existing buffer into itself but to append to the string by starting to write from the memory address immediately behind the current content of the buffer.


PromoFaux (approved):
I couldn't quite work out what was going on here, so I compiled a quick and dirty mini version to see whether or not it still works. It still works, but I still don't have my head around it.

But, it defintely works.

https://onlinegdb.com/BJ3Y6jNBv

PR #888, opened 2020-09-14
Add no-id flag that works like NO_ID compile flag, disable .info and …

rndnoise:

**How familiar are you with the codebase?:**  2

***

In v2.81, `dnsmasq` added a compile-time flag `NO_ID` that disables answers to RFC4892 zones `.server` and `.info`. This pull request adds a runtime option `no-id` that has the same effect, with the benefit of not needing to recompile FTL to enable it.

### DIFF
--- a/src/dnsmasq/dnsmasq.h
+++ b/src/dnsmasq/dnsmasq.h
@@ -269,7 +269,8 @@ struct event_desc {
 #define OPT_IGNORE_CLID    59
 #define OPT_SINGLE_PORT    60
 #define OPT_LEASE_RENEW    61
-#define OPT_LAST           62
+#define OPT_NO_ID          62
+#define OPT_LAST           63
 
 #define OPTION_BITS (sizeof(unsigned int)*8)
 #define OPTION_SIZE ( (OPT_LAST/OPTION_BITS)+((OPT_LAST%OPTION_BITS)!=0) )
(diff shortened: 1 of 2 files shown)


DL6ER:
Thank you for your suggestion. Unfortunately, we cannot accept it as such as we have a policy to not accept changes to the `dnsmasq´ code, unless it cannot be avoided (like adding Pi-hole-specific changes). In your case, however, this is a pure `dnsmasq` change which should be submitted to the mother project. This will not only ensure everyone can benefit from it, but also it will avoid/reduce future merge conflicts when updating `dnsmasq` in our own project.

To show how much we appreciate your submission, I converted your suggested change into a patch that can be directly submitted as attachment to dnsmasq-discuss@lists.thekelleys.org.uk ([Info page](http://lists.thekelleys.org.uk/mailman/listinfo/dnsmasq-discuss), there is no option to open PRs to `dnsmasq` otherwise). Your authorship will be preserved by this patch: [0001-Add-no-id-flag-that-works-like-NO_ID-compile-flag.patch.zip](https://github.com/pi-hole/FTL/files/5216256/0001-Add-no-id-flag-that-works-like-NO_ID-compile-flag.patch.zip)

I'm sorry for this but I hope you will understand our motives and why this is good for everyone.

rndnoise:
Thanks, I understand. I appreciate the instructions on where to submit the patch, since I wasn't able to figure out how to submit changes upstream.

PR #898, opened 2020-09-30
Allocate shared memory instead of only reserving it

DL6ER:

## 10

---

Use `fallocate()` instead of `ftruncate()` when resizing and/or creating shared memory objects.
This ensures we reserve the requested memory exclusively for ourselves. With `ftruncate` we only announce we want the memory, however, other processes can still take the memory when the `/dev/shm` is close to being full.

This has the effect that this PR ensures that FTL does not crash when `/dev/shm` is getting full.

In addition, even when the shared memory arena is full, there will be a better error message, like

```
[2020-10-01 00:18:42.812 13052M] Creating shared memory with name "FTL-lock" and size 48 (/dev/shm: 2.0GB used, 2.0GB total)
[2020-10-01 00:18:42.812 13052M] WARNING: More than 90% of /dev/shm is used
[2020-10-01 00:18:42.812 13052M] FATAL: create_shm(): Failed to resize "FTL-lock" (4) to 48: No space left on device (-1)
```

### DIFF
--- a/src/shmem.c
+++ b/src/shmem.c
@@ -441,14 +441,15 @@ SharedMemory create_shm(const char *name, const size_t size)
 		exit(EXIT_FAILURE);
 	}
 
-	// Resize shared memory file
-	ret = ftruncate(fd, size);
-
-	// Check for `ftruncate` error
-	if(ret == -1)
+	// Allocate shared memory object to specified size
+	// Using fallocate() will ensure that there's actually space for
+	// this file. Otherwise we end up with a sparse file that can give
+	// SIGBUS if we run out of space while writing to it.
+	ret = fallocate(fd, 0, 0U, size);
+	if(ret != 0)
 	{
-		logg("FATAL: create_shm(): ftruncate(%i, %zu): Failed to resize shared memory object \"%s\": %s",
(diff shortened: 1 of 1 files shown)


dschaper (approved):

src/shmem.c
```diff
@@ -584,7 +589,7 @@ bool realloc_shm(SharedMemory *sharedMemory, const size_t size, const bool resiz
 		logg("FATAL: realloc_shm(): mremap(%p, %zu, %zu, MREMAP_MAYMOVE): Failed to reallocate \"%s\": %s",
 		     sharedMemory->ptr, sharedMemory->size, size, sharedMemory->name,
 		     strerror(errno));
-		exit(EXIT_FAILURE);
+		return false;
```
dschaper, line 592: I assume there doesn't need to be a change to the code that calls `realloc_shm` to expect returns instead of the function exiting on it's own?

DL6ER: Thanks for your comment. There isn't really a proper way of handling these kinds of out-of-memory issues. I'll undo this to `exit()` because it's easier than ensuring down the road that everything is always properly cascading it downwards.

dschaper (approved):

PR #901, opened 2020-10-08
Add new query retried status

DL6ER:

## 10

---

Add new status `RETRIED` (12) to be used for queries which were retried. If a query was retried five times until we received a reply form upstream, queries 1-4 will be marked as `RETRIED` and only query 5 will stay in status `FORWARDED`. This does not affect the statistics because all five queries where send upstream, so five upstream packages are counted, even when only one query stays in status `FORWARDED`.

**edit** Add another new status `RETRIED_DNSSEC` (13) to be used for queries for which which automatic DNSSEC queries were retried. If we've already got an answer to a query, but we're awaiting keys for validation, there's no point retrying the query, so we're retrying the key query instead.

The web interface will need updating to support this status. This will be done in a follow-up PR.

### DIFF
--- a/src/dnsmasq/forward.c
+++ b/src/dnsmasq/forward.c
@@ -347,7 +347,7 @@ static int forward_query(int udpfd, union mysockaddr *udpaddr,
 	start = daemon->servers; /* at end of list, recycle */
       header->id = htons(forward->new_id);
 
-      FTL_forwarding_failed(forward->sentto);
+      FTL_forwarding_retried(forward->sentto, forward->log_id, daemon->log_id);
     }
   else 
     {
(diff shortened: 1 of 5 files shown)


DL6ER:
The failed `arm-qemu` build is expected because the container was updated meanwhile and the cross-build doesn't use the versioned containers. This will be fixed by #903 and is of no concern here.

The failed CodeFactor issue is something I will address separately, however, the mentioned [possibly security-relevant race condition](https://cwe.mitre.org/data/definitions/362.html) doesn't apply to our code at all. It is only meant for ensuring that the ownership is correct when FTL is started as `root` and later drops to `pihole` as a security measure.

This PR can go in as-is. The two :red_circle: can be ignored.

dschaper (approved):

gie69:
Pada tanggal Kam, 8 Okt 2020 22.08, DL6ER <notifications@github.com>
menulis:

PR #902, opened 2020-10-12
Make arm-qemu builds ARMv4T compliant (was ARMv6) to support more devices

DL6ER:

## 10

---

Explicitly specify the building host to achieve true ARM Debian compliance with the cross-compiled build. We remove concurrent building of the dependencies as this can lead to compiler segmentation faults on many-core machines with a lot of concurrency in the emulator.

Even though this build only requested ARMv5 compliance, the produce code actually generated a binary that is even ARMv4T complaint:
```
Attribute Section: aeabi
File Attributes
  Tag_CPU_name: "4T"
  Tag_CPU_arch: v4T
  Tag_ARM_ISA_use: Yes
  Tag_THUMB_ISA_use: Thumb-1
  Tag_ABI_PCS_wchar_t: 4
  Tag_ABI_FP_rounding: Needed
  Tag_ABI_FP_denormal: Needed
  Tag_ABI_FP_exceptions: Needed
  Tag_ABI_FP_number_model: IEEE 754
  Tag_ABI_align_needed: 8-byte
  Tag_ABI_enum_size: int
```

Even when this is not the expected outcome, any ARMv5TE processor will be able to run ARMv4T code. And as the compiler generated ARMv4T code while it was allowed to generate ARMv5TE means that, despite the fact that all compiler optimizations are enabled, no ARMv5TE-specific instructions are useful here and the generation of ARMv4T-compliant code is not slower by any means (due to missed optimizations).

The PR updating the FTL builds for ARMv4T-compliance will also address the changed architecture in the tests. This has to come *after* this PR was merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,16 +4,11 @@ version: 2
   steps:
     - checkout
     - run:
-        name: "Setup"
-        command: |
-          if [[ $CIRCLE_JOB == *"qemu"* ]] ; then sudo docker run --rm --privileged multiarch/qemu-user-static:register --reset ; fi
-    - run:
         name: "Build"
         no_output_timeout: 30m
         command: |
           BRANCH=$([ -z "$CIRCLE_TAG" ] && echo "$CIRCLE_BRANCH" || echo "master")
-          [[ $CIRCLE_JOB == *"qemu"* ]] && DOCKERIFNEEDED="docker run --rm -v $(pwd):/workspace -w /workspace pihole/ftl-build:v1.5-arm-qemu "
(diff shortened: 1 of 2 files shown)


DL6ER:
Putting this on hold right now, it seems we might have found a variant without the `arm-qemu` build. This does not only reduce the number of builds (and, hence, code and management complexity) but also reduce the overall compile time on the CI from some 15-20 minutes down to less than 3 which is very nice.

DL6ER:
Superseded by https://github.com/pi-hole/FTL/pull/903

Issue #924, opened 2020-11-13
hostnames do not match IP after RPi upgrade


Marooned-MB:
### Versions

- Pi-hole:  v5.1.2 (Latest: v5.1.2)
- AdminLTE:  v5.1.1 (Latest: v5.1.1)
- FTL:  v5.2 (Latest: v5.2)

### Platform

- OS and version:  Linux raspberrypi 5.4.72+ pi-hole/pi-hole#1356 Thu Oct 22 13:56:00 BST 2020 armv6l GNU/Linux, 10 (buster)

- Platform:  Raspberry Pi Model B Rev 2 (512MB)

### Expected behavior

I had pi-hole running on old Raspbian and all was good. Few days ago I decided to wipe out my SD card and install newest soft for my RPi. I've installed latest OS and other programs. After setting up pi-hole I've noticed that it assigned wrong hostnames to some IPs (see attachment - green ones are correct, red one are not).
What is probably worth noting, I copied `/etc/.pihole` and `/etc/pihole` from old system to restore the backup and all my settings/white/blacklists.

### Actual behavior / bug

pi-hole assigns wrong hostnames to some IP resulting in some strange results like the vacuum cleaner calling Facebook etc

Also, color of the background in `/admin/network.php` suggest that those devices were not seen in >24h which is not true as `/admin/queries.php` shows proper and updated list (just wrong hostnames).

### Steps to reproduce

Just checking different pages in pi.hole/admin

## Debug Token

- URL: https://tricorder.pi-hole.net/6i4xc8ma1n

### Screenshots

![pi-hole ip](https://user-images.githubusercontent.com/716018/99062939-06919300-25a4-11eb-94f3-51fcb28c81cb.jpg)

DL6ER:
What is the output of
```
ip neigh show
```
?

Marooned-MB:
`ip neigh show` shows a list of IPs and MAC address and they match the list from the router, so the proper one.

DL6ER:
Then something odd may have happened in between. Please try flushing your Pi-hole's address assignments using the command below. You will not loose any other details like the first seen timestamps, etc.

```
sudo sqlite3 /etc/pihole/pihole-FTL.db "DELETE FROM network_addresses;"
```

Then wait some minutes and check again what the Pi-hole table shows to you.

Marooned-MB:
[...]

PR #935, opened 2020-11-25
Keep upper case characters in host names

DL6ER:

## 10

---

Keep upper case characters in host names because they may make them more readable (like `FritzBox`, `WDMyCloud`, or `VacuumRobot`). Inspired by https://github.com/pi-hole/FTL/issues/924

We make the corresponding API filters case-insensitive to preserve the current behavior (where everything was converted to lowercase) and as capitalization does not play a role in domain names.

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -737,7 +737,7 @@ void getAllQueries(const char *client_message, const int *sock)
 				// (if available) their host names
 				if((strcmp(getstr(forward->ippos), serv_addr) == 0 ||
 				   (forward->namepos != 0 &&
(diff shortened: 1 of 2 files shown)


dschaper:
Will this affect, or be affected by 0x20 encoding?

DL6ER:
Domain names stay untouched, this is only about host names as results of `PTR` queries. 0x20 isn't a thing here to the best of my knowledge.

dschaper:
Okay, I'll trust you on that.

dschaper (approved):

PR #936, opened 2020-11-25
Tweaks to disabling the FTL database

DL6ER:

## 10

---

With modern Pi-hole we cannot easily disable the database because it is used for a lot more than simple long-term query logging. Instead, it also contains the entire network table wisdom (and hence an essential source for how to identify clients in groups), messages about warnings and errors such as config errors or regex syntax issues, alias-client definitions and a lot more.

The recommended way to disable the database is setting `MAXDBDAYS=0`.
However, as the documentation mentions that `DBFILE=` is also possible, we handle this case by using the default path if we find an empty path but setting `MAXDBDAYS=0` internally in this case. This allows old configurations to work as before without introducing any issues.

### DIFF
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -74,10 +74,6 @@ bool dbopen(void)
 		return true;
 	}
 
-	// Do not open database if it is not to be used
-	if(!use_database())
-		return false;
-
 	if(config.debug & DEBUG_LOCKS)
 		logg("Locking FTL database");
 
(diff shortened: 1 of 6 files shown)


src/database/common.c
```diff
@@ -622,20 +614,4 @@ long get_lastID(void)
...
-	if(FTLfiles.FTL_db == NULL ||
-	   strlen(FTLfiles.FTL_db) == 0 ||
-	   config.maxDBdays == 0)
-	{
-		return false;
-	}
-
-	return true;
-}
+}
```
dschaper, line 617: Just for my personal information, do you care if the file has ending newlines?

DL6ER: No, I don't care and changed to whether there is a newline at the end or not are not really intended but seem unavoidable to me when working with different editors from different systems.

dschaper: Okay, good to know.

dschaper (approved):

PR #943, opened 2020-11-29
Fix compatibility with GCC 10.

DDoSolitary:

**How familiar are you with the codebase?:** 1

---

GCC 10 defaults to -fno-common, which requires "extern" when defining global variables in headers.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -41,7 +41,8 @@ set(SQLITE_DEFINES "-DSQLITE_OMIT_LOAD_EXTENSION -DSQLITE_DEFAULT_MEMSTATUS=0 -D
 # -Wl,-z,defs: Detect and reject underlinking (phenomenon caused by missing shared library arguments when invoking the linked editor to produce another shared library)
 # -Wl,-z,now: Disable lazy binding
 # -Wl,-z,relro: Read-only segments after relocation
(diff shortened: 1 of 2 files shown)


DDoSolitary:
Tests failed unfortunately... I'll take a look later.

DL6ER:
Tests are all green after restarting the failed ones. There exists a race-collision in the tests leading to false-negative results. I haven't had the time to look at this so far, also because it doesn't happen very often.

DL6ER (changes requested):
Thanks for your submission. I'm slightly surprised I forgot this in only one spot :-)

Can I ask you to add
```
-fno-common
```
also to this line:
https://github.com/pi-hole/FTL/blob/e1db31de09a5177459690118f55513f9d8fad2bd/src/CMakeLists.txt#L44
and add a descriptive comment above like
```
# -fno-common: Emit globals without explicit initializer from `.bss` to `.data`. This causes GCC to reject multiple definitions of global variables. This is the new default from GCC-10 on.
```
to enforce this also in the future to avoid anyone making such a "mistake" again?

DDoSolitary:
@DL6ER Done.

DL6ER:
Thanks, the failing tests should get fixed by https://github.com/pi-hole/FTL/pull/945

DL6ER (approved):

PR #953, opened 2020-12-01
Fix for numerous hourly PTR requests seen in network with many (dead) IPv6 addresses

DL6ER:

## 10

---

 Do not try to resolve IPs for records without hostnames in the `network_addresses` table to resolve a flood of PTR requests if there are many IPv6 addresses (probably due to "privacy extension" (PE)) without host names.

The issue itself comes from the growing number of temporary IPv6 addresses having recently being enabled by default in Windows and (some?) Mac devices. The PR adds a new option `REFRESH_HOSTNAMES` which which you can change how the hourly PTR requests are made. I'm currently thinking about three options:

- `REFRESH_HOSTNAMES=NONE` - Don't do any hourly PTR lookups
   This means we look host names up exactly once (when we first see a client) and never again. This means we will miss future changes of host names.
- `REFRESH_HOSTNAMES=IPV4` - Do the hourly PTR lookups only for IPv4 addresses
   This is **the new default** and should do away with the issue.
- `REFRESH_HOSTNAMES=ALL` - Do the hourly PTR lookups for all addresses
   This is the same as what we're doing with FTL v5.3(.1) and which seems to create a lot of PTR queries for some

This should be the most satisfactory solution because many IPv6 hosts won't even have a name associated to it (PE IPs surely have no names by construction).


dschaper:
Needs to be documented on `docs.pi-hole.net` and in the release notes.

dschaper (approved):

DL6ER:
Docs: https://github.com/pi-hole/docs/pull/424
Release notes: Done

PR #965, opened 2020-12-08
Automate IP blocking mode

DL6ER:

## 10

---

Until now, FTL's [Full IP](https://docs.pi-hole.net/ftldns/blockingmode/#pi-holes-full-ip-blocking) and [IP (IPv6 NODATA)](https://docs.pi-hole.net/ftldns/blockingmode/#pi-holes-ip-ipv6-nodata-blocking) blocking modes sourced the IP to deliver on a blocked domain from the `setupVars.conf` values `IPV4_ADDRESS` and `IPV6_ADDRESS`. This is, however, quite a limitation, especially if the device running Pi-hole has more than one interface.

This PR implements an automated IP blocking. Instead of reading the two addresses  from `setupVars.conf`, we now determine the address of the interface a query arrived on. We then use this IP address in the blocked reply. This does not only reduce maintenance (`IPV4_ADDRESS` and `IPV6_ADDRESS` can now be removed from `setupVars.conf`) but also localizes blocked queries.

Example on the Pi-hole itself:
```
$ dig A doubleclick.net +short
127.0.0.1
$ dig AAAA doubleclick.net +short
::1
```
and from my desktop machine:
```
$ dig A doubleclick.net +short
192.168.2.10
$ dig AAAA doubleclick.net +short
fe80::a81c:cafe:dead:beef
```

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -480,6 +480,44 @@ void read_FTLconf(void)
 	else
 		logg("   RATE_LIMIT: Disabled");
 
+	// REPLY_ADDR4
+	// Use a specific IP address instead of automatically detecting the
+	// IPv4 interface address a query arrived on
+	// defaults to: not set
+	config.reply_addr.overwrite_v4 = false;
+	config.reply_addr.v4.s_addr = 0;
+	buffer = parse_FTLconf(fp, "REPLY_ADDR4");
+	if(buffer != NULL && inet_pton(AF_INET, buffer, &config.reply_addr.v4))
+		config.reply_addr.overwrite_v4 = true;
+
+	if(config.reply_addr.overwrite_v4)
+	{
+		char addr[INET_ADDRSTRLEN] = { 0 };
+		inet_ntop(AF_INET, &config.reply_addr.v4, addr, INET_ADDRSTRLEN);
+		logg("   REPLY_ADDR4: Using IPv4 address %s in IP blocking mode", addr);
+	}
+	else
+		logg("   REPLY_ADDR4: Automatic interface-dependent detection of address");
+
+	// REPLY_ADDR6
+	// Use a specific IP address instead of automatically detecting the
+	// IPv6 interface address a query arrived on
+	// defaults to: not set
+	config.reply_addr.overwrite_v6 = false;
+	memset(&config.reply_addr.v6, 0, sizeof(config.reply_addr.v6));
+	buffer = parse_FTLconf(fp, "REPLY_ADDR6");
+	if(buffer != NULL && inet_pton(AF_INET6, buffer, &config.reply_addr.v6))
+		config.reply_addr.overwrite_v6 = true;
+
+	if(config.reply_addr.overwrite_v6)
+	{
+		char addr[INET6_ADDRSTRLEN] = { 0 };
+		inet_ntop(AF_INET6, &config.reply_addr.v6, addr, INET6_ADDRSTRLEN);
+		logg("   REPLY_ADDR6: Using IPv6 address %s in IP blocking mode", addr);
+	}
+	else
+		logg("   REPLY_ADDR6: Automatic interface-dependent detection of address");
+
 	// Read DEBUG_... setting from pihole-FTL.conf
 	read_debuging_settings(fp);
 
--- a/src/config.h
+++ b/src/config.h
@@ -19,6 +19,8 @@
 #include <idn-int.h>
 // assert_sizeof
 #include "static_assert.h"
+// struct in_addr, in6_addr
+#include <netinet/in.h>
 
 void getLogFilePath(void);
 void read_FTLconf(void);
@@ -59,8 +61,14 @@ typedef struct {
 	} rate_limit;
 	enum debug_flags debug;
 	time_t DBinterval;
+	struct {
+		bool overwrite_v4 :1;
+		bool overwrite_v6 :1;
+		struct in_addr v4;
+		struct in6_addr v6;
+	} reply_addr;
 } ConfigStruct;
-ASSERT_SIZEOF(ConfigStruct, 64, 56, 56);
+ASSERT_SIZEOF(ConfigStruct, 88, 80, 80);
 
 typedef struct {
 	const char* conf;
--- a/src/dnsmasq/dnsmasq.c
+++ b/src/dnsmasq/dnsmasq.c
@@ -1789,11 +1789,12 @@ static void check_dns_listeners(time_t now)
   for (listener = daemon->listeners; listener; listener = listener->next)
     {
 
-      /******************** Pi-hole modification *******************/
-      FTL_next_iface(listener->iface ? listener->iface->name : NULL,
-                     listener->iface ? listener->iface->index : -1,
-                     daemon->listeners);
-      /*************************************************************/
+      /***************************** Pi-hole modification ****************************/
+      // Get the interface here only if we know the listeners are bound to specific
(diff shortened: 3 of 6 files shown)


DL6ER:
This PR is in principle ready but awaits further testing by @PromoFaux

PromoFaux:
Maybe just a WSL quirk, not sure yet, but...

image

```text
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: bond0: <BROADCAST,MULTICAST,MASTER> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether f6:69:0d:36:0a:f1 brd ff:ff:ff:ff:ff:ff
3: dummy0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 66:41:ed:7e:f0:1c brd ff:ff:ff:ff:ff:ff
4: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:15:5d:4b:8b:75 brd ff:ff:ff:ff:ff:ff
    inet 172.28.60.222/20 brd 172.28.63.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::215:5dff:fe4b:8b75/64 scope link
       valid_lft forever preferred_lft forever
5: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default qlen 1000
    link/sit 0.0.0.0 brd 0.0.0.0
```

PromoFaux:
OK, not just WSL, can reproduce with a droplet - too.

image

and from home to the droplet:

image

DL6ER:
@PromoFaux Can you try again with
```
DEBUG_QUERIES=true
DEBUG_NETWORKING=true
```
in `pihole-FTL.conf` ?

This will result in a lot of output (once for each address on each interface per second). Of interesting will be the interface details immediately *before* your PTR is recorded in the log.

PromoFaux:
Not sure if this is relevant:
```
[2020-12-10 18:31:28.976 14721M] No interface data available at this point (interface is "-")
[2020-12-10 18:31:28.976 14721M] v---- interface scan starting, reported interface is -
[2020-12-10 18:31:28.976 14721M] Skipping NULL interface pointer
[2020-12-10 18:31:28.976 14721M] Skipping NULL interface pointer
[2020-12-10 18:31:28.976 14721M] ^---- interface scanning done
[2020-12-10 18:31:28.976 14721M]
[2020-12-10 18:31:28.976 14721M] No interface data available at this point (interface is "-")
[2020-12-10 18:31:28.976 14721M] v---- interface scan starting, reported interface is -
[2020-12-10 18:31:28.976 14721M] Skipping NULL interface pointer
[2020-12-10 18:31:28.976 14721M] Skipping NULL interface pointer
[2020-12-10 18:31:28.976 14721M] ^---- interface scanning done
```

image

DL6ER:
Yes it is, so `dnsmasq` does not really have any details about your network interface. Is this in your embedded Windows Linux hybrid or is it on a droplet (or both)?

PromoFaux:
I've not yet tried it again on a droplet, this was in WSL

PromoFaux:
Same thing on a droplet:

image

DL6ER:
Converted to draft PR until I have the time to look into this in more detail

DL6ER:
@PromoFaux It didn't work for you before because I had `bind-interfaces` enabled from a previous experiment and the algorithm worked only in this case. I generalized this now (and rebased on latest `development`) so it should work with any possible configuration.

PromoFaux:
image

This time it is returning `127.0.0.1`

Some `pihole-FTL.conf` lines here: [debug.txt](https://github.com/pi-hole/FTL/files/6173165/debug.txt)

DL6ER:
Okay, thanks. This is expected. The automatic detection is much more flexible than the single IP we had before. Your query came over `lo` which is why FTL served `::1` for the blocked `A` record.

Please test also
- `dig AAAA google.com +short @127.0.0.1` (should return `::1`)
- `dig A google.com +short @192.168.0.241` (should return `192.168.0.241`)
- `dig AAAA google.com +short @192.168.0.241` (should return `fe80:what:ever:else` *or* `fdd6:20e2:something:more`)

If a single interface has more than one address (like in the IPv6 case of your `eth0` interface), we return the first address we find. It is undefined which one this will be. Maybe we should scan all IPv6 addresses and favor GUA over ULA and use LL only as last-resort. Opinions?

PromoFaux:
Oh! I very much like that behaviour. So it returns the IP address of the interface that is queried, basically? 

And, I can't really answer the IPv6 question, I don't understand that in the slightest

DL6ER:
> So it returns the IP address of the interface that is queried, basically?

Yeah, exactly that's my idea.

Imagine different interfaces like also a VPN. Without proper routes, nobody from the internal network can reach the IP of the Pi-hole in the VPN and nobody from the VPN can reach an IP in the local network. Same for VLANs, etc.
This all wasn't possible so far.

DL6ER:
I'll convert this back to draft mode to add the IPv6 address preference when I can get down to it.

DL6ER:
@dschaper This PR changes the `IP` blocking mode to always respond with the correct addresses of the interface the query arrived on. While things are simple for IPv4, we can have multiple addresses for IPv6 on the same interface. The algorithm works by preferring ULA addresses over GUA and use LL only as last-resort.

Do you think this is the correct order or should GUA be preferred here? My idea about favoring ULA is that this may work better when the upstream connection is down.

dschaper:
Makes sense to me to order ULA > GUA > LL. Though if this is going to be interface specific then LL might be prioritized since that is a link specific address. May want to get some other input from others to gather more feedback.

Can this be overridden? There are some users that may want a specific IP to be returned and not one that may be variable (like docker).

DL6ER:
@dschaper I was thinking about DNS-relays like your router doing the requests on your behalf. The router may be on a different link than the others. May it be WiFi as an additional network segment, ow maybe a VPN the router itself hosts or VLANs, etc. Here, LL may not be reachable (by design) while ULA should be.

We can add a setting to overwrite this if you think there are real-world applications out there.

dschaper:
> We can add a setting to overwrite this if you think there are real-world applications out there.

I think that's a good idea.

DL6ER:
@dschaper Added. Users can now configure
```
REPLY_ADDR4=127.1.2.3
REPLY_ADDR6=fd00::52:1ff:ae21:2ab1
```
to overwrite IPv4/IPv6 automatic detection. They can specify either no, one or both of the options. When they specify only the IPv4 address, the IPv6 will still be detected automatically and vice versa.

dschaper:
> Users can now configure

🥇

dschaper (approved):

dschaper:
This change affects more than just IP mode blocking. `pi.hole` resolves to the detected IP, in most cases this is fine but should be noted. I think the docs state that it affects IP mode only. 

This is what caused the docker images to respond that `pi.hole` was the internal address. We've fixed the docker images to set the FTL config override to handle this.

PR #984, opened 2020-12-17
Improve compatibility with old (ancient) kernels

### LINKED ISSUE #977: pihole-FTL unable to reisze shared memory at create_shm()
### Versions

- Pi-hole:  v5.2.1
- AdminLTE:  v5.2.1
- FTL:  v5.3.2-dirty

### Platform

- OS and version:  Debian 10

- Platform:  Zyxel NAS, armv7l

### Expected behavior

pihole-FTL should start without error.

### Actual behavior / bug

pihole-FTL stops with error: 
FATAL: create_shm(): Failed to resize "FTL-lock" (4) to 28: Operation not supported (-1)

### Steps to reproduce
just normal startup of the service

not relevant

## Debug Token
not needed

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Additional context
[shmem_falloc_5.3.2.tar.gz](https://github.com/pi-hole/FTL/files/5684561/shmem_falloc_5.3.2.tar.gz)
Problem seems to be my ancient kernel, which doesn't reply with the expected result to the fallocate() system function call (may be this function call is not even supported). Unfortunately I can not update as there is no newer kernel available for my hardware. 
Since pihole-FTL v5.2 was working without issues, I checked the difference and found that ftruncate() is used there instead of fallocate() in shmem.c. The attached path does revert the change and now pihole-FTL is working for me. Could you check if using ftruncate() instead of fallocate() would be possible for future pihole-FTL versions?

DL6ER:

## 10

---

Fall back to using `ftruncate()` when `fallocate()` returns with `Operation not supported`. This may happen if the kernel is older than 2.6.23 or `glibc` older than 2.10. `ftruncate()` has its own disadvantages, however, it is POSIX compliant (POSIX.1-2001) so should be supported even by ancient kernels.

Fixes #977

### DIFF
--- a/src/shmem.c
+++ b/src/shmem.c
@@ -511,12 +511,35 @@ SharedMemory create_shm(const char *name, const size_t size, bool create_new)
 	// Using fallocate() will ensure that there's actually space for
 	// this file. Otherwise we end up with a sparse file that can give
 	// SIGBUS if we run out of space while writing to it.
-	const int ret = fallocate(fd, 0, 0U, size);
+	int ret = fallocate(fd, 0, 0U, size);
+	const int fallocate_errno = errno;
 	if(ret != 0)
 	{
-		logg("FATAL: create_shm(): Failed to resize \"%s\" (%i) to %zu: %s (%i)",
-		     sharedMemory.name, fd, size, strerror(errno), ret);
-		exit(EXIT_FAILURE);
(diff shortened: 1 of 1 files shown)


DL6ER:
Confirmed working by the original issue reporter in #977

PromoFaux (approved):

PR #985, opened 2020-12-17
Add UNKNOWN option for REFRESH_HOSTNAME

DL6ER:

## 10

---

This PR adds the new parameter `REFRESH_HOSTNAMES=UNKNOWN` to support only refreshing recently active clients without hostnames.
There will be no PTR queries made for clients where hostnames are known. This also means that known hostnames will never be updated. This is intentional for this option. Other behavior can be achieved with the other available values for this config option.

In addition, we improve the behavior of real-time signal 4: Whenever the signal is arrived, we force a hostname renewal for all clients matching the current `REFRESH_HOSTNAMES` setting. "Forcing" in this context means that the "only resolve recently active clients" rule is ignored.


src/resolve.c
```diff
@@ -386,7 +386,7 @@ static void resolveClients(const bool onlynew)
 
 		// Only try to resolve host names of clients which were recently active if we are re-resolving
 		// Limit for a "recently active" client is two hours ago
-		if(onlynew == false && client->lastQuery < now - 2*60*60)
+		if(!force_refreshing && onlynew == false && client->lastQuery < now - 2*60*60)
```
PromoFaux, line 389:
```suggestion
		if(!force_refreshing && !onlynew && client->lastQuery < now - 2*60*60)
```

Just for consistency, or

```suggestion
		if(force_refreshing == false && onlynew == false && client->lastQuery < now - 2*60*60)
```

PromoFaux (approved):

PR #1007, opened 2020-12-27
Added missing NS query type to getQueryTypes()

Daxtorim:
Signed-off-by: Fabian Preuß <preuss_fabian@gmx.de>

---

**What does this PR aim to fix?:**

When the Query Types chart tries to fetch the query types, it does not receive info about the "NS" type. This results in two issues:

1. The "NS" type is completely missing from the chart. Tooltips do not add up to 100%.
2. The legend of the Query Types chart on the dashboard allows you to filter the Query Log for a specific query type. This works fine for all types except "OTHER". Clicking on "OTHER" filters the Query Log for the type "NS". There is a mismatch between the query type id that the chart forwards to the Query Log and the one FTL will ultimately send back.

What FTL uses:
["A (IPv4)", "AAAA (IPv6)", "ANY", "SRV", "SOA", "PTR", "TXT", "NAPTR", "MX", "DS", "RRSIG", "DNSKEY", "NS", "OTHER"]
What the chart receives:
["A (IPv4)", "AAAA (IPv6)", "ANY", "SRV", "SOA", "PTR", "TXT", "NAPTR", "MX", "DS", "RRSIG", "DNSKEY", "OTHER"]

Added query type "NS" to getQueryTypes().

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -620,11 +620,11 @@ void getQueryTypes(const int *sock)
 		ssend(*sock, "A (IPv4): %.2f\nAAAA (IPv6): %.2f\nANY: %.2f\nSRV: %.2f\n"
 		             "SOA: %.2f\nPTR: %.2f\nTXT: %.2f\nNAPTR: %.2f\n"
 		             "MX: %.2f\nDS: %.2f\nRRSIG: %.2f\nDNSKEY: %.2f\n"
-		             "OTHER: %.2f\n",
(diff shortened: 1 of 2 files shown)


DL6ER (approved):
Ah yes, thanks!

Note: There are plans to add more generic support for query types but it is not yet clear what to do with database entries (converting the entire database would likely take quite some time).

PromoFaux:
Ah... that reminds me

PR #1013, opened 2021-01-03
Implement support for displaying exact type instead of just "OTHER"

DL6ER:

## 10

---

Implement support for displaying exact type instead of the catch-them-all category `OTHER`. The `OTHER` category is still used when it comes to computing statistics to ensure your chart's legend does not explode.

Example:
```
dig -t TYPE65 google.de
dig -t TYPE66 google.de
```
results in:
Screenshot from 2021-01-03 21-52-16
(note that type 65 is correctly translated into the named type `HTTPS` rather than showing `TYPE65`)

And the database contains the expected (`16` = `HTTPS` mapped status, `166` = `TYPE66` with offset 100)
``` sql
SELECT * FROM queries WHERE domain = 'google.de' ORDER BY id DESC LIMIT 2;
```
``` plain
id          timestamp   type        status      domain      client      forward         additional_info
----------  ----------  ----------  ----------  ----------  ----------  --------------  ---------------
10910413    1609706951  166         2           google.de   127.0.0.1   127.0.0.1#5353                 
10910412    1609706943  16          2           google.de   127.0.0.1   127.0.0.1#5353 
```

No changed to core of AdminLTE are necessary for this to work. A small change to the long-term queries page will be needed to handle these offset types still as `OTHER`.

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -878,6 +878,14 @@ void getAllQueries(const char *client_message, const int *sock)
 			continue;
 		// Get query type
 		const char *qtype = querytypes[query->type];
+		char othertype[12] = { 0 }; // Maximum is "TYPE65535" = 10 bytes
+		if(query->type == TYPE_OTHER)
+		{
+			// Format custom type into buffer
+			sprintf(othertype, "TYPE%u", query->qtype);
+			// Replace qtype pointer
+			qtype = othertype;
+		}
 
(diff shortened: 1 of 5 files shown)


PromoFaux (approved):
Reluctantly approving this. It feels dirty, but having exhausted all other options - it is indeed the most efficient way forward.

PR #1040, opened 2021-01-24
Update dnsmasq to v2.84

DL6ER:

## 10

---

According to Simon Kelley, it looks like if two queries are combined (because they ask the same question) then dnsmasq can get confused when it comes to return the answer, the reply to the second query can be sent via the socket that the first one arrived on. That's normally OK, but if the first query arrives via IPv4 and the second via IPv6, for instance, then the bug is triggered.

In addition, FTL can get confused in a similar way because `dnsmasq-v2.83+` forwards multiple queries to the same destination once and stores the other queries as duplicates. They do receive the answer later on, however, this is usually not logged (when `log-queries=extra` is enabled, there will be a warning about the duplicate). Commit 6bb025f in this PR handles such duplicates and introduces a new reply type 14 = "already forwarded" and fixes https://github.com/pi-hole/AdminLTE/issues/1713

### DIFF
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -2280,7 +2280,7 @@ void FTL_duplicate_reply(const int id, int *firstID)
 
 	// The original query may have been blocked during CNAME inspection,
 	// correct status in this case
-	if(duplicated_query->status != QUERY_FORWARDED)
+	if(source_query->status != QUERY_FORWARDED)
 		duplicated_query->status = source_query->status;
 	duplicated_query->CNAME_domainID = source_query->CNAME_domainID;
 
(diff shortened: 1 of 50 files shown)


src/dnsmasq_interface.c
```diff
@@ -2278,6 +2278,12 @@ void FTL_duplicate_reply(const int id, int *firstID)
 	duplicated_query->dnssec = source_query->dnssec;
 	duplicated_query->flags.complete = true;
 
+	// The original query may have been blocked during CNAME inspection,
+	// correct status in this case
+	if(duplicated_query->status != QUERY_FORWARDED)
```
Daxtorim, line 2283: If I understand this correctly `source_query` is the query that actually got forwarded and `duplicated_query` is the query that was "ignored" for the time being. Should this then not be `if(source_query->status != QUERY_FORWARDED)` since the "ignored" query's status will never be `QUERY_FORWARDED`?

DL6ER: Yes.

DL6ER:
Superseded by https://github.com/pi-hole/FTL/pull/1043

PR #1047, opened 2021-01-31
Fix incorrect "FATAL" error message during garbage collection

DL6ER:

## 10

---

Fixes a harmless incorrect `FATAL` message in FTL's log.

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -72,16 +72,12 @@ void *GC_thread(void *val)
 				if(query->timestamp > mintime)
 					break;
 
-				// Adjust client counter
+				// Adjust client counter (total and overTime)
 				clientsData* client = getClient(query->clientID, true);
-				if(client != NULL)
-					change_clientcount(client, -1, 0, 0, 0);
-
-				// Adjust total counters and total over time data
 				const int timeidx = query->timeidx;
 				overTime[timeidx].total--;
 				if(client != NULL)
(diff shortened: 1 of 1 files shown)


src/gc.c
```diff
@@ -72,24 +72,19 @@ void *GC_thread(void *val)
...
 				clientsData* client = getClient(query->clientID, true);
-				if(client != NULL)
-					change_clientcount(client, -1, 0, 0, 0);
-
-				// Adjust total counters and total over time data
 				const int timeidx = query->timeidx;
 				overTime[timeidx].total--;
 				if(client != NULL)
-					change_clientcount(client, 0, 0, timeidx, -1);
+					change_clientcount(client, -1, 0, timeidx, -1);
```
DL6ER, line 80: This reduces the number of functions calls from two to one.

src/gc.c
```diff
@@ -104,8 +99,12 @@ void *GC_thread(void *val)
 						// Forwarded to an upstream DNS server
 						// Adjust counters
 						counters->forwarded--;
-						if(upstream != NULL)
-							upstream->count--;
+						if(query->upstreamID > -1)
+						{
+							upstreamsData* upstream = getUpstream(query->upstreamID, true);
```
DL6ER, line 104: This is the actual fix: Do not try to access the data of an upstream destination with ID < 0 (this is causing the wrong warning here).

PromoFaux (approved):

PR #1052, opened 2021-02-03
Add per-client rate-limiting

DL6ER:

## 10

---

Add per-client rate-limiting. Rate-limited queries are answered with a `REFUSED` reply and not further processed by FTL Even when they are logged in `pihole.log`, they will not contribute to the overall statistics nor enter the Query Log or the database.
This serves the purpose of a real rate-limit and ensures that abnormally behaving clients hammering FTL with thousands of queries per second cannot lead to a denial-of-service failure.

Rate-limiting is very customizable, it defaults to allowing not more than 1000 queries in 60 seconds. Both numbers can be changed by the user.

It is important to note that rate-limiting is happening on a *per-client* basis. Other clients can continue to use FTL while rate-limited clients are short-circuited at the same time.

Rate-limiting can be disabled by setting `RATE_LIMIT=0/0`.

One might argue that rate-limiting should best be realized with a firewall. However, we do not want to touch user firewalls and this effectively does the same thing (albeit better because we don't simply drop but reply with a proper `REFUSED` message).

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -461,6 +461,25 @@ void read_FTLconf(void)
 		logg("   REFRESH_HOSTNAMES: Periodically refreshing IPv4 names");
 	}
 
+	// RATE_LIMIT
+	// defaults to: 1000 queries / 60 seconds
+	config.rate_limit.count = 1000;
+	config.rate_limit.interval = 60;
+	buffer = parse_FTLconf(fp, "RATE_LIMIT");
+
+	unsigned int count = 0, interval = 0;
+	if(buffer != NULL && sscanf(buffer, "%u/%u", &count, &interval) == 2)
+	{
+		config.rate_limit.count = count;
+		config.rate_limit.interval = interval;
+	}
+
+	if(config.rate_limit.count > 0)
+		logg("   RATE_LIMIT: Rate-limiting client making more than %u queries in %u second%s",
+		     config.rate_limit.count, config.rate_limit.interval, config.rate_limit.interval == 1 ? "" : "s");
+	else
+		logg("   RATE_LIMIT: Disabled");
+
 	// Read DEBUG_... setting from pihole-FTL.conf
 	read_debuging_settings(fp);
 
@@ -523,7 +542,7 @@ static char *parse_FTLconf(FILE *fp, const char * key)
 
 	// Go to beginning of file
 	fseek(fp, 0L, SEEK_SET);
-	
+
 	if(config.debug & DEBUG_EXTRA)
 		logg("initial: conflinebuffer = %p, keystr = %p, size = %zu", conflinebuffer, keystr, size);
 
--- a/src/config.h
+++ b/src/config.h
@@ -53,10 +53,14 @@ typedef struct {
 	int dns_port;
 	unsigned int delay_startup;
 	unsigned int network_expire;
+	struct {
+		unsigned int count;
+		unsigned int interval;
+	} rate_limit;
 	enum debug_flags debug;
 	time_t DBinterval;
 } ConfigStruct;
-ASSERT_SIZEOF(ConfigStruct, 56, 48, 48);
+ASSERT_SIZEOF(ConfigStruct, 64, 56, 56);
 
 typedef struct {
 	const char* conf;
--- a/src/database/query-table.c
+++ b/src/database/query-table.c
@@ -432,7 +432,7 @@ void DB_read_queries(void)
 			query->type = TYPE_OTHER;
 			query->qtype = type - 100;
 		}
-		
+
 		query->status = status;
 		query->domainID = domainID;
 		query->clientID = clientID;
--- a/src/datastructure.h
+++ b/src/datastructure.h
@@ -79,6 +79,7 @@ typedef struct {
 	int blockedcount;
 	int aliasclient_id;
 	unsigned int id;
+	unsigned int rate_limit;
 	unsigned int numQueriesARP;
 	int overTime[OVERTIME_SLOTS];
 	size_t groupspos;
@@ -88,7 +89,7 @@ typedef struct {
 	time_t lastQuery;
 	time_t firstSeen;
 } clientsData;
-ASSERT_SIZEOF(clientsData, 688, 664, 664);
+ASSERT_SIZEOF(clientsData, 696, 668, 668);
 
 typedef struct {
 	unsigned char magic;
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -190,8 +190,8 @@ static bool _FTL_check_blocking(int queryID, int domainID, int clientID, const c
 			// as something along the CNAME path hit the whitelist
 			if(!query->flags.whitelisted)
 			{
-				query_blocked(query, domain, client, QUERY_BLACKLIST);
 				force_next_DNS_reply = dns_cache->force_reply;
+				query_blocked(query, domain, client, QUERY_BLACKLIST);
 				return true;
 			}
 			break;
@@ -210,8 +210,8 @@ static bool _FTL_check_blocking(int queryID, int domainID, int clientID, const c
 			// as sometving along the CNAME path hit the whitelist
 			if(!query->flags.whitelisted)
 			{
-				query_blocked(query, domain, client, QUERY_GRAVITY);
 				force_next_DNS_reply = dns_cache->force_reply;
+				query_blocked(query, domain, client, QUERY_GRAVITY);
 				return true;
 			}
 			break;
@@ -532,19 +532,9 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
 		return false;
 	}
 
-	// Lock shared memory
-	lock_shm();
-
-	// Ensure we have enough space in the queries struct
-	memory_check(QUERIES);
-	const int queryID = counters->queries;
-
 	// If domain is "pi.hole" we skip this query
 	if(strcasecmp(name, "pi.hole") == 0)
-	{
-		unlock_shm();
 		return false;
-	}
 
 	// Convert domain to lower case
 	char *domainString = strdup(name);
@@ -574,9 +564,49 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
 	   (strcmp(clientIP, "127.0.0.1") == 0 || strcmp(clientIP, "::1") == 0))
 	{
 		free(domainString);
+		return false;
+	}
+
+	// Lock shared memory
+	lock_shm();
+
+	// Find client IP
+	const int clientID = findClientID(clientIP, true, false);
+
+	// Get client pointer
+	clientsData* client = getClient(clientID, true);
+	if(client == NULL)
+	{
+		// Encountered memory error, skip query
+		// Free allocated memory
+		free(domainString);
+		// Release thread lock
 		unlock_shm();
 		return false;
 	}
+
+	// Check rate-limit for this client
+	if(config.rate_limit.count > 0 &&
+	   ++client->rate_limit > config.rate_limit.count)
+	{
+		if(config.debug & DEBUG_QUERIES)
+		{
+			logg("Rate-limiting %s %s query \"%s\" from %s:%s",
+			     proto == TCP ? "TCP" : "UDP",
+			     types, domainString, next_iface, clientIP);
+		}
+
+		// Block this query
+		force_next_DNS_reply = REFUSED;
+
+		// Do not further process this query, Pi-hole has never seen it
+		unlock_shm();
+		return true;
+	}
+
+	// Ensure we have enough space in the queries struct
+	memory_check(QUERIES);
+	const int queryID = counters->queries;
 
 	// Log new query if in debug mode
 	if(config.debug & DEBUG_QUERIES)
@@ -606,9 +636,6 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
 
 	// Go through already knows domains and see if it is one of them
 	const int domainID = findDomainID(domainString, true);
-
-	// Go through already knows clients and see if it is one of them
-	const int clientID = findClientID(clientIP, true, false);
 
 	// Save everything
 	queriesData* query = getQuery(queryID, false);
@@ -661,19 +688,6 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
 
 	// Update overTime data
 	overTime[timeidx].total++;
-
-	// Get client pointer
-	clientsData* client = getClient(clientID, true);
-	if(client == NULL)
-	{
-		// Encountered memory error, skip query
-		logg("WARN: No memory available, skipping query analysis");
-		// Free allocated memory
-		free(domainString);
-		// Release thread lock
-		unlock_shm();
-		return false;
-	}
 
 	// Update overTime data structure with the new client
 	change_clientcount(client, 0, 0, timeidx, 1);
@@ -761,11 +775,19 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
 void _FTL_get_blocking_metadata(union all_addr **addrp, unsigned int *flags, const char* file, const int line)
 {
 	// Check first if we need to force our reply to something different than the
-	// default/configured blocking mode For instance, we need to force NXDOMAIN
-	// for intercepted _esni.* queries
+	// default/configured blocking mode. For instance, we need to force NXDOMAIN
+	// for intercepted _esni.* queries.
 	if(force_next_DNS_reply == NXDOMAIN)
 	{
 		*flags = F_NXDOMAIN;
+		// Reset DNS reply forcing
+		force_next_DNS_reply = 0u;
+		return;
+	}
+	else if(force_next_DNS_reply == REFUSED)
+	{
+		// Empty flags result in REFUSED
+		*flags = 0;
 		// Reset DNS reply forcing
 		force_next_DNS_reply = 0u;
 		return;
@@ -1662,7 +1684,7 @@ static void save_reply_type(const unsigned int flags, const union all_addr *addr
                             queriesData* query, const struct timeval response)
 {
 	// Iterate through possible values
-	if(flags & F_NEG)
+	if(flags & F_NEG || force_next_DNS_reply == NXDOMAIN)
 	{
 		if(flags & F_NXDOMAIN)
 		{
@@ -1694,15 +1716,15 @@ static void save_reply_type(const unsigned int flags, const union all_addr *addr
 		// TXT query
 		query->reply = REPLY_RRNAME;
 	}
-	else if(flags & F_RCODE && addr != NULL)
+	else if((flags & F_RCODE && addr != NULL) || force_next_DNS_reply == REFUSED)
 	{
-		const unsigned int rcode = addr->log.rcode;
-		if(rcode == REFUSED)
+		if((addr != NULL && addr->log.rcode == REFUSED)
+		   || force_next_DNS_reply == REFUSED )
 		{
 			// REFUSED query
 			query->reply = REPLY_REFUSED;
 		}
-		else if(rcode == SERVFAIL)
+		else if(addr != NULL && addr->log.rcode == SERVFAIL)
 		{
 			// SERVFAIL query
 			query->reply = REPLY_SERVFAIL;
--- a/src/gc.c
+++ b/src/gc.c
@@ -23,29 +23,46 @@
 
 bool doGC = false;
 
-time_t lastGCrun = 0;
+static void reset_rate_limiting(void)
+{
+	for(int clientID = 0; clientID < counters->clients; clientID++)
+	{
+		clientsData *client = getClient(clientID, true);
+		if(client != NULL)
+			client->rate_limit = 0;
+	}
+}
+
 void *GC_thread(void *val)
 {
 	// Set thread name
 	prctl(PR_SET_NAME,"housekeeper",0,0,0);
 
-	// Save timestamp as we do not want to store immediately
-	// to the database
-	lastGCrun = time(NULL) - time(NULL)%GCinterval;
+	// Remember when we last ran the actions
+	time_t lastGCrun = time(NULL) - time(NULL)%GCinterval;
+	time_t lastRateLimitCleaner = time(NULL);
 	while(!killed)
 	{
-		if(time(NULL) - GCdelay - lastGCrun >= GCinterval || doGC)
+		const time_t now = time(NULL);
+		if((unsigned int)(now - lastRateLimitCleaner) >= config.rate_limit.interval)
+		{
+			lastRateLimitCleaner = now;
+			lock_shm();
+			reset_rate_limiting();
+			unlock_shm();
+		}
+		if(now - GCdelay - lastGCrun >= GCinterval || doGC)
 		{
 			doGC = false;
 			// Update lastGCrun timer
-			lastGCrun = time(NULL) - GCdelay - (time(NULL) - GCdelay)%GCinterval;
+			lastGCrun = now - GCdelay - (now - GCdelay)%GCinterval;
 
 			// Lock FTL's data structure, since it is likely that it will be changed here
 			// Requests should not be processed/answered when data is about to change
 			lock_shm();
 
 			// Get minimum time stamp to keep
-			time_t mintime = (time(NULL) - GCdelay) - MAXLOGAGE*3600;
+			time_t mintime = (now - GCdelay) - MAXLOGAGE*3600;
 
 			// Align to the start of the next hour. This will also align with
 			// the oldest overTime interval after GC is done.


PromoFaux:
> it defaults to allowing not more than 1000 queries in 60 seconds.

I wonder, would a default of `off` be better?

DL6ER:
I guess we want this protection to be enabled by default. Use cases are clients going crazy because of some defect and/or DNS loops between a router and the Pi-hole. We can add a warning to the Pi-hole diagnosis system if you like so users are aware of this.

My first attempt was showing rate-limited queries in the Query Log, however, this would not effectively help against a DoS attack as the memory needed to hold them would still quickly grow.

We should rather stress in the change log that this is something you can disable. The numbers I put in should not even be close to be triggered by any correctly working client. 10,000 queries per minute still allow 14,4 million queries **per-client** in 24 hours. We may even want to reduce this number but `off` by default seems wrong to me.

PromoFaux (approved):

LordSimal:
Would be nice if this new rate limiting setting can be adjusted via the Web GUI.
I was wondering why my backup script wasn't working anymore since 2 weeks and just recently stumbled upon this new feature.
Currently I added the setting into the .conf file as described in the doc here and now everything works again.
https://docs.pi-hole.net/ftldns/configfile/#rate_limit

DL6ER:
@LordSimal Why is your backup script getting rate-limited in the first place? And do you consider this healthy behavior?

LordSimal:
I use rclone to sync my server data to PCloud. I cant decide/change any way how rclone connects to the server. I guess every request, and therefore ever file/folder, is creating a new DNS query.
I know the above comments are saying you can adjust that ratelimiting behaviour per client but where/how do i do that?

DL6ER:
Rate-limiting is only measured and applied per-client. You cannot set different levels for different devices. The proper fix for this behavior seems to be a local DNS chance on the server. Like what modern Ubuntu ships with `systemd-resolved`

LordSimal:
Well I am running on Ubuntu 20 so I will look into that.

Still I think it would be nice if there would be at least a notice in the backend when a client is being rate limited (not only in debug mode).
Or at least notify the user about that new feature when updating (which now is a bit late i would say)

On the client side I only got "server misbehaving" (which is OK of course) and in the pihole.log the Query was logged with REFUSED. Can that at least be adapted to REFUSED/RATELIMITED or something that leads to this feature?
Or perhaps a notification of some sort to the system admin that a specific client is being rate limited right now?

jfb-pihole:
"Or at least notify the user about that new feature when updating (which now is a bit late i would say)"

In all of our release announcements, we have stressed the need to read the release notes for each new release.  This new feature, in particular, was thoroughly discussed in the V5.7 release notes:

https://pi-hole.net/2021/02/16/pi-hole-ftl-v5-7-and-web-v5-4-released/#page-content

LordSimal:
Im sorry, I didn't look at the blog/release notes 🙇🏻

anno006:
Or/and first make a web-gui instell page nd make it possible to enter it per client. Before you make a drastic short mindbreak.So next time make sure to have everything ready before you inplement stuff. as a router, fer example pfsense use pihole to. And the router has a lot of dns quiries from other devices..... smart!!!

DL6ER:
@anno006 Smart people read the release notes or blog posts before updating. We mentioned the option to disable rate-limiting and you could have done so even before updating to ensure there would have been no downtime.

In your case, your router (pfSense) seems to be configured incorrectly which is the cause for this. When using ECS (EDNS0 Client Subnet), the Pi-hole can tell apart your clients even when the router makes all the queries on their behalf. We are preparing a pfSense + Pi-hole step-by-step to make this more obvious in the future.

As you can see above, this feature was merged and released more than half a year ago. In the meantime, it *always* turned out that there is an underlying problem (rouge client) whenever rate limiting was triggered. Even for those where the router is the single point of contact. The default value is rather permissive: 1000 queries every 60 seconds equals up to 1.4 million queries a day. If your network is larger than this, you should clearly have your router configured accordingly to stop things like these from happening.

anno006:
> @anno006 Smart people read the release notes or blog posts before updating. We mentioned the option to disable rate-limiting and you could have done so even before updating to ensure there would have been no downtime.
> 
> In your case, your router (pfSense) seems to be configured incorrectly which is the cause for this. When using ECS (EDNS0 Client Subnet), the Pi-hole can tell apart your clients even when the router makes all the queries on their behalf. We are preparing a pfSense + Pi-hole step-by-step to make this more obvious in the future.
> 
> As you can see above, this feature was merged and released more than half a year ago. In the meantime, it _always_ turned out that there is an underlying problem (rouge client) whenever rate limiting was triggered. Even for those where the router is the single point of contact. The default value is rather permissive: 1000 queries every 60 seconds equals up to 1.4 million queries a day. If your network is larger than this, you should clearly have your router configured accordingly to stop things like these from happening.

Really smart people will accept there mistakes and learn from them instead of backfire. As i said it is possible to overrun you limet. Next time make the step by step before activating...

dschaper:
@anno006 Please let me know your receipt number and I'll be more than happy to refund your purchase.

DL6ER:
It wasn't my intention to "backfire". Not in the slightest. Rereading my own post, I can see how you can have seen it differently. You have my sincere apologies for that. I just wanted to point out that you can improve your router configuration to get an improvement which also influences rate-limiting.

I can also assure you that we have no issues with undoing a change if it turned out to be incorrect. We have at least a few thousand Pi-hole users out there and I've heard less than five complaints about it. Typically, users were even thankful because it revealed misbehaving clients and they were able to do something against it. And I have seen at least one report that helped a user to keep the rest of their network responsive when one client went nuts and started requesting millions of queries per minute.

What do you consider to be the mistake exactly? Rate-limiting in the first place or do you consider the default value too low? Would it still be too low in your network when pfSense would be configured such that clients can be kept apart? Or is the problem rather that the value is not modifiable via the web dashboard? If the latter, please take into account that the vast majority of advanced settings is currently not editable on the wbe interface.

anno006:
I did search for the terms you posted. But this is also pretty new. it has the same problem, not inplemented yet or atleast understandable/findable for the peaple involved in the devoloping. not for people like me. don't understand me wrong i don't dislike it, only the way you inplement it.

1. " Rate-limiting in the first place or do you consider the default value too low?" you do have good point to have a form of rate-limitation, so no with the aguments. i can see that it is for some situations the value to low. 
2. "Would it still be too low in your network when pfSense would be configured such that clients can be kept apart?" i did read the forum posts about EDNS and if it is a but further in development and i can inplement it than this will be a good workaround/ solution.
3. "Or is the problem rather that the value is not modifiable via the web dashboard?" if the default value is for whatever reason not working or good that it is indeed best to have it in the webur. Or even i can put commands in, but need to know with. And the only info i can find is: https://docs.pi-hole.net/ftldns/configfile/ i do miss where i need to change the RATE_LIMIT value.
4. "If the latter, please take into account that the vast majority of advanced settings is currently not editable on the wbe interface" I thought so, but this is the first time i actually run into a problem that is so new there is amost no information to find. Exept all of you (cryptic) posts... 

i do like what i read about the EDNS and you explenation that it can help with misbehaving. I will look out for the step by step.

And last some background info about my system. i am running a pfsense (own hardware) run some vm's for homeautomation and pihole, have about 200000 queries in 24 hours.

DL6ER:
> the only info i can find is: https://docs.pi-hole.net/ftldns/configfile/ i do miss where i need to change the RATE_LIMIT value.

The link to the documentation is directly there on the web dashboard. The place where to put the options is at the very top of this linked document:

> You can create a file /etc/pihole/pihole-FTL.conf that will be read by FTLDNS on startup.
>
> Possible settings (the option shown first is the default):
> [...]

I'm open to suggestions how to improve the situation.

> i can see that it is for some situations the value to low.

Yes. We've had a lot of discussions about the default value on our Discourse forum and this is what we concluded on. The motivation behind this is that Pi-hole is typically employed in a regular household with typically not more than a dozen devices. If Pi-hole is run on an Raspberry Pi and only one of the clients goes near the rate limiting limit of 1.4 million queries a day, the Pi-hole will continue to work. However, if only two clients go close to this query rate, the Pi-hole will eventually stop working correctly because the memory will be used up. Hence, we didn't want to set the default even higher. Of course, there are much larger networks and Pi-hole may be running on much beefier hardware with more than 1 GB of memory available, but this is rather edge-case than a default setup. This just to give some background of why we chose this value.

> I will look out for the step by step.

You are not the first one asking for this, we're currently preparing a blog post about this. @dschaper might be able to give some more info about the current status of it (I do not use pfSense myself).

> i am running a pfsense (own hardware) run some vm's for homeautomation and pihole, have about 200000 queries in 24 hours.

So this means roughly 140 queries per minute so about 7 times lower than the default limit. So we're talking about peak load here. We could easily change the default from 1,000 queries in 1 minute to, say, 10,000 queries in 10 minutes. Even when this would not change the upper limit, it'd relax the peak load issue. *However*, this would also mean that clients will only later be rate-limited and blocking them will take longer. This is the price to pay. @dschaper thoughts?

dschaper:
I have most of the guide written for OPNsense. If pfSense uses `dnsmasq` then it's a few lines of configuration. Bonus is if you can run `unbound` as well then the Sense stays as DHCP server and sole DNS server for the network while it forwards queries along with ECS/EDNS0 information to Pi-hole. No need for Conditional Forwarding. 

The basic idea is to add 
```
add-mac
add-subnet=32,128
```
as a `dnsmasq` config snippet and then adjust the DNS servers accordingly after including DHCP lease names in `dnsmasq`. (You may also need to check if rebind protection is something pfSense does, I know OPNsense has it on by default.)

dschaper:
@anno006 Please read https://pi-hole.net/2021/09/30/pi-hole-and-opnsense/ and see if this works on pfsense as well. Thanks!

blakev:
Awful feature that cost me hours of lost work. Not even exposed in the GUI.

blakev:
I was too frustrated yesterday because it basically ruined my whole day.

The feature itself is good, and has a lot of merit. I think I was most frustrated by not seeing anything about rate limiting in the GUI. If there was a checkbox in the admin interface that was broad `Enable rate limiting (see docs)` or something that would have saved hours on my debug time. I did not notice the warning icon because I'm color blind, and I didn't think the pihole was the problem because my main workstation is a heavily configured Linux box. 

In the screenshot I posted on reddit the maximum DNS queries were 2x'd because my machine's IP was throttled after 1k requests, then it went to the router's default DNS and exhausted its 1k. So, to me, it looked like my internet was dying randomly (wireless ISP on a rainy day) or there was something wrong with my desktop.

In an ideal world I think it would be nice to expose "random" configuration files in the GUI by loading them 1-for-1 in a textbox. At least then I would have noticed there was a setting that needed tweaking.

FWIW after I read up on the documentation I did throttle my IoT devices and bumped my lab to 10k/req per 30s.

DL6ER:
Thanks for elaborating on this. I've been working on a high-contrast like (black & white) theme for some time but never really finished it because support for this kind of styling is pretty bad in many of the used standard libraries like those used to generate the dashboard graphics. Anyway, what would you say to this example (currently work-in-progress and unlikely to be the final result)?
pi hole_admin_index php (1)

Would you say you'd have noticed the :warning: better in this screenshot or is it maybe too small in general?
Having said that I'm still not sure I like it very much (see lack of support for distinct colors in external components mentioned above).

Also, I'll revisit adding a setting for this feature. We planned to have a completely redesigned settings page for Pi-hole v6.0 and I stopped adding new settings because this always introduces new merge conflicts with the ongoing development, but if this is needed now then this may be.

Z3r0Auth:
Having run into the same challenge others are mentioning above (apps hitting rate limiting) I acknowledge it's not a problem where applications and systems are properly configured to cache DNS queries locally. But we don't live in a perfect world, apps aren't perfect, and this can be an absolute head scratcher when those same apps provide absolutely trash error reporting to even start sending me the direction of DNS being the issue.   
  
So the goal of instituting a default rate limit is to prevent denial of service issues should a local client go absolutely off the rails, right...? so should the limit not be decided around the capabilities of the minimum hardware spec of a pihole device?  
  
What's the current maximum query response for in 1 minute for a Raspberry Pi B+, the lowest spec Pi that meets the min spec 512mb of ram. Hell.. what is it on an original A?  
  
The default value should be expressed proportionally to the capability of the lowest spec hardware, to accomplish the desired role of maintaining quality of life for other devices relying upon the same device for DNS queries, instead of after reading this thread what feels like an arbitrary "this feels like a lot of requests"  
  
I also concur with others who have said the controls for this being config file only definitely made this one a bigger pita to troubleshoot.

dschaper:
> So the goal of instituting a default rate limit is to prevent denial of service issues should a local client go absolutely off the rails, right...?

And to prevent log spam, database size explosive growth, web interface extreme slowness...

While your idea is interesting, it's not all that feasible. What if the hardware is beefy but there  are a lot of other applications running? We'd need to benchmark the server to see what it's capable of, and then periodically test again to see if the hardware is still capable of the load.

Also, the other concerns for why the rate limiting was implemented aren't really processor specific. Storage size, disk I/O, memory and other things come in to play.

Z3r0Auth:
I think using a tool like [DNSBlast ](https://github.com/jedisct1/dnsblast)would achieve a reasonable approximation of the maximum performance of a baseline device, and then set a standard based on that number, to leave overhead.  
  
True, you can't accommodate for anything else someone may also be running on the same device, but I don't think setting a 1m max that represents a 20-40% load approximation of the worst case scenarios (an app literally designed to shit out un-cached DNS queries) wouldn't be a better answer for 99% of valid app use cases that are badly coded then "eh, a thousand seems like a lot, right?", and you'd certainly be on a better footing when the "why that value?" question pops up.

A device bumping up against that 40% number getting shut down would likely have an almost unnoticeable impact on other services/applications thanks to said device getting rate limited in under a minute.

dschaper:
You make assumptions that we didn't do any kind of testing. Though we've had this fork https://github.com/pi-hole/dnsblast since 2017.

You also ignored most of what I responded. You're only factoring in one specific metric. CPU. So you have a 16 core server that can handle billions of queries per minute. So you allow it to have a very high rate limit. Then it crashes when the log fills the storage, the database grows to enormous size and any attempt to view the query log via the web interface takes 4 hours to load a page. 

It wasn't just a "Hey, 1k sound like a great number, at least that's what the dartboard said." 

I understand that this may have caused you issues in troubleshooting, but the rate limiting is logged and warnings are produced. I know when I troubleshoot something the first thing I look for are the daemon logs to see what is being said by the daemon for guidance on where to look to start the process.

Z3r0Auth:
I welcome enlightenment on what metrics were used to decide on the default 1000/1m rate limit setting.

DL6ER:
As FTL is very efficient on CPU load, the main limitation is memory. We had a quite extensive discussion about this somewhere on our Discourse forum, I'd have to look this up later. Anyway, it is easily explained why memory is the issue here:

FTL uses a time-memory trade-off. You have to decide whether you either want to compute statistics from data on disk (which is *horribly* slow on Raspberry Pi's) or hold an in-memory copy of the relevant stuff making Pi-hole as fast and usable as it is right now.

When a client is close to the limit of issuing 1,000 queries per minute, it alone would be making 1.4 million queries in 24 hours. Currently, one query consumes about 56 bytes of memory, a bit less on ARM (see [here](https://github.com/pi-hole/FTL/blob/master/src/datastructure.h#L55)) plus some bookkeeping memory on top of this (we store the domain strings, etc. separately), making this one client consuming about 100 MB of RAM on your Pi-hole's RAM and also on disk. In one month, this client alone will make your database grow by 3 GB. If you have several rouge clients, this multiplies the numbers.

100 MB of RAM *per client* is something we have found to be something like the upper limit to avoid resource exhaustion especially on the devices that have only 1 GB of RAM available, may be running other things besides Pi-hole and, obviously, also have to put their operating system's stuff somewhere into this memory, too. And don't forget about other clients using the Pi-hole at the same time, consuming extra money at the same time.

It just seemed to be the most sensible solution given resource constraints on low-end products. If you would, say, tenfold the per-client limit (`10000/1m`), you'd immediately end up at 1 GB per client next to the operating system which would already be too much for some systems. If you make the limit `5000/1m`, two clients could easily kill the system, etc.

In the end, the real numbers suggested something like `1056/1m`. I guess that wouldn't have made all that much of a difference to the "nice number" of `1000/1m` we are using right now. In the end, even `2000/1m` would likely not improve all that much on rouge clients. And we specifically implemented this *per client* to ensure other clients in the network will still operate when one device goes crazy.

To bring the issue to the user's attention, we drop a message into the Pi-hole diagnosis system whenever a rate-limitation happens. In the next release, we make the icon highlighting that there is something you should look at twice as large and animate it to catch more attention. Hopefully, this will finally do away with users wasting time to find out about this feature. Despite the users expressing their dislike here, I'm still convinced it is a good feature and I also don't think the numbers we chose are off. But I can most certainly be convinced of the opposite and we can change anything when there are clear arguments to do so.

bretonium:
I found this feature when software was trying to resolve hostnames that could be resolved only inside corporate VPN. It led to a weird loop: i got ratelimited, could not resolve my VPN hostname, programs could not resolve their hostnames and were retrying and it led to ratelimiting, preventing resolving VPN hostname. I could not google this problem (could not resolve google) with the rate-limited computer and had to use another one.

Although "retry-like-crazy" is probably incorrect behavior for software, nobody can change it, and it led to an unpleasant situation. I would prefer ratelimiting being opt-in rather than opt-out, because it was like that for many years.

bretonium:
i am wondering, could ratelimiting be not only per-client, but also per-client-and-success? It would let me connect to the VPN or to google (or to slack, teams and the rest of the corporate world) and mute the only-inside-vpn software

Raineer:
> I welcome enlightenment on what metrics were used to decide on the default 1000/1m rate limit setting.

Same here. I've hit it twice today doing nothing out of the ordinary. One was building a package in Arch (the script happened to download Imagemagick libraries in many different chunks), the other was downloading ts chunks of a video stream to concatenate. These were both serial requests yet, apparently, easily triggered this default.

The error was presented in the GUI, I was able to google a fix, and I'm up and running. But I heavily caution so many in this thread thinking the only way to get the default 1000/60 limit is a mistake. This is too conservative by at least an order of magnitude.

dschaper:
I'd counter that with Arch being broken if it's querying 1000 different hosts to get imagemagick libraries in one minute. At least it should be caching some of the domain queries.

DL6ER:
>>  I welcome enlightenment on what metrics were used to decide on the default 1000/1m rate limit setting.
>
> Same here

Please have a look at [my reply](https://github.com/pi-hole/FTL/pull/1052#issuecomment-997177481) to the initial request of enlightenment.

> This is too conservative by at least an order of magnitude.

I disagree for the reasons already mentioned above:

> If you would, say, tenfold the per-client limit (10000/1m), you'd immediately end up at 1 GB **per client** next to the operating system which would already be too much for some systems.

Maybe we can add a step in the installer script that asks the user to specify a custom limit (or accept the default) so you can guess what would be appropriate for your setup.

bretonium:
> I disagree for the reasons already mentioned above:
> 
> > If you would, say, tenfold the per-client limit (10000/1m), you'd immediately end up at 1 GB **per client** next to the operating system which would already be too much for some systems.

I appreciate the level of care taken about system resources, especially in the context of rpi's constrained RAM.

However, i feel that the approach taken here is not the commonly accepted one. We have many well-known mechanisms in the OS to tackle the problem of lack of memory. If i don't have enough RAM, OOM kicks in, and it is loud, and it is known where to look for its signs, and it will probably yell at me as soon as i ssh to the rpi. If i am ok with some slowness, i add swap. If i want to limit RAM, i use cgroups. And those who care about system resources monitor them with grafana or something else.

Firefox does not refuse to open the 101-st tab. Messenger does not refuse to accept n-th message. Terminal does not refuse to execute m-th command. They just try to do it.

I still think that rate-limiting should be opt-in instead of opt-out.

DL6ER:
The comparison isn't valid because Firefox spawn individual processes that can by killed independently without taking the entire application down. Neither do messagers, etc. keep all their stuff in memory. Pi-hole keeps all queries in memory intentionally so you can do quick filtering and requesting. In the end, Pi-hole always pays attention to work fast even on the low-end of hardware and this means we cannot do a ton of disk lookups over the slow SD interface.

I also disagree about OOM being "loud". You will not notice a single bit of its action if you are not connected to a live terminal that will show "Killed." or if you are not used to reading the system logs manually. I would argue that Pi-hole's rate-limiting is a lot louder (for the inexperienced user) as it will be shown prominently on the dashboard and an in the log files, too. I perfectly see how power users want to have more control and that us giving them the control via a config options isn't obvious enough.

This and 

> I still think that rate-limiting should be opt-in instead of opt-out.

is why I suggested

> Maybe we can add a step in the installer script that asks the user to specify a custom limit (or accept the default) so you can guess what would be appropriate for your setup.

This will explicitly ask you what to do. It will neither by an opt-in or opt-out. Instead it will be explicitly asking. This ensures we are not actually making any assumptions for what is good for the user.

okaestne:
I also just found out about this feature, by restoring my last firefox session with like 15 tabs. Reading this thread I agree on the rate limiting itself, even if it might be too conservative per default.
The actual reason I comment is: how long does the rate limit last? How to configure it?

yubiuser:
> The actual reason I comment is: how long does the rate limit last? How to configure it?

https://docs.pi-hole.net/ftldns/configfile/#rate_limit

okaestne:
> > The actual reason I comment is: how long does the rate limit last? How to configure it?
> 
> https://docs.pi-hole.net/ftldns/configfile/#rate_limit

I meant the time of being limited.

PromoFaux:
It's all there in the docs.

>...
>The default settings for FTL's rate-limiting are to permit no more than `1000` queries in `60` seconds. 
...
>For this setting, both numbers, the maximum number of queries within a given time, and the length of the time interval (seconds) have to be specified.
...
>  For instance, if you want to set a rate limit of 1 query per hour, the option should look like `RATE_LIMIT=1/3600`

okaestne:
> It's all there in the docs.

Correct me if I'm wrong, but all I see there is the time and the amount of queries needed to **get limited**. But I want to set the time of **being rate limited**. Even after 5 minutes I was unable to use anything on my PC because of being limited. Therefore, I would like to know, how to configure the time of being in the state "rate limited" or at least how to un-limit again. Not setting the bounds of before getting limited.

yubiuser:
@okaestne 

I agree it's not that obvious and we need to update the documentation. Since https://github.com/pi-hole/FTL/pull/1199
the rate limiting works the following way:
1) FTL uses a fixed 1 minute counting interval (starting straight at the minute)
2) if a client exceeds the set limit it will be blocked **until the end of the counting interval** (it will let you know in `/var/log/pihole-FTL.log` something like `Rate-limiting 10.0.1.39 for at least 44 seconds`)
3) if the client will exceed the limit while being blocked, it will also be blocked in the next interval (minute) as well.

So far, there is no way to 
> set the time of being rate limited

It's always until the end of the interval (straight minute) or until the end of the next interval

okaestne:
@yubiuser 

Ah I see, thanks. So likely systemd-resolve needs to be reconfigured to don't retry too often..
```[2022-01-16 13:09:32.091 1314M] Rate-limiting 192.168.178.10 for at least 14 seconds
[2022-01-16 13:09:46.579 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 1007 queries
[2022-01-16 13:10:46.819 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 11928 queries
[2022-01-16 13:11:46.055 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 5976 queries
[2022-01-16 13:12:46.295 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 14856 queries
[2022-01-16 13:13:46.535 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 1176 queries
[2022-01-16 13:14:46.775 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 1392 queries
[2022-01-16 13:15:46.011 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 1824 queries
[2022-01-16 13:16:46.251 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 7176 queries
[2022-01-16 13:17:46.491 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 6955 queries
[2022-01-16 13:18:46.731 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 5933 queries
```

DL6ER:
If you surpass the limit at any point, the client gets blocked. This means they were able to do 1000 queries within a minute and then only blocked *thereafter*. A client is unblocked as soon as it made less than 1000 queries per minute. 

> [2022-01-16 13:09:46.579 1314/T1319] Still rate-limiting 192.168.178.10 as it made additional 1007 queries

Here your client was only slightly above the limit and almost got de-limited. The client could immediately have done 1000 queries thereafter.

> Maybe instead of blocking all requests all together, can't we just reject all requests above the limit?

Because this doesn't look like a proper solution. Imagine the typical usecase of the rate-limitation that could be a DNS loop between, e.g., the router and the Pi-hole or a client going rouge. Allowing 1000 queries per minute leads to significant load on the Pi-hole - consider also the ever growing queries database eventually eating up all space on disk.
Take now the example from your log above (11928 within 60 seconds, this is not the worst number). Let us simplify a bit by assuming the queries were uniformly distributed over time, i.e., we can say that roughly 200 queries were done per second. In your suggestion we would allow these queries (5 seconds until we hit 10,000) and then block the rest of the 55 seconds.
Your Internet would be working 8% of the time and *some* content can be downloaded. I guess that's a situation that is *much* harder to understand and debug for the user than a straight loss of Internet connection altogether.

yubiuser:
I adjusted my above answer slightly as it was a bit incorrect regarding the interval. The interval steps correspond to the set rate-limiting interval. They also do not correspond to the straight minute but relative to when FTL has finished starting (so start of the daemon + possible delay by DELAY_STARTUP).

> Edit: Maybe instead of blocking all requests all together, can't we just reject all requests above the limit?

This was the behavior before https://github.com/pi-hole/FTL/pull/1199
We intentionally changed this, because it seems a but inconsistent to block a client going wild (constantly generating queries) after 1000 queries, but then allow 1000 queries, block again, unblock, block....

vseven:
> What would you suggest? The current limit equates to rate-limiting when a single client issues more than 1.4 million queries a day. I don't think it should be increased.

I route all our clients to our in house DNS then have the pi-hole setup as the first forwarder address.  So in my case the in house DNS server, serving 100+ clients, can easily hit 1000 requests in 60 seconds and does so often.  I increased my rate limit in the config file to 2000/60 and it rarely hits that but it would be nice if this was a GUI setting as others have mentioned.

MarkusNiGit:
Feedback from user land:
Setup 
2 piholes in HA
Synology 820+
I enabled Cloud Sync on the Synology, a software package delivered by Synology and part of every box they ship. It took me 4 days to figure out the source of slow down/stop of the Cloud Sync. Turns out there was a little exclamation point on my pihole dashboard blinking. It took me a day to read up on what the message "rate limit" meant. The dashboard, btw, gives no information about the fact that this is not a one-time occurance but that there are other messages that, in my case, went on for days that the limit is still active because the Synology tried again within 60 sec. I had to go digging through logs via SSH into the pihole.
I was finally able to bum up the limit to 4000/60 after I read all the documentation I could get my hands on. From the "still blocking the Synology message" I was able to deduce what "my" appropriate limit is.
Synology now races like a charm and is able to keep several thousands of files in sync between my cloud and on-prem.
It took me another day to find this github where this variable/functionality originated.
A couple of thoughts:
- this situation was caused by a professional piece of hardware, with a professional piece of software on it (Synology after all), not some fly by night, rogue client. 
- Even for an somewhat advanced user this was way too much work to identify the source of the problem. And the source of the problem is NOT my Synology box but pihole. In this case simply doubling the artificial limit to 4000/60 solved the issue and was easily handled by a Raspberry Pi 4 w/2GB
- It took way to long to find steps for problem resolution: the warning message in Pi-Hole diagnosis did not let me know: 1) it is an ongoing problem and all outbound DNS request coming from the Synology were being blocked (I finally caught it when my son told me he cannot get on the Minecraft server anymore) 2) the original message did not instruct me how to fix it in case I want to

My take-away here is that this is a suboptimal failsafe for a problem that, granted, I barely understand but keeps off -the rack software from well-known vendors from working as designed AND discoverability of the source of the issue is super low AND ease of removing the blocker is super complicated.
The team here doesn't own me anything and I owe the usage of pihole to this team of creators. My recommendation as a user would be: please address the three main issues outlined.

MarkusNiGit:
I should also add:
- This was the inital sync so transfer rate was higher than "normal changes" in cloud synch over time
- I did not see that the rate limit was a throttling the client but flat out stopped it completely from any further DNS queries. Perhaps my observation here was wrong.
- Based on what I am seeing, the rate limit was in place with cached DNS query results not new ones. See the picture after I increased the limit:
image

DL6ER:
@MarkusNiGit Sorry for the inconvenience this feature caused.

We've been discussing ways to improve the user experience but this is difficult. For instance, the jumping triangle used to be larger and jump more heavily in a previous version but users complained its too catchy so it was reduced in size to be less obtrusive. This may or not be the reason why you didn't see it before.

My initial idea to add a dialogue to the installer that explicitly asks you to configure the rate limit was discussed by then dismissed because it had a couple of drawbacks. The main being that there are no dialogues on upgrades, only on fresh installs, as the former is a semi-automated process that shouldn't require user interaction.

So I meant to add a dedicated page to the dashboard's settings page but, you know, family duties took over and, eventually, I forgot about it, so I'll use this as a reminder to hopefully get this done, soonish.

> this situation was caused by a professional piece of hardware, with a professional piece of software on it (Synology after all), not some fly by night, rogue client

I know you said that a few times but I disagree on this bit. Let me say why: Synology might be a bigger company (about 600 employees according to Wikipedia), but this doesn't mean they are a global player like Google and similar. While the latter surely have a quality assurance center, I don't think this is true for Synology. At least not for their software or, at least, not for this part of it. Why do I say this? Because opening a dedicated connection *per file* (which is what I deduce from your explanation) is not good software design by any means. Imagine you are synchronizing a large number of really small files. The three-way handshake of TCP may easily be more traffic than the synchronization itself. Even more when TLS (HTTPS) is also in use. On top comes the thousands of DNS queries that, in a typical setup without a Pi-hole (absence of a caching DNS resolver) bounce back and forth between you and, say, Google DNS and cause additional delays and traffic all over the place. This all does not tell me that this is a good software.

*snip*

Concerning your three points: I disagree on the quality of the software and explained why (point 1), point 2 is unfortunate but it seems we cannot have a solution that suits all users, point 3: I'll work on this as said above.

MarkusNiGit:
I used the word "professional" and not "quality" to describe the Synology solution deliberately. I cannot judge the quality of the software though even I can tell that multiple DNS requests per file to be transferred might appear inefficient but there might be very good reasons such as restrictions on the API put in place by the cloud provider. My point is that you cannot judge either whether this is "good software" and this vendor created facts by selling the solution as it is currently designed.

Based on your answer you seem to be equally concerned about "rogue clients" and enforcing quality on the software of vendors that sell professional solutions.
While I find it admirable that you are introducing a protection of the ecosystem, I wonder how many hits you are getting for each: a) number of rogue clients stopped vs. b) number of professional pieces of software stopped functioning. And what is the the right ratio? Is one rogue client acceptable? Is one professional piece of software and hardware stopped from functioning completely acceptable?
BTW, this sounds way more combative than I intend it to be. I think we agreed, at a minimum, on increasing discoverability and ease of changing for the end-user.

WoLpH:
This "feature" cost me quite a bit of time to track down... for the future it would be nice if features like these would be asked about during an upgrade instead of being silently enabled.

Either that or only enable features like these on new installations.

I understand that 1000 queries per minute seems like a reasonable limit, but pihole has enough power users and/or quirky setups that can run into issues.

jfb-pihole:
"it would be nice if features like these would be asked about during an upgrade instead of being silently enabled."

This is why we write and publish detailed release notes, and in our release post we remind readers to read the notes prior to upgrading. 

https://pi-hole.net/blog/2021/02/16/pi-hole-ftl-v5-7-and-web-v5-4-released/

At user request, activation of rate limits was added to the diagnostic messages, and this change was also covered in release notes:

https://pi-hole.net/blog/2021/09/11/pi-hole-ftl-v5-9-web-v5-6-and-core-v5-4-released/

A later release did some tweaking to rate limits:

https://pi-hole.net/blog/2021/10/23/pi-hole-ftl-v5-11-web-v5-8-and-core-v5-6-released/

WoLpH:
> "it would be nice if features like these would be asked about during an upgrade instead of being silently enabled."
> 
> This is why we write and publish detailed release notes, and in our release post we remind readers to read the notes prior to upgrading.

I know, and I did read them. I just never assumed I would hit the limit because at first glance it seemed reasonable.

Only after a bunch of random issues over te last few days did I notice that pi-hole was the culprit. 

That's the issue with these types of changes. Everything might seem fine and dandy initially but break things down the line at some seemingly random moment when some service or cron job suddenly does a burst of requests.

FreedomFaighter:
[bf52156](https://github.com/pi-hole/FTL/pull/1052/commits/bf52156ae0ad2755fdced005a86ea2e82f24d236) seems to have altered the default rate limit of queries for pihole and if this should be opened in a new issue I can do that but in what basis is that number generated? I can't see how this limit would even support a normal network of a household of four. This is probably in the same concern others have expressed in this pull request.

yubiuser:
@FreedomFaighter 

The commit you linked did not change any rate limit but introduced it in the first place. Before there was no rate-limit. This is also not a new commit, but over 1 year old. No need to open an new issue about it.

lexcyn:
Just chiming in I think this should be an option in the web GUI or opt-in. I was pulling my hair out why my network was grinding to a halt, but if I rebooted things it would work for a while then stop. Well, turns out the corporate VPN I am using (I work from home, no way around it) likes to randomly spam thousands of DNS requests for literally every network endpoint at random times, which rate limits me and blocks all further connections from my main router. I ended up disabling rate limiting on both my pihole's to resolve it, but it would be nice if there was just a toggle to turn it on/off, so I could turn it back on after my VPN is finished having its fit.

jfb-pihole:
"turns out the corporate VPN I am using (I work from home, no way around it) likes to randomly spam thousands of DNS requests for literally every network endpoint at random times"

It seems unusual that a corporate VPN would send DNS queries to a local DNS server, and not send the DNS through the VPN.  Consider discussing this with corporate IT and moving the DNS traffic to the tunnel.

lexcyn:
> "turns out the corporate VPN I am using (I work from home, no way around it) likes to randomly spam thousands of DNS requests for literally every network endpoint at random times"
> 
> It seems unusual that a corporate VPN would send DNS queries to a local DNS server, and not send the DNS through the VPN. Consider discussing this with corporate IT and moving the DNS traffic to the tunnel.

I agree - this has happened before, but our network team could not figure out what was going on. We have split tunneling enabled for our VPN client, so internet traffic goes through our local connection but internal goes through the VPN, and in this case, the DNS queries start 'bleeding' to local, causing the massive spike in queries. This morning there were over 260k requests after I removed rate limiting, so now I will re-enable and hope it doesn't happen again - but still, would be nice if there was a GUI option! If you're curious, we are using Global Protect by Palo Alto.

D33M0N:
Apparently `steam-for-linux` manages to hit the default limit 1000/60 easily. Sadly the "per client" blocking kills the pi-hole server entirely for everyone, when you run it in the same workstation as your steam client.
Could you exclude the server itself (127.0.0.1) from the per-client blockage??

yubiuser:
Not sure why we abandoned https://github.com/pi-hole/FTL/pull/1468 but maybe we could revive it at some point when all the other PRs are merged.

kodeMunky:
Bumping - I recently ran into the the rate limit doing nothing special , and nothing misbehaving particularly - I mentioned this  to a friend who also uses a pi-hole and on closer inspection he realized he was also hitting the limit. I suggest bumping up the default by 50% until [#1468](https://github.com/pi-hole/FTL/pull/1468) can be implemented. I suspect this affecting may users without them realizing.

PR #1058, opened 2021-02-15
Retain pseudoheader in blocked queries

DL6ER:

## 10

---

When requesting blocked queries through `sytemd-resolved`, there may be an additional delay of 5 seconds because `systemd-resolved` jejects the first 23 replies before finally accepting the 24th (despite all of them being identical). This can be resolved from our side by retaining EDNS0 bits from incoming queries in blocked replies. This is what this PR implements.

This is a bug as [RFC 6891, Scn. 6.1.1](https://tools.ietf.org/html/rfc6891#section-6.1.1) specifies that:
> An OPT pseudo-RR (sometimes called a meta-RR) MAY be added to the additional data section of a request. [...]
> **If an OPT record is present in a received request, compliant responders MUST include an OPT record in their respective responses.**


PromoFaux (approved):
Conditional approve from me: Appears to not break anything else on my system, but would like to see that it fixes the reported issue for the OP before merging :)

DL6ER:
Conditions for merging met.

PR #1065, opened 2021-02-17
Add E2E tests for pihole-FTL commands

Beanow:
This demonstrates issues where daemons are running unexpectedly.
Can be used to validate #1064 is fixed.

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -99,6 +99,10 @@ fi
 
 # Kill pihole-FTL after having completed tests
 kill $(pidof pihole-FTL)
+sleep 2
+
+# Now the daemon has stopped, run e2e tests.
+test/libs/bats/bin/bats "test/e2e_test_suite.bats"
 
 # Restore umask
 umask $OLDUMASK
--- a/test/e2e_test_suite.bats
+++ b/test/e2e_test_suite.bats
@@ -1,0 +1,69 @@
+#!./test/libs/bats/bin/bats
+
+monitor_process() {
+  TERM_TIMEOUT="3s"
+  KILL_TIMEOUT="5s"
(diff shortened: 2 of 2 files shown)


test/run.sh
```diff
@@ -99,6 +99,10 @@ fi
 
 # Kill pihole-FTL after having completed tests
 kill $(pidof pihole-FTL)
+sleep 2
+
+# Now the daemon has stopped, run e2e tests.
+test/libs/bats/bin/bats "test/e2e_test_suite.bats"
```
Beanow, line 105: Looks like we need to explicitly track the return code, as the build passes while these tests fail.

```
1..5
ok 1 Running without arguments doesn't block and starts a daemon
ok 2 Running with 'no-daemon' blocks untill killed
not ok 3 Running with 'test' doesn't block and exits immediately
# (in test file test/e2e_test_suite.bats, line 47)
#   `[[ ${lines[1]} == "PID: -" ]]' failed
# Jobs:
# PID: 2508
# Jobs:
# PID: -
not ok 4 Running with '-f test' blocks but exits immediately
# (in test file test/e2e_test_suite.bats, line 57)
#   `[[ ${lines[1]} == "PID: -" ]]' failed
# Jobs:
# [1]+  Running                 timeout -k 5s -s TERM 3s /home/pihole/pihole-FTL -f test &> /dev/null &
# PID: 2533
# Jobs:
# PID: -
ok 5 Running with 'dnsmasq-test' doesn't block and exits immediately

CircleCI received exit code 0
```

DL6ER: Yeah, look at what we do above for the other `bats` test. (`RET` variable later used in `exit $RET`).

DL6ER:
Some further tests have been done on branch `test/e2e`, however, I don't have the time currently to continue working on this.

PR #1080, opened 2021-03-12
Update SQLite from 3.34.0 to 3.35.0

DL6ER:

## 10

---

This PR updates the embedded SQLite engine to v3.35.0 released today. [CHANGELOG](https://www.sqlite.org/releaselog/3_35_0.html)

Highlights for us:
- Math functions (disabled by default as we don't need them right now)
- Added support for [`ALTER TABLE DROP COLUMN`](https://www.sqlite.org/lang_altertable.html#altertabdropcol)
- Improvements for [`UPSET`](https://www.sqlite.org/lang_upsert.html)
  `UPSERT` is a clause added to `INSERT` that causes the `INSERT` to behave as an `UPDATE` or a no-op if the `INSERT` would violate a uniqueness constraint.
- Add support for the [`RETURNING`](https://www.sqlite.org/lang_returning.html) clause on `DELETE`, `INSERT`, and `UPDATE` statements. 
  The `RETURNING` clause is not a statement itself, but a clause that can optionally appear near the end of top-level `DELETE`, `INSERT`, and `UPDATE` statements. The effect of the `RETURNING` clause is to cause the statement to return one result row for each database row that is deleted, inserted, or updated.
- Various query planner/optimizer improvements
- CLI improvements
- Three bugfixes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-# Object files output by the assembler
-/build
-
 # Generated binary
 pihole-FTL
 
@@ -11,13 +8,16 @@ version*
 /cmake/
 /cmake-build-debug/
 /cmake-build-release/
-/cmake/
 
 # IDE files
 .idea/
 *.sw*
 .vscode/
+/build/
 
 # MAC->Vendor database files
 tools/manuf.data
 tools/macvendor.db
+
+# Test dependencies
+/node_modules/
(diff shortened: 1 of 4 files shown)


DL6ER:
The query planner optimzation with the largest impact for us will like be to use an index, if appropriate, on `IS NOT NULL` expressions in the `WHERE` clause. Also, attempting to process `EXISTS` operators in the `WHERE` clause as if they were `IN` operators, in cases where this is a valid transformation and seems likely to improve performance, is a promising idea.

PromoFaux (approved):

Issue #1081, opened 2021-03-04
pihole-FTL hangs / DNS unresponsive, daemon times out during restart


takerukoushirou:
### Versions
-  Pi-hole version is v5.2.4 (Latest: v5.2.4)
-  AdminLTE version is v5.4 (Latest: v5.4)
-  FTL version is v5.7 (Latest: v5.7)

### Platform
- OS and version:  Debian 10.8
- Platform:  Raspberry Pi 4 Model B Rev 1.1, 4 GB RAM

### Issue

DNS _randomly_ stops working. Devices can no longer make a connection to pihole-FTL; connection attempts hang.

Last pihole-FTL log messages:
```
[2021-03-04 15:15:07.893 26913M] Resizing "FTL-dns-cache" from 159744 to (10240 * 16) == 163840 (/dev/shm: 6.4MB used, 2.0GB total, FTL uses 6.4MB)
[2021-03-04 15:56:04.649 26913M] Resizing "FTL-dns-cache" from 163840 to (10496 * 16) == 167936 (/dev/shm: 6.4MB used, 2.0GB total, FTL uses 6.4MB)
[2021-03-04 16:00:01.267 26913/T26917] Notice: Database size is 4345.02 MB, deleted 6532 rows
```

`pihole restartdns` fails as pihole-FTL does not stop within the time limit. systemd journal:
```
Mar 04 16:17:04 gateway-hub systemd[1]: Stopping LSB: pihole-FTL daemon...
Mar 04 16:17:13 gateway-hub pihole-FTL[9257]: .....
Mar 04 16:17:13 gateway-hub pihole-FTL[9257]: Not stopped; may still be shutting down or shutdown may have failed, killing now
Mar 04 16:17:09 gateway-hub systemd[1]: pihole-FTL.service: Control process exited, code=exited, status=1/FAILURE
Mar 04 16:17:09 gateway-hub systemd[1]: pihole-FTL.service: Failed with result 'exit-code'.
Mar 04 16:17:09 gateway-hub systemd[1]: Stopped LSB: pihole-FTL daemon.
Mar 04 16:17:09 gateway-hub systemd[1]: Starting LSB: pihole-FTL daemon...
Mar 04 16:17:15 gateway-hub pihole-FTL[9274]: .....
Mar 04 16:17:15 gateway-hub pihole-FTL[9274]: Not stopped; may still be shutting down or shutdown may have failed, killing now
[...]

PR #1084, opened 2021-03-19
Various enhancements and a few memory-leak fixes

DL6ER:

## 10

---

This PR adds a few technical improvements and fixes a (rarely seen but reproducible) possibility of freezing under excessive TCP query load. Furthermore, we fix a few possible memory leaks. Most of them actually cannot result in memory being actually leaked (memory is freed either way when a fork terminates), however, they made reading `valgrind` logs somewhat harder.

Overview over the individual improvements and fixes contained here:
- Avoid jump depending on uninitialized bytes when only a few nameservers are stored in  `resolv.conf` (this only happened with `DEBUG_RESOLVER=true`)
- Join canceled threads on exit to ensure they exited properly before we exit from the main process. This includes waiting for them to clean up their own stack memory, etc. This makes reading `Valgrind` log outputs a lot simpler.
- Ensure we close FTL database connection when exiting the main process. Again, this has no consequences else than silencing some memory-lost complaints by `valgrind`,
- Ensure shared memory strings bucket is large enough when locking. Do not resize it when we are holding the lock. This avoids an obscure deadlock in case of many forking TCP workers. Also, optimize FTL-domains size
- Don't finalize gravity statements two times (makes the process faster)
- More fine-grained locking in network table processing should decrease delays in DNS resolution on very slow machines
- Reduce rate-limiting checking to once per second (rather than every 100 msec) to stop filling the logs too quickly when `DEBUG_LOCKS=true` is set
- Simplify locking during network table processing and generalize spacial handling for virtual interfaces (`hwaddr 00:00:00:00:00:00`)
- Simplify signal handling and catch `SIGABRT` in addition 
- Do not keep a globally opened SQLite3 connection to the long-term database but open and close connections when needed.

Documentation updated related to this change: https://github.com/pi-hole/docs/pull/485

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ version*
 # MAC->Vendor database files
 tools/manuf.data
 tools/macvendor.db
+node_modules
--- a/src/FTL.h
+++ b/src/FTL.h
@@ -144,11 +144,4 @@
 #define str(x) # x
 #define xstr(x) str(x)
 
-extern pthread_t telnet_listenthreadv4;
-extern pthread_t telnet_listenthreadv6;
-extern pthread_t socket_listenthread;
-extern pthread_t DBthread;
-extern pthread_t GCthread;
-extern pthread_t DNSclientthread;
-
 #endif // FTL_H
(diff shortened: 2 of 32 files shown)


DL6ER:
Confirmed working by two bug-affected users in https://github.com/pi-hole/FTL/issues/1081#issuecomment-803672503 and https://github.com/pi-hole/FTL/issues/1081#issuecomment-803673630

PromoFaux (approved):

dschaper (approved):

PR #1090, opened 2021-03-27
Delay earlier if requested

DL6ER:

## 10

---

If `DELAY_STARTUP` is set, we can delay earlier to have this option being useful for misbehaving fake hwclocks as well.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -71,6 +71,12 @@ int main (int argc, char* argv[])
 	if(strcmp(username, "pihole") != 0)
 		logg("WARNING: Starting pihole-FTL as user %s is not recommended", username);
 
+	// Delay startup (if requested)
+	// Do this before reading the database to make this option not only
+	// useful for interfaces that aren't ready but also for fake-hwclocks
+	// which aren't ready at this point
+	delay_startup();
+
 	// Initialize query database (pihole-FTL.db)
 	db_init();
 
@@ -85,8 +91,7 @@ int main (int argc, char* argv[])
 	// immediately before starting the resolver.
 	check_capabilities();
 
-	// Start the resolver, delay startup if requested
-	delay_startup();
+	// Start the resolver
 	startup = false;
 	if(config.debug != 0)
 	{


dschaper:
The related Discourse issue was closed because it was cause by a user error.

DL6ER:
It still makes sense to add the delay at the very start of FTL instead of somewhere in between (after reading the database). I see this change useful even without a related Discourse discussion.

dschaper:
The entire reason for the delay is for networking. I think it would be quite efficient to load the database during any delay or pause.

I don't know the feasibility of that, nor unintended consequences however.

DL6ER:
But then the delay is artificial itself. If a user specifies FTL should wait for 20 seconds, they just pick 20 because it is a nice number. I doubt they would mind if the delay was actually 20.5 seconds.

All this PR does is shifting the delay earlier so all routines replying on a properly set clock are covered as well. This seems useful to have.

src/main.c
```diff
@@ -71,6 +71,12 @@ int main (int argc, char* argv[])
 	if(strcmp(username, "pihole") != 0)
 		logg("WARNING: Starting pihole-FTL as user %s is not recommended", username);
 
+	// Delay startup (if requested)
+	// Do this before reading the database to make this option not only
+	// useful for interfaces that aren't ready but also for fake-hwclocks
+	// which aren't ready at this point
+	delay_startup();
```
PromoFaux, line 78: Is there any value in putting this before `read_FTLconf` or even `init_shmem`, or even the first line of `main()`?

DL6ER: We need it after `read_FTLconf` because this is where we read the value how long to wait ;-)

DL6ER:
Rebased on lastest `development`

dschaper:
Thinking this over more, I'm going to change my mind. I still think that it would be more efficient to load the database during the delay but having the database be time dependent and the fact that the RPi boards lack a real time clock makes a delay like this logical.

And really, how much time does it take to load the database?

dschaper (approved):

PR #1091, opened 2021-03-27
Ensure FTL can be compiled from static tarballs

### LINKED ISSUE #928: Failed to build 5.2 release on ubuntu 20.04
I downloaded 5.2 from github and untar-ed it to a directory, run build.sh and I got this

```
inlined from ‘getVersion’ at /home/test/FTL-5.2/src/api/api.c:1062:2:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:34:10: error: ‘__builtin_memcpy’ forming offset [2, 7] is out of the bounds [0, 1] [-Werror=array-bounds]
   34 |   return __builtin___memcpy_chk (__dest, __src, __len, __bos0 (__dest));
```

I then git clone the git repository and it built fine.

The problem is with src/version.h.in, for released files those variables are empty, and gcc/memcpy does not like that.

maybe in the release version.in.h file it should be pre-filled release info manually? after I manually put stuff in I can build fine.

DL6ER:

## 10

---
Ensure FTL can be compiled from source archives offered by GitHub for each release. Fixes #928

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -1174,8 +1174,8 @@ void getVersion(const int *sock)
 	const char *version = get_FTL_version();
 
 	// Extract first 7 characters of the hash
-	char hash[8];
-	memcpy(hash, commit, 7); hash[7] = 0;
+	char hash[8] = { 0 };
+	memcpy(hash, commit, min((size_t)7, strlen(commit)));
 
 	if(strlen(tag) > 1) {
 		if(istelnet[*sock])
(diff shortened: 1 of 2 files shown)


dschaper:
How does this fix the issue? Allow for an FTL build that has no version string?

DL6ER:
I just changed this so tarball builds will log
```
$ pihole-FTL -v
UNKNOWN (not a GIT build)
```

I'm really not a fan of manually updating a version string in the source. This will be forgotten (`dnsmasq` does this and it has happened in the past) and this has only once been reported as an issue so it seems to be really an edge-case.

dschaper:
That's perfect. My concern is that there would be a blank version. This clearly indicates it's not a CI build.

dschaper (approved):

PR #1113, opened 2021-04-19
Circle CI: SHA-256 hashes for integrity checks

bershanskiy:
Signed-off-by: Anton Bershanskiy <45960703+bershanskiy@users.noreply.github.com>

---

Publish SHA-256 hashes of built binaries along with SHA-1 hashes. This would allow users to eventually migrate away from SHA-1 to SHA-256. If you would be interested, I can file a matching PR for the official install script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ version: 2
         name: "Compute checksum"
         command: |
           mv pihole-FTL "${BIN_NAME}"
-          sha1sum pihole-FTL-* > ${BIN_NAME}.sha1
+          sha256sum pihole-FTL-* > ${BIN_NAME}.sha256
     - run:
         name: "Upload binary to binary bucket"
         command: |
@@ -36,8 +36,8 @@ version: 2
           mkdir download
           cd download
           wget "https://ftl.pi-hole.net/${DIR}/${BIN_NAME}"
-          wget "https://ftl.pi-hole.net/${DIR}/${BIN_NAME}.sha1"
-          sha1sum -c "${BIN_NAME}.sha1"
+          wget "https://ftl.pi-hole.net/${DIR}/${BIN_NAME}.sha256"
+          sha256sum -c "${BIN_NAME}.sha256"
           cd ..
     - run:
         name: "Test"


DL6ER:
Yes, the encryption hash used in SHA2 is significantly stronger and not subject to the same vulnerabilities as SHA1.

There is no harm in using it for our purposes (file integrity checks). I feel like I should stress that the vulnerability known in SHA-1 (for years) is entirely irrelevant here. We use the hash solely to verify the file on-disk equals the binary on-web without having to download it. The hash is not meant as any kind of preventing MITM attacks or whatsoever. This is done by HTTPS which is the exclusive protocol used for downloading and is much better at this job.

I highly doubt a hard-drive defect can cause a malicious hash which equals the on-web one. Having it is fine, replacing it would be even better. Do we know `sha256sum` is available on all systems out there where users install Pi-hole on?
(Yes I know this is a separate thing from this PR and the discussion should be continued on the PR for the installer script).

TL;DR: We should not have both. I don't see striving arguments from switching, however, so far, I also haven't seen ones against. But I don't think we should have two competing hashes for the very same thing.

bershanskiy:
I probably should have clarified in the original comment that use of SHA-256 over SHA-1 in this case does not improve security since the integrity of hash file itself is not verified.

> Do we know sha256sum is available on all systems out there where users install Pi-hole on?

I will check. As far as I know, on modern environments `sha256sum` is always available if `sha1sum` is available.

> We should not have both.

Why not? Would it be OK to only publish both hashes but use `sha1sum` in the install script?

DL6ER:
I just stressed this because migrating from SHA-1 to SHA-2 is typically a security argument - which is very valid for security-related certificate hashes, etc. For mere data integrity checks, using even MD5 is still okay and maybe even advantageous as the algorithm is a lot faster. I've seen reports that MD5 is (often) double as fast as SHA-1. It will be even worse for SHA-2. We still selected SHA-1 to avoid possible discussions in this direction.

> > We should not have both.
>
> Why not?

Because it increases the number of uploaded files from 18 to 27 without any real improvement. It'd create unnecessary redundancy as we have two files for the exact same purpose.

> Would it be OK to only publish both hashes but use sha1sum in the install script?

We should first check if we can switch to using `sha256sum`. I'm not in favor of having both (file number and redundancy argument).

bershanskiy:
> I just stressed this because migrating from SHA-1 to SHA-2 is typically a security argument - which is very valid for security-related certificate hashes, etc.

My main motivation is to clamp down on obsolete hashes (MD5, SHA-1) system-wide. It's just a peace of mind consideration.

> We should first check if we can switch to using `sha256sum`.

As a rule of thumb, since both `sha1sum` and `sha256sum` come from GNU Coreutils, they are either both available or neither one is. According to [Coreutils release notes](https://github.com/coreutils/coreutils/blob/master/NEWS), `sha256sum` first appeared in release 6.0 in 2006. I went ahead and checked that all configurations listed in [official list of supported platforms](https://docs.pi-hole.net/main/prerequisites/#supported-operating-systems) actually ship with more recent Coreutils. If needed, I can post sources for all these here.

Funny enough, CentOS and Raspbian explicitly ask to check SHA-256 of downloaded images before upgrading. Debian provides MD-5, SHA-1, SHA-256 and SHA-512.

Also I realize that this is just a first chunk of work necessary to switch to SHA-256, I can prepare corresponding PR for install script if you are interested in this switch.

DL6ER:
Yes, please provide a core PR and link it here. If you don't find time to do this, I'll do it beginning of next week.

bershanskiy:
> Yes, please provide a core PR and link it here. If you don't find time to do this, I'll do it beginning of next week.

By "core" you mean [pi-hole/pi-hole](https://github.com/pi-hole/pi-hole), right? I'll draft and test a PR over the weekend.

How do you want me to handle backwards compatibility? Should install script try to download `*.sha256` and if that's missing falls back to `*.sha1` to support older releases? Or would it make sense to just upload SHA-256 hashes for past releases?

bershanskiy:
Also, I pushed a commit removing SHA-1 hash upload on Circle CI. If that is desired, I'll squash two commit stogether (and ensure there is a proper summary and sign-off comment).

DL6ER:
Replacing send to be the only same thing to keep file count reasonable. Also pinging @dschaper for his opinion.

dschaper:
Honestly? I don't see a need to change anything. SHA1 is fine for our use.

bershanskiy:
OK, thank you for your time.

PR #1108, opened 2021-04-15
Fix database update to version 7 reporting error

DL6ER:
**By submitting this pull request, I confirm the following:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 

## 10

---

Fix database update to version 7 reporting error when there is none:
``` plain
Column additional_info not initialized, database not available
```

This is not a critical bug as the issue resolves itself on the next start of FTL.

PR #1116, opened 2021-04-19
Pi-hole FTL v5.8.1

DL6ER:
Pi-hole FTL hotfix release

- Don't try to terminate threads when we never launched them #1115 
- Give threads a bit more time to shut down properly #1112
- Fix database update to version 7 reporting error #1108


dschaper:
Do we need a `release/v5.8.1` branch?

DL6ER:
Release branches before the tag on `master` never pick up the latter. Hence, an FTL version built on the `release/v5.8.1` branch still shows `vDev-12-v5.8` (intermediate version). The solution can be either to merge the tagged `master` commit into the release branch after releasing it - or - to create the branch only after we merged into `master`. Or - leave the branch entirely (because users will always be able to check out a specific version using a tag).

dschaper (approved):

PromoFaux (approved):

PR #1133, opened 2021-05-11
Fix possible crash during shutdown

DL6ER:

## 10

---

Cancel still running API threads at shutdown. This avoids a possible crash when a long-running API task is still running when FTL closes database connections on shut down.

### DIFF
--- a/src/api/socket.c
+++ b/src/api/socket.c
@@ -392,6 +392,10 @@ static void *socket_connection_handler_thread(void *socket_desc)
 	sprintf(threadname, "socket-%i", sock);
 	prctl(PR_SET_NAME, threadname, 0, 0, 0);
 
+	// Ensure this thread can be canceled at any time (not only at
+	// cancellation points)
+	pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, NULL);
+
 	// Store TID of this thread
 	lock_shm();
 	unsigned int tid;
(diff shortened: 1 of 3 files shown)


src/api/socket.c
```diff
@@ -367,6 +392,29 @@ static void *socket_connection_handler_thread(void *socket_desc)
...
+		{
+			api_threads[tid] = pthread_self();
+			api_tids[tid] = gettid();
+			break;
+		}
+	}
+	unlock_shm();
+	if(tid == MAX_API_THREADS)
+	{
+		logg("Not able to spawn new API thread, limit reached.");
```
dschaper, line 414: Can this include the value of `MAX_API_THREADS` instead of just reporting that the limit was reached? 

Something like "`limit of n threads reached.`"

dschaper (changes requested):

src/api/socket.c
```diff
@@ -312,10 +315,29 @@ static void *telnet_connection_handler_thread(void *socket_desc)
...
+	for(tid = 0; tid < MAX_API_THREADS; tid++)
+	{
+		if(api_threads[tid] == 0)
+		{
+			api_threads[tid] = pthread_self();
+			break;
+		}
+	}
+	unlock_shm();
+	if(tid == MAX_API_THREADS)
```
PromoFaux, line 334: Is this code supposed to be a duplicate (ish) of what Dan commented on below? (lines 412-416)

DL6ER: Thanks, updated.

DL6ER:
Rebased on lastest `development`

dschaper (approved):

PR #1096, opened 2021-04-02
Cisco Umbrella Remote IP/Device ID support

tresni:
**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 5

---

This pull request add support for Cisco Umbrella/OpenDNS Device ID and remote IP reporting.  This is based on the information at https://docs.umbrella.com/umbrella-api/docs/identifying-dns-traffic and https://docs.umbrella.com/umbrella-api/docs/identifying-dns-traffic2 . Using `--umbrella` by itself will enable Remote IP reporting. This can not be used for any policy filtering in Cisco Umbrella/OpenDNS. Additional information can be supplied using specific option specifications, multiple can be separated by a comma:

`--umbrella=orgid:1234,deviceid=0123456789abcdef`

Specifies that you want to report organization 1234 using device 0123456789abcdef.  For Cisco Umbrella Enterprise, see [Register (Create) a device](https://docs.umbrella.com/umbrella-api/docs/create-a-device) how to get a Device ID  and [Organization ID endpoint](https://docs.umbrella.com/umbrella-api/docs/organization-endpoint) to get organizations ID. For OpenDNS Home Users, there is no organization, see [Registration API endpoint](https://docs.umbrella.com/umbrella-api/docs/registration-api-endpoint2) for how to get a Device ID.  Asset ID should be ignored unless specifically instructed to use by support.

PR #1143, opened 2021-06-22
Update dnsmasq to v2.86

DL6ER:

## 10

---

New FTL features in this version:

- Update to `dnsmasq` v2.86 (see `dnsmasq` changelog below for the highlights)
- Support Cisco Umbrella/OpenDNS Device ID & Remote IP  (see https://github.com/pi-hole/FTL/pull/1096)
- Automatically generated queries (typically DNSSEC `DNSKEY` and `DS`) are now included in the Query Log and dashboard.
  They can be suppressed by setting `SHOW_DNSSEC=false` in `/etc/pihole/pihole-FTL.conf`

CHANGELOG:

> **Handle `DHCPREBIND` requests in the DHCPv6 server code.**
> Thanks to Aichun Li for spotting this ommision, and the initial patch.
> 
> **Fix bug which caused `dnsmasq` to lose track of processes forked to handle TCP DNS connections under heavy load.** The code checked that at least one free process table slot was available before listening on TCP sockets, but didn't take into account that more than one TCP connection could arrive, so that check was not sufficient to ensure that there would be slots for all new processes. It compounded this error by silently failing to store the process when it did run out of slots. Even when this bug is triggered, all the right things happen, and answers are still returned. Only under very exceptional circumstances, does the bug manifest itself: see https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2021q2/014976.html
> Thanks to Tijs Van Buggenhout for finding the conditions under which the bug manifests itself, and then working out exactly what was going on.
> 
> **Major rewrite of the DNS server and domain handling code.** This should be largely transparent, but it drastically improves performance and reduces memory foot-print when configuring large numbers domains of the form `local=/adserver.com/` or `local=/adserver.com/#` Lookup times now grow as log-to-base-2 of the number of domains, rather than greater than linearly, as before. The change makes multiple addresses associated with a domain work `address=/example.com/1.2.3.4` `address=/example.com/5.6.7.8` It also handles multiple upstream servers for a domain better; using the same try/retry alogrithms as non domain-specific servers. This also applies to DNSSEC-generated queries. Finally, some of the oldest and gnarliest code in dnsmasq has had a significant clean-up. It's far from perfect, but it _is_ better.
> 
> **Revise resource handling for number of concurrent DNS queries.** This used to have a global limit, but that has a problem when using different servers for different upstream domains. Queries which are routed by domain to an upstream server which is not responding will build up and trigger the limit, which breaks DNS service for all other domains which could be handled by other servers. The change is to make the limit per server-group, where a server group is the set of servers configured for a particular domain. In the common case, where only default servers are declared, there is no effective change.
> 
> **Improve efficiency of DNSSEC.** The sharing point for DNSSEC RR data used to be when it entered the cache, having been validated. After that queries requiring the KEY or DS records would share the cached values. There is a common case in dual-stack hosts that queries for A and AAAA records for the same domain are made simultaneously. If required keys were not in the cache, this would result in two requests being sent upstream for the same key data (and all the subsequent chain-of-trust queries.) Now we combine these requests and elide the duplicates, resulting in fewer queries upstream and better performance. To keep a better handle on what's going on, the "extra" logging mode has been modified to associate queries and answers  for DNSSEC queries in the same way as ordinary queries. The requesting address and port have been removed from DNSSEC logging lines, since this is no longer strictly defined.
> 
> **Connection track mark based DNS query filtering.** Thanks to Etan Kissling for implementing this It extends query filtering support beyond what is currently possible with the `--ipset` configuration option, by adding support for: 1) Specifying allowlists on a per-client basis, based on their    associated Linux connection track mark. 2) Dynamic configuration of allowlists via Ubus. 3) Reporting when a DNS query resolves or is rejected via Ubus. 4) DNS name patterns containing wildcards. Disallowed queries are not forwarded; they are rejected with a REFUSED error code.


DL6ER:
This PR is superseded by the running beta branch `release/v5.9` which contains all commits from here.

PR #1144, opened 2021-06-23
Update SQLite engine to 3.36.0

DL6ER:

## 10

---

SQLite Release 3.36.0 On 2021-06-18

1.  Improvement to the [EXPLAIN QUERY PLAN](https://sqlite.org/eqp.html) output to make it easier to understand.
2.  Byte-order marks at the start of a token are skipped as if they were whitespace.
3.  An error is raised on any attempt to access the [rowid](https://sqlite.org/lang_createtable.html#rowid) of a VIEW or subquery. Formerly, the rowid of a VIEW would be indeterminate and often would be NULL. The -DSQLITE_ALLOW_ROWID_IN_VIEW compile-time option is available to restore the legacy behavior for applications that need it.
4.  The [sqlite3_deserialize()](https://sqlite.org/c3ref/deserialize.html) and [sqlite3_serialize()](https://sqlite.org/c3ref/serialize.html) interfaces are now enabled by default. The -DSQLITE_ENABLE_DESERIALIZE compile-time option is no longer required. Instead, there is is a new [-DSQLITE_OMIT_DESERIALIZE](https://sqlite.org/compile.html#omit_deserialize) compile-time option to omit those interfaces.
5.  The "memdb" VFS now allows the same in-memory database to be shared among multiple database connections in the same process as long as the database name begins with "/".
6.  Back out the EXISTS-to-IN optimization (item 8b in the [SQLite 3.35.0 change log](https://sqlite.org/releaselog/3_35_0.html)) as it was found to slow down queries more often than speed them up.
7.  Improve the [constant-propagation optimization](https://sqlite.org/optoverview.html#constprop) so that it works on non-join queries.
8.  The [REGEXP extension](https://sqlite.org/src/file/ext/misc/regexp.c) is now included in [CLI](https://sqlite.org/cli.html) builds.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,7 +27,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
 # SQLITE_DEFAULT_FOREIGN_KEYS=1: This macro determines whether enforcement of foreign key constraints is enabled or disabled by default for new database connections.
 # SQLITE_DQS=0: This setting disables the double-quoted string literal misfeature.
 # SQLITE_ENABLE_DBPAGE_VTAB: Enables the SQLITE_DBPAGE virtual table. Warning: writing to the SQLITE_DBPAGE virtual table can very easily cause unrecoverably database corruption.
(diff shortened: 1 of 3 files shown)


PromoFaux (approved):
Oh, and then I see another SQLite engine upgrade.

DL6ER:
Whoops, you are right, I forgot about the other one.

PR #1145, opened 2021-06-25
feat(examples): Create example pihole-FTL.conf

thompsonbrian:

**How familiar are you with the codebase?:** _2_

---

_Add an example of the pihole-FTL.conf file.  Having example
configuration files is nice for the end users to have so that they don't
need to constantly consult the web documentation or man pages._

### DIFF
--- a/examples/pihole-FTL.conf
+++ b/examples/pihole-FTL.conf
@@ -1,0 +1,261 @@
+# This is an example configuration file for pihole-FTL.conf that was
+# derived from https://docs.pi-hole.net/ftldns/configfile/.
+# 
+# Author: Brian Thompson <git@hashvault.io>
+# Created: June 25, 2021
+#
+#####
+### DNS settings
+#####
+##
+## How should FTL reply to blocked queries?
+#BLOCKINGMODE=NULL|IP-NODATA-AAAA|IP|NXDOMAIN
+#
+## Use this option to disable deep CNAME inspection.  This might be
+## beneficial for very low-end devices.
+#CNAME_DEEP_INSPECT=true|false
+#
+## Encrypted Server Name Indication (ESNI) is certainly a good step into
+## the right direction to enhance privacy on the web.  It prevents
+## on-path observers, including ISPs, coffee shop owners and firewalls,
+## from intercepting the TLS Server Name Indication (SNI) by encrypting
+## it.  This prevents the SNI from being used to determine which
+## websites users are visiting.
+#BLOCK_ESNI=true|false
+#
+## Should we overwrite the query source when client information is provided
+## through EDSN0 client subnet (ESCS) information?  This allows Pi-hole to
+## obtain client IPs even if they are hidden behind the NAT of a router.
+#EDNS0_ECS=true|false
+#
+## Control FTL's query rate-limiting.  Rate-limited queries are
+## answered with a REFUSED reply and not further processed by FTL.
+#RATE_LIMIT=1000/60
+##
+#####
+### Statistics settings
+#####
+##
+## Up to how many hours of queries should be imported from the database
+## and logs?  Values greater than the hard-coded maximum of 24h need a
+## locally compiled FTL with a changed compile-time value.
+#MAXLOGAGE=24.0
+#
+## Which privacy level is used? (0, 1, 2, or 3)
+#PRIVACYLEVEL=0|1|2|3
+#
+## Should FTL ignore queries coming from the local machine?
+#IGNORE_LOCALHOST=no|yes
+#
+## Should FTL analyze AAAA queries?  The DNS server will handle AAAA
+## queries the same way, regardless of this setting.  All this does
+## is ignoring AAAA queries when computing the statistics in Pi-hole.
+## This setting is considered obsolete and may be removed in a future
+## version.
+#AAAA_QUERY_ANALYSIS=no/yes
+#
+## Should FTL only analyze A and AAAA queries?
+#ANALYZE_ONLY_A_AND_AAAA=false|true
+##
+#####
+### Other settings
+#####
+##
+## Listen only for local socket connections or permit all connections.
+#SOCKET_LISTENING=localonly|all
+#
+## On which port should FTL be listening?
+#FTLPORT=4711
+#
+## Should FTL try to resolve IPv6 addresses to hostnames?
+#RESOLVE_IPV6=yes|no
+#
+## Should FTL try to resolve IPv4 addresses to hostnames?
+#RESOLVE_IPV4=yes|no
+#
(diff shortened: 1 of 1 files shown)


DL6ER:
Thanks for putting effort into this, but I just want to point out that we already have something similar coded for the upcoming Pi-hole v6.0 (where the config file format will also change to TOML). The file will be created by FTL itself and looks similar to what you proposed. Currently, it looks like this:
``` toml
# This file is managed by pihole-FTL
#
# Do not edit the file while FTL is
# running or your changes may be overwritten
#
# Last update: 2021-06-10 12:09:06

[dns]
  # How should FTL reply to blocked queries?
  # Possible values are: [ "NULL", "IP-NODATA-AAAA", "IP", "NXDOMAIN" ]
  blockingmode = "NULL"

  # Should FTL walk CNAME paths?
  CNAMEdeepInspect = true

  # Should _esni. subdomains be blocked by default?
  blockESNI = true

  # Should FTL analyze possible ECS information to obtain client IPs hidden behind NATs?
  EDNS0ECS = true

  # Should FTL hide queries made by localhost?
  ignoreLocalhost = false

  [dns.ipBlocking]
    # Use a specific IPv4 address in IP blocking mode
    # Possible values are: <valid IPv4 address> or empty string ("")
    IPv4 = ""

    # Use a specific IPv6 address in IP blocking mode
    # Possible values are: <valid IPv6 address> or empty string ("")
    IPv6 = ""

  [dns.rateLimit]
    # How many queries are permitted...
    count = 0 ### CHANGED, default = 1000

    # ..in the set interval before rate-limiting?
    interval = 0 ### CHANGED, default = 60

[resolver]
  # Should FTL try to resolve IPv4 addresses to hostnames?
  resolveIPv4 = true

  # Should FTL try to resolve IPv6 addresses to hostnames?
  resolveIPv6 = true

  # Try to obtain client names from the network table
  networkNames = true

  # How (and if) hourly PTR lookups should be made
  # Possible values are: [ "IPV4_ONLY", "ALL", "UNKNOWN", "NONE" ]
  refresh = "IPV4_ONLY"

[database]
  # Should FTL load information from the database on startup to be aware of the most recent history?
  DBimport = true

  # How much history should be imported from the database [seconds]? (max 24*60*60 = 86400)
  maxHistory = 86400

  # How long should queries be stored in the database [days]?
  maxDBdays = 365

  # How often do we store queries in FTL's database [seconds]?
  DBinterval = 60

  [database.network]
    # Should FTL anaylze the local ARP cache?
    parseARP = true

    # How long should IP addresses be kept in the network_addresses table [days]?
    expire = 365

[http]
  # Does local clients need to authenticate to access the API?
  localAPIauth = true

  # Should FTL insert extra spaces to prettify the API output?
  prettyJSON = false

  # How long should a session be considered valid after login [seconds]?
  sessionTimeout = 300

  # On which domain is the web interface served?
  # Possible values are: <valid domain>
  domain = "pi.hole"

  # Webserver access control list.
  # Possible values are: <valid ACL>
  acl = "+0.0.0.0/0"

  # Ports to be used by the webserver
  # Possible values are: list of <[ip_address:]port>
  port = "8080,[::]:8080"

  [http.paths]
    # Server root on the host
    # Possible values are: <valid path>
    webroot = "/var/www/html"

    # Sub-directory of the root containing the web interface
    # Possible values are: <valid subpath>, both slashes are needed!
    webhome = "/admin/"

[files]
  # The location of FTL's log file
  # Possible values are: <any writable file>
  log = "/var/log/pihole-FTL.log"

  # The location of FTL's PID file
  # Possible values are: <any writable file>
  pid = "/run/pihole-FTL.pid"

  # The location of FTL's long-term database
  # Possible values are: <any FTL database>
  database = "/etc/pihole/pihole-FTL.db"

  # The location of Pi-hole's gravity database
  # Possible values are: <any gravity database>
  gravity = "/etc/pihole/gravity.db"

  # The database containing MAC -> Vendor information for the network table
  # Possible values are: <any macvendor database>
  macvendor = "/etc/pihole/macvendor.db"

  # The config file of Pi-hole
  # Possible values are: <any setupVars file>
  setupVars = "/etc/pihole/setupVars.conf"

  # The log file used by the webserver
  # Possible values are: <any writable database>
  HTTPinfo = "/var/log/pihole/HTTP_info.log"

  # The log file used by the dynamic interpreter PH7
  # Possible values are: <any writable file>
  PH7error = "/var/log/pihole/PH7.log"

[misc]
  # Privacy level
  privacyLevel = 0

  # Set niceness of pihole-FTL (can be disabled by setting to -999)
  nice = -30 ### CHANGED, default = -10

  # Artificially delay FTL's startup (0 to 300 seconds)
  delayStartup = 0

[debug]
  # Temporarily enable all debug flags
  all = false

  # Enable extra logging of database actions
  database = false

  # Enable extra logging of detected interfaces
  networking = false

  # Enable extra logging of shared memory lock actions
  locks = false

  # Print extensive query information
  queries = true ### CHANGED, default = false

  # Print flags of queries received by the DNS hooks
  flags = true ### CHANGED, default = false

  # Print information about shared memory buffers
  shmem = true ### CHANGED, default = false

  # Print information about garbage collection
  gc = false

  # Print information about ARP table processing
  arp = false

  # Enable extra logging of regex matching details
  regex = false

  # Enable extra logging of API activities
  api = true ### CHANGED, default = false

  # Print information about overTime memory operations
  overtime = false

  # Enable extra logging of query status changes
  status = true ### CHANGED, default = false

  # Print information about capabilities granted to the pihole-FTL process
  caps = false

  # Enable extra logging in dnsmasq's log routine
  dnsmasq_lines = false

  # Print vector operation details
  vectors = false

  # Extensive information about hostname resolution like which DNS servers are used
  resolver = true ### CHANGED, default = false

  # Print EDNS(0) debugging information
  edns0 = false

  # Enable extra client detail logging
  clients = false

  # Print aliasclient details
  aliasclients = false

  # Log information about processed internal events
  events = true ### CHANGED, default = false

  # Enable logging of script helper activity
  helper = false

  # Print config parsing details
  config = true ### CHANGED, default = false
```

You will note that when a user decides to change a value, FTL will **automatically** append a comment like `### CHANGED, default = false` to point out what was changed.

Any existing `pihole-FTL.conf` will be scanned, and the settings will be used to create a new TOML-style config file (which will preserve all your settings).

Hence, I'm not sure it's actually worth the effort to create such a config file in the FTL repo. Also, and especially, because this repo (with the source code) is never cloned onto user systems so this example file would only live online in Github.

thompsonbrian:
Thank you for your consideration.  I'll close this PR since you already have this in the works; just thought I'd share the idea.  Cheers!

PR #1148, opened 2021-07-04
Implement special handling of the Mozilla canary domain to disable Firefox automatic DoH

DL6ER:

## 10

---

Ensure FTL always replies with `NXDOMAIN` to `A` and `AAAA` queries of `use-application-dns.net`
This is following the recommendation on https://support.mozilla.org/en-US/kb/configuring-networks-disable-dns-over-https

Enforcing `NXDOMAIN` for `use-application-dns.net` can be disabled by setting `MOZILLA_CANARY=false` in `/etc/pihole/pihole-FTL.conf`

The Query Log will display this as reply answered from cache:
Screenshot from 2021-07-04 13-52-05

We could also introduce a new "special domain" status, but I don't see this necessary for now (and it'd require another change to the AdminLTE repo as well).

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -537,6 +537,17 @@ void read_FTLconf(void)
 	else
 		logg("   SHOW_DNSSEC: Disabled");
 
+	// MOZILLA_CANARY
+	// Should FTL handle use-application-dns.net specifically and always return NXDOMAIN?
+	// defaults to: true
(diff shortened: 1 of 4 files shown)


PromoFaux (approved):
I think this is a much cleaner way of doing things. It also gives the user the option to opt out of the canary handling if they wish

PR #1156, opened 2021-08-23
Customizable locking while database is busy

DL6ER:

## 10

---

When the gravity database is locked/busy, Pi-hole allows all queries because it has no way of detecting if they should be blocked or not. However, for very large lists and/or on very busy networks and/or on very (!) slow devices, this may open a windows of up to several seconds where undesired queries may get through, ads and unwanted telemetry could be the result.

This PR adds a config option (`REPLY_WHEN_BUSY`) to customize what happens to queries while the database is locked:
1. `ALLOW` - as before, just allow all queries when the database is busy
2. `BLOCK` - block all queries when the database is busy. This uses the configured `BLOCKINGMODE` (default `NULL`)
3. `REFUSE` - refuse all queries which arrive while the database is busy
4. `DROP` - just drop the queries, i.e., never reply to them at all.

In my opinion, option 4 is the best one because it will typically cause clients to retry the query after some timeout of typically a few seconds. I.e., the DNS resolution will only be delayed but we do never hand out incorrectly either allowed or blocked queries.
`REFUSE` first sounded like a good option, too, however, it turned out that many clients will just immediately retry, causing up to several thousands of queries per second. This does not happen in `DROP` mode.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -570,6 +570,32 @@ void read_FTLconf(void)
 	else
 		logg("   ADDR2LINE: Disabled");
 
+	// REPLY_WHEN_BUSY
+	// How should FTL handle queries when the gravity database is not available?
+	// defaults to: BLOCK
+	buffer = parse_FTLconf(fp, "REPLY_WHEN_BUSY");
+
+	if(buffer != NULL && strcasecmp(buffer, "DROP") == 0)
+	{
+		config.reply_when_busy = BUSY_DROP;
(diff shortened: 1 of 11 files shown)


DL6ER:
Connected web PR: https://github.com/pi-hole/AdminLTE/pull/1869

yubiuser:
Connected doc PR: https://github.com/pi-hole/docs/pull/554

dschaper:
Sounds like a great idea to me.

dschaper (approved):

PR #1162, opened 2021-09-12
Change FTL database permission to 664

yubiuser:

1

---
Sets the file permission of the `pihole-FTL.db` to `664` to allow write permission to the `pihole` group. This is necessary for https://github.com/pi-hole/AdminLTE/pull/1886 as it allows the web interface write permission to delete items from the `message` table

### DIFF
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -212,9 +212,9 @@ static bool db_create(void)
 	// Close database handle
 	dbclose(&db);
 
-	// Explicitly set permissions to 0644
-	// 644 =            u+w       u+r       g+r       o+r
-	const mode_t mode = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH;
+	// Explicitly set permissions to 0664
+	// 664 =            u+w       u+r       g+w       g+r       o+r
+	const mode_t mode = S_IWUSR | S_IRUSR | S_IWGRP | S_IRGRP | S_IROTH;
 	chmod_file(FTLfiles.FTL_db, mode);
 
 	return true;
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -212,11 +212,6 @@ static bool db_create(void)
 	// Close database handle
 	dbclose(&db);
 
-	// Explicitly set permissions to 0664
-	// 664 =            u+w       u+r       g+w       g+r       o+r
-	const mode_t mode = S_IWUSR | S_IRUSR | S_IWGRP | S_IRGRP | S_IROTH;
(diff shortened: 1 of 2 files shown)


yubiuser:
I've been thinking about how to change permissions on already existing databases. My idea would be to move the whole section 

https://github.com/pi-hole/FTL/blob/b90ab8b1e78d7aece11c8fee1c3c00817761de52/src/database/common.c#L172-L175

from `db_create` to `db_init` right before the database is going to be opened

https://github.com/pi-hole/FTL/blob/b90ab8b1e78d7aece11c8fee1c3c00817761de52/src/database/common.c#L211

This would set the permissions every time the database in initialized (same for newly created db or existing databases). What do you think?

DL6ER:
Yes, this sounds like a workable solution. There should be no harm involved in (re)setting the permissions once on every FTL start. FTL may not be `root` at the point where you want to do the `chown`, however, this shouldn't be a concern here because FTL has `CAP_CHOWN` which should give it the power of changing ownership of files even when not `root`.

**edit 1**

Thinking again about this, we might as well want to do it here:

https://github.com/pi-hole/pi-hole/blob/2673c2c0720fb3871a4c3cae6a3f38f9ccede24e/advanced/Templates/pihole-FTL.service#L29

because we never ensure the permissions are right after the file has been created.

yubiuser:
Like that https://github.com/pi-hole/pi-hole/pull/4328?

src/database/common.c
```diff
@@ -261,6 +256,11 @@ void db_init(void)
 		}
 	}
 
+		// Explicitly set permissions to 0664
+		// 664 =            u+w       u+r       g+w       g+r       o+r
+		const mode_t mode = S_IWUSR | S_IRUSR | S_IWGRP | S_IRGRP | S_IROTH;
+		chmod_file(FTLfiles.FTL_db, mode);
+
```
DL6ER, line 263: Sorry, one last but, then we are ready to merge: Indentation is broken here.
Screenshot from 2021-09-23 20-51-52
```suggestion
	// Explicitly set permissions to 0664
	// 664 =            u+w       u+r       g+w       g+r       o+r
	const mode_t mode = S_IWUSR | S_IRUSR | S_IWGRP | S_IRGRP | S_IROTH;
	chmod_file(FTLfiles.FTL_db, mode);

```

DL6ER (dismissed):

DL6ER (approved):

DL6ER:
Signed commits locally and updated the fork branch as discussed with @yubiuser

PR #1171, opened 2021-09-21
Add special handling of iCloud Private Relay domains

DL6ER:

## 10

---

Implement special handling of Apple iCloud Private Relay domains to prevent Applie devices from bypassing Pi-hole.

> Some enterprise or school networks might be required to audit all network traffic by policy, and your network can block access to Private Relay in these cases. The user will be alerted that they need to either disable Private Relay for your network or choose another network.
>
> The fastest and most reliable way to alert users is to **return a negative answer from your network’s DNS resolver, preventing DNS resolution for the following hostnames used by Private Relay traffic**. Avoid causing DNS resolution timeouts or silently dropping IP packets sent to the Private Relay server, as this can lead to delays on client devices.
> 
> `mask.icloud.com`
> `mask-h2.icloud.com`

Source: https://developer.apple.com/support/prepare-your-network-for-icloud-private-relay

The new option `BLOCK_ICLOUD_PR=true|false` can be used to control the new behavior.

This is being discussed and tested in the Discourse topic linked below.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -621,6 +621,18 @@ void read_FTLconf(void)
 	else
 		logg("   BLOCK_TTL: %u seconds", config.block_ttl);
 
+	// BLOCK_ICLOUD_PR
+	// Should FTL handle the iCloud privacy relay domains specifically and
+	// always return NXDOMAIN??
+	// defaults to: true
+	buffer = parse_FTLconf(fp, "BLOCK_ICLOUD_PR");
+	config.special_domains.icloud_private_relay = read_bool(buffer, true);
+
+	if(config.special_domains.icloud_private_relay)
+		logg("   BLOCK_ICLOUD_PR: Enabled");
+	else
+		logg("   BLOCK_ICLOUD_PR: Disabled");
+
 	// Read DEBUG_... setting from pihole-FTL.conf
 	read_debuging_settings(fp);
 
--- a/src/config.h
+++ b/src/config.h
@@ -50,6 +50,7 @@ typedef struct {
 	bool addr2line :1;
 	struct {
 		bool mozilla_canary :1;
+		bool icloud_private_relay :1;
 	} special_domains;
 	enum privacy_level privacylevel;
 	enum blocking_mode blockingmode;
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -1086,6 +1086,30 @@ static bool special_domain(const queriesData *query, const char *domain)
 		return true;
 	}
 
+	// Apple iCloud Private Relay
+	// Some enterprise or school networks might be required to audit all
+	// network traffic by policy, and your network can block access to
+	// Private Relay in these cases. The user will be alerted that they need
+	// to either disable Private Relay for your network or choose another
+	// network.
+	// The fastest and most reliable way to alert users is to return a
+	// negative answer from your network’s DNS resolver, preventing DNS
+	// resolution for the following hostnames used by Private Relay traffic.
+	// Avoid causing DNS resolution timeouts or silently dropping IP packets
+	// sent to the Private Relay server, as this can lead to delays on
+	// client devices.
+	// > mask.icloud.com
+	// > mask-h2.icloud.com
+	// https://developer.apple.com/support/prepare-your-network-for-icloud-private-relay
+	if(config.special_domains.icloud_private_relay &&
+	   (strcasecmp(domain, "mask.icloud.com") == 0 ||
+	    strcasecmp(domain, "mask-h2.icloud.com") == 0))
+	{
+		blockingreason = "Apple iCloud Private Relay domain";
+		force_next_DNS_reply = REPLY_NXDOMAIN;
+		return true;
+	}
+
 	return false;
 }
 


DL6ER:
@nol166 My understanding is that you cannot use Privacy Relay and Pi-hole at the same time as the former always bypasses the latter. Pinging @yubiuser, he tested the feature and had an Apple device he can run tests on.

yubiuser:
@nol166 
I'm not sure I fully understand your question. If private relay is enabled and the those two domains are not blocked the device will start sending all DNS traffic to one of apples DNS servers, hence bypassing Pi-hole. If PR is enabled, but two domains are blocked, the device is going to ask the user, if it want's to disable PR for that particular network (wifi). If you choose to disable it, the local DNS server is used.

In effect you can have PR enabled in general but shut off at home were your local Pi-hole is blocking those two domains.

nol166:
@yubiuser i think I misunderstood the spec. After reading the docs again it sounds like Apple recommends returning a negative response for the two domains used for private relay, thus prompting an alert to the user.

This PR just implements that negative response in FTL, is that correct?

DL6ER:
Rebased onto latest `development` and squashed commits into one.

jpgpi250:
also mentioned in [this](https://old.reddit.com/r/pihole/comments/px136i/enforcing_safe_search_is_not_working/) reddit topic. Mentioned both solutions (the special branch and the server= solution), not all users want to switch to specific branches or even dev...

PromoFaux (approved):

wAuner:
Hi, after doing some tests I wanted to share my observations with you: When I use the new option `BLOCK_ICLOUD_PR=true` (with its default value), it works as expected, I'm not able to use private relay. 
However, I found that turning it off is much more useful (same behaviour with an older pihole version): this way my Apple device uses pihole as  my dns resolver but still routes Safari traffic through Apple's private relay network. This can easily be verified by looking at the external IP while surfing. So this seams like the best of both worlds for me right now.

I'm not sure why this is working, because it probably shouldn't just use the local dns provider. Maybe this a beta bug in private relay or some other PR domains are blocked by some blocklist that yields to this behaviour?

wAuner:
@yubiuser that's right, but what is not expected (at least for me) is that private relay does still work in Safari, but uses my pihole DNS. If that would be a consistent behaviour, then there wouldn't be a reason for this option if I'm not mistaken. 
With this option you're forcing the user to disable private relay entirely, although that does not seem to be necessary. 

But I'm not sure why exactly this works for me. I was expecting private relay to be an all-or-nothing deal. When I first tested it on my network, private relay did use its own DNS servers (including ads). So I'm not sure how I did get it to work to still use private relay for Safari but using my pihole for DNS.

My setup is like this: 
* my router tells its clients to use the pihole ip for dns
* my pihole instance (Synology docker with port forwarding) has my router configured as upstream dns
* my router uses DoT for DNS

wAuner:
@nol166 good question. Now that you've mentioned it everything works as expected. The ad blocking on my iOS device was due to Firefox Clear and not pihole. I forgot that I had that extension installed. The only queries I see form the private relay devices are `mask.apple-dns.net`. 
I thought the ad-blocking was due to pihole which I couldn't explain. Thanks, I hope I didn't waste anyone's time.

DL6ER:
> I wonder what others think the default behaviour for PiHole should be?

I wanted to say that Apple users are the minority of users, however, then I realized that this is a home-made issues and non-Apple users are not affected at all by this.

> Behaviour is that within the Mail app the message “Unable to load remote content privately” is shown. The user can then choose to load remote content.

Sounds reasonable. What guarantees you that downloading the content over the Apple service is any more "privately"? I mean, it is them encrypting it for you to have it encrypted during transit. However, this also means they see it in plain. As the vast majority of the web is using HTTPS and mailsservers are almost all using TLS encryption, too, I would say **this feature may even reduce privacy** when enabled. Am I wrong?

If people say "Google has my data anyway and I don't expect them to do anything harmful" and that's okay for them, I can agree. I could, too, for Apple, I guess, but, so far, something still seems to be missing here.

Question: What kind of "remote content" are we talking about? Images? Attachments? I think the point they're making about "private" downloads is mood given the ubiquity of HTTPS these days.

Write:
> 
> 
> But the email marketers will know the unique image that corresponds to the email that was sent to you was viewed and opened. So they know you read it.

That doesn't matter since all remote content will be loaded at random time (or instantly, I don't remember) & trough an Apple IP. So the data is spoofed, it'll be marked as open even if you never opened it. 

More info : https://www.apple.com/legal/privacy/data/en/mail-privacy-protection/

> Unless the underlying method is that Apple will read your email and open attachments and load images without your interaction?
Apple can't see the content. https://www.apple.com/legal/privacy/data/en/mail-privacy-protection/

dschaper:
> By using these features, you agree and consent to Apple’s and its subsidiaries’ and agents’ transmission, collection, maintenance, processing, and use of this information as described above.

jfb-pihole:
I have had no connectivity issues nor have I heard of any.  I have a variety of Apple devices with both IOS and MacOS and zero problems.  With the Pi-hole default behavior, a client trying to use Private Relay will show a screen telling you that the network does not support it.  The device continues to work normally, but you cannot use Private Relay.

The Apple developers specifically provided the correct response to the two iCloud domains and included the following comments in their discussion. 

> Avoid causing DNS resolution timeouts or silently dropping IP packets sent to the Private Relay server, as this can lead to delays on client devices.

https://developer.apple.com/support/prepare-your-network-for-icloud-private-relay

The reply of NXDOMAIN triggers the IOS warning which avoids connectivity issues.  If we  silently blocked the domain with no reply, then we might have caused connectivity issues, but we do not.

In my opinion, enabling this by default in Pi-hole is the correct behavior.  When you install a network-wide ad blocker, you expect it to work network wide.  Disabling a potential DNS bypass (and doing so by specifically following the Apple recommendations) is consistent with how we handled the Firefox DoH canary domain several releases ago.

andyxheli:
any status on this ?

rdwebdesign:
This issue was closed more than a year ago. Also, the last comment was posted more than a year ago.

If you need help with configuration, please open a topic on [Pi-hole Discourse Forum](https://discourse.pi-hole.net/).
If you want to suggest or request a new feature, please open a request on [Pi-hole Discourse Forum - Feature Requests](https://discourse.pi-hole.net/c/feature-requests/8).
If you found a new bug, please open a [new issue](https://github.com/pi-hole/FTL/issues).

PR #1173, opened 2021-09-24
Handle TTL for blocked queries seperately 

DL6ER:

## 10

---

So far, we used `local-ttl=2` hard-coded in dnsmasq's config because this ensured clients won't cache blocked queries for longer times, possibly rendering whitelisting ineffective.

This PR implements a **new TTL for blocked queries only**. This allows users to set `local-ttl` to any value they prefer without this being overwritten in `01-pihole.conf` after each update. This seems useful in context of locally used hostnames that are known to stay constant over long times (e.g., printers, etc.).

The new value defaults to `2` seconds (no change in behavior) and can be controlled using the new FTL config option `BLOCK_TTL=2`

### DIFF
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -77,6 +77,7 @@ static union all_addr null_addrp = {{ 0 }};
 static enum reply_type force_next_DNS_reply = REPLY_UNKNOWN;
 static int last_regex_idx = -1;
 static struct ptr_record *pihole_ptr = NULL;
+#define HOSTNAME "Pi-hole hostname"
 
 // Fork-private copy of the interface data the most recent query came from
 static struct {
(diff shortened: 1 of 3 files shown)


src/dnsmasq_interface.c
```diff
@@ -77,6 +77,7 @@ static union all_addr null_addrp = {{ 0 }};
 static enum reply_type force_next_DNS_reply = REPLY_UNKNOWN;
 static int last_regex_idx = -1;
 static struct ptr_record *pihole_ptr = NULL;
+#define HOSTNAME "Pi-hole hostname"
```
dschaper, line 80: Is this from the hostname PR or is this related to this PR?

DL6ER: It is intended here, now that we need the string in two places, I made it a preprocessor define.

dschaper: Great.

dschaper (approved):

yubiuser:
Documentation needed.

DL6ER:
See https://github.com/pi-hole/docs/pull/573/commits/60195eadd304ccb0da05d50c3e18cfca45bc20cb in https://github.com/pi-hole/docs/pull/573

PR #1177, opened 2021-09-29
Github Actions

PromoFaux:
Ready GHA to take over from circle.

This still leaves circle in place to build and push branch builds to ftl.pi-hole.net

However, this will automatically attach binaries built by GHA to the release when we release. I have tested this over on my own fork.  If, for whatever reason, the binaries that GHA builds are duff, we can easily replace them with the ones circle builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ name: Build and test
 on:
   push:
     branches:
-      - '*'
+      - '**'
   pull_request:
   release:
     types: [published]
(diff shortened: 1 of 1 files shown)


Daxtorim:
Is there a way to check for account/repo name before uploading binaries to your server?
Something like `if: ${{ repo_name == 'pi-hole/FTL' }}` (pseudo code, haven't looked at any docs yet)?

Right now, all jobs will fail on my fork because I obviously don't have credentials for "Pi-hole server".
https://github.com/Daxtorim/FTL/runs/3748611677?check_suite_focus=true

PromoFaux:
Yeah, we should be able to add something like that. I'm currently looking at a (hacky workaround) to get around GHA's lack of retrying a single job vs retrying the lot. After which I will look into that.

At least for now you can see the tests are passing, it's just failing at the upload step

dschaper (approved):

PR #1182, opened 2021-09-30
Do not fail hard when rev-server has a non-zero final address part

### LINKED ISSUE #1181: FTL fails to start with conditional forwarding enabled
### Versions

  Pi-hole version is v5.5 (Latest: v5.5)
  AdminLTE version is v5.7 (Latest: v5.7)
  FTL version is release v5.10.1 (Latest: v5.10.1)

### Platform

- OS and version:  Ubuntu 20.04.3

- Platform:  VM on ESXi

### Expected behavior

Conditional forwarding works / FTL doesn't fail to start.

### Actual behavior / bug

`FTL failed to start due to address part not zero at line 44 of /etc/dnsmasq.d/01-pihole.conf`

### Steps to reproduce

1. Enable conditional forwarding
2. Click "Save."
Disabling conditional forwarding fixes the issue.

## Debug Token

- URL:  https://tricorder.pi-hole.net/dt2GmB7x/
Note: this is without conditional forwarding on, it would not upload with FTL failed

DL6ER:

## 10

---

Fixes #1181 by `dnsmasq v2.78test2` enforcing a special CIDR notation where the remaining bits have to be set to zero (e.g., `192.168.2.0/24` is fine but `192.168.2.1/24` fails). This breaks a lot of existing installations. This PR adds a patch I also sent upstream to `dnsmasq` to, instead of unnecessarily failing hard, just sets those bits to zero. This seems the right thing to do in any case as I do not see any reason for this break in backwards compatibility.
In case the patch is not accepted upstream - or modified - we can always revert this change and insert the fix added by Simon.

### DIFF
--- a/src/dnsmasq/option.c
+++ b/src/dnsmasq/option.c
@@ -967,13 +967,9 @@ static char *domain_rev4(int from_file, char *server, struct in_addr *addr4, int
   
   if (size > 32 || size < 1)
     return _("bad IPv4 prefix length");
-  
-  for (i = 0; i < addrbytes; i++)
-    if (((u8 *)addr4)[3-i] != 0)
-      break;
-  
-  if (i != addrbytes || (((u8 *)addr4)[3-addrbytes] & ((1 << addrbits) - 1)) != 0)
-    return _("address part not zero");
+
+  /* Zero out last address bits according to CIDR mask */
+  ((u8 *)addr4)[3-addrbytes] &= ~((1 << addrbits)-1);
   
   size = size & ~0x7;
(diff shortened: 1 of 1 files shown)


yubiuser (approved):
Works.

```Sep 30 12:16:43 dnsmasq[30568]: using nameserver 10.0.1.1#53 for domain 1.0.10.in-addr.arpa 
Sep 30 12:16:43 dnsmasq[30568]: using nameserver 10.0.1.1#53 for domain fritz.box 
```

PR #1183, opened 2021-09-30
Pi-hole FTL v5.10.2

DL6ER:

## 10

---

Update embedded `dnsmasq` version to 2.87test3 including a fix we submitted upstream to fix https://github.com/pi-hole/FTL/issues/1181. We also fix an issue with `REPLY_ADDR{4,6}` not being used for replies to `pi.hole`.

### DIFF
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -794,6 +794,7 @@ void _FTL_iface(struct irec *recviface, const char *file, const int line)
 	if(config.reply_addr.overwrite_v4)
 	{
 		memcpy(&next_iface.addr4, &config.reply_addr.v4, sizeof(config.reply_addr.v4));
+		next_iface.haveIPv4 = true;
 
 		if(config.debug & DEBUG_NETWORKING)
 		{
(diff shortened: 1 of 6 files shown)


yubiuser (approved):
Tested both changes: `--rev-server` ending with non zero and `REPLY_ADDR4` are working

src/dnsmasq_interface.c
```diff
@@ -800,6 +800,7 @@ void _FTL_iface(struct irec *recviface, const char *file, const int line)
 			char buffer[ADDRSTRLEN+1] = { 0 };
 			inet_ntop(AF_INET, &next_iface.addr4, buffer, ADDRSTRLEN);
 			logg("Configuration overwrites IPv4 address: %s", buffer);
+			next_iface.haveIPv4 = true;
```
yubiuser, line 803: This needs to go 6 lines up I think. (thanks to @jpgpi)

jpgpi250:
please don't release yet

When the debug flags are set, it works
when the debug flags are removed, it stops working

confirmed by @yubiuser 

he says:

```
Guess it is because the added line is within an if(config.debug & DEBUG_NETWORKING) condition.
```

src/dnsmasq_interface.c
DL6ER: Yes, I see I pushed the wrong commit (this was from earlier testing) but the tests complained anyway so nothing would have happened :-)

Fixed now. Binaries are currently building.

yubiuser: Working with the new builds even without the debug flags

jpgpi250: version  is now vDev-ccf8d15, tried again, appears to be working.
Thanks.

dschaper:
I like squash.

dschaper (approved):

PR #1196, opened 2021-10-03
Fix network byte order confusion in local lookups

### LINKED ISSUE #1195: Pihole asking 1.0.0.127 - Cloudflare for RFC1918 arpa info
### Versions
  Pi-hole version is v5.5 (Latest: v5.5)
  AdminLTE version is v5.7 (Latest: v5.7)
  FTL version is v5.10.2 (Latest: v5.10.2)  

### Platform
$ cat os-release 
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian

### Expected behavior
I have a setup where my piholes are forwarding DNS queries to my two linux bind9 servers (on same subnet)
No other DNS serves are specified (or selected) in the pihole-setup.

The Pi-Holes have always behaved well , and i have never seen any DNS requests to other servers.
I have blocked (pfSense firewall) DNS queries from pihole ip to any other DNS servers.

After me updating my two PI-Holes to latest , PI-Hole had begun to send DNS queries to 1.0.0.127 (Cloudflare)

### Actual behavior / bug

I'm beginning to see these block lines in my pfSense , after i updated yesterday.
On both my piholes.

`    Oct 2 08:00:50 	Pihole-Vlan 	Deny Outside access DNS from local IP's (1522486856) 	Pihole-IP:51765		1.0.0.127:53		UDP
	Oct 2 08:00:45 	Pihole-Vlan 	Deny Outside access DNS from local IP's (1522486856) 	Pihole-IP:38610		1.0.0.127:53		UDP
	Oct 2 08:00:40 	Pihole-Vlan 	Deny Outside access DNS from local IP's (1522486856) 	Pihole-IP:38610		1.0.0.127:53		UDP `

Maybe see:
https://forum.netgate.com/topic/166933/pi-hole-new-version-leaking/1

This is not a major issue - But why have it begun doing that ???
Pi-Hole has always been well behaved.

My pihole network is setup to a static ip , and use "localhost" as dns resolver.
Technically it could be the Debian OS doing this query , i can just see it comming from the pihole ip.

`$ cat /etc/resolv.conf 
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
nameserver 127.0.0.1
search mydomain.org
$ 
`

After a bit more digging , it seems to be reverse resolving of the "Top Clients" that causes the queries to 1.0.0.127
It seems there is a resolver job running every hour at 00.

Firewall log - Easy to see the hourly job - It began yesterday after pihole update to latest
[pihole-block.txt.zip](https://github.com/pi-hole/pi-hole/files/7271141/pihole-block.txt.zip)

pfSense pcap trace of the actions today at 10:00 (just match on 1.0.0.127 UDP 53)
[pihole-1.0.0.127-dns-packetcapture.cap.zip](https://github.com/pi-hole/pi-hole/files/7271142/pihole-1.0.0.127-dns-packetcapture.cap.zip)

### Steps to reproduce

Unknown .. It has never done that before.
Well capture DNS on the hour 00 , and see if your Local (rfc1918) Top Clients also gets resolved at some external server 

## Debug Token

NA
pihole-forwarders
pihole-ver

DL6ER:

## 10

---

This PR fixes #1195 by ensuring local queries are sent to `127.0.0.1` instead of `1.0.0.127`. The issue comes from the networking constant `INADDR_LOOPBACK` being in host byte order whereas `struct in_addr` expects an address in network byte order so we need to convert the static constant. This was not mentioned in [`man 7 ip`](https://man7.org/linux/man-pages/man7/ip.7.html), however, once we've seen the symptoms, it became clear that this is what is meant with
```
           /* Internet address */
           struct in_addr {
               uint32_t       s_addr;     /* address in network byte order */
           };
```
There is no extra information attached to the constant itself that it'd be converted:
```
       There are several special addresses: INADDR_LOOPBACK (127.0.0.1)
       always refers to the local host via the loopback device;
       INADDR_ANY (0.0.0.0) means any address for binding;
       INADDR_BROADCAST (255.255.255.255) means any host and has the
       same effect on bind as INADDR_ANY for historical reasons.
```
You'd naively expect that a constant defined to be used in this context should be defined in a way that doesn't need prior conversion.

### DIFF
--- a/src/resolve.c
+++ b/src/resolve.c
@@ -255,7 +255,9 @@ char *resolveHostname(const char *addr)
 		res_initialized = true;
 	}
 
-	struct in_addr FTLaddr = { INADDR_LOOPBACK };
+	// INADDR_LOOPBACK is in host byte order, however, in_addr has to be in
+	// network byte order, convert it here if necessary
+	struct in_addr FTLaddr = { htonl(INADDR_LOOPBACK) };
 	in_port_t FTLport = htons(config.dns_port);
 
 	// Set FTL as system resolver only if not already the primary resolver


DL6ER:
Confirmed working by two independent testers in #1195

PromoFaux (approved):
> Confirmed working by two independent testers in #1195

👍

Issue #1191, opened 2021-10-02
Pi-hole not responding to clients asking to Pi-hole hostname


Mic-Pod:
Hi,
Before recent updates clients were able to resolve Pi-hole server's hostname with IP. Currently client can't resolve Pi-hole server's hostname.
In web GUI, there are no entries in Query Log when doing something like `dig @192.168.1.100 server1` but there are entries in /var/log/pihole.log (details below).

### Before:
`dig +short @192.168.1.100 server1`
`192.168.1.100`

### After:
`dig +short @192.168.1.100 server1`

### Versions

- Pi-hole:  5.5
- AdminLTE:  5.7
- FTL:  5.10.2

### Platform

- OS and version:  Ubuntu Server 18.04.6 LTS

- Platform:  

### Expected behavior

Resolving pi-hole hostname with IP

### Actual behavior / bug

No answer with IP when asking for pi-hole hostname

### Steps to reproduce

`dig @192.168.1.100 server1`
no ANSWER SECTION
entry in /var/log/pihole.log:
`Oct  1 22:45:27 dnsmasq[23259]: query[A] server1 from 192.168.1.11
Oct  1 22:45:27 dnsmasq[23259]: Pi-hole hostname server1 is Pi-hole hostname`

DL6ER:
Thanks for your report and sorry for the inconvenience caused by this. You should be able to workaround this by manually setting a suitable `REPLY_ADDR4` and/or `REPLY_ADDR6` in `/etc/pihole/pihole-FTL.conf`. See also the discussion in our Discourse forum posted by @pralor-bot directly above my comment.

Could you please add
```
DEBUG_QUERIES=true
DEBUG_NETWORKING=true
```
to `/etc/pihole/pihole-FTL.conf` and send me the snippet from `/var/log/pihole-FTL.log` when doing the `dig pi.hole` ?
There should be a line like
```
Interfaces: Called from ...
```
and one like
```
**** new UDP IPv4 query[A] query "pi.hole" from ...
```

Ideally, there is content in between. If there is noting in between these lines, Pi-hole isn't able to read the interfaces on your system. On which kind of hardware are your system running? Do you use any kind of virtualization?

DL6ER:
@Mic-Pod @AlphaCluster Please also run

pihole checkout ftl tweak/iface_detection

and repeat the `dig pi.hole` and send me what you see in `pihole-FTL.log`.

[...]

PR #1198, opened 2021-10-03
Improve interface detection

DL6ER:

## 10

---

Fixes https://github.com/pi-hole/FTL/issues/1191 by extending interface analysis to also support configurations defining `bind-interface`. So far, we have only seen users having this config line enabled unintentionally (auxiliary `ubuntu-fan` or `lxd` config files placed by the system into `/etc/dnsmasq.d`), however, this does not mean we should not support it. A second goal of the PR is to largely extend the `DEBUG_NETWORKING` toolset to ease debugging in complex network environments.

The bug fix as been confirmed working.

### DIFF
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -778,7 +778,8 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
 	return blockDomain;
 }
 
-void _FTL_iface(struct irec *recviface, const char *file, const int line)
+void _FTL_iface(struct irec *recviface, const union all_addr *addr, const sa_family_t addrfamily,
+                const char *file, const int line)
 {
 	// Invalidate data we have from the last interface/query
 	// Set addresses to 0.0.0.0 and ::, respectively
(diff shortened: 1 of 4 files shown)


src/dnsmasq_interface.c
```diff
@@ -820,29 +821,98 @@ void _FTL_iface(struct irec *recviface, const char *file, const int line)
...
+	    (addrfamily == AF_INET6 && addr->addr6.s6_addr[0] != 0)))
+	{
+		if(config.debug & DEBUG_NETWORKING)
+		{
+			char addrstr[INET6_ADDRSTRLEN] = { 0 };
+			if(addrfamily == AF_INET)
+				inet_ntop(AF_INET, &addr->addr4, addrstr, INET6_ADDRSTRLEN);
+			else // if(addrfamily == AF_INET6)
+				inet_ntop(AF_INET6, &addr->addr6, addrstr, INET6_ADDRSTRLEN);
+			logg("Identifying interface (looking for %s):", addrstr);
```
yubiuser, line 836: This part is only executed while debug flags are set. But more than logging happens. Should it be like that?

yubiuser (approved):

PR #1206, opened 2021-10-07
Tweak to GHA workflow

PromoFaux:

- [] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.

Tweaks to the tests to make things a bit more efficient.  Duplicates things a little for now, but this is an ever evolving configuration as we learn what works and what doesn't...

 Split workflow up into three distinct parts, each depending on the previous part
     1 - Build and Test x86_64/32 binaries
     2 - Build arm/aarch64 binaries (test script is hardcoded to not test these, anyway)
     3 - If relevant, upload binaries via SSH and/or attach to release

~~This also means that when there are transient errors with DNS, causing the tests to fail - we can just restart the workflow and "only" have to rebuild 3 arch's rather than the whole lot~~ Sorted - we now use pdns and all DNS tests are performed "offline"

Temporary artifact storage is used to pass the binaries through to the final step, at which point they are uploaded

~~I would also like to get for https://github.com/pi-hole/docker-base-images/pull/40 to be merged and tagged as v11 so that we can remove the "update git" step~~ Done

An example of the result of this:

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,61 +9,49 @@ on:
     types: [published]
 
 jobs:
-  build:
-
-    strategy:
-      matrix:
-        include:
-         - arch: armv4t
-           bin_name: pihole-FTL-armv4-linux-gnueabi
-         - arch: armv5te
-           bin_name: pihole-FTL-armv5-linux-gnueabi
-         - arch: armv6hf
-           bin_name:  pihole-FTL-armv6-linux-gnueabihf
-         - arch: armv7hf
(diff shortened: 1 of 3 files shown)


PromoFaux:
Merge conflicts make sense, will rebase on development once `tweak/tests` is merged

DL6ER:
Does the `v1.13` container have `pdns-doc` for `alpine` ?
If so, we can also remove

https://github.com/pi-hole/FTL/blob/4c4224662bddc48c4496f29496b64daca708ebc9/test/pdns/setup.sh#L19-L20

PR #1213, opened 2021-10-10
Fix incorrect response times for complex CNAMEs

DL6ER:

## 10

---

Sometimes long-lived `CNAME`s are in our cache but we still need to forward upstream because there are shorter-lived children such as other `CNAME`s, `A`, or `AAAA` records. In this case, we need to start measuring again because we are interested in the time the upstream takes to reply because this is the delay the users sees (we can only serve a DNS reply when we have everything ready). The delay needed for the initial cache lookup is not interesting here (this is always very fast).


DL6ER:
Confirmed fixed by original reporter in https://discourse.pi-hole.net/t/pi-hole-no-longer-blocks-via-cname/50078/34?u=dl6er

PR #1215, opened 2021-10-10
[Internal] Whoops

PromoFaux:
Updates the version of the build container used in the circle builds. Tests had the `apk install pdns-doc` removed, as it was added directly into the build container.


PromoFaux:
image

PR #1238, opened 2021-10-31
Use REPLY_ADDR4 to overwrite the DNS server in DHCPOFFERs

DL6ER:

## 10

Use IP address specified by `REPLY_ADDR4` (if any), to overwrite the automatically added `DNS` server in `DHCPOFFER`s.

### DIFF
--- a/src/dnsmasq/rfc2131.c
+++ b/src/dnsmasq/rfc2131.c
@@ -15,6 +15,7 @@
 */
 
 #include "dnsmasq.h"
+#include "../dnsmasq_interface.h"
 
 #ifdef HAVE_DHCP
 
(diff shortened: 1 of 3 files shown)


DL6ER:
I rather think this should be achieved with an extra `dnsmasq` config line hard-coding the DNS server to broadcast via option `6` like
```
dhcp-option=6,192.168.1.2
```

This should reliably overwrite the automatic address which may be wrong inside a `docker` container. I think this extra step is preferred over the other (somewhat obscure/"magic") way of doing it behind the scenes, pretending to know what is "good" for the user. This feeling is stengthened by the fact that users may use `REPLY_ADDR4` to point to a server that is serving some kind of blocking page and may not be running a DNS server at all.

PR #1241, opened 2021-11-05
Improve virtual interface analysis

DL6ER:

## 10

---

Allow alias interfaces to be analyzed by making a bug in `dnsmasq` consistent. More information can be found in the linked Discourse discussion.

### DIFF
--- a/src/dnsmasq/dnsmasq.h
+++ b/src/dnsmasq/dnsmasq.h
@@ -640,6 +640,9 @@ struct irec {
   struct in_addr netmask; /* only valid for IPv4 */
   int tftp_ok, dhcp_ok, mtu, done, warned, dad, dns_auth, index, multicast_done, found, label;
   char *name; 
+  /* Pi-hole modification */
+  char *slabel;
+  /************************/
   struct irec *next;
 };
 
--- a/src/dnsmasq/network.c
+++ b/src/dnsmasq/network.c
@@ -492,11 +492,11 @@ static int iface_allowed(struct iface_param *param, int if_index, char *label,
     }
   
   if (addr->sa.sa_family == AF_INET &&
-      !iface_check(AF_INET, (union all_addr *)&addr->in.sin_addr, label, &auth_dns))
+      !iface_check(AF_INET, (union all_addr *)&addr->in.sin_addr, ifr.ifr_name, &auth_dns))
     return 1;
 
   if (addr->sa.sa_family == AF_INET6 &&
-      !iface_check(AF_INET6, (union all_addr *)&addr->in6.sin6_addr, label, &auth_dns))
+      !iface_check(AF_INET6, (union all_addr *)&addr->in6.sin6_addr, ifr.ifr_name, &auth_dns))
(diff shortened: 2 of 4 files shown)


DL6ER:
Confirmed working by original reporter in https://discourse.pi-hole.net/t/finding-pi-hole-ip-addresses-after-v5-4-update/49712/60

DL6ER:
Note: This change should be sent upstream to `dnsmasq` at some point in the future to improve also the upstream project's handling of virtual interfaces/interface aliases. This will be independent to this fix and will only be done as soon as Simon Kelley returned.

yubiuser:
So this PR makes the bug consistent, but is only a workaround until the bug is really fixed. Fixing the underlying bug is what your linked `dnsmasq` PR does?

DL6ER:
The "make the bug consistent" I did in here is going to be sent to `dnsmasq`. I don't think the bug will ever be really fixed due to technical limitations of the Linux networking stack.

Background: When we receive a reply, we get two details about the receiving interface: The interface index (from the kernel) and the address on which the packet arrived on (from the DNS packet header). Unfortunately, alias/virtual interfaces are still the same *physical* device and, hence, `dnsmasq` cannot distinguish based on the interface index alone.
FTL does this differently, but it means extra work. Instead of relying on the kernel-provided interface index, we compile a list of IP addresses on the machine and then iterate over them until we find a match for the address we received the query on. This allows us to see if the address is bound to an alias/virtual interface.

`dnsmasq` has two options: Either they implement the same iterating over all interface address we do now *inside FTL* or they acknowledge that there is a bug and make the bug consistent (this means "alias/virtual" interfaces are indistinguishable from "real" ones in the config). This is wrong but less code change.

We'll see.

DL6ER:
Sorry for the long text above. I'll try again shorter and from another perspective:

1. The bug is that `dnsmasq` also listens on `eth0:0` when the config asks it to listen only on `eth0` (reasons above)
2. As we should not be listening on `eth0:0`, FTL has no information about this interface.
3. Because of this, the local address in not available when a query arrives, and
4. we fail to provide the correct response.

When we fix no. 1 in `dnsmasq` (stop accepting queries on the virtual interface), nothing of this is actually needed. However, fixing no. 1 does not straightforward at this time.

yubiuser (approved):

yubiuser:
Thanks for the explanation. I have little hope `dnsmasq` will pick up the iteration but maybe include this PR.

DL6ER:
They already have such a code, the implementation is, however, a bit incorrect.

1. Config `interface=eth0`
Queries on `eth0` and `eth0:0` are accepted because `dnsmasq` only compares the physical interface name string.

2. Config `interface=eth0:0`
Queries on `eth0` and `eth0:0` are rejected (at first!) because of the mismatch. However, there is another check "`label_expection()`" that does said iteration and would lead to accepting the `eth0:0` query. The `eth0` query is correctly rejected.

The fix would be to generalize the `label_exception()` to be always be used *instead* of the interface name string.

PR #1244, opened 2021-11-14
Ensure dbclose() is not called multiple times

DL6ER:

## 10

---

Reduce code duplication by using labels + ensure we do not run `dbclose()` multiple times on database issues (e.g., when database is locked). The double-closing can cause some warnings in the FTL log even when this is harmless, for instance,
```
[2021-11-13 09:52:03.250 875/T879] ERROR: SQL query "END TRANSACTION" failed: database is locked
[2021-11-13 09:52:03.251 875/T879] WARNING: Storing devices in network table failed: database is locked
[2021-11-13 09:52:03.251 875/T879] SQLite3 message: API call with invalid database connection pointer (21)
[2021-11-13 09:52:03.251 875/T879] SQLite3 message: misuse at line 166280 of [5c9a6c0687] (21)
[2021-11-13 09:52:03.251 875/T879] Error while trying to close database: bad parameter or other API misuse
```
with this PR, a locked database will instead log only the following lines:
```
[2021-11-13 09:52:03.250 875/T879] ERROR: SQL query "END TRANSACTION" failed: database is locked
[2021-11-13 09:52:03.251 875/T879] WARNING: Storing devices in network table failed: database is locked
```

### DIFF
--- a/src/database/aliasclients.c
+++ b/src/database/aliasclients.c
@@ -259,8 +259,8 @@ void reset_aliasclient(sqlite3 *db, clientsData *client)
 	// Find corresponding alias-client (if any)
 	client->aliasclient_id = get_aliasclient_ID(db, client);
 
-	if(db_opened)
-		dbclose(&db);
+	// Close the database if we opened it here
+	if(db_opened) dbclose(&db);
 
 	// Skip if there is no responsible alias-client
 	if(client->aliasclient_id == -1)
@@ -360,6 +360,5 @@ void reimport_aliasclients(sqlite3 *db)
 	}
 
 	// Close the database if we opened it here
-	if(db_opened)
-		dbclose(&db);
+	if(db_opened) dbclose(&db);
 }
--- a/src/daemon.c
+++ b/src/daemon.c
@@ -18,8 +18,6 @@
 #include "api/socket.h"
 // gravityDB_close()
 #include "database/gravity-db.h"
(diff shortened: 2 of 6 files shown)


yubiuser:
Why is the `goto XXXX` only used on the `message` and `query` table, but not on the `network` table?

src/database/aliasclients.c
```diff
@@ -259,8 +259,8 @@ void reset_aliasclient(sqlite3 *db, clientsData *client)
 	// Find corresponding alias-client (if any)
 	client->aliasclient_id = get_aliasclient_ID(db, client);
 
-	if(db_opened)
-		dbclose(&db);
+	// Close the database if we opened it here
+	if(db_opened) dbclose(&db);
```
yubiuser, line 263: Changing it to be on one line is purely stylisic?

DL6ER: Yes, it is made so a `grep dbclose` in the source code immediately reveals that this is not an unconditional call to this function.

DL6ER:
> Why is the `goto XXXX` only used on the `message` and `query` table, but not on the `network` table?

Because it is used to avoid code duplication (always the same error handling) in the former two cases. However, for the `network` table, the error handling is a bit more complex and is not always the same, i.e., there is no common target we could `goto`.

yubiuser (approved):

PR #1255, opened 2021-12-16
Optimize queries table

DL6ER:

## 10

---

This PR optimizes the `queries` table by using link-tables with IDs instead of storing the `domain`, `client` and `forward` as strings over and over. This is *not* a breaking change to users as we provide a `VIEW queries` in place of the old `TABLE queries` that will give the same results for external queries as before:

`VIEW queries`
```
$ sqlite3 /etc/pihole/pihole-FTL.db -header "SELECT * FROM queries ORDER BY id DESC LIMIT 1;"                                                                                           
id|timestamp|type|status|domain|client|forward|additional_info
19391970|1639654453|2|2|forum.arduino.cc|192.168.2.206|127.0.0.1#5353|
```

`TABLE query_storage` (holding the real data)
```
$ sqlite3 /etc/pihole/pihole-FTL.db -header "SELECT * FROM query_storage ORDER BY id DESC LIMIT 1;"                                                                                           
id|timestamp|type|status|domain|client|forward|additional_info
19391970|1639654453|2|2|114|2|1|
```

In the example above, storage required *per query* before:
```
   4    sizeof(id)
+  4    sizeof(timestamp)
+  1    sizeof(type)
+  1    sizeof(status)
+ 16    strlen("forum.arduino.cc") + 1
+ 14    strlen("192.168.2.206") + 1
+ 15    strlen("127.0.0.1#5353") + 1
+  1    sizeof(NULL)
= 56 bytes per record
```
Storage required *per query* with the optimization of this PR:
```
   4    sizeof(id)
+  4    sizeof(timestamp)
+  1    sizeof(type)
+  1    sizeof(status)
+  1    sizeof(domain)
+  1    sizeof(client)
+  1    sizeof(forward)
+  1    sizeof(additional_info)
= 14 bytes per record
```
This reduction by factor 4 easily outperforms the tiny extra amount of space we need for the newly introduced link-tables. We chose this hybrid approach through the `VIEW` because we cannot afford rewriting the entire `queries` table as this would literally take forever on low-end devices such as Raspberry Pis from the early days.

This also lays the foundation for a requested feature that is availability of host names in the long-term database as the linked client ID are in fact a pair of `(ip,hostname)` in the database. This allows us to store both without any really noticeable overhead.

This PR comes with a tiny drawback (that isn't really one, though): After the database has been optimized (upgraded to version 11), older versions of FTL will not be able to write new queries to the database. It is possible to revert the upgrade of the database without data-loss, however, it is a multiple-step task and should not be done copy-paste-like. That's why I'm not giving  the precise instructions to do this here.

**edit** We furthermore save some space on the `additional_information` field that con contain domains when blocking during deep CNAME inspection.


DL6ER:
Putting this back to draft mode as we might want to explore another link-table for `additional_info`

DL6ER:
Ready for review. The `additional_info` column is now also linked to a corresponding table giving us the flexibility to add multiple properties to a query in the future.

src/database/common.c
```diff
@@ -246,17 +246,6 @@ void db_init(void)
...
-	// SQLITE_DBCONFIG_DEFENSIVE
-	// Disbale language features that allow ordinary SQL to deliberately corrupt
-	// the database file. The disabled features include but are not limited to
-	// the following:
-	//
-	// - The PRAGMA writable_schema=ON statement.
-	// - The PRAGMA journal_mode=OFF statement.
-	// - Writes to the sqlite_dbpage virtual table.
-	// - Direct writes to shadow tables.
-	sqlite3_config(SQLITE_DBCONFIG_DEFENSIVE, true);
```
yubiuser, line 258: Why did you remove this? I see, that `PRAGMA writable_schema` is now set if needed and switched off afterwards. But what's about the other options set by the flag?

DL6ER: It is not possible to set `writable_schema` at all in defensive mode. It is neither possible to disable defensive mode when it was activated once. As we are in full control of all SQL ran against the database and users cannot inject malicious code anywhere, this is superfluous and we should just remove it.

DL6ER:
Branch updated from latest `development`

dschaper (approved):

src/database/common.c
```diff
@@ -419,6 +408,36 @@ void db_init(void)
...
+			logg("Queries table not optimized, database not available");
+			dbclose(&db);
+			return;
+		}
+		// Get updated version
+		dbversion = db_get_int(db, DB_VERSION);
+	}
+
+	// Update to version 11 if lower
+	if(dbversion < 11)
```
yubiuser, line 427: Do we need a two step upgrade here? It's all done in one PR so both could be done by upgrading to version 10.

src/database/gravity-db.c
```diff
@@ -155,15 +155,15 @@ bool gravityDB_open(void)
 	//            matches 'google.de' and all of its subdomains but
 	//            also other domains starting in google.de, like
 	//            abcgoogle.de
-	rc = sqlite3_prepare_v2(gravity_db,
+	rc = sqlite3_prepare_v3(gravity_db,
```
yubiuser, line 158: > sqlite3_prepare_v3() has an extra "prepFlags" option that is used for special purposes.

Do we need the extra prepFlag? Are we having special purposes?

yubiuser:
After switching to that branch

```
[2022-01-09 21:00:00.718 2828/T2832] SQLite3 message: cannot modify queries because it is a view in "DELETE FROM queries WHERE timestamp <= 1633982400" (1)
[2022-01-09 21:00:00.719 2828/T2832] ERROR: SQL query "DELETE FROM queries WHERE timestamp <= 1633982400" failed: SQL logic error
[2022-01-09 21:00:00.719 2828/T2832] delete_old_queries_in_DB(): Deleting queries due to age of entries failed
```

yubiuser:
This will need some documentation changes in https://docs.pi-hole.net/database/ftl/

E.g. `addinfo_by_id` meaning of `type`

src/database/common.c
DL6ER: Some already checked out version 10 which modified their database. If I'd have squashed the other updated into this, they would have ended in an unsupported database.

src/database/gravity-db.c
DL6ER: Yes, see the extra new flag `SQLITE_PREPARE_PERSISTENT`. Explantation can be found in the commit message of https://github.com/pi-hole/FTL/pull/1255/commits/d134a4943ae7e937792f8ee1c6f9147d9a9cfa3a

DL6ER:
@yubiuser Thanks for the error message, you indeed found the spot where the renaming `queries -> query_storage` is missing. I pushed an update.

yubiuser:
Error is fixed
`[2022-01-10 07:00:00.352 24148/T24152] Notice: Database size is 149.30 MB, deleted 2955 rows
`

yubiuser (approved):
Please remember the documentation. 

Thanks for that update.

Do we need any modification on `web`? I know it will use the `queries` view now automatically, but generating the view can take some time. We might not need all columns from there, e.g. wenn computing "Long term graph"

yubiuser:
Just an idea: now that you shrank the size per record tremendously, should we save the reply as well?

DL6ER:
I will use this PR as reminder by only merging when I had time to work on the documentation changes.

We are not requesting these columns through the VIEW, hence, they are also not looked up. For instance, take
```

```
as used here:

```

**Comment by DL6ER** [comment]
I will use this PR as reminder by only merging when I had time to work on the documentation changes.

> Do we need any modification on web? I know it will use the queries view now automatically, but generating the view can take some time. We might not need all columns from there, e.g. wenn computing "Long term graph"

We are not requesting these columns through the VIEW, hence, they are also not looked up. For instance, take https://github.com/pi-hole/AdminLTE/blob/682a174330d37c4bb8222ff371a3a4b7d363fafa/api_db.php#L264
``` sql
SELECT COUNT(timestamp) FROM queries;
```
compiles to the following database engine bytecode:
```
addr  opcode         p1    p2    p3    p4             p5  comment      
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     11    0                    00  Start at 11  
1     Null           0     1     2                    00  r[1..2]=NULL 
2     OpenRead       1     2     0     8              00  root=2 iDb=0; query_storage
3     Rewind         1     7     0                    00               
4       Column         1     1     3                    00  r[3]=query_storage.timestamp
5       AggStep        0     3     1     count(1)       01  accum=r[1] step(r[3])
6     Next           1     4     0                    01               
7     AggFinal       1     1     0     count(1)       00  accum=r[1] N=1
8     Copy           1     4     0                    00  r[4]=r[1]    
9     ResultRow      4     1     0                    00  output=r[4]  
10    Halt           0     0     0                    00               
11    Transaction    0     0     22    0              01  usesStmtJournal=0
12    Goto           0     1     0                    00
```

Querying `query_storage` does not change anything:
``` sql
SELECT COUNT(timestamp) FROM query_storage;
```
```
addr  opcode         p1    p2    p3    p4             p5  comment      
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     11    0                    00  Start at 11  
1     Null           0     1     2                    00  r[1..2]=NULL 
2     OpenRead       1     4     0     k(2,,)         00  root=4 iDb=0; idx_queries_timestamps
3     Rewind         1     7     3     0              00               
4       Column         1     0     3                    00  r[3]=query_storage.timestamp
5       AggStep        0     3     1     count(1)       01  accum=r[1] step(r[3])
6     Next           1     4     0                    01               
7     AggFinal       1     1     0     count(1)       00  accum=r[1] N=1
8     Copy           1     4     0                    00  r[4]=r[1]    
9     ResultRow      4     1     0                    00  output=r[4]  
10    Halt           0     0     0                    00               
11    Transaction    0     0     22    0              01  usesStmtJournal=0
12    Goto           0     1     0                    00
```
as the former already uses the real storage. The same holds true for all other queries.

> Just an idea: now that you shrank the size per record tremendously, should we save the reply as well?

For v6.0, I already implemented storing everything visible on the Query Log (also reply *time*, and DNSSEC status). We can do this already now, that's a good idea. Something for a follow-up PR.

PR #1259, opened 2021-12-23
Fix locale issue in Telnet API

DL6ER:

## 10

---

The recent `dnsmasq` update included a commit (https://github.com/pi-hole/FTL/commit/95473e7696bd9408a413c0a17123dfc21321edab) that forced the entire `pihole-FTL` process to adhere to the system-wide defined locale. In the majority of non-English locales (German, French, Italian, and Spanish, ...), a comma is used as decimal separator instead of a dot.

This caused troubles in the PHP API scripts that are not aware of the system-wide locale as the comma is not interpreted as decimal separator leading to malfunctions of the statistics on the dashboard.

This PR fixes this by ensuring that the the numeric part of the locale, responsible for things such as decimal separators, is set to `C` which is the standard setting and was used by FTL before the commit mentioned above.

### DIFF
--- a/src/dnsmasq/dnsmasq.c
+++ b/src/dnsmasq/dnsmasq.c
@@ -80,6 +80,9 @@ int main_dnsmasq (int argc, char **argv)
 
 #if defined(HAVE_IDN) || defined(HAVE_LIBIDN2) || defined(LOCALEDIR)
   setlocale(LC_ALL, "");
+  /*** Pi-hole modification ***/
+  setlocale(LC_NUMERIC, "C");
+  /****************************/
 #endif
 #ifdef LOCALEDIR
   bindtextdomain("dnsmasq", LOCALEDIR); 


yubiuser (approved):

DL6ER:
Confirmed working here: https://discourse.pi-hole.net/t/nach-update-keine-counter-mehr/51812/11

I force-pushed an easier fix that was confirmed working there as well.

dschaper (approved):

yubiuser (approved):

r2ixuz:
I think at least for Docker 2021.12.1 it still doesn't look correctly:

screenshot

If I switch back to 2021.12 it looks still the same. The notation with a dot would be pretty for germans ;-).

yubiuser:
@r2ixuz

The bug is fixed, before you had no output in those counters at all.
But you are right, Pi-hole does not localize the separators so far.

See here for more information 
https://discourse.pi-hole.net/t/change-decimal-and-thousands-separators-in-dashboard-according-to-local-settings/31168/10

DL6ER:
What you observe is expected as this PR intentionally *removes* any internationalization from the number printing. Commas as thousands separator and periods as decimal separator are the international standard and we use them to avoid interoperability issues between the involved parties: FTL, PHP and Javascript.

Having said that, the currently still heavily under development v6.0 version of Pi-hole actually improves on this by offloading all the number formatting into your browser. If you browser is honest about the preferred language (not all are, some consider this a privacy leak...), you should get locale-aware separators in the not so distant future.

DL6ER:
@r2ixuz On the other hand ... why wait so long if the code is already written. See https://github.com/pi-hole/AdminLTE/pull/2029

PR #1274, opened 2022-01-07
Return zero domains if no gravity domains are available

yubiuser:

3

If no domains are available on the gravity table the gravity count will error out leaving `counters->gravity` undefined. This in turn will be displayed at the web interface as `-2`.
This PR explicitly returns zero if there were no domains on the gravity table.
See: https://github.com/pi-hole/pi-hole/issues/4519

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -1127,6 +1127,7 @@ int gravityDB_count(const enum gravity_tables list)
 			logg("Count of gravity domains not available. Please run pihole -g");
 			// set the gravity count to zero if no gravity domains are available
 			const int result = 0;
+			return result;
 		}
 		gravityDB_finalizeTable();
 		gravityDB_close();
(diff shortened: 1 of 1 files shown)


src/database/gravity-db.c
```diff
@@ -1125,6 +1125,10 @@ int gravityDB_count(const enum gravity_tables list)
 		if(list == GRAVITY_TABLE)
 		{
 			logg("Count of gravity domains not available. Please run pihole -g");
+			// set the gravity count to zero if no gravity domains are available
+			const int result = 0;
+			logg("Setting gravity count to zero");
+			return result;
```
DL6ER, line 1131: We cannot return here. If we do without calling the `finalizeTable` and `close` routines, we leak notable amount of memory.

I'd propose changing the `if` above like
``` diff
+if(rc == SQLITE_DONE)
+{
+		gravityDB_finalizeTable();
+		gravityDB_close();
+		return 0;
+}
-	if(rc != SQLITE_ROW){
+	else if(rc != SQLITE_ROW){
```
to check for the `DONE` status as this is what should be there when there is no row available but also no error.

(code above is fiction, I haven't tested it)

DL6ER: Alternative is to ensure `gravity_count` cannot be absent. This would likely fix the core issue as well.

yubiuser: ... yeah you're right. The proper solution has to take place in core.

PR #1281, opened 2022-01-12
Update embedded dnsmasq to v2.87test8

DL6ER:

## 10

---

Highlights:
- DNSSEC improvements:
   - When searching for an in-flight DNSSEC query to use (rather than starting a new one), compare the already sent query, rather than using the hash of the query. This is probably faster (no hash calculation) and eliminates having to worry about the consequences of a hash collision.
   - Check for dependency loops in DNSSEC validation, say validating A requires DS B and validating DS B requires DNSKEY C and validating DNSKEY C requires DS B. This should never happen in correctly signed records, but it's likely the case that sufficiently broken ones can cause our validation code requests to exhibit cycles. The result is that the ->blocking_query list can form a cycle, and under certain circumstances that can lock us in an infinite loop. Instead we transform the situation into an ABANDONED state.
   - Fix DNSSEC failure to validate unsigned NoDATA replies.  A reply with an empty answer section would not always be checked for either suitable NSEC records or proof of non-existence of the relevant DS record.
- Log port numbers in server addresses when non-standard ports in use (**Pi-hole contribution**)
- Fix header of cache dump (**Pi-hole contribution**)
- Extend cache dump: "!" as type for non-terminals, new flag "C" for config-provided and log source when applicable. (**Pi-hole contribution**)
- Log source of ignored query when local-service is used. (**Pi-hole contribution**)
- Minimum safe size is recommended to be 1232. (**Pi-hole contribution**)
- Extend packet dump system to RA, DHCP and TFTP
- Handle changing interface indexes when binding DHCP sockets.
- Add `--conf-script` to generate dnsmasq config using a script rather than static files

The warning
- `Ignoring query from non-local network`
has been changed to
- `ignoring query from non-local network <ADDRESS>`


regae:
with this pull ipset is not working properly, not adding a new ip.

yubiuser (approved):
Needs documentation change in https://docs.pi-hole.net/ftldns/dnsmasq_warn/

PR #1284, opened 2022-01-19
Add others section to upstream servers pie chart

DL6ER:

## 10

---

There are queries which are neither sent upstream nor are they cached or blocked. These are incorrectly ignored in the upstream server pie chart leading to a mismatch of the blocked percentage compared to the value reported at the top. Even when both are correct (their difference is caused by different reference frames), this is confusing. To compensate, we add a new type `other` which summarizes all these queries, causing the two reference frames to be identical.

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -542,7 +542,7 @@ void getUpstreamDestinations(const char *client_message, const int *sock)
 			name = ip;
 
 			if(counters->queries > 0)
-				// Whats the percentage of locked queries on the total amount of queries?
+				// Whats the percentage of blocked queries on the total amount of queries?
 				percentage = 1e2f * blocked_queries() / counters->queries;
 		}
 		else if(i == -2)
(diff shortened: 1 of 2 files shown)


yubiuser (commented):
I think we need to add something here:
https://github.com/pi-hole/FTL/blob/7387e81981760c835148b7f5f61302b91ee83e7d/src/api/api.c#L723-L726

yubiuser:
If adding "-3" here: I promised **Promo** https://github.com/pi-hole/AdminLTE/pull/1594

DL6ER:
This shouldn't make any difference, the order is not changed. `Other` will use the color the first upstream had before, the first upstream gets the color for the former seconds, etc.

rdwebdesign:
@yubiuser 

The code you highlighted is just to ensure that "BL" will be the first item (with index -2) and the second item will be cache (with index -1).

The rest remains in their original positions (from 0 forward).

yubiuser (approved):

src/api/api.c
```diff
@@ -525,34 +525,45 @@ void getUpstreamDestinations(const char *client_message, const int *sock)
...
-		if(i == -2)
+		if(i == -3)
 		{
 			// Blocked queries (local lists)
 			ip = "blocklist";
 			name = ip;
 
-			if(totalqueries > 0)
+			if(counters->queries > 0)
 				// Whats the percentage of locked queries on the total amount of queries?
```
yubiuser, line 545:
```suggestion
				// Whats the percentage of blocked queries on the total amount of queries?
```

yubiuser (approved):

PR #1285, opened 2022-01-19
Keep Query Log details across FTL restart

DL6ER:

## 10

---

When restarting FTL, it imports history from the database to repopulate the history. So far, this was only possible with a reduced set of variables to keep the database from growing forever. Recent database optimizations https://github.com/pi-hole/FTL/pull/1255 reduced the database size significantly allowing us to store more data in the database.

This PR adds storing all the details necessary for the query log so a restart goes (almost) invisible and without any data loss.

*) almost because some queries may not have been completed due to the forced restart and according service interruption.

### DIFF
--- a/src/database/query-table.c
+++ b/src/database/query-table.c
@@ -812,10 +812,10 @@ void DB_read_queries(void)
 		}
 
 		int reply_type = REPLY_UNKNOWN;
-		if(sqlite3_column_type(stmt, 7) == SQLITE_INTEGER)
+		if(sqlite3_column_type(stmt, 8) == SQLITE_INTEGER)
 		{
 			// The field has been added for database version 12
-			reply_type = sqlite3_column_int(stmt, 7);
+			reply_type = sqlite3_column_int(stmt, 8);
 			if(reply_type < REPLY_UNKNOWN || reply_type >= QUERY_REPLY_MAX)
 			{
 				logg("DB warn: REPLY value %i is invalid, %lli", reply_type, (long long)queryTimeStamp);
@@ -824,10 +824,12 @@ void DB_read_queries(void)
 		}
 
 		double reply_time = 0.0;
-		if(sqlite3_column_type(stmt, 8) == SQLITE_FLOAT)
+		bool reply_time_avail = false;
+		if(sqlite3_column_type(stmt, 9) == SQLITE_FLOAT)
 		{
 			// The field has been added for database version 12
-			reply_time = sqlite3_column_double(stmt, 8);
+			reply_time = sqlite3_column_double(stmt, 9);
+			reply_time_avail = true;
 			if(reply_time < 0.0)
 			{
 				logg("DB warn: REPLY_TIME value %f is invalid, %lli", reply_time, (long long)queryTimeStamp);
@@ -836,10 +838,10 @@ void DB_read_queries(void)
 		}
 
 		int dnssec = DNSSEC_UNSPECIFIED;
-		if(sqlite3_column_type(stmt, 9) == SQLITE_INTEGER)
+		if(sqlite3_column_type(stmt, 10) == SQLITE_INTEGER)
 		{
 			// The field has been added for database version 12
-			dnssec = sqlite3_column_int(stmt, 9);
+			dnssec = sqlite3_column_int(stmt, 10);
 			if(dnssec < DNSSEC_UNSPECIFIED || dnssec >= DNSSEC_ABANDONED)
 			{
 				logg("DB warn: DNSSEC value %i is invalid, %lli", dnssec, (long long)queryTimeStamp);
@@ -877,7 +879,7 @@ void DB_read_queries(void)
 		query->upstreamID = upstreamID;
 		query->id = 0;
 		query->response = 0;
-		query->flags.response_calculated = false;
+		query->flags.response_calculated = reply_time_avail;
 		query->dnssec = dnssec;
 		query->reply = reply_type;
 		query->response = reply_time * 1e4; // convert to tenth-millisecond unit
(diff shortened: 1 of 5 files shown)


test/test_suite.bats
```diff
@@ -587,16 +587,16 @@
...
   [[ "${lines[@]}" == *"CREATE INDEX idx_queries_timestamps ON \"query_storage\" (timestamp);"* ]]
   [[ "${lines[@]}" == *"CREATE TABLE ftl (id INTEGER PRIMARY KEY NOT NULL, value BLOB NOT NULL);"* ]]
   [[ "${lines[@]}" == *"CREATE TABLE counters (id INTEGER PRIMARY KEY NOT NULL, value INTEGER NOT NULL);"* ]]
   [[ "${lines[@]}" == *"CREATE TABLE IF NOT EXISTS \"network\" (id INTEGER PRIMARY KEY NOT NULL, hwaddr TEXT UNIQUE NOT NULL, interface TEXT NOT NULL, firstSeen INTEGER NOT NULL, lastQuery INTEGER NOT NULL, numQueries INTEGER NOT NULL, macVendor TEXT, aliasclient_id INTEGER);"* ]]
   [[ "${lines[@]}" == *"CREATE TABLE IF NOT EXISTS \"network_addresses\" (network_id INTEGER NOT NULL, ip TEXT UNIQUE NOT NULL, lastSeen INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), name TEXT, nameUpdated INTEGER, FOREIGN KEY(network_id) REFERENCES network(id));"* ]]
   [[ "${lines[@]}" == *"CREATE TABLE aliasclient (id INTEGER PRIMARY KEY NOT NULL, name TEXT NOT NULL, comment TEXT);"* ]]
-  [[ "${lines[@]}" == *"INSERT INTO ftl VALUES(0,11);"* ]] # Expecting FTL database version 11
+  [[ "${lines[@]}" == *"INSERT INTO ftl VALUES(0,12);"* ]] # Expecting FTL database version 11
```
yubiuser, line 597:
```suggestion
  [[ "${lines[@]}" == *"INSERT INTO ftl VALUES(0,12);"* ]] # Expecting FTL database version 12
```

yubiuser:
I tried it but it did not really work for me. I got a lot of

```
[2022-01-31 20:07:14.640 24751M] DB warn: REPLY value 52 is invalid, 1643656012
[2022-01-31 20:07:14.640 24751M] DB warn: REPLY value 116 is invalid, 1643656013
[2022-01-31 20:07:14.640 24751M] DB warn: REPLY value 116 is invalid, 1643656013
[2022-01-31 20:07:14.640 24751M] DB warn: REPLY value 52 is invalid, 1643656017

```
But there is no `reply_type` with 52 or 116 in the database.

I also queried a test domain which was successfully stored in the database with all information, including reply_type 1. After a restart, the query log displays `Reply` `N/A`

src/database/query-table.c
```diff
@@ -767,6 +811,42 @@ void DB_read_queries(void)
 			upstreamID = findUpstreamID(serv_addr, (in_port_t)serv_port);
 		}
 
+		int reply_type = REPLY_UNKNOWN;
+		if(sqlite3_column_type(stmt, 7) == SQLITE_INTEGER)
+		{
+			// The field has been added for database version 12
+			reply_type = sqlite3_column_int(stmt, 7);
+			if(reply_type < REPLY_UNKNOWN || reply_type >= QUERY_REPLY_MAX)
```
yubiuser, line 819: According to https://github.com/pi-hole/docs/pull/655
Max reply_type is 13 (a.k.a `blob`). Where does `QUERY_REPLY_MAX` come from?

src/database/query-table.c
```diff
@@ -767,6 +811,42 @@ void DB_read_queries(void)
...
+			// The field has been added for database version 12
+			reply_time = sqlite3_column_double(stmt, 8);
+			if(reply_time < 0.0)
+			{
+				logg("DB warn: REPLY_TIME value %f is invalid, %lli", reply_time, (long long)queryTimeStamp);
+				continue;
+			}
+		}
+
+		int dnssec = DNSSEC_UNSPECIFIED;
```
yubiuser, line 838: `unknown`? (As default to zero).
https://github.com/pi-hole/docs/pull/655

src/database/query-table.c
DL6ER: It is the maximum value that is currently supported (so one higher than `BLOB`). This ensures this code stays valid also in the future when new reply types are added.

DL6ER:
Thanks for your review, not sure how I missed. We also have to set the `response_time_calculated` property to `true` or the reply will not be shown on the dashboard. The API still presented the data but the Javascript skipped it as it thought: When we don't have a response time, we cannot have a reply type either (as there is not response so far). Should be fixed.

src/database/query-table.c
```diff
@@ -662,7 +706,7 @@ void DB_read_queries(void)
 	// Get time stamp 24 hours in the past
 	const time_t now = time(NULL);
 	const time_t mintime = now - config.maxlogage;
-	const char *querystr = "SELECT * FROM queries WHERE timestamp >= ?";
+	const char *querystr = "SELECT id,timestamp,type,status,domain,client,forward,additional_info,reply_type,reply_time,dnssec FROM queries WHERE timestamp >= ?";
```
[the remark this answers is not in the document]
DL6ER: In case the database is further extended in the future, this ensures we only source those fields we know. Also, this ensures we get the on the order we expect them. Until now, we just relied on the database reporting them in the order of their definition (which seems logic, too). It is no extra work for SQLite3 only the initial byte compilation process is a bit longer but this is done only once over the entire lifetime of FTL.

yubiuser:
Thanks for the update.

There are no `REPLY value 52 is invalid` warnings any more. Also the  `reply time` and `reply type` dashboard issue is fixed.

yubiuser (approved):

PR #1288, opened 2022-01-23
Add a Dev container

PromoFaux:

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [x] I have made only one major change in my proposed changes.
- [ ] I have commented my proposed changes within the code.
- [x] I have tested my proposed changes, and have included unit tests where possible.
- [x] I am willing to help maintain this change if there are issues with it later.
- [x] I give this submission freely and claim no ownership.
- [x] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [x] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---

Adds a devcontainer, along with some suggested extensions for vscode - this should make it easier to build and test for those that, for example, are nervous about setting up the build and test dependencies - this "just works"

As its kind of related, I have also reworked the build script a little to take arguments at any position (except the -D* ones as I don't really know what to do about them)

I am not sure why the compiled binaries are not uploading here - I have rebuilt several times. I am not sure it's the build script changes as the commit before failed, too.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,10 @@
+{
+  "name": "FTL x86_64 Build Env",  
+  "image": "ghcr.io/pi-hole/ftl-build:x86_64", 
+  "extensions": [
+    "jetmartin.bats",
+    "ms-vscode.cpptools",
+    "ms-vscode.cmake-tools",
+    "eamodio.gitlens"
+  ],  
+}
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ tools/macvendor.db
 
 # Test dependencies
 /node_modules/
+
+# Test Leftovers
+dig.log
+ptr.log


build.sh
```diff
@@ -34,8 +47,17 @@ cmake --build . -- -j $(nproc)
...
+    ${SUDO} make install
 else
+    echo "Copying compiled pihole-FTL binary to repository root"
     cp pihole-FTL ../
 fi
+
+if [[ -n "${test}" ]]; then
+    cd ..
+    ./test/run.sh
+    cd cmake
```
DL6ER, line 62: Why this final `cd`?

build.sh
```diff
@@ -12,10 +12,23 @@
...
+done
+
 # Prepare build environment
-if [[ "${1}" == "clean" ]]; then
+if [[ -n "${clean}" ]]; then
+    echo "Cleaning build environment"
     rm -rf cmake/
-    exit 0
+    if [[ -n ${nobuild} ]]; then
+        exit 1
```
DL6ER, line 30: This reports an error when the script actually did what is was asked to do, is this intentional?

PromoFaux: Ah no, typo.. was supposed to be `exit 0`
```suggestion
        exit 0
```

build.sh
PromoFaux:
```suggestion
```

Typo again, initially I had the test section before the install/copy section

PromoFaux:
I'll split the build container/test script stuff out to separate PRs. I'm also going to need to do some stuff with the build containers, but that will need to come _after_ the CI tweaks go into `development` (and then `special/CI_development`)

dschaper (approved):

PR #1296, opened 2022-02-02
Do not warn about less-than-full addresses when analyzing EDNS(0) packets

DL6ER:

## 10

---

See title. This error is incorrectly shown when a client is sending *sparse* ECS data, e.g., `add-subnet=1.2.3.0/24` instead of `add-subnet=1.2.3.4/32`. The important bit here is the less-than-32 bits. Likewise for IPv6.


DL6ER:
Confirmed working by OP [here](https://discourse.pi-hole.net/t/edns-with-24-subnet/53238/9?u=dl6er)

yubiuser (approved):

PR #1297, opened 2022-02-05
Prevent excessive reply times in the Query Log after restarting

DL6ER:

## 10

---

The `query->reply_time` variable serves two purposes: (1) When sending a query upstream, we store the current time in this field. (2) When the reply arrives, we compute the difference and store it in the same field. We memorize the transition from (1) to (2) by setting the flag `query->flags.response_calculated` to `true`.

Currently (broken in `development`, not in `master`), we store the reply_time field unconditionally, hence, for a few queries which are still being processed upstream at restart, we store the real time instead of the response. FTL incorrectly interprets the (homogeneous!) timestamp as difference and telly you the query needed quadrillions of milliseconds to get answered.

We fix this small bug by instead storing `NULL` in the `REPLY_TIME` column when the reply time is not available. FTL already handles this case well when loading from the database: the default of `query->flags.response_calculated` is `false`. A `NULL` value in this field does not change this default.

### DIFF
--- a/src/database/query-table.c
+++ b/src/database/query-table.c
@@ -431,8 +431,11 @@ int DB_save_queries(sqlite3 *db)
 		// REPLY_TYPE
 		sqlite3_bind_int(query_stmt, 10, query->reply);
 
-		// REPLY_TIME (stored in units of seconds)
-		sqlite3_bind_double(query_stmt, 11, 1e-4*query->response);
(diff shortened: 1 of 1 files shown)


DL6ER:
No documentation change is required. The current documentation already says that the field is allowed to by "empty" which means `NULL` for SQL with column type `REAL`.

yubiuser (approved):

PR #1310, opened 2022-03-01
Reload blocking mode independently of incoming queries

DL6ER:

## 10

---

So far, FTL defers reading the blocking mode until the new query arrived, however, on very quiet systems, there can be a notable gap in between changing the status and FTL picking it up on the next query. This optimization has the side-effect of reporting an incorrect blocking status after changing the status until the next query arrives. This PR changes lazy loading of the blocking status (enabled/disabled) to instantaneous.

The constant `RELOAD_BLOCKINGMODE` is renamed to `RELOAD_BLOCKINGSTATUS` because this is what it actually does.


src/database/database-thread.c
```diff
@@ -116,8 +118,14 @@ void *DB_thread(void *val)
...
-		// Sleep 1 sec
-		thread_sleepms(DB, 1000);
+		// Inspect setupVars.conf to see if Pi-hole blocking is enabled
+		if(get_and_clear_event(RELOAD_BLOCKINGSTATUS))
+			check_blocking_status();
+
+		BREAK_IF_KILLED();
+
+		// Sleep 0.1 sec
+		thread_sleepms(DB, 100);
```
yubiuser, line 128: Why do we need a sleep here?

src/dnsmasq_interface.c
```diff
@@ -1775,7 +1769,7 @@ void FTL_dnsmasq_reload(void)
 
 	// Request reload the privacy level and blocking mode
 	set_event(RELOAD_PRIVACY_LEVEL);
-	set_event(RELOAD_BLOCKINGMODE);
+	set_event(RELOAD_BLOCKINGSTATUS);
```
yubiuser, line 1772: L1770 comment 
 Request reload the privacy level and blocking **status**

src/database/database-thread.c
DL6ER: There was a sleep here before, too (see deleted line 120). Without a sleep, this thread will start spinning at 100% checking the flags millions of times every second. No need to waste CPU resources on this bit so we say it's safe for the thread to sleep for 100 msec until it should recheck all the flags again.

yubiuser (approved):

PR #1312, opened 2022-03-06
Fix initOverTime's calculation of last slot timestamp

icyflame:

---

initOverTime calculate's the last slot timestamp with the assumption that `OVERTIME_INTERVAL` will
allow atleast `OVERTIME_SLOTS` number of slots between now and the last slot in the current
hour. This leads it to calculate the timestamp of the last slot, and then assign timestamps to each
slot counting down from that timestamp.

This logic is not compatible with changes in the OVERTIME_INTERVAL and the MAXLOGAGE variables. Here is a demonstration of a case where this can be shown clearly:

```
OVERTIME_INTERVAL = 60s (Each slot stores 1 minute of data)
MAXLOGAGE = 600s (Store date for only 10 minutes)
OVERTIME_SLOTS = 11 (MAXLOGAGE/OVERTIME_INTERVAL)

GCInterval = 90 (Any value less than MAXLOGAGE is okay)
```

**Current logic:**

- Current time: 10:35:04
- First slot time: 10:48:30
- Last slot time: 10:59:30

**This PR's proposal:**

- Current time: 10:35:04
- First slot time: 10:35:30
- Last slot time: 10:46:30

Note that this logic *does* work as expected with the default minimum of 1 hour for the MAXLOGAGE
configuration variable.

However, I believe that there is no need to make this part of the code dependent on this minimum. If this part of the code is changed, the macros for OVERTIME_INTERVAL and MAXLOGAGE can be freely changed by users.

### DIFF
--- a/src/overTime.c
+++ b/src/overTime.c
@@ -70,23 +70,22 @@ void initOverTime(void)
 	// Get current timestamp
 	time_t now = time(NULL);
 
-	// The last timestamp (overTime[149]) should be the last interval of this hour
-	// If the current time is 09:35, the last interval is 09:50 - 10:00 (centered at 09:55)
-	time_t timestamp = now - now % 3600 + 3600 - (OVERTIME_INTERVAL / 2);
+	// Get the centered timestamp of the current timestamp
+	time_t first_slot_ts = now - now % OVERTIME_INTERVAL + (OVERTIME_INTERVAL / 2);
+	time_t last_slot_ts = first_slot_ts + (OVERTIME_SLOTS-1) * OVERTIME_INTERVAL;
 
 	if(config.debug & DEBUG_OVERTIME)
 		logg("initOverTime(): Initializing %i slots from %llu to %llu",
 		     OVERTIME_SLOTS,
-		     (long long)timestamp-OVERTIME_SLOTS*OVERTIME_INTERVAL,
-		     (long long)timestamp);
+		     (long long)first_slot_ts,
+		     (long long)last_slot_ts);
 
 	// Iterate over overTime
-	for(int i = OVERTIME_SLOTS-1; i >= 0 ; i--)
+	for(int i = 0; i < OVERTIME_SLOTS; i++)
 	{
-		// Initialize onerTime slot
-		initSlot(i, timestamp);
-		// Prepare for next iteration
-		timestamp -= OVERTIME_INTERVAL;
+		time_t this_slot_ts = first_slot_ts + OVERTIME_INTERVAL * i;
+		// Initialize overTime slot
+		initSlot(i, this_slot_ts);
 	}
 }
 


DL6ER:
Thank you for your contribution and sorry for the delay in reviewing. Things are moving a lot slower with the attention being closer to a war that is going on only a few hours by car from here. I'll come back to you PR and do a proper review when time allows.

src/overTime.c
```diff
@@ -70,23 +70,22 @@ void initOverTime(void)
...
+	time_t first_slot_ts = now - now % OVERTIME_INTERVAL + (OVERTIME_INTERVAL / 2);
+	time_t last_slot_ts = first_slot_ts + (OVERTIME_SLOTS-1) * OVERTIME_INTERVAL;
 
 	if(config.debug & DEBUG_OVERTIME)
 		logg("initOverTime(): Initializing %i slots from %llu to %llu",
 		     OVERTIME_SLOTS,
-		     (long long)timestamp-OVERTIME_SLOTS*OVERTIME_INTERVAL,
-		     (long long)timestamp);
+		     (long long)first_slot_ts,
+			 (long long)last_slot_ts);
```
DL6ER, line 81:
```suggestion
		     (long long)last_slot_ts);
```

DL6ER (commented):
Thanks for your contribution. I only have a very small comment concerning indentation of a line you changed and then we're good to g on this one.

icyflame:
Thank you for the review, I fixed the indentation, and pushed again with GPG verification for each commit.

For piHole 6.0's development / plan / roadmap / timeline, I was wondering where I could follow the development. Is it everything on the `development` branch of this repository? I can't find anything related to the next version on [Discourse][1]. A search for `6.0` does not yield anything (https://discourse.pi-hole.net/search?q=6.0)

[1]: https://discourse.pi-hole.net/

yubiuser:
>  Is it everything on the development branch of this repository

It's in `new/http`. To get a glance at the API use https://ftl.pi-hole.net/new/http/docs/#overview or after you checked out that branch `http://pi.hole:8080/api/docs/

DL6ER (approved):

src/overTime.c
```diff
@@ -70,23 +70,22 @@ void initOverTime(void)
 	// Get current timestamp
 	time_t now = time(NULL);
 
-	// The last timestamp (overTime[149]) should be the last interval of this hour
-	// If the current time is 09:35, the last interval is 09:50 - 10:00 (centered at 09:55)
-	time_t timestamp = now - now % 3600 + 3600 - (OVERTIME_INTERVAL / 2);
+	// Get the centered timestamp of the current timestamp
+	time_t first_slot_ts = now - now % OVERTIME_INTERVAL + (OVERTIME_INTERVAL / 2);
+	time_t last_slot_ts = first_slot_ts + (OVERTIME_SLOTS-1) * OVERTIME_INTERVAL;
```
DL6ER, line 75: The logic in this PR is broken, unfortunately, I didn't see this before. We'll revert the PR for now. The idea can be re-applied later when we have a solution and proper time for testing.

The issue here is, basically, that `first_slot_ts` should be the one 24 hours in the past, however, it is a timestamp centered at the current timestamp (as the comment correctly says). The `last_slot_ts` is then 24 hours *in the future* which is wrong.

This wasn't an issue in the code before the change as the old code was going *backwards* from now. The new code tries to go forward but start at *now* causing the history to incorrectly love into the future rather than into the past.

PR #1313, opened 2022-03-06
 Replace reference to hard-coded default GCinterval value

icyflame:

---

This hard-coded GCinterval value makes any change to the macro defining the GCinterval ineffective
when the GC runs.

For e.g., if I reduce the GC from once an hour to once every 3 minutes, the GC runs every 3 minutes,
but the mintime is set to the start of the next hour, which wipes everything from the overTime data.

I believe that this change is safe and can be merged and shipped to existing users because the
GCinterval value is not configurable and users who don't build the PiHole version that they are
running will not be affected by this change.

Signed-off-by: Siddharth Kannan <mail@siddharthkannan.in>

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -156,10 +156,10 @@ void *GC_thread(void *val)
 			// Get minimum timestamp to keep (this can be set with MAXLOGAGE)
 			time_t mintime = (now - GCdelay) - config.maxlogage;
 
-			// Align to the start of the next hour. This will also align with
-			// the oldest overTime interval after GC is done.
-			mintime -= mintime % 3600;
-			mintime += 3600;
+			// Align the start time of this GC run to the GCinterval. This will also align with the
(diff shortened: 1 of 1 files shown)


DL6ER:
The same comment as in #1312 applies here. Most of this was not really designed to run at any custom intervals and the lack of complains (so far) seems to have proven that basically nobody even tried it. In the ongoing Pi-hole v6.0 development, we're looking at using an in-memory SQLite3 database instead of relying heavily on our own datastructure to gain some more flexibility. This should also have taken care of these two bugs.

DL6ER (approved):

DL6ER:
> The base branch requires all commits to be signed. [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

Do you want to do this? If not, we can add an exception for your commits.

DL6ER:
Merged unsigned commit using admin privileges.

PR #1328, opened 2022-04-18
Remove PID and PORT file during shutdown

DL6ER:

## 10

---

This PR tries to remove the PID file during shutdown. We still empty the file above to ensure it is at least empty when it cannot be removed. because removing files on Linux is actually `unlink`ing them. If any processes still have the file open, it will remain in existence until the last file descriptor referring to it is closed.

### DIFF
--- a/src/api/socket.c
+++ b/src/api/socket.c
@@ -54,8 +54,18 @@ void saveport(int port)
 	}
 	else
 	{
-		// FTL is terminating: Leave file truncated
+		// FTL is terminating: Truncate file and remove it (if possible)
 		fclose(f);
+		// We also try to remove the file. We still empty the file above
+		// to ensure it is at least empty when it cannot be removed.
+		// because removing files on Linux is actually unlinking them.
+		// If any processes still have the file open, it will remain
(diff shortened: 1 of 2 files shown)


yubiuser:
There is another file `FTL` is writing: `pihole-FTL.port`. Should FTL remove this file as well?

DL6ER:
Is `pihole-FTL.port` still use anywhere?

yubiuser:
Yes. We allow to configure the file location:

https://docs.pi-hole.net/ftldns/configfile/#file_PORTFILE

And get the port from that file for `core` and `web`
https://github.com/pi-hole/pi-hole/blob/2735481da8fac3cb521338b7af2259da7b621e88/advanced/Scripts/utils.sh#L76

https://github.com/pi-hole/AdminLTE/blob/6c320a42574f9f0944ded71d84b7331557dbc6ea/scripts/pi-hole/php/FTL.php#L29

DL6ER:
@yubiuser Sorry, I forgot about this. I added removing the `pihole-FTL.port` file now as well.

yubiuser (approved):

PR #1329, opened 2022-04-18
Check struct sizes on the CI instead of during compile-time

DL6ER:

## 10

---

Check struct sizes on the CI (in code) instead of during compile-time (using the preprocessor). This should resolve an issue reported on [Discourse](https://discourse.pi-hole.net/t/static-assert-h-error-compiling-from-source-alpine-linux/54915).

It adds an (undocumented) new flag `pihole-FTL --check-structs` that is supposed to be used by the tests. The exit code will tell if everything worked as expected (`bash`-compatible: exit code 0 is success).


dschaper:
Thanks!

src/config.h
```diff
@@ -17,10 +17,12 @@
...
 // typedef uni32_t
 #include <stdint.h>
-// assert_sizeof
-#include "static_assert.h"
 // struct in_addr, in6_addr
 #include <netinet/in.h>
+// type bool
+#include <stdbool.h>
+// type FILE
+#include <stdio.h>
```
yubiuser, line 25: Why are those two `include` needed now?

src/datastructure.h
```diff
@@ -15,8 +15,9 @@
 
 // enum privacy_level
 #include "enums.h"
-// assert_sizeof
-#include "static_assert.h"
+
+// Definitions like OVERTIME_SLOT
+#include "FTL.h"
```
yubiuser, line 20: Why is this needed now?

src/config.h
DL6ER: We remove the `assert.h` dependency which causes some automatic "behind-the-scenes" includes to not happen. Being more explicit what we want seems a better idea in general.

src/datastructure.h
DL6ER: See https://github.com/pi-hole/FTL/pull/1329#discussion_r857076488

DL6ER:
It's worth pointing out that we will only test struct sizes on platform where we run the tests for (`x86`, `x86-musl`, `i386`). However, this is *not* a drawback as we are checking on both 32bit and a 64bit architectures. (also note that there is *no* difference between `arm`, `mips` and `i386` so everything is still covered).

yubiuser (approved):

PR #1332, opened 2022-04-20
5.15 Release

PromoFaux:
Release PR


PromoFaux:
We can ignore the `deploy` fail in this case - it's because the `release` dir does not exist on the remote server - we don't need the binaries there anyway

dschaper (approved):

PR #1341, opened 2022-04-25
Make REPLY_WHEN_BUSY=DROP the new default (was ALLOW)

DL6ER:

## 10

---

The current default is in conflict with the description in https://github.com/pi-hole/FTL/pull/1156 which says we use `DROP` as the default. This is changed here.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -688,13 +688,13 @@ void read_FTLconf(void)
 
 	// REPLY_WHEN_BUSY
 	// How should FTL handle queries when the gravity database is not available?
-	// defaults to: BLOCK
+	// defaults to: DROP
 	buffer = parse_FTLconf(fp, "REPLY_WHEN_BUSY");
 
-	if(buffer != NULL && strcasecmp(buffer, "DROP") == 0)
+	if(buffer != NULL && strcasecmp(buffer, "ACCEPT") == 0)
 	{
-		config.reply_when_busy = BUSY_DROP;
-		logg("   REPLY_WHEN_BUSY: Drop queries when the database is busy");
+		config.reply_when_busy = BUSY_ALLOW;
+		logg("   REPLY_WHEN_BUSY: Permit queries when the database is busy");
 	}
 	else if(buffer != NULL && strcasecmp(buffer, "REFUSE") == 0)
 	{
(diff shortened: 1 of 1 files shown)


dschaper:
If the database is locked because of a broken install will this change mean all replies are dropped? (And badly behaving clients then slam the hell out of FTL with retries?)

Not that we can do much about bad clients but would like to know for troubleshooting purposes if we see retry floods.

yubiuser (approved):

DL6ER:
@dschaper You are right. Maybe this is the argument *against* this change? And maybe even why it is like it is right now.

dschaper:
I agree with the change, my only concern is how to handle the queries in that situation. I think a rejection or SRVERR if possible?

yubiuser:
I think this is a target conflict: we can not set the default to drop to preserve privacy when the database is busy and at the same time prevent dropping all queries in case we do have a database malfunction. 
If DB is busy all the time and users loose internet connection they certainly will ask for help. In their `pihole-FTL.log` will be hundreds of `REPLY_WHEN_BUSY: Drop queries when the database is busy` which should give us a good troubleshooting guidance what the culprit is.

DL6ER:
> hundreds of `REPLY_WHEN_BUSY:` [...]

Yeah, until the SD card is full (can definitely happen in less than a day). Also, we don't know if they can even ask for help without working DNS.

Something tells me this solution here is too simple and will cause too many issues - only for very few users but they'll be affected heavily.

Say we could reliability tell apart a busy database from a broken one, what should we do? Allow on broken and drop on busy? What is a third party script is keeping the database busy all the time (okay, this may go too far)?

yubiuser:
On the other hand: if the database is broken or a script is keeping the database busy, dropping the queries (and thereby blocking DNS) will signal "something is wrong" to the user. If we just allow the queries, the might never notice...

DL6ER:
After having slept on it for a few nights, I tend to agree on this change. Yes, a few users will notice an issue with their database, this taking down DNS resolution and will will come here and complain, possibly loudly. In the end that seems acceptable for the benefit of ensuring that database issues will not cause the Pi-hole to stop filtering.

We will make it very clear in the upcoming release notes that the default value changes here and everyone wanting to preserve the current behavior can easily do this be explicitly adding the setting `REPLY_WHEN_BUSY=ALLOW` to the config file.

dschaper:
SGTM

Issue #1335, opened 2022-04-22
"Total queries over last 24 hours" chart does not draw after clearing query history


Sassafras76:

yubiuser:
PR #1345 fixed the issue for me.
@Sassafras76 
If you want you can try it yourself by checking out the custom branch and apply your test.

`pihole checkout ftl fix/overTimeEmptyDatabase`

If you want to go back on track and before upgrading your Pi-hole to the next version please run `pihole checkout master`

Sassafras76:
> 

tested now and i do see the charts look better 

![image](https://user-images.githubusercontent.com/89309860/169628048-4b66c1a8-a941-4e8f-8248-bb6d1bdcedbf.png)

yubiuser:
Fixed with https://github.com/pi-hole/FTL/releases/tag/v5.16

PR #1345, opened 2022-05-01
Always return full overTime data

DL6ER:

## 10

---

The current version of FTL tries to be clever when returning overtime data and returns *trimmed* content, i.e., if there is nothing before 9am and nothing after 11am, it returns only thee two hours of data instead of a full 24 hours window. While looking at what caused https://github.com/pi-hole/FTL/issues/1335, I realized that this may be too much hand-holding and may actually even be confusing because it starts from super-wide bars (new installed and newly Initialized FTL database) progressing slowly to the well-known narrow bars after 24 hours.

This PR removes the "smart" reduction of the reported time span and ensures we always return all overtime data when requested. Even when it is partially empty, e.g. after a new install (or with a fresh database):

Screenshot from 2022-05-20 22-50-40

Another, indeed connected, change introduced in this PR is to run the garbage collection - recomputing and adjusting statistics - at the same interval as the overtime interval length. This makes the algorithms a bit simpler to understand and even saves some memory (we had to store 25 hours of data before, now we are really only storing 24 hours). All this should be invisible inside the Pi-hole world, however, when grabbing the data and displaying it in some tool that isn't aware of this, you'll see the statistics adjustment every 10 minutes now (instead of only once per hour). This is expected and correct.

Fixes https://github.com/pi-hole/FTL/issues/1335

### DIFF
--- a/src/FTL.h
+++ b/src/FTL.h
@@ -52,8 +52,8 @@
 #define SOCKETBUFFERLEN 1024
 
 // How often do we garbage collect (to ensure we only have data fitting to the MAXLOGAGE defined above)? [seconds]
-// Default: 3600 (once per hour)
-#define GCinterval 3600
+// Default: 600 (10 minute intervals)
+#define GCinterval 600
 
 // Delay applied to the garbage collecting [seconds]
 // Default: -60 (one minute before a full hour)
@@ -69,8 +69,8 @@
 #define MAXLOGAGE 24
 
 // Interval for overTime data [seconds]
-// Default: 600 (10 minute intervals)
-#define OVERTIME_INTERVAL 600
+// Default: same as GCinterval
+#define OVERTIME_INTERVAL GCinterval
 
 // How many overTime slots do we need?
 // (24+1) hours * number of intervals per hour
--- a/src/FTL.h
+++ b/src/FTL.h
@@ -73,11 +73,14 @@
 #define OVERTIME_INTERVAL GCinterval
 
 // How many overTime slots do we need?
-// (24+1) hours * number of intervals per hour
-// We need to be able to hold 25 hours as we need some reserve
-// due to that GC is only running once an hours so the shown data
-// can be 24 hours + 59 minutes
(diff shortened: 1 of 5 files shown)


yubiuser:
I do like the idea a lot. However, it does not fix https://github.com/pi-hole/FTL/issues/1335 (completely) for me. It still stops to send data at the 10 min "border".

Bildschirmfoto zu 2022-05-01 17-13-17

(See it's already 17:13)

yubiuser:
The current state of the PR does fix issue 1335 for me.

yubiuser (approved):
The scope/behavior of this PR has changed during it's development. Please adjust the PR introduction text.

yubiuser:
No, this should not happen. It might not be related to this issue. Please enable `DEBUG_OVERTIME` (https://docs.pi-hole.net/ftldns/configfile/#debug_overtime) and see what is logged before/after a restart.

thomasmerz:
Ok, this clarifies my question 👍🏻 You told me that my problem that I asked for in the discourse topic is (or may be?) related to this PR and issue #1312 . So should I open a new issue for this with the requested logging?

yubiuser:
My comment above was not considering you are using docker. I thought you run bare metal with the `development` branch and still saw that log entries. So I take back what I said above and tell the opposite.
If you want to try it, there should be a docker tag `dev` which should include the development versions of all Pi-hole components.

thomasmerz:
@yubiuser
With currently `latest` version I can't see `Found database entries in the future` anymore in any of my docker installations.
> Docker Tag [2022.07.1](https://github.com/pi-hole/docker-pi-hole/releases/2022.07.1) Pi-hole [v5.11.4](https://github.com/pi-hole/pi-hole/releases/v5.11.4) FTL [v5.16.1](https://github.com/pi-hole/FTL/releases/v5.16.1) Web Interface [v5.13](https://github.com/pi-hole/AdminLTE/releases/v5.13)

PR #882, opened 2020-09-05
logs from /var/log/ -> /var/log/pihole/

PromoFaux:
- Changes references from `/var/log/pihole*` -> `/var/log/pihole/pihole*`

See https://github.com/pi-hole/pi-hole/pull/3732

PR #1346, opened 2022-05-12
Change default log facility to /var/log/pihole/pihole-FTL.log

yubiuser:
- Changes references from `/var/log/pihole*` -> `/var/log/pihole/pihole*`

See https://github.com/pi-hole/pi-hole/pull/3732

This is an up-to-date version of https://github.com/pi-hole/FTL/pull/882

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -78,7 +78,7 @@ void getLogFilePath(void)
 	}
 
 	// Read LOGFILE value if available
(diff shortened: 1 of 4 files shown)


DL6ER (approved):
Approved but do not merge without the core PR being merged first to ensure everything is prepared for the changed default log path.

PR #1354, opened 2022-06-06
New API endpoint ">gateway"

DL6ER:

## 10

---

Add API endpoint `>gateway` that returns the IP address and interface of the default gateway.

This PR has been made for https://github.com/pi-hole/AdminLTE/pull/2218

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -40,6 +40,8 @@
 #include "../database/aliasclients.h"
 // get_edestr()
 #include "api_helper.h"
+// RTF_UP, RTF_GATEWAY
+#include <linux/route.h>
 
 // defined in src/dnsmasq/cache.c
 extern char *querystr(char *desc, unsigned short type);
@@ -1461,6 +1463,7 @@ void getGateway(const int *sock)
 	// Get IPv4 default route gateway and associated interface
 	in_addr_t gw = 0;
 	long dest_r = 0, gw_r = 0;
+	int flags = 0, metric = 0, minmetric = __INT_MAX__;
 	char iface[IF_NAMESIZE] = { 0 };
 	char iface_r[IF_NAMESIZE] = { 0 };
 	char buf[1024] = { 0 };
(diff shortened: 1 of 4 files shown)


yubiuser (commented):
What happens if FTL can't read `/proc/net/route`? We should log an error.
What is returned in such a case? What is returned if no non-default rout is found?
____
Documentation needed for https://docs.pi-hole.net/ftldns/telnet-api/

dschaper:
> What happens if FTL can't read `/proc/net/route`?

It will always be there if the kernel has a network stack.

yubiuser:
This should be the case of a device that is meant to be a DNS server ;-)
___

We should add a test for this new endpoint to the test_suite

DL6ER:
I thought about a test case, too, however, we cannot statically predict what the interface name and default gateway would be on the CI. At least not as static string - so the test will need to be a dynamic one. It'll take a few days until I can sit down and properly work this out.

IP address:
```
ip -4 route show default | awk '{print $3}'
```

Interface:
```
ip -4 route show default | awk '{print $5}'
```

---

> What is returned if no non-default route is found?

This can happen when no network connection is up. The returned value will be the defaults in the variables:
```
0.0.0.0 
```
(empty interface string)

dschaper:
> What is returned if no non-default route is found?

And sorry, missed this one. 

`/proc/net/route` is the route table in kernel. It's the information used by `route` or `ip route`. Is it possible to not have a destination of 0 (the default route)? I'm not sure. 

It is possible to have multiple gateways but only one per interface. I think this PR's code expects only a single gateway globally and that may result in the wrong gateway being selected if the first interface is not the used interface. (If `wan0` has 0000 destination and 001A for gateway and `eth0` has 0000 for destination and 002C for gateway. `wan0`'s gateway may be selected while `eth0`'s gateway is the one really used.)

rdwebdesign:
> (If `wan0` has 0000 destination and 001A for gateway and `eth0` has 0000 for destination and 002C for gateway. `wan0`'s gateway may be selected while `eth0`'s gateway is the one really used.)

If we know the interface we can use `ip -4 route show default dev "$INTERFACE"`.

dschaper:
I think we've found a good way to narrow the range of possible interfaces, there's a discussion on Team for it.

yubiuser (approved):

DL6ER:
I'd strongly prefer if we don't do these style changes in the bats file. We don't gain anything from them but they'll for sure cause a lot of complex merge conflicts in the next development -> v6.0 branch merger.

dschaper:
Yeah, the style stuff wasn't my intention.

yubiuser (approved):

yubiuser:
:point_up: 
> Documentation needed for https://docs.pi-hole.net/ftldns/telnet-api/

yubiuser:
https://github.com/pi-hole/docs/pull/721

DesktopECHO:
> > What is returned if no non-default route is found?
> 
> This can happen when no network connection is up. The returned value will be the defaults in the variables:
> 
> ```
> 0.0.0.0 
> ```
> 
> (empty interface string)

This can happen when you use a [quad-zero](https://www.cisco.com/c/en/us/support/docs/ip/routing-information-protocol-rip/16448-default.html) gateway.

In this sitution, the API responds like so:

```
>gateway
0.0.0.0
---EOM---
```

Problem with that is, the web interface is expecting someting like:

```
>gateway
10.13.12.11 eth0
---EOM---
```

...as a result, PHP cries about the missing gateway interface:

mod_fastcgi.c.487) FastCGI-stderr:PHP Warning:  array_combine(): Both parameters should have an equal number of elements in /var/www/html/admin/scripts/pi-hole/php/func.php on line 638
mod_fastcgi.c.487) FastCGI-stderr:PHP Notice:  Trying to access array offset on value of type bool in /var/www/html/admin/settings.php on line 77
mod_fastcgi.c.487) FastCGI-stderr:PHP Notice:  Undefined offset: 1 in /var/www/html/admin/settings.php on line 434
mod_fastcgi.c.487) FastCGI-stderr:PHP Notice:  Undefined offset: 2 in /var/www/html/admin/settings.php on line 434
mod_fastcgi.c.487) FastCGI-stderr:PHP Notice:  Undefined offset: 1 in /var/www/html/admin/settings.php on line 435
mod_fastcgi.c.487) FastCGI-stderr:PHP Notice:  Undefined offset: 2 in /var/www/html/admin/settings.php on line 435

Which is not such a big deal in PHP 7.4; the web interface continutes to run.  But in PHP 8.0+, apparently events like this are handled differently and the website glitches out with an empty settings page.

I guess this is a bug in some situations?

PR #1355, opened 2022-06-16
Add interface info for PADD

DL6ER:

## 10

---

Adds a bit of network information to be used in [PADD](https://github.com/pi-hole/PADD).

```
$ echo ">interfaces" | ncat 127.0.0.1 4711
enp2s0 UP 1000 3.4GB 5.1GB
lo UP -1 54.4MB 54.4MB
docker0 DOWN -1 0.0B 0.0B
wg0 UP -1 855.6MB 137.4MB
sum UP 0 4.3GB 5.3GB
---EOM---
```

Column definitions are:
1. Interface name
2. UP/DOWN status
3. Link speed in MBit/s (-1 means "Not available" (like link down) or "Not applicable" (like virtual interface))
4. TX bytes
5. RX bytes

The default interface (the one connected to the (favorite) gateway) will always be the first one so PADD can just `head -n1` it. The `sum` will always be the last one - even if you have (for whatever reason) an interface called `sum`.

Caveat about link speed: It won't work for most WiFi interfaces are the speed is not known at the kernel level. Instead, the drivers manage them dynamically depending on package loss, signal strength, etc. - in this case, you'll see link speed `-1` as well.

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -1458,13 +1458,11 @@ void getMAXLOGAGE(const int *sock)
 	ssend(*sock, "%d\n", config.maxlogage);
 }
 
-void getGateway(const int *sock)
+static bool getDefaultInterface(char iface[IF_NAMESIZE], in_addr_t *gw)
 {
 	// Get IPv4 default route gateway and associated interface
-	in_addr_t gw = 0;
 	long dest_r = 0, gw_r = 0;
 	int flags = 0, metric = 0, minmetric = __INT_MAX__;
-	char iface[IF_NAMESIZE] = { 0 };
 	char iface_r[IF_NAMESIZE] = { 0 };
 	char buf[1024] = { 0 };
 
@@ -1491,16 +1489,26 @@ void getGateway(const int *sock)
 			if(metric < minmetric)
 			{
 				minmetric = metric;
-				gw = gw_r;
+				*gw = gw_r;
 				strcpy(iface, iface_r);
 
 				if(config.debug & DEBUG_API)
 					logg("Reading interfaces: flags: %i, addr: %s, iface: %s, metric: %i, minmetric: %i",
-					     flags, inet_ntoa(*(struct in_addr *) &gw), iface, metric, minmetric);
+					     flags, inet_ntoa(*(struct in_addr *) gw), iface, metric, minmetric);
 			}
 		}
 		fclose(file);
 	}
 
+	// Return success based on having found the default gateway's address
+	return gw != 0;
+}
+
+void getGateway(const int *sock)
+{
+	in_addr_t gw = 0;
+	char iface[IF_NAMESIZE] = { 0 };
+
+	getDefaultInterface(iface, &gw);
 	ssend(*sock, "%s %s\n", inet_ntoa(*(struct in_addr *) &gw), iface);
 }
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -1567,17 +1567,6 @@ static bool listInterfaces(struct if_info **head, char default_iface[IF_NAMESIZE
 		if((f = fopen(fname, "r")) == NULL || fscanf(f, "%i", &(new->speed)) != 1)
 			new->speed = -1;
 
-		// Extract hardware address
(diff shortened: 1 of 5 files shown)


DL6ER:
Example for accessing the information from `bash`:
``` bash
ifaces="$(echo ">interfaces >quit" | ncat 127.0.0.1 4711)"
def_iface_data="$(head -n1 <<< "${ifaces}")"
iface_name="$(awk '{print $1}' <<< "${def_iface_data}")"
def_tx_bytes="$(awk '{print $4}' <<< "${def_iface_data}")"
def_rx_bytes="$(awk '{print $5}' <<< "${def_iface_data}")"
echo "Interface ${iface_name}: TX: ${tx_bytes}, RX: ${rx_bytes}"
```
``` plain
Interface eth0: TX: 3.4GB, RX: 5.2GB
```

yubiuser:
What is the aggregation time for `TX bytes` and `RX bytes`? I see several GB on a freshly booted system for `eth0`.

src/api/api.c
```diff
@@ -1491,16 +1489,259 @@ void getGateway(const int *sock)
...
+			tail->next = new;
+		tail = new;
+
+		tx_sum += new->tx_bytes;
+		rx_sum += new->rx_bytes;
+	}
+
+	freeifaddrs(ifap);
+
+	// Create sum entry only if there is more than one interface
```
yubiuser, line 1670: I would not create sum only if more than one interface is found. The API output should stay constant, even if the number of interfaces changes. Users might search for `sum` and wonder why their scripts fail if only one interface is up.

DL6ER:
> What is the aggregation time for `TX bytes` and `RX bytes`? I see several GB on a freshly booted system for `eth0`.

This is surprising and the answer is ... it depends. From the kernel documentation:

> What:		/sys/class/<iface>/statistics/rx_bytes
> Date:		April 2005
> KernelVersion:	2.6.12
> Contact:	netdev@vger.kernel.org
> Description:
> 		Indicates the number of bytes received by this network device.
> 		See the network driver for the exact meaning of when this
> 		value is incremented.

What is the output of
```
sudo ethtool -i eth0 | grep driver
```
on your machine?

The values are in the low GB regime for me which seems plausible on a server with an uptime of a few weeks serving DNS and some Wireguard. It is close to zero on my laptop which I just restarted.

src/api/api.c
DL6ER: It's not about interfaces being up but existing. You should always have at least `lo` in addition to your normal interface. The current simplification is to keep the code a bit simpler.

yubiuser:
Please scratch this
>  I see several GB on a freshly booted system for eth0.

I confused two terminals.

src/api/api.c
```diff
@@ -1491,16 +1489,259 @@ void getGateway(const int *sock)
...
+	if(head == NULL)
+		return true;
+
+	struct if_info *new = calloc(1, sizeof(struct if_info));
+	new->name = strdup("sum");
+	new->carrier = true;
```
yubiuser, line 1676: Are `carrier` and `speed` really properties of `sum`?

src/api/api.h
```diff
@@ -29,6 +29,7 @@ void getDBstats(const int *sock);
 void getUnknownQueries(const int *sock);
```
yubiuser, line 29: `getUnknownQueries`

was added by https://github.com/pi-hole/FTL/pull/174/commits/541d86bc744c7f46a69738c27ad06faa6b7a0fb9.

Does it still serve a purpose? If so we should document it - or did you still remember the function exists ;-)

src/api/api.c
DL6ER: It seems odd to give `sum` a different column definition than the other interfaces.

src/api/api.h
DL6ER: Yes, I do remember it still and have used it not too long ago in debugging a DNSSEC issue (which turned out to be a misconfiguration in the user's upstream `bind` server). That's also exactly the purpose I see for it and don't think there is a general necessity to document it.

yubiuser (approved):

PR #1377, opened 2022-07-06
Add supervised mode

DL6ER:

## 10

---

Add supervision mode. FTL running as supervisor starts a child process which is automatically restarted in case of any *abnormal* termination including crashes and uncatchable signals such as `SIGKILL`.

This is also the first step into the direction of supporting updates via the web frontend as this allows FTL to restart itself without having to rely on a third-party.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -149,6 +149,8 @@ set(sources
         signals.h
         struct_size.c
         struct_size.h
+        supervisor.c
+        supervisor.h
         timers.c
         timers.h
         vector.c
--- a/src/args.c
+++ b/src/args.c
@@ -43,9 +43,10 @@ static inline bool strEndsWith(const char *input, const char *end){
 	return strcmp(input + strlen(input) - strlen(end), end) == 0;
 }
 
-void parse_args(int argc, char* argv[])
+bool parse_args(int argc, char* argv[])
 {
 	bool quiet = false;
+	bool supervised = false;
 	// Regardless of any arguments, we always pass "-k" (nofork) to dnsmasq
 	argc_dnsmasq = 3;
 	argv_dnsmasq = calloc(argc_dnsmasq, sizeof(char*));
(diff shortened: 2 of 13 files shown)


DL6ER:
Moving the termination into the tests (to check if the supervisor behaves as expected) revealed a crash happening only on `musl`. On our "normal" builds, canceling a non-existing thread is a simple no-op whereas, on `musl`, this leads to a segmentation fault. The fix is to avoid canceling non-existing threads ([3c38420/src/daemon.c#R234-R237](https://github.com/pi-hole/FTL/pull/1377/commits/3c384203ea01f15d950e0d30b6e081dc79bd110f#diff-27afef919bde7bf81114f48a9b3959b094de6d866c60fa473921ad45a43cf16dR234-R237)).

yubiuser (changes requested):
Needs documentation in `pihole-FTL --help` and probably https://docs.pi-hole.net/ftldns/in-depth/

MichaIng:
Is there a mechanism which prevents an endless restart loop? I cannot see it in code currently. Probably a restart limit, optionally adjustable with a CLI option, makes sense?

DL6ER:
There is a forced delay of 1sec between restarts to avoid CPU spinning by endless restarts in a very short time. Others than that, there is no counter of recently failed attempts. Is there such a limit with `systemd` and it's `Restart=on-failure` feature? Even when it was easy, I would still like to avoid reinventing the wheel and use a proper `systemd` unit with said `Restart=on-failure` instead of this PR.

dschaper:
https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=

My recollection is that systemd will only try 5 times by default and then journal an error that says "retried too many times" or "tried to restart too quickly". There are limits that are discussed above as well as alternatives to 'on-failure'.

MichaIng:
The related (re)start limits can be set in the `[Unit]` section: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval

I see, I need to start doing the systemd integration 😅. I'm sorry for not doing so far what I promised, new full-time job has reduced my space time dramatically.

DL6ER:
I plan to look at another `systemd` attempt, too, sometime next week. Maybe the busy two of us manage to do it together :-)

DL6ER:
Superseded by native `systemd` unit ( https://github.com/pi-hole/pi-hole/pull/4924 )

PR #1380, opened 2022-07-07
Log new and terminating telnet connections only when debug flag is set

yubiuser:

5

PR https://github.com/pi-hole/FTL/pull/1370 changed the telnet API error handling. It also logs when new telnet connections were made. However, the web interface makes a lot of connections via the API, all of them get logged. This creates logspam. 

This PR changes the behavior to only log if the debug flag is set.

### DIFF
--- a/src/api/socket.c
+++ b/src/api/socket.c
@@ -519,8 +519,8 @@ void *telnet_listening_thread_IPv4(void *args)
 			logg("WARNING: Unable to open telnet processing thread: %s", strerror(errno));
 		}
 	}
-
-	logg("Terminating IPv4 telnet thread");
+	if(config.debug & DEBUG_API)
+		logg("Terminating IPv4 telnet thread");
 	return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


DL6ER:
It seems to be there already on `master`:

https://github.com/pi-hole/FTL/blob/37e605201bf76c9143171770c8d7dfd0ae367697/src/api/socket.c#L505-L506

DL6ER (changes requested):
Should also be here:

https://github.com/pi-hole/FTL/blob/37e605201bf76c9143171770c8d7dfd0ae367697/src/api/socket.c#L523

yubiuser:
You did not link `master` but the current branch. On master
https://github.com/pi-hole/FTL/blob/bc5077d56e9dd75b6ac99c096d35926783e1e859/src/api/socket.c#L504-L505.
__
Pushed fix for `terminating`

DL6ER (approved):

PR #1383, opened 2022-07-08
Fix status of blocked special domains when they are in FTL's DNS cache

DL6ER:

## 10

---

When querying special domains (such as `use-application-dns.net`), they are getting logged with status 16 (`Blocked (special domain)`). However, subsequent queries to the same domain incorrectly get status `OK (cached)`. This PR fixes these subsequent queries to also show status 16. This big is a mere displaying issue, blocking works as expected.

Thanks to @jpgpi250 for pointing this out.


yubiuser (approved):
The first two are on the `development` the latter on the current branch

Bildschirmfoto zu 2022-07-08 22-45-17

PR #1385, opened 2022-07-10
Add automatic database optimization 

DL6ER:

## 10

---

Add automatic database optimization using `PRAGMA optimize` before closing database connections as recommended by the SQLite3 maintainers: 

> To achieve the best long-term query performance without the need to do a detailed engineering analysis of the application schema and SQL, it is recommended that applications run "PRAGMA optimize" (with no arguments) just before closing each [database connection](https://www.sqlite.org/c3ref/sqlite3.html). Long-running applications might also benefit from setting a timer to run "PRAGMA optimize" every few hours.
> 
> This pragma is usually a no-op or nearly so and is very fast. However if SQLite feels that performing database optimizations (such as running [ANALYZE](https://www.sqlite.org/lang_analyze.html) or creating new indexes) will improve the performance of future queries, then some database I/O may be done.


DL6ER:
I'll close this. after having used it for some time on a spare Pi-hole (still receiving all traffic from a normal household) it doesn't seem to improve database access times compared to a copy of the database where I manually strip all the `ANALYZE` results before running test queries. Further analysis suggests this is caused by an assumption made by `ANALYZE` (accumulate usage information over runtime) which we do not fulfill on purpose: We are opening the FTL database connection only momentarily as all Pi-holes are always expecting power losses without proper shutdowns and past experiences on database connection that stay on for a long time were far worse than what we have now (data losses basically don't happen at all these days for users).

PR #1396, opened 2022-07-31
Major rewrite of the socket processing routines

DL6ER:

## 10

---

Major rewrite of the socket processing routines of FTL's Telnet-like API.

Before, we launched a new thread for each incoming connection (more than 40 threads concurrent were never allowed) to handle incoming connections independent from another and serve content concurrently. However, this could potentially be used to DoS FTL by opening and closing telnet sessions in very quick succession.
The new behavior is to, instead, launch five (compile-time setting) threads per type (independent for telnet over IPv4, IPv6, and Unix socket communication = 15 threads in total). Each of them accepts incoming connections in a FIFO manner (they always pick the first connection in the waiting connection queue). If too many requests are sent to FTL at once, they will simply be queued by the system until FTL is ready to accept them.

This is meant as a bugfix related to https://github.com/pi-hole/PADD/issues/252,

### DIFF
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -2851,11 +2851,11 @@ static char *get_ptrname(struct in_addr *addr)
 }
 
 // int cache_inserted, cache_live_freed are defined in dnsmasq/cache.c
-void getCacheInformation(const int *sock)
+void getCacheInformation(const int sock)
 {
 	struct cache_info ci;
 	get_dnsmasq_cache_info(&ci);
-	ssend(*sock,"cache-size: %i\ncache-live-freed: %i\ncache-inserted: %i\nipv4: %i\nipv6: %i\nsrv: %i\ncname: %i\nds: %i\ndnskey: %i\nother: %i\nexpired: %i\nimmortal: %i\n",
+	ssend(sock, "cache-size: %i\ncache-live-freed: %i\ncache-inserted: %i\nipv4: %i\nipv6: %i\nsrv: %i\ncname: %i\nds: %i\ndnskey: %i\nother: %i\nexpired: %i\nimmortal: %i\n",
 	            daemon->cachesize,
 	            daemon->metrics[METRIC_DNS_CACHE_LIVE_FREED],
 	            daemon->metrics[METRIC_DNS_CACHE_INSERTED],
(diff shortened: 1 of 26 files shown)


yubiuser (commented):
Why is `istelnet` missing in `/api/reques.c` for all messages `>cacheinfo` up to `>interfaces`

DL6ER:
The Unix socket interface was introduced after the Telnet was already present. It was supposed to become the interface between FTL and the at that time drafted (and now abandoned) HTTP API written in another language. As the Telnet-like replying using ASCII-encoded numbers is pretty inefficient for machine-machine communication, we implemented `MessagePack` instead (https://github.com/pi-hole/FTL/pull/75) which allows for more efficient streaming.

After it got clear that the HTTP API isn't going to make it, we abandoned the development of the Unix socket part and new API endpoints did not get a different output than "ordinary" Telnet requests (this is the answer to your qeustion). We could (and maybe should) have removed the Unix socket sometime in between as I'm not sure anyone on this planet is actually using it. However, we still have it. All this code is removed for `new/http` where we will only provide a standardized REST interface instead.

src/dnsmasq_interface.c
```diff
@@ -3133,15 +3117,6 @@ void FTL_TCP_worker_created(const int confd)
...
 	gravityDB_forked();
-
-	// Children inherit file descriptors from their parents
-	// We don't need them in the forks, so we clean them up
-	if(config.debug != 0)
-		logg("Closing Telnet socket for this fork");
-	close_telnet_socket();
-	if(config.debug != 0)
-		logg("Closing Unix socket for this fork");
-	close_unix_socket(false);
```
DL6ER, line 3144: I just want to leave some details where why we don't need this in the new version. This code is called when FTL forks (creates a dedicated worker for DNS TCP requests). When a multi-threaded process forks, only the calling thread is cloned in the child process - the other threads of the father process are not forked: the child process is mono-threaded until it creates new threads on its side. However, we are only listening on the sockets in the respective threads of the main application (non-forked UDP process). There is nothing in the forked process that would be listening so we can be sure that no API requests are accidentally replied to from forks.

yubiuser:
> We could (and maybe should) have removed the Unix socket sometime in between as I'm not sure anyone on this planet is actually using it.

Maybe this PR is a good place to remove all fragments of the Unix socket?

yubiuser (approved):

dschaper:
My self-assignment isn't a blocker for merge.

PR #1389, opened 2022-07-15
Change configuration option for REPLY_WHEN_BUSY ACCEPT > ALLOW

sticksen:
Signed-off-by: Marc <marc@deyan7.de>

**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] I have checked that [another pull request](https://github.com/pi-hole/FTL/pulls) for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 1

---
Please also check https://github.com/pi-hole/docs/pull/742#issuecomment-1185285675  

The configuration option for REPLY_WHEN_BUSY did not parse the correct value `ALLOW`. Instead `ACCEPT` was parsed, resulting in the docs being out of sync with the implementation and possibly leading to unwanted behavior of Pihole.

PR #1400, opened 2022-08-13
The expected value for REPLY_WHEN_BUSY is ALLOW not ACCEPT

CTassisF:

**How familiar are you with the codebase?:** _1_

---

According to the [documentation](https://docs.pi-hole.net/ftldns/configfile/#reply_when_busy), the expected value for `REPLY_WHEN_BUSY` is `ALLOW`, not `ACCEPT`.

This fixes an issue where `REPLY_WHEN_BUSY` is set to the default value `DROP` even when the user manually sets to `ALLOW`.

Introduced in #1341.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -691,7 +691,7 @@ void read_FTLconf(void)
 	// defaults to: DROP
 	buffer = parse_FTLconf(fp, "REPLY_WHEN_BUSY");
 
-	if(buffer != NULL && strcasecmp(buffer, "ACCEPT") == 0)
+	if(buffer != NULL && strcasecmp(buffer, "ALLOW") == 0)
 	{
 		config.reply_when_busy = BUSY_ALLOW;
 		logg("   REPLY_WHEN_BUSY: Permit queries when the database is busy");


src/config.c
```diff
@@ -691,7 +691,7 @@ void read_FTLconf(void)
 	// defaults to: DROP
 	buffer = parse_FTLconf(fp, "REPLY_WHEN_BUSY");
 
-	if(buffer != NULL && strcasecmp(buffer, "ACCEPT") == 0)
+	if(buffer != NULL && strcasecmp(buffer, "ALLOW") == 0)
```
DL6ER, line 694:
```suggestion
	if(buffer != NULL && (strcasecmp(buffer, "ALLOW") == 0 || strcasecmp(buffer, "ACCEPT") == 0))
```

We should accept both here to ensure we are not breaking existing configs where users found it they have to set `ACCEPT` instead of `ALLOW`.

DL6ER (changes requested):

yubiuser:
I think you are working on an outdated `development` branch. We recently made exactly this change.
https://github.com/pi-hole/FTL/pull/1389

CTassisF:
> I think you are working on an outdated `development` branch. We recently made exactly this change. #1389

I'm sorry, you're correct. I didn't check if it was already fixed on `development` branch. I did the change on `master` and then requested to merge into `pi-hole:development`, as required by the template.

PR #1412, opened 2022-08-21
Ensure compatibility with gcc 11.2.0 (Ubuntu 11.2.0-19ubuntu1)

DL6ER:

## 10

---

Add missing function attribute to syscall to ensure compatibility with gcc 11.2.0 (Ubuntu 11.2.0-19ubuntu1) which was installed on my main development system after the upgrade from Ubuntu 20.04.4 LTS to Ubuntu 22.04.1 LTS.

The function `FTLfopen()` is a wrapper around `fopen()` which is, in turn, equipped with the very same attribute:

Screenshot from 2022-08-21 12-02-51

https://github.com/bminor/glibc/blob/f7b0fc5cc61301461e3c1a278240ce78701bb9a8/libio/stdio.h#L256-L260

Screenshot from 2022-08-21 12-03-03

https://github.com/bminor/glibc/blob/d055481ce39d03652ac60de5078889e15b6917ff/misc/sys/cdefs.h#L277-L284

### DIFF
--- a/src/syscalls/fopen.c
+++ b/src/syscalls/fopen.c
@@ -15,7 +15,7 @@
 static uint8_t already_writing = 0;
 
 #undef fopen
-FILE *FTLfopen(const char *pathname, const char *mode, const char *file, const char *func, const int line)
+FILE * __attribute__ ((__malloc__)) FTLfopen(const char *pathname, const char *mode, const char *file, const char *func, const int line)
 {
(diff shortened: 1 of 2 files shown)


yubiuser (commented):
Why is it set once before `FTLfopen` and once after?

yubiuser (approved):

Issue #1368, opened 2022-06-27
[Tracking] add .editorconfig


yubiuser:
Add `.editorconfig` if feasible

yubiuser:
WIP

```
# EditorConfig is awesome: https://editorconfig.org/

# top-most EditorConfig file
root = true

# GNU coding style is used for FTL
[*]
end_of_line = lf
insert_final_newline = true
indent_style = tab
indent_size = tab
tab_width = 8
charset = utf-8
trim_trailing_whitespace = true
# The left part of the curly bracket should be on the next line
curly_bracket_next_line = true
spaces_around_operators = true
spaces_around_brackets = none
indent_brace_style = Allman

[src/dnsmasq/**]
indent_style = space
indent_size = 2
insert_final_newline = false
trim_trailing_whitespace = false
```

DL6ER:
Related branch: https://github.com/pi-hole/FTL/tree/new/editorconfig

yubiuser:
Yeahh... it's complaining a lot...

DL6ER:
Closing this for now. It seems we cannot specify the code style used not only by us but also by `dnsmasq` and `SQLite3` that is mixed space and tab indentation. `dnsmasq` is doing crazy things such as "every eight spaces are replaced by a tab", so
```
ab()
SSab()
SSSSab()
T     ab()
T     SSab()
T     SSSSab()
T     SSSSSSab()
TT            ab()
```

(where `S` = space and `T` = tab)

and we are doing some alignment using spaces *after* tabs, too like
```
T       some(function, with, very
T       SSSSSmany, parameters)
```

and SQLite3 is doing some other things, too.

PR #1415, opened 2022-08-31
Add warning about inaccessible adlists to message table (Pi-hole diagnosis)

yubiuser:

**How familiar are you with the codebase?:** _10_

---

Adds a warning to the `message` table (Pi-hole diagnosis) if adlists have been inaccessible during the last gravity run. We already save `status` in the `adlists` table in the `gravity.db`. This status is set during the gravity run and indicates if the "download" was successful (https://docs.pi-hole.net/database/gravity/#adlist-table-adlist). Status type `3` and `4` indicate that the adlist was inaccessible. 
This PR checks on each `FTL` reload if `gravity.db` contains adlists with `status=[3,4]` (and are enabled) and adds a warning if applicable.

Corresponding web interface PR: https://github.com/pi-hole/AdminLTE/pull/2320

Discourse thread: https://discourse.pi-hole.net/t/gravity-update-timeout/57391

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -1409,7 +1409,8 @@ bool gravityDB_get_regex_client_groups(clientsData* client, const unsigned int n
 	return true;
 }
 
-void check_inaccessible_adlists(void){
+void check_inaccessible_adlists(void)
+{
 
 	// check if any adlist was inaccessible in the last gravity run
 	// if so, gravity stored `status` in the adlist table with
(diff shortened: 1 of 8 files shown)


src/database/gravity-db.c
```diff
@@ -1407,3 +1408,42 @@ bool gravityDB_get_regex_client_groups(clientsData* client, const unsigned int n
 
 	return true;
 }
+
+void check_inaccessible_adlists(void){
+
```
DL6ER, line 1413:
```suggestion
void check_inaccessible_adlists(void)
{
```

src/database/gravity-db.c
```diff
@@ -1407,3 +1408,42 @@ bool gravityDB_get_regex_client_groups(clientsData* client, const unsigned int n
...
+		logg("check_inaccessible_adlists(): %s - SQL error prepare: %s", querystr, sqlite3_errstr(rc));
+		gravityDB_close();
+		return;
+	}
+
+    // Perform query
+    while((rc = sqlite3_step(query_stmt)) == SQLITE_ROW)
+    {
+        int id = sqlite3_column_int(query_stmt, 0);
+        const char *address = strdup((const char*)sqlite3_column_text(query_stmt, 1));
```
DL6ER, line 1441: This currently leads to a memory leak (memory is allocated by `strdup()` but never `free`d).

```suggestion
        const char *address = (const char*)sqlite3_column_text(query_stmt, 1);
```

src/database/gravity-db.c
```diff
@@ -1407,3 +1408,42 @@ bool gravityDB_get_regex_client_groups(clientsData* client, const unsigned int n
...
+
+        // log to the message table
+        logg_inaccessible_adlist(id, address);
+    }
+
+    // Finalize statement
+    sqlite3_finalize(query_stmt);
```
DL6ER, line 1448: This section uses spaces instead of tabs (used anywhere else in this file).

DL6ER (changes requested):

src/database/gravity-db.c
yubiuser: I'll change that. Maybe a good point to think about https://github.com/pi-hole/FTL/issues/1368 again.

DL6ER (approved):

PR #1430, opened 2022-09-12
Replace nonstandard `which`

a1346054:

**How familiar are you with the codebase?:** 6

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -49,7 +49,7 @@ cmake --build . -- -j $(nproc)
 # Otherwise, we simply copy the binary one level up
 if [[ -n "${install}" ]]; then
     echo "Installing pihole-FTL"
-    SUDO=$(which sudo)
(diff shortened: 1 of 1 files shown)


DL6ER:
Why are you replacing `which` by `command -v`? It's the former not available on your system? (which system?)

a1346054:
> Why are you replacing `which` by `command -v`? It's the former not available on your system? (which system?)

On invocation, I get this on an older version of debian testing:

> /usr/bin/which: this version of `which' is deprecated; use `command -v' in scripts instead.

It is going away, eventually, in debian https://salsa.debian.org/debian/debianutils/-/blob/5aeca1e3b14c08b5ab92995e91efcc2b3806a639/debian/NEWS#L7-16

It might still get reverted before release, but it's good to replace nonstandard things with standard things, and `command -v` is POSIX

PR #1437, opened 2022-09-17
Only log disk space warning when usage increased

yubiuser:

**How familiar are you with the codebase?:**  5

---

We check two locations for free disk space and warn if disk usage >90% (https://github.com/pi-hole/FTL/pull/1249). Warning consists of adding a message in the `message_table` and logging to `FTL.log`
One location we monitor is the filesystem where the log files are located. We already ensure that we only log to the message table once, however, we continue to log to `FTL.log` every time we check (300 sec). This leads to log spam which further decreases the available space for the log files. 
See this discourse thread with a log warning every 5 minutes: https://discourse.pi-hole.net/t/disk-shortage-var-log-pihole-ftl-log-ahead-99-used/57905/10?u=yubiuser

This PR adds a check that the available disk space is less then during the previous check and only executes the logging in such a case.

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -75,17 +75,21 @@ time_t get_rate_limit_turnaround(const unsigned int rate_limit_count)
 	return (time_t)config.rate_limit.interval*how_often - (time(NULL) - lastRateLimitCleaner);
 }
 
-static void check_space(const char *file)
+static int check_space(const char *file, int LastUsage)
 {
 	if(config.check.disk == 0)
-		return;
+		return 0;
 
 	int perc = 0;
 	char buffer[64] = { 0 };
 	// Warn if space usage at the device holding the corresponding file
-	// exceeds the configured threshold
-	if((perc = get_filepath_usage(file, buffer)) > config.check.disk)
+	// exceeds the configured threshold and current usage is higher than
+	// usage in the last run (to prevent log spam)
+	perc = get_filepath_usage(file, buffer);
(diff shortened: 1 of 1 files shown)


src/gc.c
```diff
@@ -138,8 +146,12 @@ void *GC_thread(void *val)
...
 		{
 			check_load();
-			check_space(FTLfiles.FTL_db);
-			check_space(FTLfiles.log);
+			int DBStorageUsage = check_space(FTLfiles.FTL_db, LastDBStorageUsage);
+			int LogStorageUsage = check_space(FTLfiles.log, LastLogStorageUsage);
+			
+			LastLogStorageUsage = LogStorageUsage;
+			LastDBStorageUsage = DBStorageUsage;
+
```
DL6ER, line 154:
```suggestion
			LastDBStorageUsage = check_space(FTLfiles.FTL_db, LastDBStorageUsage);
			LastLogStorageUsage = check_space(FTLfiles.log, LastLogStorageUsage);
```
No need for intermediate variables.

yubiuser: Nice. I wasn't sure if this is supported and went for the safer way first.

DL6ER: The way it works now is that it takes the value at the time of invoking the function (old value) and overwrites it only once the function returns (new value). Could be even fancier and pass the pointer to the memory location and do all the changing in the function, but no need for special fanciness. This works equally well.

DL6ER (approved):

PR #1439, opened 2022-09-18
FTL v5.18.1

PromoFaux:


yubiuser (dismissed):

DL6ER:
Waiting for final confirmation of #1438 on https://discourse.pi-hole.net/t/log-entry-for-dns-type-type5335-with-reply-of-blob/57908

PromoFaux:
Keep the internal label, as that prevents it going into the auto generated release notes.

DL6ER (approved):

DL6ER:
The fix has again been confirmed, this is good to go.

yubiuser (approved):

PR #1445, opened 2022-09-21
Do not create `FTL.port` file anymore

yubiuser:

**How familiar are you with the codebase?:**  5

---

`FTL` used to write its API port to `/run/FTL.port` (or whatever was configured in `pihole-FTL.conf` as path of `PORTFILE`). This was meant as a way to allow `core` and `web` to get the current/configured port. However, this is a unnecessary step, because `FTL` will use the default port `4711` unless `FTLPORT` is set in `pihole-FTL.conf`. If this is set, `web` and `core` can simply take the value from this config file instead of doing a detour to `FTL.port`.

Corresponding PRs: **XXX** and **YYY**

Documentation at: https://github.com/pi-hole/docs/pull/778

___

While removing the code for the port file, I noticed we have an old `FTLtest.sh` which was not used within FTL but as a guidance for users how to interact with the API. I don't think it's much of a use today.


DL6ER (approved):

PromoFaux:
> and YYY

@yubiuser did you already do the core changes but forget to PR? I also need to do update docker stuff, too.

yubiuser:
I have the code and the branch, but did not open a PR yet because I have `black` installed locally and it "formatted" the edited `test*.py` file, making the review tougher than nececcary with a few hundred changed lines. Wanted to get the `black`PR in first

MichaIng:
Tiny regression: When the port is changed in the config file, web UI and others will get a wrong port until FTL is restarted. `ss` would be another option to read out the actually used port, but requires some `awk`/`sed` action.

rdwebdesign:
I notice the opposite:
Using the file as Port number source was causing a problem for the web interface.
When web interface tries to read the file after a restart, the file is available only a few milliseconds after the restart (FTL doesn't write the file right away).
The result is: web interface never finds the port immediately after a restart.

yubiuser:
> Tiny regression: When the port is changed in the config file, web UI and others will get a wrong port until FTL is restartet.

Good catch. However, I think we can tolerate this. I've never seen anyone changing the port in the wild.

PR #1460, opened 2022-10-16
Remove deprecated set-output in build.yml

yubiuser:
Github considers the `set-output` command to be deprecated https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This PR changes our `build.yml` to use the recommended environmental `$GITHUB_OUTPUT` instead of `set-output` (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)


DL6ER (approved):
Thank you!

Issue #1463, opened 2022-10-29
FTL crashes after system reboot


matthiasstrauch:
### Versions

-  Pi-hole version is v5.13 (Latest: v5.13)
-  AdminLTE version is v5.16 (Latest: v5.16)
-  FTL version is v5.18.2 (Latest: v5.18.2)

### Platform

- OS and version:  Ubuntu 20.04.5 LTS

- Platform:  x86_64 bare metal host

### Expected behavior

After a reboot, Pihole services should start regularly after the defined sleep interval.

### Actual behavior / bug

Pihole-FTL crashes after waiting the defined sleep interval. This has happened every time after updating to the latest version.

### Steps to reproduce

Unknown.

## Debug Token

- URL:  https://tricorder.pi-hole.net/IgqdTBEc/

### Screenshots

-

### Additional context

This started happening after updating to the latest version, it worked fine before.  Unfortunately, I'm not sure which version I was on earlier.

[2022-10-29 02:30:33.494 2038M] Finished config file parsing
[2022-10-29 02:30:33.494 2038M] Sleeping for 20 seconds as requested by configuration ...
[2022-10-29 02:30:48.870 1459M] Done sleeping, continuing startup of resolver...
[2022-10-29 02:30:48.884 1459M] Database version is 12
[2022-10-29 02:30:48.884 1459M] Resizing "FTL-strings" from 40960 to (81920 * 1) == 81920 (/dev/shm: 1.4MB used, 4.0GB total, FTL uses 700.8KB)
[2022-10-29 02:30:48.884 1459M] Imported 0 alias-clients
[2022-10-29 02:30:48.884 1459M] Database successfully initialized
[2022-10-29 02:30:48.894 1459M] New upstream server: 8.8.4.4:53 (0/512)
[2022-10-29 02:30:48.914 1459M] New upstream server: 8.8.8.8:53 (1/512)
[2022-10-29 02:30:48.919 1459M] New upstream server: 192.168.188.1:53 (2/512)
[2022-10-29 02:30:48.939 1459M] Resizing "FTL-domains" from 12288 to (1024 * 24) == 24576 (/dev/shm: 1.5MB used, 4.0GB total, FTL uses 741.7KB)
[2022-10-29 02:30:48.939 1459M] !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
[...]

PR #1464, opened 2022-11-03
Improve shared memory protection

DL6ER:

## 10

---

**Store and check shared memory ownership before resizing shared memory objects.**

The scenario we are resolving here is the following: FTL is already running and someone(/-thing) deletes all the shared memory objects. If another instance of FTL is then started, it doesn't know a previous process is running and creates new SHM objects for itself. Both processes can - in theory - run just fine without touching each other as they are both actually using *different* shared memory objects. You are right when you think this is bizarre. The reason is that the first one has the shared memory objects still mapped into memory, i.e., "deleting" them from disk only removes the visible file handles. The next FTL instance creates files with the very same name, however, they are also new and distinct files, i.e., their memory will point elsewhere. Both instances can now run in parallel just fine *until* one of them needs to resize a shared memory objects. If one of the shared memory events now gets resized to a size larger than it was before BUT smaller than what the other FTL instance is expecting, the other instance will instantaneously crash with a `SIGSEGV` (a bus error to be precise).

This PR resolves this by storing the PID of the SHM object creator in the settings object.

Each FTL instance reloads this shared memory instance now before performing any potentially dangerous operation and checks if the shared memory files on disk are still owned by this process. If this is not the case, we are in serious trouble and exit immediately. This should allow the second instance (you could call it the "rightful owner" of the current existing SHM objects) a fairly good chance to never even notice this and continue to operate just fine.

Resolves the root causes behind #1463, #1448 and possibly other similar issues in the past by preventing a certain "mis"configuration to have the bad effect it had for these users.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -55,6 +55,9 @@ int main (int argc, char* argv[])
 	logg("########## FTL started on %s! ##########", hostname());
 	log_FTL_version(false);
 
+	// Process pihole-FTL.conf
+	read_FTLconf();
+
 	// Catch signals not handled by dnsmasq
 	// We configure real-time signals later (after dnsmasq has forked)
 	handle_signals();
@@ -67,9 +70,6 @@ int main (int argc, char* argv[])
 		check_running_FTL();
 		return EXIT_FAILURE;
 	}
-
-	// Process pihole-FTL.conf
-	read_FTLconf();
 
 	// pihole-FTL should really be run as user "pihole" to not mess up with file permissions
 	// print warning otherwise
--- a/src/main.c
+++ b/src/main.c
@@ -60,7 +60,7 @@ int main (int argc, char* argv[])
 	handle_signals();
 
 	// Initialize shared memory
-	if(!init_shmem(true))
+	if(!init_shmem())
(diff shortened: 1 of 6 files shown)


src/main.c
```diff
@@ -55,27 +55,31 @@ int main (int argc, char* argv[])
 	logg("########## FTL started on %s! ##########", hostname());
 	log_FTL_version(false);
 
+	// Process pihole-FTL.conf
```
yubiuser, line 58: Why did you move reading the conf file before initializing the shared memory?

DL6ER: This has been mentioned in the commit message of https://github.com/pi-hole/FTL/pull/1464/commits/fc303aeba52cd2e742ea6cc2b62e3d6c8a5f13ac. Sorry, I should have included this in the PR description:

> Read FTL config *before* initializing shared memory to ensure we are aware of a possible DEBUG_SHMEM flag when creating the shared memory segments

Without this, any `DEBUG_SHMEM` code was useless in `init_shmem()`.

yubiuser (approved):

PR #1468, opened 2022-11-06
Move rate-limiting from per-client to per-client-per-domain

DL6ER:

Let me be clear here: Rate-limit is there it *protect* you. It is not meant as a method to cause pain. What are we protecting against? Basically against [Denial of Service (DoS)](https://de.wikipedia.org/wiki/Denial_of_Service) "attacks" either by a malicious attacker or by some client that has a bug or crashed or ... (you never know, we have seen so may things). If one attacker or or misbehaving client can take down your entire network because your DNS resolver runs out of memory - that doesn't sound like something you'd want, right? - and this is where FTL's rate-limiting kicks in, simply discarding queries when there are too many. The default settings are suitable both for home and enterprise, small and huge networks, however, they can always be adjusted (or even disabled) to your liking if you feel you need no DoS-protection.

This PR proposes to change FTL's rate-limiting mechanism from a pure "per-client" to a "per-client *and* per-domain" scheme. If a client queries `domain.com` like hell, it will get rate-limited as before. However, any *other* domain (say `example.com`) can still resolved (except, obviously, it it gets queried too often, too).

This reduces the "there is not Internet connection" effect some users reported at the costs of a much less effective rate-limiting as clients and/or malicious attackers can simply evade it by querying varying domains. We have seen this by buggy clients querying seemingly random subdomain - each of them would be counted on their own.

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** ***to be done***

---

## Checklist:

### DIFF
--- a/src/database/message-table.c
+++ b/src/database/message-table.c
@@ -27,7 +27,7 @@
 #include "../gc.h"
 
 static const char *message_types[MAX_MESSAGE] =
-	{ "REGEX", "SUBNET", "HOSTNAME", "DNSMASQ_CONFIG", "RATE_LIMIT", "DNSMASQ_WARN", "LOAD", "SHMEM", "DISK", "ADLIST" };
+	{ "REGEX", "SUBNET", "HOSTNAME", "DNSMASQ_CONFIG", "RATE_LIMIT", "DNSMASQ_WARN", "LOAD", "SHMEM", "DISK", "ADLIST", "DOMAIN_RATE_LIMIT_MESSAGE" };
 
 static unsigned char message_blob_types[MAX_MESSAGE][5] =
 	{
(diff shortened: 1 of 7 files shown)


DL6ER:
Rebased on latest `development`

src/database/message-table.c
```diff
@@ -26,8 +26,38 @@
...
+	{
+	case REGEX_MESSAGE:
+		return "REGEX";
+	case SUBNET_MESSAGE:
+		return "SUBNET";
+	case HOSTNAME_MESSAGE:
+		return "HOSTNAME";
+	case DNSMASQ_CONFIG_MESSAGE:
+		return "DNSMASQ_CONFIG";
+	case RATE_LIMIT_MESSAGE:
```
yubiuser, line 41: Why do we keep this here? `RATE_LIMIT_MESSAGE` is not used anymore.

src/dnsmasq_interface.c
```diff
@@ -1550,17 +1553,51 @@ bool _FTL_CNAME(const char *dst, const char *src, const int id, const char* file
...
+	clientsData *client = getClient(clientID, true);
+	if(client == NULL)
+	{
+		logg("WARN: No memory available, skipping query analysis (%s:%d)",
+		     short_path(__FILE__), __LINE__);
+		free(child_domain_str);
+		return false;
+	}
+
+	// Get child's DNS cache entry (we eed to create a new cache entry here
```
yubiuser, line 1587:
```suggestion
	// Get child's DNS cache entry (we need to create a new cache entry here
```

src/enums.h
```diff
@@ -211,6 +211,7 @@ enum message_type {
 	SHMEM_MESSAGE,
```
yubiuser, line 211: Still need L208 "RATE_LIMIT_MESSAGE"?

DL6ER:
The `unique` change we added recently won't work here as uniqueness is now composed out of three fields for this new message type (type + client + domain). Needs more work.

PR #1496, opened 2022-11-27
Update embedded dnsmasq to v2.88rc5

DL6ER:

Two noteworthy changes:
- `use-stale-cache`  got a freely configurable maximum stale time defaulting one day (**patch sent by Pi-hole**)
- `dnsmasq` used to return malformed queries from upstream to the clients downstream and just refused to cache them. This is now changed to instead serve `SERVFAULT` for obvious errors like more than one question or no requested name in the DNS header. Other reasons for invalidity are too short packets (announced to be longer than what we received data for) or existing but empty `TXT` records (this is not allowed).
  The tests aren't very strict and this is considered a feature as old `dnsmasq` version in the future (think of this release in ten years time) shouldn't reject query types that are currently unknown as errors when they are only defined in the future.

---

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,6 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(PIHOLE_FTL C)
 
-set(DNSMASQ_VERSION pi-hole-v2.88rc4)
+set(DNSMASQ_VERSION pi-hole-v2.88rc5)
 
 add_subdirectory(src)
--- a/src/dnsmasq/cache.c
+++ b/src/dnsmasq/cache.c
@@ -1937,8 +1937,10 @@ void dump_cache(time_t now)
 char *record_source(unsigned int index)
 {
   struct hostsfile *ah;
+#ifdef HAVE_INOTIFY
   struct dyndir *dd;
-
+#endif
+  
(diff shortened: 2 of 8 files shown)


yubiuser:
Should we add test to our test suite?

1) Add a DNS record with a short TTL (1sec) and use `use-stale-cache`. Query two times and see if it is returned from cache
2) Add a DNS record with a short TTL (1sec) and use `use-stale-cache=2`. Query two times (diff_t >2sec) and see if the query is send upstream again.
3) Add an existing but empty TXT records and see if `dnsmasq` returns `SERVFAULT`

DL6ER:
> Should we add test to our test suite?

No, we shouldn't start adding tests for `dnsmasq` features we do not touch to our testing suite. Instead, `dnsmasq` should get their own test suite. It can be very largely a copy-and-paste of the test suite we have for FTL, we should even be able to reuse the same containers. I guess what you want could be set up in less than an hour but I have the strong feeling that this should be happening in the `dnsmasq` realm. May this be on `pi-hole/dnsmasq` or somewhere else. But not here (`pi-hole/FTL`). TBH, we'd just add it here because it'd be easier to do (as the underlying machinery is already running).

yubiuser (approved):

PR #1501, opened 2022-12-09
Improve -vv and --help

DL6ER:

This PR adds colors (if supported by the terminal) to verbose version output and completely revamps the help function.
When the "consumer" is not a terminal (e.g. a pipe), no escape sequences will be added.

---

**`pihole-FTL -vv` BEFORE**

vv-old

**`pihole-FTL -vv` NOW**

vv-new

---

**`pihole-FTL -h` BEFORE**

help-old

**`pihole-FTL -h` NOW**

help-new
 
---

---

6. I am willing to help maintain this change if there are issues with it later.

## Checklist:

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -475,7 +475,7 @@ void parse_args(int argc, char* argv[])
 
 			printf("%sEmbedded SQLite3 shell:%s\n", yellow, normal);
 			printf("\t%ssql %s[-h]%s, %ssqlite3 %s[-h]%s        FTL's SQLite3 shell\n", green, gray, normal, green, gray, normal);
-			printf("\tthe optional %s-h%s starts a special %shuman-readable mode%s\n\n", gray, normal, bold, normal);
+			printf("\t%s-h%s starts a special %shuman-readable mode%s\n\n", gray, normal, bold, normal);
 
 			printf("    Usage: %spihole-FTL sqlite3 %s[-h] %s[OPTIONS] [FILENAME] [SQL]%s\n\n", green, gray, 	cyan, normal);
 			printf("    Options:\n\n");
--- a/src/args.c
+++ b/src/args.c
@@ -477,7 +477,7 @@ void parse_args(int argc, char* argv[])
 			printf("\t%ssql %s[-h]%s, %ssqlite3 %s[-h]%s        FTL's SQLite3 shell\n", green, gray, normal, green, gray, normal);
 			printf("\tthe optional %s-h%s starts a special %shuman-readable mode%s\n\n", gray, normal, bold, normal);
 
-			printf("    Usage: %spihole-FTL sqlite3 %s[OPTIONS] [FILENAME] [SQL]%s\n\n", green, cyan, normal);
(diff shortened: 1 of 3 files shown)


src/args.c
```diff
@@ -335,33 +428,79 @@ void parse_args(int argc, char* argv[])
...
+			const char *blue = cli_color(COL_BLUE);
+			const char *cyan = cli_color(COL_CYAN);
+			const char *gray = cli_color(COL_GRAY);
+			const char *green = cli_color(COL_GREEN);
+			const char *yellow = cli_color(COL_YELLOW);
+			const char *purple = cli_color(COL_PURPLE);
+
+			printf("%sThe Pi-hole FTL engine - %s%s\n\n", bold, get_FTL_version(), normal);
+			printf("Typically, pihole-FTL runs as a system service and is controlled\n");
+			printf("by %ssudo service pihole-FTL %s<action>%s where %s<action>%s is one out\n", green, purple, normal, purple, normal);
```
yubiuser, line 442: Should we go for `systemd`'s `systemctl` here? All supported distros run `systemd` by default.

DL6ER: `service` is a high-level wrapper script and common to all Linux'es. It wraps `/etc/init.d` scripts, `Upstart`'s `initctl` command, and `systemctl` as well. In several ways its similar to `apt` which is also just a wrapper script that will direct commands to the appropriate tool, automatically elevating itself with `sudo` if need-be. I think using `service` is appropriate here.

yubiuser (commented):
`--hash`, `d`,  handling of `--`(`dnsmasq` options), `-q` are missing from the help section. 

(Some more as well, but I think the above should be part of the help text)

rdwebdesign (commented):
This is one thing I don't like in codespell.

You had to change a perfectly correct line of code (and removed some text: "the optional") just to please the checker.

DL6ER:
Thanks for your comments. `-q` makes only sense in combination with `regex-test` which is why I mention it only there.

I added the missing bits with the exception of the `--` pass-through. It seems a dangerous idea and I don't think we have ever advertised to be able to do it. Really, all configuration should be done using config files, instead. It was added in [v5.0](https://github.com/pi-hole/FTL/commit/a30e5f3aff0e22b1030289260c00d5e9711f403a) when we added the drop-in binary support but this `--` was probably just a bad idea.

I'll leave the code here in case you still think we need it:
``` c

			printf("\t%s-- args...%s          Collect arguments behind %s--%s for\n", green, normal, green, normal);
			printf("\t                    the embedded dnsmasq server\n");
```

I also don't think we should always include all variants of an option, e.g, `--test` as alternative to `dnsmasq-test` as this may cause further confusion with the option option `test`. We can still leave it in there for any scripts that may be using it. 

Furthermore, I replaced `-- --help dhcp` and `-- --help dhcp6` by ``--list-dhcp` and `--list-dhcp6`, respectively. I checked Discourse, our docs and also asked Google whether the old syntax is mentioned anywhere and found no results. This isn't something you'd be using in a script so the change should be without any adverse consequences.

---

Updated screenshot:

Screenshot from 2022-12-10 13-57-31

yubiuser (approved):

PR #1509, opened 2022-12-21
Pi-hole FTL v5.20

PromoFaux:
Release PR


DL6ER (approved):

DL6ER (changes requested):
We should update dnsmasq from rc5 to the final release before pushing this, working on it now.

DL6ER (approved):

dschaper (approved):

PR #1537, opened 2023-03-11
Update embedded SQLite to 3.41.1

DL6ER:

Update the embedded SQLite3 engine. Pi-hole will benefit from improvements to the [PRAGMA integrity_check](https://sqlite.org/pragma.html#pragma_integrity_check) command and miscellaneous performance enhancements.

We do not add [SQLITE_ENABLE_STMT_SCANSTATUS](https://sqlite.org/compile.html#enable_stmt_scanstatus), the SQLite3 help says:
>  Those interfaces are normally omitted from the build **because they imposes a performance penalty**, even on statements that do not use the feature. 

### 2023-02-21 (3.41.0)

  1. Query planner improvements: 
    1. Make use of [indexed expressions](https://sqlite.org/expridx.html) within an aggregate query that includes a GROUP BY clause. 
    2. The query planner has improved awareness of when an index is a [covering index](https://sqlite.org/queryplanner.html#covidx) and adjusts predicted runtimes accordingly. 
    3. The query planner is more aggressive about using [co-routines](https://sqlite.org/optoverview.html#coroutines) rather than materializing subqueries and views. 
    4. Queries against the built-in table-valued functions [json_tree()](https://sqlite.org/json1.html#jtree) and [json_each()](https://sqlite.org/json1.html#jeach) will now usually treat "ORDER BY rowid" as a no-op. 
    5. Enhance the ability of the query planner to use [indexed expressions](https://sqlite.org/expridx.html) even if the expression has been modified by the [constant-propagation optimization](https://sqlite.org/optoverview.html#constprop). (See [forum thread 0a539c7](https://sqlite.org/forum/forumpost/0a539c76db3b9e29).) 
  2. Add the built-in [unhex()](https://sqlite.org/lang_corefunc.html#unhex) SQL function. 
  3. Add the base64 and base85 application-defined functions as an extension and include that extension in the [CLI](https://sqlite.org/cli.html). 
  4. Add the [sqlite3_stmt_scanstatus_v2()](https://sqlite.org/c3ref/stmt_scanstatus.html) interface. (This interface is only available if SQLite is compiled using [SQLITE_ENABLE_STMT_SCANSTATUS](https://sqlite.org/compile.html#enable_stmt_scanstatus).) 
  5. In-memory databases created using [sqlite3_deserialize()](https://sqlite.org/c3ref/deserialize.html) now report their filename as an empty string, not as 'x'. 
  6. Changes to the [CLI](https://sqlite.org/cli.html): 
    1. Add the new base64() and base85() SQL functions 
    2. Enhanced [EXPLAIN QUERY PLAN](https://sqlite.org/eqp.html) output using the new [sqlite3_stmt_scanstatus_v2()](https://sqlite.org/c3ref/stmt_scanstatus.html) interface when compiled using [SQLITE_ENABLE_STMT_SCANSTATUS](https://sqlite.org/compile.html#enable_stmt_scanstatus). 
    3. The "`.scanstats est`" command provides query planner estimates in profiles. 
    4. The continuation prompt indicates if the input is currently inside of a string literal, identifier literal, comment, trigger definition, etc. 
    5. Enhance the --safe command-line option to disallow dangerous SQL functions. 
    6. The [double-quoted string misfeature](https://sqlite.org/quirks.html#dblquote) is now disabled by default for CLI builds. Legacy use cases can reenable the misfeature at run-time using the "`.dbconfig dqs_dml on`" and "`.dbconfig dqs_ddl on`" commands. 
  7. Enhance the [PRAGMA integrity_check](https://sqlite.org/pragma.html#pragma_integrity_check) command so that it detects when text strings in a table are equivalent to but not byte-for-byte identical to the same strings in the index. 
  8. Enhance the [carray table-valued function](https://sqlite.org/carray.html) so that it is able to bind an array of BLOB objects. 
  9. Added the [sqlite3_is_interrupted()](https://sqlite.org/c3ref/interrupt.html) interface. 
  10. Long-running calls to [sqlite3_prepare()](https://sqlite.org/c3ref/prepare.html) and similar now invoke the [progress handler callback](https://sqlite.org/c3ref/progress_handler.html) and react to [sqlite3_interrupt()](https://sqlite.org/c3ref/interrupt.html). 
  11. The [sqlite3_vtab_in_first()](https://sqlite.org/c3ref/vtab_in_first.html) and [sqlite3_vtab_in_next()](https://sqlite.org/c3ref/vtab_in_first.html) functions are enhanced so that they reliably detect if they are invoked on a parameter that was not selected for multi-value IN processing using [sqlite3_vtab_in()](https://sqlite.org/c3ref/vtab_in.html). They return SQLITE_ERROR instead of SQLITE_MISUSE in this case. 
  12. The parser now ignores excess parentheses around a subquery on the right-hand side of an IN operator, so that SQLite now works the same as PostgreSQL in this regard. Formerly, SQLite treated the subquery as an expression with an implied "LIMIT 1". 
  13. Added the [SQLITE_FCNTL_RESET_CACHE](https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlresetcache) option to the [sqlite3_file_control()](https://sqlite.org/c3ref/file_control.html) API. 
  14. Makefile improvements: 
    1. The new makefile target "sqlite3r.c" builds an [amalgamation](https://sqlite.org/amalgamation.html) that includes the [recovery extension](https://sqlite.org/recovery.html). 
    2. New makefile targets "devtest" and "releasetest" for running a quick developmental test prior to doing a check-in and for doing a full release test, respectively. 
  15. Miscellaneous performance enhancements. 

### 2023-03-10 (3.41.1)

  1. Provide compile-time options -DHAVE_LOG2=0 and -DHAVE_LOG10=0 to enable SQLite to be compiled on systems that omit the standard library functions log2() and log10(), repectively. 
  2. Ensure that the datatype for column t1.x in "`CREATE TABLE t1 AS SELECT CAST(7 AS INT) AS x;`" continues to be INT and is not NUM, for historical compatibility. 
  3. Enhance [PRAGMA integrity_check](https://sqlite.org/pragma.html#pragma_integrity_check) to detect when extra bytes appear at the end of an index record. 
  4. Fix various obscure bugs reported by the user community. See the [timeline of changes](https://sqlite.org/src/timeline?from=version-3.41.0&to=version-3.41.1) for details. 

---

6. I am willing to help maintain this change if there are issues with it later.

## Checklist:

### DIFF
--- a/src/database/shell.c
+++ b/src/database/shell.c
@@ -126,6 +126,8 @@ typedef unsigned char u8;
 #endif
 #include <ctype.h>
 #include <stdarg.h>
+// print_FTL_version()
+#include "../log.h"
 
 #if !defined(_WIN32) && !defined(WIN32)
 # include <signal.h>
@@ -26943,7 +26945,7 @@ static char *cmdline_option_value(int argc, char **argv, int i){
 #endif
 
 #if SQLITE_SHELL_IS_UTF8
-int SQLITE_CDECL main(int argc, char **argv){
+int SQLITE_CDECL sqlite3_shell_main(int argc, char **argv){
 #else
 int SQLITE_CDECL wmain(int argc, wchar_t **wargv){
   char **argv;
@@ -27465,6 +27467,7 @@ int SQLITE_CDECL wmain(int argc, wchar_t **wargv){
       char *zHome;
       char *zHistory;
       int nHistory;
+      print_FTL_version();
       printf(
         "SQLite version %s %.19s\n" /*extra-version-info*/
         "Enter \".help\" for usage hints.\n",
--- a/src/database/shell.c
+++ b/src/database/shell.c
@@ -39,7 +39,7 @@ typedef unsigned short int u16;
 
 /*
 ** Optionally #include a user-defined header, whereby compilation options
-** may be set prior to where they take effect, but after platform setup. 
+** may be set prior to where they take effect, but after platform setup.
 ** If SQLITE_CUSTOM_INCLUDE=? is defined, its value names the #include
 ** file. Note that this macro has a like effect on sqlite3.c compilation.
 */
@@ -126,12 +126,10 @@ typedef unsigned char u8;
 #endif
 #include <ctype.h>
 #include <stdarg.h>
-// print_FTL_version()
-#include "../log.h"
 
 #if !defined(_WIN32) && !defined(WIN32)
 # include <signal.h>
-# if !defined(__RTP__) && !defined(_WRS_KERNEL)
+# if !defined(__RTP__) && !defined(_WRS_KERNEL) && !defined(SQLITE_WASI)
 #  include <pwd.h>
 # endif
 #endif
@@ -186,6 +184,14 @@ typedef unsigned char u8;
 # define SHELL_USE_LOCAL_GETLINE 1
 #endif
 
+#ifndef deliberate_fall_through
+/* Quiet some compilers about some of our intentional code. */
+# if defined(GCC_VERSION) && GCC_VERSION>=7000000
+#  define deliberate_fall_through __attribute__((fallthrough));
+# else
+#  define deliberate_fall_through
+# endif
+#endif
 
 #if defined(_WIN32) || defined(WIN32)
 # if SQLITE_OS_WINRT
@@ -212,7 +218,7 @@ typedef unsigned char u8;
  /* Make sure isatty() has a prototype. */
  extern int isatty(int);
 
-# if !defined(__RTP__) && !defined(_WRS_KERNEL)
+# if !defined(__RTP__) && !defined(_WRS_KERNEL) && !defined(SQLITE_WASI)
   /* popen and pclose are not C89 functions and so are
   ** sometimes omitted from the <stdio.h> header */
    extern FILE *popen(const char*,const char*);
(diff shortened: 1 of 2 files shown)


DL6ER:
Codespell needs to be overwritten, this is no error and not even related to this PR:

Screenshot from 2023-03-11 13-17-49

yubiuser:
> Added the [sqlite3_is_interrupted()](https://sqlite.org/c3ref/interrupt.html) interface.

Could this be something for v6 - if users select a very long time range and query runs longer than 5 seconds display a modal and offering cancleation? 

> The [double-quoted string misfeature](https://sqlite.org/quirks.html#dblquote) is now disabled by default for CLI builds.

Are we affected by this? Did you check that we don't use double-quotes string as string literals?

DL6ER:
> > Added the [sqlite3_is_interrupted()](https://sqlite.org/c3ref/interrupt.html) interface.
>
> Could this be something for v6 - if users select a very long time range and query runs longer than 5 seconds display a modal and offering cancleation?

It is actually not trivial at all to do this. The thread that does the long computation is blocking. The interrupt routines are meant for the case when exactly this thread gets interrupted, e.g., by <kbd>Ctrl</kbd> + <kbd>C</kbd>. This is not something we could do (or even emulate) over a HTTP connection.

So, the only thing we could do is sending such a cancellation request via *another* HTTP connection but how does it know *which* process to cancel (there may be several long-running queries at the same time)? We'd need to somehow maintain an array with database pointers in global memory with some smart IDs usable to cancel requests. This would be ... let's say ... not a very trivial endeavour. I also see at least one possible race condition with a delayed cancellation request for some query that is already finished so garbage collection on this global array is another complex thing.

> > The [double-quoted string misfeature](https://sqlite.org/quirks.html#dblquote) is now disabled by default for CLI builds.
>
> Are we affected by this? Did you check that we don't use double-quotes string as string literals?

We are not affected. We disabled this misfeature already [back in 2019](https://github.com/pi-hole/FTL/commit/61157ebfa9f32cf41336dbb0cfc6abb06429ff34) *intentionally*. It just now became the default for the CLI `sqlite3` command compilation. This doesn't change anything on your side as our compile-time options already enforced this also in the CLI application (and not a single user ever complained about this).

yubiuser (approved):

PR #1541, opened 2023-03-19
Fix spell checker action

DL6ER:

## 10

---

Fixes the spell checker action by adding a few unknown words to the ignore dictionary. Can be verified locally using
```
act -j spell-check
```

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,4 +1,0 @@
-ede
-edn
-nd
-tre
--- a/.github/.codespellignore
+++ b/.github/.codespellignore
@@ -1,0 +1,4 @@
+ssudo
+tre
+ede
+nd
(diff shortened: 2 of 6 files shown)


yubiuser (dismissed):
(Re)move the already existing file?

https://github.com/pi-hole/FTL/blob/master/.codespellignore

DL6ER:
For whatever reason, I really did not see this file... Force-pushed away. Having it inside `.github` seems cleaner compared to cluttering the root dir. In the end it is not part of FTL but meant for the CI.

yubiuser (commented):
@PromoFaux was implementing this at the beginning. I think he discovered an issue with `spellcheck` where the ignorelist ignored items unless they were written all lower-case. I don't know if something changed on their side, but to be safe I would go for all lower-case. (which removes the duplicate `EDE/ede`) as well.

DL6ER:
Done

PR #1551, opened 2023-04-07
Use EDE data when "proxy-dnssec" is used

DL6ER:

## 10

---

When the `dnsmasq` option `proxy-dnssec`

> **proxy-dnssec**
> Copy the DNSSEC Authenticated Data bit from upstream servers to downstream clients. This is an alternative to having dnsmasq validate DNSSEC, but it depends on the security of the network between dnsmasq and the upstream servers, and the trustworthiness of the upstream servers. Note that caching the Authenticated Data bit correctly in all cases is not technically possible. If the AD bit is to be relied upon when using this option, then the cache should be disabled using `--cache-size=0`. In most cases, enabling DNSSEC validation within dnsmasq is a better option. See `--dnssec` for details. 

is used, FTL uses the AD bit to determine the `INSECURE`/`SECURE` status of queries. In cases where the upstream resolver includes EDE data for errors (e.g. `unbound 1.16.0+`), FTL also analyses this data to tell apart `SERVFAIL`s caused by a "real" server failure from crafted `SERVFAIL`s due to failed DNSSEC validation.

See linked Discourse Discussion for further details and extensive testing done by @jpgpi250 

---

To get `proxy-dnssec` working, you'll need to add
```
edns: yes
```
to `unbound`s `server` section and
```
proxy-dnssec
add-cpe-id=01234
```
to a custom `dnsmasq` config file. The `add-cpe-id` options is necessary to query EDE data from the upstream destination in case a client does not signal EDNS0 support.

---

All in all, this PR also improves the internal handling of DNSSEC status and, e.g., ensures ignored queries (because they are duplicated) are not stamped `INSECURE` when this is not actually true (they are actually "nothing" as they are ignored). This makes this PR an enhancement even for the vast majority that will never use `proxy-dnssec`.

### DIFF
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -128,8 +128,6 @@ void FTL_hook(unsigned int flags, const char *name, union all_addr *addr, char *
 		if(!config.show_dnssec)
 			return;
 
-		const ednsData edns = { 0 };
-
 		// Type is overloaded with port since 2d65d55, so we have to
 		// derive the real query type from the arg string
 		unsigned short qtype = type;
@@ -158,7 +156,7 @@ void FTL_hook(unsigned int flags, const char *name, union all_addr *addr, char *
 			arg = (char*)"dnssec-unknown";
 		}
 
-		_FTL_new_query(flags, name, NULL, arg, qtype, id, &edns, INTERNAL, file, line);
+		_FTL_new_query(flags, name, NULL, arg, qtype, id, INTERNAL, file, line);
 		// forwarded upstream (type is used to store the upstream port)
 		FTL_forwarded(flags, name, addr, type, id, path, line);
 	}
@@ -464,7 +462,7 @@ static bool is_pihole_domain(const char *domain)
 bool _FTL_new_query(const unsigned int flags, const char *name,
                     union mysockaddr *addr, char *arg,
                     const unsigned short qtype, const int id,
-                    const ednsData *edns, const enum protocol proto,
+                    const enum protocol proto,
                     const char* file, const int line)
 {
 	// Create new query in data structure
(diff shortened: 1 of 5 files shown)


jpgpi250:
ftl crash, using this branch, see https://discourse.pi-hole.net/t/dnssec-discussion-support-for-proxy-dnssec/62217/12 for details, FTL.log with debug flags enabled attached in the topic.

jpgpi250:
build failure?

DL6ER:
PR ready for review + merge. PR text edited to reflect the lastest state

src/dnsmasq_interface.c
```diff
@@ -2171,6 +2184,10 @@ static void FTL_reply(const unsigned int flags, const char *name, const union al
 		//   name = google-public-dns-a.google.com
 		// Hence, isExactMatch is always false
 
+		// Set DNSSEC status to INSECURE if it is still unknown
```
yubiuser, line 2187: Why do you have the exact same code block here and around L2154?

jpgpi250:
don't know if this can be / needs correcting.
If a domain is blocked by gravity or domainlist entry, the dnssec value is set to 2 (insecure). since no upstream query has been submitted, I think it makes more sense to set the dnssec value to 0 (unknown).
This **doesn't affect the web interface** (status isn't shown for blocked entries, however, it **falsifies reports**, much more insecure, hardly any unknown entries.

NOT a big deal, but if it can be changed easily...

examples:
image

image

image

src/dnsmasq_interface.c
DL6ER: This part of the code handles reserve (PTR) lookups, L2154 is the code region where "forward" queries (such as A, AAAA, ...) are handled

DL6ER:
@jpgpi250 This is expected and agreement with queries in "normal" DNSSEC mode. Queries are either `SECURE` (= DNSSEC signed, validation successful), `BOGUS` (= DNSSEC signed, validation failed) or `INSECURE` (no DNSSEC signature). As gravity and all locally defined records are not DNSSEC signed, stamping them `INSECURE` is the logical option. This was previously broken but will now be fixed by this PR.

yubiuser (approved):

PR #1559, opened 2023-05-14
Add list parsing feature for gravity

DL6ER:

This PR implements a list parsing feature into FTL that is intended to be used with gravity. Its main advantage is a stream-lined approach directly writing everything directly into the new database during analysis. There is no detour over huge allocated memory blocks or temporary files.

---

## Checklist:

### DIFF
--- a/src/gravity-tools.c
+++ b/src/gravity-tools.c
@@ -50,12 +50,13 @@ int gravity_parseList(const char *infile, const char *outfile, const char *adlis
 	const char *info = cli_info();
 	const char *tick = cli_tick();
 	const char *cross = cli_cross();
+	const char *over = cli_over();
 
 	// Open input file
 	FILE *fpin = fopen(infile, "r");
 	if(fpin == NULL)
 	{
-		printf("\r  %s Unable to open %s for reading\n", cross, infile);
+		printf("%s  %s Unable to open %s for reading\n", over, cross, infile);
 		return EXIT_FAILURE;
 	}
 
@@ -64,7 +65,7 @@ int gravity_parseList(const char *infile, const char *outfile, const char *adlis
 	sqlite3_stmt *stmt = NULL;
 	if(sqlite3_open_v2(outfile, &db, SQLITE_OPEN_READWRITE, NULL) != SQLITE_OK)
 	{
-		printf("\r  %s Unable to open database file %s for writing\n", cross, outfile);
+		printf("%s  %s Unable to open database file %s for writing\n", over, cross, outfile);
 		fclose(fpin);
 		return EXIT_FAILURE;
 	}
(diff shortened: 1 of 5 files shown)


rdwebdesign (dismissed):
Tested inside a container and it is working as expected.

rdwebdesign (approved):

src/gravity-tools.c
```diff
@@ -0,0 +1,402 @@
...
+		total_read += read;
+		lineno++;
+
+		// Remove leading whitespace
+		// isspace() matches spaces, tabs, form feeds, line breaks,
+		// carriage returns, and vertical tabs
+		while(isspace(line[0]))
+			line++;
+
+		// Skip comments
```
yubiuser, line 166: Here is some duplicate code L166-186. This is still part of `gravity.sh` 
`gravity_ParseFileIntoDomains()`

https://github.com/pi-hole/pi-hole/blob/73733308ba0f1c4e2cc5fb0c4aac6a9027349e48/gravity.sh#L639-L664

Does not hurt to keep it, just wanted to report.

DL6ER: If it is duplicated, we should remove it from FTL for minimal speed gain.

DL6ER: Actually, rethinking this, we can also remove `gravity_ParseFileIntoDomains()` from `gravity.sh` and use the code here that should be a lot faster.

yubiuser: Then this would be something for https://github.com/pi-hole/pi-hole/pull/5275

rdwebdesign:
@DL6ER 

I think I approved too soon.

I noticed the output here is saying `ignored 13 non-domain entries`, but only one entry is shown. 

```
  [i] Target: https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
  [✓] Status: Retrieval successful
  [✓] Parsed 183821 exact domains and 0 ABP-style domains (ignored 13 non-domain entries)
      Sample of non-domain entries:
        - "0.0.0.0"
```

Comparing to `master`, should be only 1 entry.
I think the other 12 lines are in the "false positive" list, but they are still counted.

DL6ER:
@rdwebdesign This is expected and correct behavior. Each non-domain is counted but then only unique ones are shown. So, if there are 13 entries with `0.0.0.0` this is exactly what you are seeing here:

```
  [✓] Parsed 183821 exact domains and 0 ABP-style domains (ignored 13 non-domain entries)
      Sample of non-domain entries:
        - "0.0.0.0"
```

I'm always open to improving wording in case this isn't clear right now.

yubiuser:
Maybe simply changing the message to

"Sample of non-domain entries (duplicates removed)"

PR #1566, opened 2023-05-27
Remove code duplication found in gravit.sh gravity_ParseFileIntoDomains()

yubiuser:

With the transition of `parseList()` from `gravity.sh` to FTL realm some code duplication occurred. Duplicated code is from `gravity_ParseFileIntoDomains()`

https://github.com/pi-hole/pi-hole/blob/50d67dce73e122d98026ab2a4b30255efd9b05fc/gravity.sh#L639-L667

No all input modifications were duplicated into FTL so we could not remove `gravity_ParseFileIntoDomains` completely. 

Removes the duplicates

P.S. It would also be possible to move the whole `gravity_ParseFileIntoDomains` also into FTL's realm, however this would require more modifications (in `core` and `FTL`

---

---

### DIFF
--- a/src/tools/gravity-parseList.c
+++ b/src/tools/gravity-parseList.c
@@ -29,9 +29,6 @@
 // See https://github.com/pi-hole/pi-hole/pull/5240
 #define ABP_DOMAIN_REXEX "\\|\\|"SUBDOMAIN_PATTERN"*"TLD_PATTERN"\\^"
 
-// Detects ABP extended CSS selectors ("##", "#!#", "#@#", "#?#") preceded by a letter
-#define ABP_CSS_SELECTORS "[a-z]#[$?@]{0,1}#"
-
 // A list of items of common local hostnames not to report as unusable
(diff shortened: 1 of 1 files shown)


src/tools/gravity-parseList.c
```diff
@@ -219,13 +199,6 @@ int gravity_parseList(const char *infile, const char *outfile, const char *adlis
 		{
 			// ABP-style match (see comments above)
 
-			// Skip lines containing ABP extended CSS selectors
-			// ("##", "#!#", "#@#", "#?#") preceded by a letter
-			// See https://github.com/pi-hole/pi-hole/pull/5247 for
-			// further information on why this is necessary
-			if(regexec(&abp_css_regex, line, 1, &match, 0) == 0)
```
DL6ER, line 226: If this is removed we should also remove the corresponding definition (`#define ...`) and the variable + compilation of the regex.

DL6ER (approved):

PR #1571, opened 2023-05-30
Fix spelling in v6

yubiuser:
Satisfy `codespell` in `FTLv6`.

Fix sepelling and set some external resources to be excluded from checking


test/api/libs/responseVerifyer.py
```diff
@@ -82,7 +82,7 @@ def verify_endpoint(self, endpoint: str):
 			elif 'application/zip' in content:
 				expected_mimetype = 'application/zip'
 				jsonData = content[expected_mimetype]
-				# Thie endpoint requires HEADER authentication
+				# The endpoint requires HEADER authentication
```
rdwebdesign, line 85: Just a comment:

I think the intended text was `This`, but I don't think this will change something here.
The same for line 129.

PR #1586, opened 2023-07-11
Fix the Stale workflow

PromoFaux:
See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -32,8 +32,8 @@ jobs:
   remove_stale:
     # trigger "stale" removal immediately when stale issues are commented on
     # we need to explicitly check that the trigger does not run on comment on a PR as
-    # 'issue_comment' triggers on issues AND PR comments
-    if: github.event_name == 'issue_comment' && ${{ !github.event.issue.pull_request }}
+    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only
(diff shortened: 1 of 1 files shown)


DL6ER (approved):

yubiuser:
Looking at the triggers and the conditionals, I think this is still triggered on schedule.

PromoFaux:
Ah, yes - you're right. I was conflating two separate issues with the config.

I also noted that it was running against Pull requests, too (which this fixes). I'll rebase on `master` once https://github.com/pi-hole/FTL/pull/1587 is merged, as that also contains the commit to fix the codespell issue

yubiuser:
>  (which this fixes)

Did you try this? I'm not sure, as we have now

`if: github.event_name == 'issue_comment' && ${{ !github.event.issue.pull_request }}`

which also contains the part `${{ !github.event.issue.pull_request }}` which will be left after your PR. Already the current state should not (but indeed does) trigger on PRs....

PromoFaux:
I did, but then I deleted the comment. Here, I did it again:

https://github.com/pi-hole/docker-pi-hole/actions/runs/5536224129 - which was triggered by: https://github.com/pi-hole/docker-pi-hole/pull/1366#issuecomment-1633184606

DL6ER (approved):

yubiuser (approved):

PR #1605, opened 2023-07-28
Fix Query Log total number without filtering

DL6ER:

## 10

---

Tweak criterion we use for deciding if we are filtering on the Query Log or not. This is necessary as DataTables also sends a query string if no filters are applied. This is a direct consequence of PR #1599 which missed this special case

### DIFF
--- a/src/api/queries.c
+++ b/src/api/queries.c
@@ -367,7 +367,11 @@ int api_queries(struct ftl_conn *api)
 		                       sqlite3_errstr(rc));
 	}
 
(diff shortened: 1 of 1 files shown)


PromoFaux:
No filter, but the response has the filtered count in it. 

image

PromoFaux:
Disregard last - I think I built the container without `--no-cache` tested again and it is fine 

image

PromoFaux (approved):

PR #1607, opened 2023-07-28
Provide regex ID for API

DL6ER:

## 10

---

See title. This was so far missing.

### DIFF
--- a/src/api/docs/content/specs/queries.yaml
+++ b/src/api/docs/content/specs/queries.yaml
@@ -210,7 +210,7 @@ components:
                     description: Time until the response was received (ms, negative if N/A)
               regex_id:
                 type: integer
-                description: ID of blocking regex (`NULL` if N/A)
+                description: ID of regex (`NULL` if N/A)
                 nullable: true
               upstream:
                 type: string
--- a/src/api/docs/content/specs/queries.yaml
+++ b/src/api/docs/content/specs/queries.yaml
@@ -208,12 +208,10 @@ components:
                   time:
                     type: number
                     description: Time until the response was received (ms, negative if N/A)
-              ttl:
-                type: integer
-                description: Query's Time-To-Live (TTL) value (`-1` if N/A)
               regex_id:
                 type: integer
-                description: ID of blocking regex (`-1` if N/A)
+                description: ID of blocking regex (`NULL` if N/A)
+                nullable: true
               upstream:
                 type: string
                 description: IP or name + port of upstream server
@@ -231,8 +229,7 @@ components:
               reply:
                 type: "IP"
                 time: 19
-              ttl: -1
-              regex_id: -1
+              regex_id: NULL
               upstream: "localhost#5353"
               dbid: 112421354
             - time: 1581907871.583821
@@ -247,8 +244,7 @@ components:
               reply:
                 type: "IP"
                 time: 12.3
-              ttl: 7
-              regex_id: -1
+              regex_id: NULL
               upstream: "localhost#5353"
               dbid: 112421355
         cursor:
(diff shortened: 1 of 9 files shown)


DL6ER:

yubiuser (commented):
Why did you drop all the `ttl` stuff?

You're creating a new column in the database for the `regex_id`. So far, this was part of `additional_info`. Why did you change that? Do you plan to migrate the existing data from `additional_info`?

DL6ER:
> Why did you drop all the `ttl` stuff?

Pi-hole handles queries by "question" (a single domain) but TTL applies to replies and there may actually be multiple. Plus things get really difficult to disentangle when CNAMEs are there as well and even worse when a query is partially replied to from cache. This little extra bit of information is hard to get, not always unique and overall complicates code too much.

That's why I decided to drop it - this code was never used (it was commented out before).

---

>  You're creating a new column in the database for the `regex_id`. So far, this was part of `additional_info`. Why did you change that? Do you plan to migrate the existing data from `additional_info`?

Yes, it is in `additional_info` but only *most of the time* not `always`. The information about the regex ID is actually lost when blocking happens during deep CNAME inspection because we then use `additional_info` to store the domains that actually led to blocking in this case. This issue can only be resolved when adding a new column. Fortunately, this will not affect the database size by much as empty columns take only one byte per row. Integers up to 127 use two bytes per row, integers up to 32767 take three bytes and so on. This seems affordable given that the database indices will typically be rather small here.

src/database/query-table.c
```diff
@@ -698,6 +698,38 @@ bool add_query_storage_columns(sqlite3 *db)
...
+	                       "CASE typeof(client) WHEN 'integer' THEN (SELECT ip FROM client_by_id c WHERE c.id = q.client) ELSE client END client,"
+	                       "CASE typeof(forward) WHEN 'integer' THEN (SELECT forward FROM forward_by_id f WHERE f.id = q.forward) ELSE forward END forward,"
+	                       "CASE typeof(additional_info) WHEN 'integer' THEN (SELECT content FROM addinfo_by_id a WHERE a.id = q.additional_info) ELSE additional_info END additional_info, "
+	                       "reply_type, reply_time, dnssec, regex_id "
+	                       "FROM query_storage q");
+
+	// Update database version to 13
+	if(!db_set_FTL_property(db, DB_VERSION, 13))
+	{
+		log_err("add_query_storage_columns2(): Failed to update database version!");
```
yubiuser, line 723:
```suggestion
		log_err("add_query_storage_column_regex_id(): Failed to update database version!");
```

yubiuser:
Needs documentation changes in https://deploy-preview-338--pihole-docs.netlify.app/database/ftl/

DL6ER:
@yubiuser See https://github.com/pi-hole/docs/pull/928

yubiuser (dismissed):

yubiuser:
> The information about the regex ID is actually lost when blocking happens during deep CNAME inspection because we then use additional_info to store the domains that actually led to blocking in this case. This issue can only be resolved when adding a new column.

Is this implemented with this PR as well?
I tried a quick check by setting up a local CNAME and used a deny regex to block the CNAME. It's not stored in the database nor shown in the web interface as blocked, but only as CNAME.
Or is this expected as it is a local CNAME not a CNAME from upstream?

src/api/docs/content/specs/queries.yaml
```diff
@@ -208,12 +208,10 @@ components:
                   time:
                     type: number
                     description: Time until the response was received (ms, negative if N/A)
-              ttl:
-                type: integer
-                description: Query's Time-To-Live (TTL) value (`-1` if N/A)
               regex_id:
                 type: integer
-                description: ID of blocking regex (`-1` if N/A)
+                description: ID of blocking regex (`NULL` if N/A)
```
yubiuser, line 213: Does it only save the regex_id in case it's a blocking regex? Are allowing regex ignored here?

yubiuser (approved):

PR #1609, opened 2023-07-29
Tweaks for the webserver port

DL6ER:

## 10

---

1. Enable SSL/TLS by default (FTL can generate a self-signed certificate with the strongest ECC curve of the NSA Suite B by itself on demand)
2. Initialized webserver ports dynamically depending on whether ports 80 and 443 are already taken by other services on the machine. This initialization is only done once - when either importing from a v5.x installation (`setupVars.conf`, `pihole-FTL.conf` and friends) or when this is a completely fresh installation. The result of this is printed to the log file (which ports are being used and written to the config file)

I have only been able to do some superficial testing myself, but it should be rather easy once an appropriate `docker` container is available.


DL6ER:

yubiuser:
This will need some changes on various places in the code to change the default (sometimes only comments) from 8080 > 80. You could cherry-pick and adjust from https://github.com/pi-hole/FTL/pull/1608

DL6ER:
I `cherry-pick`ed the mentioned commit but reverted the changes to [src/webserver/civetweb/civetweb.c](https://github.com/pi-hole/FTL/pull/1608/files#diff-4955b6ae5e480166d688757a1514d00d4ef3ef683465954e233955ec03a2b1d3) as this is an external file where changing the comments would anyway be undone in future upgrades.

yubiuser (approved):

PR #1646, opened 2023-09-23
Add performance tests and prevent online-brute forcing

DL6ER:

Add `pihole-FTL --perf` suitable to measure both memory- as well as compute-performance in the context of password hashing.

Context (source [Wikipedia EN](https://en.wikipedia.org/wiki/Balloon_hashing)):
> Balloon hashing is a [key derivation function](https://en.wikipedia.org/wiki/Key_derivation_function) presenting proven memory-hard password-hashing and modern design. It was created by [Dan Boneh](https://en.wikipedia.org/wiki/Dan_Boneh), Henry Corrigan-Gibbs (both at [Stanford University](https://en.wikipedia.org/wiki/Stanford_University)) and Stuart Schechter ([Microsoft Research](https://en.wikipedia.org/wiki/Microsoft_Research)) in 2016. It is a recommended function in [NIST password guidelines](https://en.wikipedia.org/wiki/Password_policy#NIST_guidelines).
>
> The authors claim that Balloon:
>
> - has proven [memory-hardness](https://en.wikipedia.org/wiki/Memory-hard_function) properties,
> - is built from standard primitives: it can use any standards non-space-hard [cryptographic hash function](https://en.wikipedia.org/wiki/Cryptographic_hash_function) as a sub-algorithm (e.g., [SHA-3](https://en.wikipedia.org/wiki/SHA-3), [SHA-512](https://en.wikipedia.org/wiki/SHA-512)),
 > - is resistant to [side-channel attacks](https://en.wikipedia.org/wiki/Side-channel_attack): the memory access pattern is independent of the data to be hashed,
 > - is easy to implement and matches the performance of similar algorithms.

Pi-hole FTL uses **BALLOON** starting in version v6.0 in combination with
- salting of **128 bits cryptographically-secure randomness** to prevent any rainbow tables, and
- the well-known **secure hashing algorithm 2** with 256 bits (SHA256) to ensure the one-way character of the resulting hash.

Even the best designed algorithm has the issue than an attacker - having once gotten his hands on the stored hash - can offline* brute-force the password until he finds a match. This new feature may be used to find the sweat-spot among the variety of devices on which users are running Pi-hole.

The current default values (s = 1024, t = 32) takes roughly 42 milliseconds on my x86_64 microserver.

*) Online brute-forcing is prevented in FTL due to (a) a certain amount of time spent in the BALLOON routine and (b) a limited number of concurrent API threads. To be on the safe side, FTL furthermore rate-limits login attempts (read as: password validation attempts) to at most 3 per second (compile-time constant).

---

## Checklist:

### DIFF
--- a/src/api/auth.c
+++ b/src/api/auth.c
@@ -509,7 +509,8 @@ int api_auth(struct ftl_conn *api)
 	// else: Login attempt
 	// - Client tries to authenticate using a password, or
 	// - There no password on this machine
-	if(empty_password ? true : verify_password(password, config.webserver.api.pwhash.v.s))
+	const enum password_result result = empty_password ? true : verify_password(password, config.webserver.api.pwhash.v.s, true);
+	if(result == PASSWORD_CORRECT)
 	{
 		// Accepted
 
@@ -603,6 +604,15 @@ int api_auth(struct ftl_conn *api)
 		{
 			log_warn("No free API seats available, not authenticating client");
 		}
+	}
+	else if(result == PASSWORD_RATE_LIMITED)
+	{
+		// Rate limited
+		log_debug(DEBUG_API, "API: Login attempt rate-limited");
+		return send_json_error(api, 429,
+					"too_many_requests",
+					"Too many requests",
+					NULL);
 	}
 	else
 	{
--- a/src/api/auth.c
+++ b/src/api/auth.c
@@ -608,11 +608,10 @@ int api_auth(struct ftl_conn *api)
 	else if(result == PASSWORD_RATE_LIMITED)
 	{
 		// Rate limited
-		log_debug(DEBUG_API, "API: Login attempt rate-limited");
 		return send_json_error(api, 429,
 					"too_many_requests",
 					"Too many requests",
-					NULL);
+					"login rate limiting");
 	}
 	else
 	{
--- a/src/api/config.c
+++ b/src/api/config.c
@@ -287,7 +287,7 @@ static const char *getJSONvalue(struct conf_item *conf_item, cJSON *elem, struct
 			char *pwhash = strlen(elem->valuestring) > 0 ? create_password(elem->valuestring) : strdup("");
 
 			// Verify that the password hash is valid
-			const bool verfied = verify_password(elem->valuestring, pwhash);
+			const bool verfied = verify_password(elem->valuestring, pwhash, false) == PASSWORD_CORRECT;
 
 			if(!verfied)
 			{
--- a/src/api/docs/content/specs/auth.yaml
+++ b/src/api/docs/content/specs/auth.yaml
@@ -79,6 +79,14 @@ components:
                   allOf:
                     - $ref: 'common.yaml#/components/errors/unauthorized'
                     - $ref: 'common.yaml#/components/schemas/took'
+          '429':
+            description: Too Many Requests
+            content:
+              application/json:
+                schema:
+                  allOf:
+                    - $ref: 'common.yaml#/components/errors/too_many_requests'
+                    - $ref: 'common.yaml#/components/schemas/took'
       delete:
         summary: Delete session
         tags:
--- a/src/api/docs/content/specs/common.yaml
+++ b/src/api/docs/content/specs/common.yaml
@@ -53,6 +53,26 @@ components:
               nullable: true
               description: "No additional data available"
               example: null
+    too_many_requests:
+      type: object
+      description: "Too many requests (rate limiting)"
+      properties:
+        error:
+          type: object
+          properties:
+            key:
+              type: string
+              description: "Machine-readable error type"
+              example: "too_many_requests"
+            message:
+              type: string
+              description: "Human-readable error message"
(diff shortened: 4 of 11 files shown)


yubiuser:
Can you explain the use of this new feature a bit more? Who is to run this performance test when? What should users do with the result?

> This new feature may be used to find the sweat-spot among the variety of devices on which users are running Pi-hole.

What sweet-spot do you mean? Do you want to gather user results and adjust FTL code?

DL6ER:
My plan is to collect this data from as many different machines as possible. So I'm also asking you to provide the data, however, as soon as the feature lands in `development-v6`, it will be easier to ask a wider audience.

And yes, the goal is to collect numbers, identify a sweet-spot (as in: sufficient but not too painful) and then, subsequently, change the FTL code to use this new default. The settings can be changed at any point in time. FTL can harden existing installations at any time (during a successful login).

The difficulty in finding the sweet spot is that an offline-attacker can obviously use much beefier hardware than a Raspberry Pi so it needs to be "painful enough".

DL6ER:
As said elsewhere, in addition to any reviews, I'm very interested in anyone trying out this branch running `pihole-FTL --perf` and reporting their results alongside a brief sketch of their used hardware (processor).

My laptop featuring a `AMD Ryzen 7 4800H` (64 bit binary):
```
Time-performance index:    2191.1 it/s (s = 1024)
Space-performance index:   2922.7 it/s (t = 256)
```
and the same with the 32 bit binary:
```
Time-performance index:    2049.2 it/s (s = 1024)
Space-performance index:    825.3 it/s (t = 256)
```

My tiny little microserver featuring an `AMD Ryzen 5 4500U` (64 bin binary only):
```
Time-performance index:    2071.7 it/s (s = 1024)
Space-performance index:   2770.1 it/s (t = 256)
```

And finally my other server (`AMD EPYC 7502P 32-Core Processor`):
```
Time-performance index:    1727.9 it/s (s = 1024)
Space-performance index:   2307.4 it/s (t = 256)
```

rdwebdesign:
@DL6ER 

Tested on a _Raspberry Pi 3B+_:
```
Time-performance index:    1858.9 it/s (s = 1024)
Space-performance index:    232.2 it/s (t = 256)

Total test time: 17.6 seconds
```

yubiuser:
`Nanopi Neo Plus 2`

```
Average time-performance index:      58543 +/- 86 (s = 512)
Average space-performance index:     58572 +/- 69 (t = 512)

Total test time: 17.6 seconds

```

yubiuser:
Somehow, this branch breaks my web interface. When I access `hostname/admin` it redirects to `/admin/login?target=%2fadmin` and I only see a white page, nothing else. (As I run two Pi-holes on my network, this one is accessed by hostname and not `pi.hole`)

DL6ER:
The spellcheck issue will be fixed in https://github.com/pi-hole/FTL/pull/1649 by https://github.com/pi-hole/FTL/pull/1649/commits/28358d45aa597f489c9bc054ea03b71638017c11

yubiuser (commented):
I'm not exactly sure which line is causing the logging of 

```
2023-10-07 23:46:27.971 [252938/T252939] INFO: Compiled 1 allow and 1 deny regex for 0 client in 1.5 msec
2023-10-07 23:46:42.997 [252938/T252946] WARNING: API: Too many requests
2023-10-07 23:46:45.092 [252938/T252946] WARNING: API: Too many requests
2023-10-07 23:46:47.185 [252938/T252946] WARNING: API: Too many requests
2023-10-07 23:46:51.171 [252938/T252949] WARNING: API: Too many requests
```

but I think the message should be improved to `WARNING: API: Too many login requests (rate limiting)`

Or is the limit a general one for the API (not only for login?) If so, 3 might be not enough, as the web interface (and/or PADD) already send more/second.

What happens within the 250ms of sleep? Is `FTL` still working? Answering DNS requests? Allow non-login API requests (from already authenticated devices?)

DL6ER:
> but I think the message should be improved to `WARNING: API: Too many login requests (rate limiting)`

The message is a general one for the API even if nothing else than the login currently is rate-limited. I added a `hint` into the JSON which will make the logged line look like:

`WARNING: API: Too many requests (login rate limiting)`

> Or is the limit a general one for the API (not only for login?) If so, 3 might be not enough, as the web interface (and/or PADD) already send more/second.

No, this is only for the login and there isn't any rate limiting if you have no password at all.

> What happens within the 250ms of sleep?

Nothing, the current web interface worker is intentionally blocked.

> Is FTL still working? Answering DNS requests?

Yes.

> Allow non-login API requests (from already authenticated devices?)

No. My idea here was to ensure this cannot be used to cause the API to use 100% CPU by bombarding it with an unbelievable amount of tries also resulting in thousands of lines added to the log file. So, yes, the API will become basically unresponsive during an online bruteforce attack, but one could also argue that this is okay as it might make users look into the log and close down port 80/443 from the Internet as consequence...

yubiuser (approved):
Feel free to fix spellchecking now or later

PR #1651, opened 2023-10-08
Simplify/fix redirection rules

DL6ER:

The PR has two connected goals:

1. Simplify redirection rules.
   Before we had an explicit rule to redirect rules ending in slashes to strip the slash, e.g. `pi.hole/admin/` to `pi.hole/admin`. However, as both URLs will work equally well, there is no need for this redirection.
2. When directly trying to access a directory, CivetWeb tries to determine the server hostname so it can send an **absolute** redirect (301). Especially when being behind a proxy, this will always fail and cause dead-end redirections.
   We solve this by patching CivetWeb to use **relative** redirections instead. This fully resolves this issue. The patch is added to our existing series of patches to ensure it can be re-applied in future CivetWeb upgrades.
   You can try this with, e.g. http://pi.hole/admin/scripts/ - the expected outcome is:
   ```
   Error 403: Forbidden
   Error: Directory listing denied
   ```
   The bug is: Redirection to *somewhere* (may be the IPv4 or IPv6 address of your Pi-hole).

As you can guess, these two were even able to play some weird ping-pong sending users with non-default deployments into the wild.

---

## Checklist:

### DIFF
--- a/patch/civetweb.sh
+++ b/patch/civetweb.sh
@@ -7,5 +7,6 @@ patch -p1 < patch/civetweb/0001-Always-Kepler-syntax-for-Lua-server-pages.patch
 patch -p1 < patch/civetweb/0001-Add-FTL-URI-rewriting-changes-to-CivetWeb.patch
 patch -p1 < patch/civetweb/0001-Add-mbedTLS-debug-logging-hook.patch
 patch -p1 < patch/civetweb/0001-Register-CSRF-token-in-conn-request_info.patch
+patch -p1 < patch/civetweb/0001-Do-not-try-to-guess-server-hostname-in-Civetweb-when.patch
 
 echo "ALL PATCHES APPLIED OKAY"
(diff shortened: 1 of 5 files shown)


yubiuser (commented):
Should such an access error be logged to `/var/log/pihole/webserver.log`? So far, it isn't.

yubiuser:
Is it expected to get redirects from `hostname/admin/scripts/` to `IP/admin/script` (where the error message is shown)?

DL6ER:
> Is it expected to get redirects from hostname/admin/scripts/ to IP/admin/script (where the error message is shown)?

No, actually the opposite should be true (redirecting `script` -> `script/`). Please try in a private browser window so we can exclude the redirect comes from cache (it happened to me quite a lot which was really confusing at first).

yubiuser (approved):
It was a browser cache issue. All good.

PR #1665, opened 2023-10-14
Add number of matches into /api/search/{domain} results

DL6ER:

During review of https://github.com/pi-hole/pi-hole/pull/5361, I noticed that there is no way for the client to know if their search request reached the given limit of results to return or if there are (possibly many) more entries to retrieve given a higher value of `N`.

The PR adds a new object having total counts not being clipped by the parameter `N`:
image

Example result for partial match of `doubleclick.net`:
image

---

## Checklist:

### DIFF
--- a/src/api/docs/content/specs/search.yaml
+++ b/src/api/docs/content/specs/search.yaml
@@ -156,10 +156,6 @@ components:
                     items:
                       type: integer
                     example: [0,1,2]
-            total:
-              type: integer
-              description: Total number of results
-              example: 2
             parameters:
               type: object
               properties:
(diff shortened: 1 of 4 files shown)


yubiuser (commented):
This is a helpful PR. I have some remarks. 

1) Can you return the sum of all results in `.results`?
2) Do we still need the (old) `total`? It's anyway clipped at `N` if more results are found, or would be the same as in 1) if less are found.
3) The help text for `N` says: "Maximum number of results to be returned (per type)" - is it really per type?

yubiuser:
Do we have a way to signal that the results exceed the hard coded limit of 10.000?

DL6ER:
> Can you return the sum of all results in .results?

Yes

> Do we still need the (old) total? It's anyway clipped at N if more results are found, or would be the same as in 1) if less are found.

Yes, too.

> The help text for N says: "Maximum number of results to be returned (per type)" - is it really per type?

Yes, `N=20` means at most 20 domains, 20 allow regex, 20 deny regex, and 20 anti- plus 20 gravity. So it could return up to 100 results in total. It could really easily be made a "20 maximum total". The reason for how it is right now is that 20 matches in gravity could have easily hidden 20 domains on the denylist.

> Do we have a way to signal that the results exceed the hard coded limit of 10.000?

Well, not really. The limit only applies to `N`, not the internal counting. (you can see in my example that > 30,000 domains match doubleclick partially on gravity)

yubiuser (approved):

PR #1667, opened 2023-10-15
Run gravity parseList without regex

DL6ER:

This PR seeks to accelerate the `gravity parseList` process by running it without regular expressions. This relaxes the tests slightly, however, even after adding many additional adlists, I did not see a single difference in reporting of invalid domains.

Context: [Discourse thread "Slow gravity update 6.0 beta"](https://discourse.pi-hole.net/t/slow-gravity-update-6-0-beta/65457)

As benchmark, I am checking how long `pihole-FTL gravity parseList` needs for
```
  [i] Target: https://raw.githubusercontent.com/mkb2091/blockconvert/master/output/domains.txt    
  [✓] Status: Retrieval successful
  [✓] Parsed 1279214 exact domains and 0 ABP-style domains (blocking, ignored 0 non-domain entries)
```

FTL `development-v6` built on Alpine:
```
real    0m22.213s
user    0m10.204s
sys     0m11.936s
```

FTL `development-v6` built on Debian:
```
real    0m0.959s
user    0m0.826s
sys     0m0.088s
```

this branch built on Alpine:
```
real    0m0.876s
user    0m0.772s
sys     0m0.060s
```

this branch built on Debian:
```
real    0m0.781s
user    0m0.662s
sys     0m0.072s
```

---

## Checklist:

### DIFF
--- a/src/tools/gravity-parseList.c
+++ b/src/tools/gravity-parseList.c
@@ -53,6 +53,7 @@ static inline bool __attribute__((pure)) valid_domain(const char *domain, const 
 		return false;
 
 	// Loop over line and check for invalid characters
+	unsigned int last_dot = 0;
 	for(unsigned int i = 0; i < len; i++)
 	{
 		// Domain must not contain any character other than [a-zA-Z0-9.-_]
@@ -75,7 +76,30 @@ static inline bool __attribute__((pure)) valid_domain(const char *domain, const 
 			if(domain[i] == '-' && domain[i-1] == '.')
 				return false;
 		}
+
+		// Individual label length check
+		if(domain[i] == '.')
+		{
+			// Label must not be longer than 63 characters
+			// (actually 64 because the dot at the end of the label
+			// is included here)
+			if(i - last_dot > 64)
+				return false;
+
+			// Label must be at least 1 character long
+			// We did already check above to not have two
+			// consecutive dots
+
+			// Update last_dot to this dot
+			last_dot = i;
+		}
 	}
+
+	// TLD length check
+	// The last label must be at least 2 characters long
+	// (len-1) because we start counting from zero
+	if((len - 1) - last_dot < 2)
+		return false;
 
 	return true;
 }
--- a/src/tools/gravity-parseList.c
+++ b/src/tools/gravity-parseList.c
@@ -10,32 +10,26 @@
 
 #include "tools/gravity-parseList.h"
 #include "args.h"
-#include <regex.h>
 #include "database/sqlite3.h"
-
-// Define valid domain patterns
-// No need to include uppercase letters, as we convert to lowercase in gravity_ParseFileIntoDomains() already
-// Adapted from https://stackoverflow.com/a/30007882
-#define TLD_PATTERN "[a-z0-9][a-z0-9-]{0,61}[a-z0-9]"
-#define SUBDOMAIN_PATTERN "([a-z0-9_-]{0,63}\\.)"
-
-// supported exact style: subdomain.domain.tld
-// SUBDOMAIN_PATTERN is mandatory for exact style, disallowing TLD blocking
-#define VALID_DOMAIN_REXEX SUBDOMAIN_PATTERN"+"TLD_PATTERN
-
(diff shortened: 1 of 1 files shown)


yubiuser:
`development-v6`

```
real	3m19,765s
user	0m0,088s
sys	0m0,087s

```

this branch

....wooww...

```
real	0m25,441s
user	0m0,033s
sys	0m0,030s


```

yubiuser (changes requested):
development-v6

```
  [i] Target: https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
  [✓] Status: Retrieval successful
  [✓] Parsed 143280 exact domains and 0 ABP-style domains (blocking, ignored 1 non-domain entries)
      Sample of non-domain entries:
        - "0.0.0.0"

```
v6/this branch

```
  [i] Target: https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
  [✓] Status: Retrieval successful
  [✓] Parsed 143293 exact domains and 0 ABP-style domains (blocking, ignored 0 non-domain entries)
  [i] List stayed unchanged
```

There is no non-domain and the number of parsed domains changed.

src/tools/gravity-parseList.c
```diff
@@ -45,6 +39,96 @@
...
 #define MAX_INVALID_DOMAINS 5
 
+// Validate domain name
+static inline bool __attribute__((pure)) valid_domain(const char *domain, const size_t len)
+{
+	// Domain must not be NULL or empty
+	if(domain == NULL || len == 0)
+		return false;
+
+	// Domain must not start or end with a hyphen or dot
```
yubiuser, line 49: Not sure about that. Isn't a final dot just indicating the root zone?

src/tools/gravity-parseList.c
```diff
@@ -45,6 +39,96 @@
 // Number of invalid domains to print before skipping the rest
 #define MAX_INVALID_DOMAINS 5
 
+// Validate domain name
+static inline bool __attribute__((pure)) valid_domain(const char *domain, const size_t len)
```
yubiuser, line 43: There is no check for the overall and each label length (~255 chars and 63 chars I think.)

src/tools/gravity-parseList.c
DL6ER: Yes, but this will never be (and never has been) matched. `dnsmasq` drops any possibly existing final dots.

src/tools/gravity-parseList.c
DL6ER: But do we really need it? I think having an overall length check if meaningful but do we really need the individual labels one? In the end I don't think we need to be overly careful (remember, domains that don't exist also aren't harmful and don't cause extra time due to the logarithmic scaling of the tree-based lookup). However, ensuring that users do not pick up megabytes-long domains does make sense.

DL6ER: The latest commit adds total and individual maximum and minimum length checks.

NittanySeaLion: I'd put in a limit. Been reading too much about buffer overflow attacks.

DL6ER: Buffer overflows cannot happen in this code, when the line is too long to get enough memory for it, the process is aborted.

src/tools/gravity-parseList.c
yubiuser: So if `dnsmasq` receives an query to `somedomain.com.` it will convert it to `somedomain.com` and send this upstream? In that case, we should not reject adlist entries with `somedomain.com.` but instead convert it to `somedomain.com` and add it to gravity. Otherwiese `somedomain.com.` would bypass Pi-hole even if `somedomain.com.` is blocked.

(And this is actually happening. Added `google.com.` to blacklist

```
Oct 15 20:46:05 dnsmasq[111250]: query[A] google.com from 10.0.1.59
Oct 15 20:46:05 dnsmasq[111250]: cached google.com is 172.217.18.110

```
)

yubiuser (approved):

PR #1709, opened 2023-10-29
Add Basic Auth authentication

DL6ER:

Add ability to create a session using [HTTP Basic Auth](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#basic_authentication_scheme) ([RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617)).

Example:
``` bash
$ curl -s 127.0.0.1/api/auth | jq .session.valid
false

$ curl -s -u pi-hole:ABC 127.0.0.1/api/auth | jq .session.valid
true
```

---

## Checklist:

### DIFF
--- a/src/api/auth.c
+++ b/src/api/auth.c
@@ -25,6 +25,8 @@
 #include "config/password.h"
 // database session functions
 #include "database/session-table.h"
+// base64_decode()
+#include "config/password.h"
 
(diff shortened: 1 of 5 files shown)


DL6ER:
I don't think this is the way we should go for the reasons pointed out in the related Discourse discussion (see link above). I will close this but still leave the branch alive for some time in the unlikely case there are striving arguments to recover this.

PR #1717, opened 2023-10-31
Ensure %used is 0.0 if no swap is used

DL6ER:

When test compiling FTL on a machine without swap, we've seen the following result in the tests:
```
GET /api/info/system (BODY auth):
  - FTL's reply (<class 'NoneType'>) does not match defined type (number) in system.memory.swap.%used
```

This is caused by a division-by-zero error which lead to a `null` being returned in the JSON. There are two obvious fixes:

1. Change the API specs to accept returning a `null` here, or
2. Ensure that in case of no swap, `0` is returned instead (0% swap used).

The latter seems superior as it needs less logic on the client side to handle this.

Even when we will never see a system with no RAM, the logic is applied here as well - just as a precaution because division-by-zero is never a good thing.

---

3. I have commented my proposed changes within the code.
5. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
6. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

## Checklist:

### DIFF
--- a/src/api/info.c
+++ b/src/api/info.c
@@ -194,7 +194,7 @@ static int get_system_obj(struct ftl_conn *api, cJSON *system)
 	// "cached", which was fine ten years ago, but is pretty much guaranteed
 	// to be wrong today."
 	JSON_ADD_NUMBER_TO_OBJECT(ram, "available", mem.avail);
-	JSON_ADD_NUMBER_TO_OBJECT(ram, "%used", 100.0*mem.used/mem.total);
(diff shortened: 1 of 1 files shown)


DL6ER:
I did just verify that this branch compiles fine without warnings on said system without swap memory.

yubiuser (approved):

PR #1732, opened 2023-11-04
Changing dns.hosts does not need a restart

DL6ER:

Changing `dns.hosts` does not need a full FTL restart but only a cache flush and re-read of `custom.list`

---

## Checklist:

### DIFF
--- a/src/api/config.c
+++ b/src/api/config.c
@@ -26,6 +26,8 @@
 #include "shmem.h"
 // hash_password()
 #include "config/password.h"
+// main_pid()
+#include "signals.h"
 
 static struct {
 	const char *name;
@@ -659,6 +661,7 @@ static int api_config_patch(struct ftl_conn *api)
 	// Read all known config items
 	bool config_changed = false;
 	bool dnsmasq_changed = false;
+	bool rewrite_hosts = false;
 	struct config newconf;
 	duplicate_config(&newconf, &config);
 	for(unsigned int i = 0; i < CONFIG_ELEMENTS; i++)
@@ -730,6 +733,10 @@ static int api_config_patch(struct ftl_conn *api)
 		if(conf_item->f & FLAG_RESTART_FTL)
 			dnsmasq_changed = true;
 
+		// Check if this item requires rewriting the HOSTS file
+		if(conf_item == &config.dns.hosts)
+			rewrite_hosts = true;
+
 		// If the privacy level was decreased, we need to restart
 		if(new_item == &newconf.misc.privacylevel &&
 		   new_item->v.privacy_level < conf_item->v.privacy_level)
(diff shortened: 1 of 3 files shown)


yubiuser (commented):
Why do we write the config file twice?

```

2023-11-04 20:42:26.374 [388970/T388981] INFO: Writing config file
2023-11-04 20:42:26.402 [388970M] INFO: Received SIGHUP, flushing cache and re-reading config
2023-11-04 20:42:26.480 [388970M] INFO: Writing config file
2023-11-04 20:42:26.517 [388970M] INFO: Blocking status is enabled
2023-11-04 20:42:26.523 [388970/T388972] INFO: Compiled 0 allow and 1 deny regex for 3 clients in 84.1 msec
```

DL6ER:
> Why do we write the config file twice?

Yeah, that's maybe not so great but I'm not sure we can solve this elegantly otherwise.

1. The first writing is clear: It happens when you modify `dns.hosts` via the API and this needs to be stored in `pihole.toml`. (see the custom thread ID 388981 indicating this is a webserver worker thread).
2. We then issue `SIGHUP` to our embedded `dnsmasq` to signal it it has to reread the config.
3. Whenever we successfully read the config file file, we "write" it. Writing in this context means that the file is rewritten (we add comments next to config options that were changed from their default values, etc.).

Step 3 does not know we already did this in step 1 as the two processes are intentionally decoupled and they actually cannot exchange any information (only the signal is passing through). Rewriting the config after reading is useful in many cases, e.g. when a user changed the config and wants to signal FTL it should re-read it. Mind that FTL needed a full restart before v6.0 to re-read the config, now always everything can be done without a full restart (and when it is needed, FTL can decide to restart by itself).

The only really safe workaround here could be defining a second signal that would have the exact same effect as `SIGHUP` but without the writing. However, this would involve quite a few changes in the `dnsmasq` code which I'd really like to avoid as handling these kind of signals is happening here. We'd also have to modify `dnsmasq`'s signal handlers to be able to pass the *actual* signal they received to FTL (this isn't currently done as only `SIGHUP` can come from there).

yubiuser (approved):
No need to rewrite `dnsmasq`. Works nicely.

yubiuser:
> The only really safe workaround here could be defining a second signal that would have the exact same effect as SIGHUP but without the writing

I've been thinking about this, esp. I looked at https://github.com/pi-hole/FTL/pull/1738. When the latest changes all required a restart, it would reduce the number of distinct backup files by half. One way around it would be: read the config on startup, save it in memory, apply all changes, construct the new toml file in memeory (e.g. add "Changed.... Default XXX"). Compare the saved and new toml settings, only write the new file if it is different from the old one, else log "No config changes detected, not writing to disk"

DL6ER:
@yubiuser You are very welcome https://github.com/pi-hole/FTL/pull/1746

PR #1733, opened 2023-11-04
Make number of maximum concurrent API sessions adjustable

DL6ER:

Make number of maximum concurrent API sessions adjustable via `webserver.api.max_sessions` defaulting to 16 (as before this change). This PR also fixes a small glitch in the API verifier that lost the ability to detect properties returned by FTL but not documented. This bug was introduced when adding the ability to check arrays as we implemented a workaround that had adverse effects. Arrays are not handled as first-class YAML citizens and can hence be checked as well.

---

## Checklist:

### DIFF
--- a/src/api/2fa.c
+++ b/src/api/2fa.c
@@ -198,7 +198,7 @@ static bool encode_uint8_t_array_to_base32(const uint8_t *in, const size_t in_le
 }
 
 static uint32_t last_code = 0;
-bool verifyTOTP(const uint32_t incode)
+enum totp_status verifyTOTP(const uint32_t incode)
 {
 	// Decode base32 secret
 	uint8_t decoded_secret[RFC6238_SECRET_LEN];
@@ -228,15 +228,15 @@ bool verifyTOTP(const uint32_t incode)
 			{
 				log_warn("2FA code has already been used (%i, %u), please wait %lu seconds",
 				         i, gencode, (unsigned long)(RFC6238_X - (now % RFC6238_X)));
-				return false;
+				return TOTP_REUSED;
 			}
 			log_info("2FA code verified successfully at %i", i);
 			last_code = gencode;
-			return true;
+			return TOTP_CORRECT;
 		}
 	}
 
-	return false;
+	return TOTP_INVALID;
 }
 
 // Print TOTP code to stdout (for CLI use)
--- a/src/api/2fa.c
+++ b/src/api/2fa.c
@@ -230,7 +230,8 @@ enum totp_status verifyTOTP(const uint32_t incode)
 				         i, gencode, (unsigned long)(RFC6238_X - (now % RFC6238_X)));
 				return TOTP_REUSED;
 			}
-			log_info("2FA code verified successfully at %i", i);
+			const char *which = i == -1 ? "previous" : i == 0 ? "current" : "next";
+			log_debug(DEBUG_API, "2FA code from %s time step is valid", which);
 			last_code = gencode;
 			return TOTP_CORRECT;
 		}
--- a/src/api/api.h
+++ b/src/api/api.h
@@ -92,7 +92,12 @@ int api_auth_session_delete(struct ftl_conn *api);
 bool is_local_api_user(const char *remote_addr) __attribute__((pure));
 
 // 2FA methods
-bool verifyTOTP(const uint32_t code);
+enum totp_status {
+	TOTP_INVALID,
+	TOTP_CORRECT,
+	TOTP_REUSED,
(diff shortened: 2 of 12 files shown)


yubiuser (commented):
1) Can we prevent this setting to be <1 - just in case? I tried with 0 and it "works" as expected.
2) Can we change the log warning to include the currently set max_sessions?
3) Can we change (likely on the `´web` repo`) the response when the max is hit? It's not a wrong password but the max_sessions limit.

Screenshot at 2023-11-04 21-16-08

DL6ER:
> Can we prevent this setting to be <1 - just in case? I tried with 0 and it "works" as expected.

I think we should instead mention what it does in the setting's description. It does make some sense to instead improve the login page message. I'll push this to the TOTP PR we still have open for review.

> Can we change the log warning to include the currently set max_sessions?

:+1:

> Can we change (likely on the ´web repo`) the response when the max is hit? It's not a wrong password but the max_sessions limit.

Yes, see my answer above (no. 1)

src/api/2fa.c
```diff
@@ -228,15 +228,15 @@ bool verifyTOTP(const uint32_t incode)
 			{
 				log_warn("2FA code has already been used (%i, %u), please wait %lu seconds",
 				         i, gencode, (unsigned long)(RFC6238_X - (now % RFC6238_X)));
-				return false;
+				return TOTP_REUSED;
 			}
 			log_info("2FA code verified successfully at %i", i);
```
yubiuser, line 233:
```
2023-11-09 21:22:20.082 [505506/T505514] INFO: 2FA code verified successfully at 0
2023-11-09 21:24:39.962 [505506/T505520] INFO: 2FA code verified successfully at -1
```

Those numbers logged here are from self-explanatory. And they also do not match the RFC mentioned above in the comment. We accept -1,0,-1 but decided to deny double use. I would suggest to remove part of the comment and change the log output to 

```
log_info("Current 2FA code verified successfully);
log_info("2FA code from previous time slot verified successfully);
log_info("2FA code from following time slot verified successfully);
```

yubiuser (changes requested):
Changing `config.webserver.api.max_sessions` (with TOTP enabled) from the web interface does strange things to the saved sessions. Reducing it to `1` logs the current user out. Increasing it back to 16 generates strange sessions.

Screenshot at 2023-11-09 21-40-11

src/api/2fa.c
DL6ER: I also made this a debug message, no real need to log this during normal operation.

DL6ER:
The last commit addresses your observation. `max_sessions` can only change in between FTL restarts. Otherwise, when it is increased, you may end up with the database being filled with random garbage which is then imported after restart causing these phantom sessions you have seen.

Reducing to 1 may still log you out in case you have more than one session being active and yours happens to be skipped during restoring (because we end after having imported 1, no matter how many the database had).

yubiuser:
> Reducing to 1 may still log you out in case you have more than one session being active and yours happens to be skipped during restoring (because we end after having imported 1, no matter how many the database had).

Is there a way to guarantee that the session triggering the config change is the one surviving for sure?

DL6ER:
> Is there a way to guarantee that the session triggering the config change is the one surviving for sure?

No, not easily. We'd need to invent a new function `invalidate_all_sessions_by_myself()` (with a better name) than then hope nobody sneaks in. I think this is enough of an edge-case to allow that a user may need to re-login (and be warned that this is not possible right now due to too few seats). What do you think?

DL6ER:
Thinking further about this, the only really useful thing we could do without adding much work on top, would be always invalidating old sessions when this value is changed. We can document this behavior and everything would be consistent, but is that really what we want?

yubiuser:
>  really useful thing we could do without adding much work on top, would be always invalidating old sessions when this value is changed

As it's a bit random which session get's kicked off, it might be more consistent if first (all) expired sessions get kicked off and if if at least one current valid session would be affected to kick off all sessions. This is at least predictable behavior.

DL6ER:
We only store valid (= non-expired) sessions already now. What you describe is still somewhat random - you don't really know what will happen unless you try or get the list of sessions and counting them yourself *before* making the API request

yubiuser:
Your correct. This leaves two options I see here: invalidate all sessions when the number decreases or leave it like it is.

yubiuser:
Giving it some thoughts, I think it's best to leave it as it is for now. If ever enough users complain about it, we can still change it. But it's the edge of an edge-case.

yubiuser (approved):

PR #1734, opened 2023-11-05
Make inclusion of /etc/dnsmasq.d/*.conf files configurable

DL6ER:

Add a new config option `misc.etc_dnsmasq_d` to allow setting whether files in `/etc/dnsmasq.d` are loaded as additional config files for Pi-hole.

This resolves two long standing issues

1. with Pi-hole not being compatible with software that installs custom files with conflicting lines in this directory (e.g.,` lxd`, [example reference](https://discourse.pi-hole.net/t/dns-service-is-not-running-ubuntu-19-10/30157/5?u=dl6er)) as well as
2. improving the coexistence of Pi-hole with an already running `dnsmasq` on the host as - by default - we won't share any config files in the future (no reference needed)

This change also goes hand-in-hand with [a recently discovered issue](https://discourse.pi-hole.net/t/ftl-failed-to-start-due-to-duplicate-cname-at-line-1-of-etc-dnsmasq-d-05-pihole-custom-cname-conf/66087) where the origin of a spurious file inside `/etc/dnsmasq.d/` could not be determined. We have reports of at least a few dozen testers which had no issues with the CNAME migration FTL applies when moving from pre-v6.0 to v6.0 for the first time.

Overall, it took me a day to realize that this is probably really the better default as it finally makes Pi-hole be able run - for the first time - independent from a possibly intentionally installed `dnsmasq` that may be listening only on some interfaces or on another port and, hence, is in no other conflict with Pi-hole on this machine.

It is important to make it crystal clear in the blog post describing the release that this setting should be switched to on for those that have custom files in there. In case they only want to add one or two custom options, they can already do so from the web interface in v6.0 and doesn't even have to create a file, improving user experience:

Screenshot from 2023-11-05 10-31-28

---

---

## Checklist:

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -1074,6 +1074,12 @@ void initConfig(struct config *conf)
 	conf->misc.addr2line.f = FLAG_ADVANCED_SETTING;
 	conf->misc.addr2line.d.b = true;
 
+	conf->misc.etc_dnsmasq_d.k = "misc.etc_dnsmasq_d";
+	conf->misc.etc_dnsmasq_d.h = "Should FTL load additional dnsmasq configuration files from /etc/dnsmasq.d/?";
+	conf->misc.etc_dnsmasq_d.t = CONF_BOOL;
+	conf->misc.etc_dnsmasq_d.f = FLAG_ADVANCED_SETTING;
+	conf->misc.etc_dnsmasq_d.d.b = false;
+
 	conf->misc.dnsmasq_lines.k = "misc.dnsmasq_lines";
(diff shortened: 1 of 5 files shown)


src/config/config.c
```diff
@@ -1074,6 +1074,12 @@ void initConfig(struct config *conf)
 	conf->misc.addr2line.f = FLAG_ADVANCED_SETTING;
 	conf->misc.addr2line.d.b = true;
 
+	conf->misc.etc_dnsmasq_d.k = "misc.etc_dnsmasq_d";
+	conf->misc.etc_dnsmasq_d.h = "Should FTL load additional dnsmasq configuration files from /etc/dnsmasq.d/?";
+	conf->misc.etc_dnsmasq_d.t = CONF_BOOL;
+	conf->misc.etc_dnsmasq_d.f = FLAG_ADVANCED_SETTING;
```
yubiuser, line 1080: This needs the restartflag as well.
```suggestion
	conf->misc.etc_dnsmasq_d.f = FLAG_RESTART_FTL | FLAG_ADVANCED_SETTING;
```

yubiuser (changes requested):

yubiuser (approved):

PR #1735, opened 2023-11-05
Fix domain modification

DL6ER:

Fix domain modification. The `UNIQUE` key in the `domainlist` table is the combination of `domain,type`, not `domain` alone so you can add the same domain, e.g. once as allowed and once as denied entries and assign them to different clients using appropriate groups.

This PR fixes the error
```
ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint in "INSERT INTO domainlist (domain,type,enabled,comment) VALUES (:item,:type,:enabled,:comment) ON CONFLICT(domain) DO UPDATE SET type = : (1)
```
also shown in https://github.com/pi-hole/web/pull/2808

---

---

## Checklist:

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -1672,7 +1672,7 @@ bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
 	{
 		*message = sqlite3_errmsg(gravity_db);
 		log_err("gravityDB_addToTable(%d, %s) - SQL error prepare (%i): %s",
-		        row->type_int, row->domain, rc, *message);
+		        row->type_int, row->item, rc, *message);
 		return false;
 	}
 
@@ -1694,7 +1694,7 @@ bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
 	{
 		*message = sqlite3_errmsg(gravity_db);
 		log_err("gravityDB_addToTable(%d, %s): Failed to bind name (error %d) - %s",
-		        row->type_int, row->name, rc, *message);
+		        row->type_int, row->item, rc, *message);
 		sqlite3_reset(stmt);
 		sqlite3_finalize(stmt);
 		return false;
@@ -1706,7 +1706,7 @@ bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
 	{
 		*message = sqlite3_errmsg(gravity_db);
 		log_err("gravityDB_addToTable(%d, %s): Failed to bind type (error %d) - %s",
-		        row->type_int, row->domain, rc, *message);
+		        row->type_int, row->item, rc, *message);
 		sqlite3_reset(stmt);
 		sqlite3_finalize(stmt);
 		return false;
@@ -1727,7 +1727,7 @@ bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
 			// Error, one is not meaningful without the other
 			*message = "Field type missing from request";
 			log_err("gravityDB_addToTable(%d, %s): type missing",
-			        row->type_int, row->domain);
+			        row->type_int, row->item);
 			sqlite3_reset(stmt);
 			sqlite3_finalize(stmt);
 			return false;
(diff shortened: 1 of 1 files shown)


DL6ER:
This bug has been reported [on Discourse](https://discourse.pi-hole.net/t/sql-err-custom-group-domain-assignments/66053)

yubiuser (approved):

yubiuser:
Using this branch in combination with https://github.com/pi-hole/web/pull/2808.

Is it expected to reset the type to the previous state on conflict and show a success message?


Peek 2023-11-09 22-46

DL6ER:
Yes, this is a success - but .... BUT what really happens is actually much worse because of the `ON CONFLICT` - and it is not even immediate obvious to me how to solve this...

Well ... what happens here is that `ON CONFLICT` we `UPDATE`. In your case you thought nothing has happened. However, this is actually not true, there was just nothing visible. Behind the scenes, you created a `CONFLICT`. This then caused the `INSERT` to behave like an `UPDATE` but it actually updates the wrong (!) record (not the one you are changing but the conflicting one). If you'd have had a comment or a custom group you'd have magically seen this being copied to the conflicting record. This is obviously wrong.

---

Note: This can only happen on `domainlist` (not the other tables) as only this table uses a combined `UNIQUE`ness.

DL6ER:
No, ignore what I said above. The solution is trivial, I did just confuse myself unnecessarily ... this is actually simple and just got broken three weeks ago in https://github.com/pi-hole/FTL/pull/1695 ... incoming ...

DL6ER:
Doing what you did above should now result in

Screenshot from 2023-11-10 20-46-03

asdesigned to happen

yubiuser:
The error is now shown as expected. However, there is an issue with the group assignment. See the screen recording. Only default an one additional group exist.

Peek 2023-11-10 21-51

yubiuser:
(The last screenshot might shown an issue on `web` rather than FTL)

DL6ER:
:thinking: I have been watching your GIF and I am not sure what the issue is - there is no text describing it. Is it that the `Apply` button is disabled?

yubiuser:
Sorry, yes. It's the disabled apply button.

rdwebdesign:
The button is only disabled when the checkbox on the same line is not selected.

rdwebdesign:
@yubiuser 

Which browser did you use for this test? Do you see the same behavior on other browsers?

yubiuser:
I'm, seeing this on Firefox and on Chrome. 

> The button is only disabled when the checkbox (used to delete) on the same line is not selected.

This is not the case here. It seems it does not matter if the line is selected or not, sometimes Apply is disabled, sometimes not.

yubiuser:
Sometimes the checkbox gets disabled when using the drop-down.

Peek 2023-11-11 20-39

DL6ER:
I'm a bit confused: are we still trying to figure of if this is related to the current PR or did we already conclude this is a web issue? Anyway, I guess it is somehow preventing this PR from being fully reviewed.

rdwebdesign:
> Sometimes the checkbox gets disabled when using the drop-down.

This is a different issue.
This is expected because the "row selection" is triggered when you click in any part of the row, [except the last cell](https://github.com/pi-hole/web/blob/9049674eb4879eedde4ce5a541daedfb4a4f7348/scripts/pi-hole/js/groups-domains.js#L291). When you click on the drop-down, you are also clicking on the row and the selection/deselection is triggered.

I think we decided to use the "everywhere" selection to make the selection easier on mobile, but this is not looking so good. I'm sure the selection/deselection is not enabled on the last cell to avoid issues when clicking on the trashcan icon. 

We could enable the behavior only on the first cell (the one containing the checkbox).

yubiuser:
All issues fixed with https://github.com/pi-hole/web/pull/2808/files/9049674eb4879eedde4ce5a541daedfb4a4f7348..ca67f3a11ef97dfbd1be936682665bbaf8b29699

yubiuser (approved):

PR #1752, opened 2023-11-12
Remove obsolet local.list

yubiuser:
As the title says. In conjunction with https://github.com/pi-hole/pi-hole/pull/5480


DL6ER (approved):
Seems it's really dead

PR #1754, opened 2023-11-13
Use hostsdir for custom.list to avoid cache flushing on changes

DL6ER:

Use `hostsdir` for `custom.list` to avoid cache flushing on changes. When adding a domain into the empty list, the expected output in `pihole.log` is:
```
Nov 13 09:56:00 dnsmasq[1761111]: inotify: /etc/pihole/hosts/custom.list new or modified
Nov 13 09:56:00 dnsmasq[1761111]: read /etc/pihole/hosts/custom.list - 2 names
```
when removing it again:
```
Nov 13 09:56:34 dnsmasq[1761111]: inotify: /etc/pihole/hosts/custom.list new or modified
Nov 13 09:56:34 dnsmasq[1761111]: inotify: flushed 2 names read from /etc/pihole/hosts/custom.list
Nov 13 09:56:34 dnsmasq[1761111]: read /etc/pihole/hosts/custom.list - 0 names
```
The DNS cache stays intact across these events.

It is expected to see two instead of one domain being added above if `dns.expandHosts` is `true` as each HOSTS entry actually adds two domains: `test` and `test.lan` (assuming you added `test` and `lan` is your `dhcp/dns.domain`).

On a fresh migration, FTL reads `custom.list` and copies it into `pihole.toml`. Then the file is renamed to `custom.list.bck`. From this point on, FTL abandons this file and will now use `/etc/pihole/hosts/custom.list`. The old `/etc/pihole/custom.list` can be removed manually. I don't think we need a cleaning strategy here as only beta users will have this file as a remnant. Users freshly upgrading will not see this file there once this PR got merged.

One other note: I removed rotation of `custom.list` as it can fully be reconstructed from `pihole.toml` - we should think about only rotating `pihole.toml` (and `dhcp.leases` on Teleporter import) in the future as the same holds true for `dnsmasq.conf`

---

---

## Checklist:

### DIFF
--- a/src/api/config.c
+++ b/src/api/config.c
@@ -778,11 +778,7 @@ static int api_config_patch(struct ftl_conn *api)
 
 		// Rewrite HOSTS file if required
 		if(rewrite_hosts)
-		{
 			write_custom_list();
-			// Reload HOSTS file
-			kill(main_pid(), SIGHUP);
-		}
 	}
 	else
 	{
@@ -976,11 +972,7 @@ static int api_config_put_delete(struct ftl_conn *api)
 
 	// Rewrite HOSTS file if required
 	if(rewrite_hosts)
-	{
 		write_custom_list();
-		// Reload HOSTS file
-		kill(main_pid(), SIGHUP);
-	}
 
 	// Send empty reply with matching HTTP status code
 	// 201 - Created or 204 - No content
--- a/src/config/cli.c
+++ b/src/config/cli.c
@@ -441,10 +441,8 @@ int set_config_from_CLI(const char *key, const char *value)
 		}
 		else if(conf_item == &config.dns.hosts)
 		{
-			// We need to rewrite the custom.list file but do not need to
-			// restart dnsmasq. If dnsmasq is going to be restarted anyway,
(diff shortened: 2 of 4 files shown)


yubiuser (commented):
Can we add one `info` log line to `FTL.log`? Currently, we only log on each modification

```
INFO: Writing config file
```

But maybe adding something like this would be useful
```
INFO: Updated /etc/pihole/hosts/custom.list
```

____

>  The old /etc/pihole/custom.list can be removed manually. I don't think we need a cleaning strategy here as only beta users will have this file as a remnant.

Why do only beta users have this file as a remnant? We are using this file in v5 already.

DL6ER:
> INFO: Writing config file

is a static string being printed when `pihole.toml` is written. We do not log this for any other files unless in `debug.config` mode.

> Why do only beta users have this file as a remnant? We are using this file in v5 already.

Yes, but when they come from v5 after this PR is merged, the file will be read during migration and gets renamed to `custom.list.bck` so no `custom.list` is left in `/etc/pihole`

yubiuser (approved):

PR #1767, opened 2023-11-21
Fix possible crash when importing legacy DHCP config

DL6ER:

Two weeks ago, https://github.com/pi-hole/FTL/pull/1731 modified the type of `dhcp.{start,end,gateway}` from string to real IP addresses but the legacy importing routines have been forgotten. This can lead to a crash on migrating v5 -> v6 Pi-holes in current `development-v6`.

This PR fixes this in three independent ways:

1. Make all the `str*` standard library functions immune against accidentally passed in `NULL` pointers. They will instead print an error with the exact location from where they were invoked with the invalid parameter (and which parameter it was for functions taking multiple).
2. Make the `get_conf_*_from_setupVars()` functions check if they are allowed to manipulate the passed in `conf_item`
3. Add a new `get_conf_ipv4_from_setupVars()` function suitable for importing said `dhcp.{...}` options (see above)

---

## Checklist:

### DIFF
--- a/src/FTL.h
+++ b/src/FTL.h
@@ -161,6 +161,22 @@
 #define pthread_mutex_lock(mutex) FTLpthread_mutex_lock(mutex, __FILE__,  __FUNCTION__,  __LINE__)
 #define fopen(pathname, mode) FTLfopen(pathname, mode, __FILE__,  __FUNCTION__,  __LINE__)
 #define ftlallocate(fd, offset, len) FTLfallocate(fd, offset, len, __FILE__,  __FUNCTION__,  __LINE__)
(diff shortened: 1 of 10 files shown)


RobertoCravallo:
Glad this old Kraut could help....

yubiuser (approved):
> Two weeks ago, https://github.com/pi-hole/FTL/pull/1731 modified the type of dhcp.{start,end,gateway} from string to real IP addresses

Should we make this change also to the other `IPs` values in the config (e.g. upstreams, revServer,....)?

DL6ER:
> Should we make this change also to the other IPs values in the config (e.g. upstreams, revServer,....)?

These are not strictly IPs because of the possible port (appended with `#5353`). That's why continue treating them as strings internally.

PR #1773, opened 2023-11-21
Add additional config validation

DL6ER:

FTL is already pretty selective when it comes to accepting config values as it has a strong filter on a config value's type. For instance, your cannot set `dhcp.start` to a number or hostname. Similarly, you canot set `dns.port` to 100,000 as such a high value does not exist or `-12` as negative ports don't exist either.

However, we have two config value types that do not allow strict testing: strings and arrays of strings.

This PR adds the facility to add validation callbacks to such config values so FTL will be able to check validity of those, too. I added validators for almost all string config options with only a few exceptions, e.g. `dns.upstreams` as `dnsmasq` allows upstreams to be defined by IP address or hostname, optionally with a port - we can leave checking the values to `dnsmasq` and will tell the user on error:
  ```
  $ sudo ./pihole-FTL --config dns.upstreams '["127.0.0.1#5335","1...1#123"]'
  New dnsmasq configuration is not valid (dnsmasq: Name or service not known at line 35 of /etc/pihole/dnsmasq.conf.temp: "server=1...1#123"), config remains unchanged
  ```

Some examples for validators we're adding herein:
- `dns.hosts`:
  ```
  $ sudo ./pihole-FTL --config dns.hosts '["127.0.0.1"]'
  Invalid value: 1st element is not in the form "IP HOSTNAME" ("127.0.0.1")
  
  $ sudo ./pihole-FTL --config dns.hosts '["1.1.1.1 cf", "127.0.0...1 hostname"]'
  Invalid value: 2nd element is neither a valid IPv4 nor IPv6 address ("127.0.0...1")
  ```
- `dns.domain` (similarly `dns.revServer.domain`, `dhcp.domain`, `webserver.domain`):
  ```
  $ sudo ./pihole-FTL --config dns.domain "----"
  Invalid value: Not a valid domain ("----")
  ```
- `dns.cnameRecords`:
  ```
  $ sudo ./pihole-FTL --config dns.cnameRecords '["cname,,1"]'
  Invalid value: 1st element contains an empty string at position 1

  $ sudo ./pihole-FTL --config dns.cnameRecords '["cname"]'
  Invalid value: 1st element is not a valid CNAME definition
  ```
- `dns.revServer.cidr`:
  ```
  $ sudo ./pihole-FTL --config dns.revServer.cidr "1.1.1.1/55"
  New dnsmasq configuration is not valid (dnsmasq: bad IPv4 prefix length at line 82 of /etc/pihole/dnsmasq.conf.temp: "rev-server=1.1.1.1/55,192.168.2.1"), config remains unchanged

  $ sudo ./pihole-FTL --config dns.revServer.cidr "1.1.1..1/24"
  Invalid value: Not a valid IPv4 nor IPv6 address ("1.1.1..1")
  ```
- `dns.revServer.target`
  ```
  $ sudo ./pihole-FTL --config dns.revServer.target "1.1.1..1"
  Invalid value: Not a valid IPv4 nor IPv6 address ("1.1.1..1")

  $ sudo ./pihole-FTL --config dns.revServer.target "1.1.1.1#500000"
  Invalid value: Not a valid port ("500000")
  ```
- `webserver.tls.cert` (similarly all paths in `files`):
  ```
  $ sudo ./pihole-FTL --config webserver.tls.cert "/etc/pihole/tls.pem?"
  Invalid value: Not a valid file path ("/etc/pihole/tls.pem?")
  ```

---

## Checklist:


src/config/validator.c
```diff
@@ -183,4 +183,53 @@ bool validate_cidr(union conf_value *val, char err[VALIDATOR_ERRBUF_LEN])
...
+}
+
+// Validate IP address optionally followed by a port (separator is "#")
+bool validate_ip_port(union conf_value *val, char err[VALIDATOR_ERRBUF_LEN])
+{
+	// Check if it's a valid IP
+	char *str = strdup(val->s);
+	char *tmp = str;
+	char *ip = strsep(&tmp, "#");
+	char *port = strsep(&tmp, "#");
```
yubiuser, line 197: This looks like a copy-paste error.
//suggestion
```
char *port = &tmp;
```

DL6ER: No, this is correct. First, we make a copy of the string as `strsep()` modifies the string internally (it is setting null-terminators where delimiters are). We them store a "working" pointer `tmp` which is updated by the individual `strsep()` calls. The calls themselves return pointers to the string in between delimiters so exactly what we want.

yubiuser: But there is only one`#` in `IP#Port`. How does the second one find anything?

DL6ER: The first one if finds is `IP`. Then there is `#` separating both. The second thing it finds is `Port`.
If the string would be `IP#Port#` it would find three strings `IP`, `Host`, and a third empty one.

So this code is still correct, but your comment just revealed that we do not check if there is such a tail string, i.e., without having tested it, I guess `1.1.1.1#53#somecrap` would currently be accepted, too.

DL6ER: Last commit:
```
$ sudo ./pihole-FTL --config dns.revServer.target "1.1.1.1#500#somecrap"
Invalid value: Not a valid IP ("1.1.1.1#500#somecrap")
```

yubiuser: Glad I could help without understanding it ;-)

But I found the missing piece:

> In case no delimiter was found, the token
       is taken to be the entire string *stringp, and *stringp is made
       NULL.
       
https://man7.org/linux/man-pages/man3/strsep.3.html

src/api/config.c
```diff
@@ -767,6 +767,20 @@ static int api_config_patch(struct ftl_conn *api)
```
yubiuser, line 767: In the line above we have `config_changed=true;`
\> should this be placed after the new value validation? We reject the new config when the value is invalid in the end.

Otherwise manual editing of `pihole.toml` can still add invalid entries.

yubiuser (commented):
Could you add a comment (in the source code), which items are validated by `dnsmasq`. Otherwise it's hard to judge if something was forgotten or left out on purpose.

src/api/config.c
DL6ER: It doesn't matter. If validation fails, this function will free the temporary `newconf` in line 776 below and return. Nobody will ever look at the value of `config_changed`

DL6ER:
> Could you please add a comment (in the source code), which items are validated by dnsmasq. Otherwise it's hard to judge if something was forgotten or left out on purpose.

`dnsmasq` validates everything causing a config rewrite (settings marked with `FLAG_RESTART_FTL`) but we are not particularly happy how it does this for these two: `dns.cnameRecords` and `dns.hosts`. 
`dnsmasq` accepts an empty target as a CNAME to the root zone. I don't think this is wrong as such bu it seems we don't want this to be valid. `dnsmasq` also "accepts" incomplete HOSTS records by simply ignoring lines not in the correct format.

DL6ER:
Should one-character domains be considered valid? In the context of `valid_domain()`  a minimum domain length of two characters is required.

yubiuser:
I think it should be two characters for external domains, but accept single characters for local hostnames.

Someone might have `a.lan`, `b.lan`,...

yubiuser:
`webserver.api.excludeDomains` could be checked for `validate_domain`

DL6ER:
https://github.com/pi-hole/FTL/pull/1832 has been changed just now to allow regex in this field

DL6ER:
New examples:
```
tre,192.168.2.0/24,192.168.2.1,fritz.box
```
Screenshot from 2024-02-11 06-57-16

```
true,692.168.2.0/24,192.168.2.1,fritz.box
```
Screenshot from 2024-02-11 07-01-39

```
true,192.168.2.0/245,192.168.2.1,fritz.box
```
Screenshot from 2024-02-11 07-01-54

```
true,192.168.2.0/24,192.168.2.1#44455585,fritz.box
```
Screenshot from 2024-02-11 07-03-14

PromoFaux:
This all looks good, do you plan on adding some unit tests?

src/config/dnsmasq_config.c
```diff
@@ -473,7 +473,7 @@ bool __attribute__((const)) write_dnsmasq_config(struct config *conf, bool test_
 
 		if(active == NULL || cidr == NULL || target == NULL || domain == NULL)
 		{
-			log_err("Invalid reverse server string: %s", revServer->valuestring);
+			log_err("Skipped invalid dns.revServers[%u]: %s", i, revServer->valuestring);
```
rdwebdesign, line 476: Should this be an error or just a warning?

```suggestion
			log_warn("Skipped invalid dns.revServers[%u]: %s", i, revServer->valuestring);
```

DL6ER: This is subtle. I think this is an error even when it is not a fatal one. It's still invalid configuration

PromoFaux (approved):
Great stuff, even greater with tests!

PR #1777, opened 2023-11-22
Fix possible crash with high client activity

DL6ER:

Fix possible race-collision leading to a theoretical out-of-bounds read

---

## Checklist:

### DIFF
--- a/src/database/network-table.c
+++ b/src/database/network-table.c
@@ -760,7 +760,7 @@ static int update_netDB_interface(sqlite3 *db, const int network_id, const char 
 
 // Loop over all clients known to FTL and ensure we add them all to the database
 static bool add_FTL_clients_to_network_table(sqlite3 *db, enum arp_status *client_status, time_t now,
-                                             unsigned int *additional_entries)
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
The race would be that `counters->clients` could be increased while the counter for `clientID` is still running?

DL6ER:
We declare an object of size `counters->clients` here:
https://github.com/pi-hole/FTL/blob/1a11413385eee8992b12b662cd37c32e55da4f6d/src/database/network-table.c#L1288-L1291

and later on use it in a loop that is running from `0` to `counters->clients`. The race is that `counters->clients` may have increased meanwhile. This is prevented by passing the `counters->clients` we have used before and use this as upper limit for the loop. It may mean we are missing a few clients, but we will pick them up shortly after when the code is run again.

DL6ER:
Unfortunately, we never got feedback from the affected users. As they were target active before we suggested this bugfix, I can only assume it fixed the issue for them. Anyway, it's a meaningful change and we can use it to push a v5.x FTL patch release.

yubiuser (approved):

PR #1785, opened 2023-11-27
Improve FTL's file comparsion algorithm

DL6ER:

Fix file comparison and add more debug logging during file comparison. So far, this should not have any negative consequences as all files we apply this function on will always differ *somewhere* before the end of the files when they are different (`custom.list` has the total number of lines on the end, `dnsmasq.conf` and `pihole.toml` have no multi-line options at the end).

---

## Checklist:

### DIFF
--- a/src/files.c
+++ b/src/files.c
@@ -607,28 +607,51 @@ bool files_different(const char *pathA, const char* pathB, unsigned int from)
 	}
 
 	// Compare both files line by line
-	char *lineA = NULL;
-	size_t lenA = 0;
-	ssize_t readA;
-	char *lineB = NULL;
-	size_t lenB = 0;
-	ssize_t readB;
+	char *lineA = NULL, *lineB = NULL;
+	size_t lenA = 0, lenB = 0;
+	ssize_t readA = 0, readB = 0;
 	bool different = false;
-	while((readA = getline(&lineA, &lenA, fpA)) != -1 &&
-	      (readB = getline(&lineB, &lenB, fpB)) != -1)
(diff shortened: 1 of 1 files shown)


DL6ER:
Bugfix confirmed working by OP on Discourse. The key issue we solve here was that `file_different()` had a small bug that did not detect if one of the files we compare to is *completely* empty.

src/files.c
```diff
@@ -607,30 +607,53 @@ bool files_different(const char *pathA, const char* pathB, unsigned int from)
...
+			log_debug(DEBUG_CONFIG, "Files %s and %s differ at line %u",
+			          pathA, pathB, lineno);
+			log_debug(DEBUG_CONFIG, "-> %s:%u = '%s'", pathA, lineno, readA < 0 ? "<EOF>" : lineA);
+			log_debug(DEBUG_CONFIG, "-> %s:%u = '%s'", pathB, lineno, readB < 0 ? "<EOF>" : lineB);
 			break;
 		}
 	}
 
+	// Check if one file has more lines than the other
+	if(!different && readA != readB)
```
yubiuser, line 648: Wouldn't it makes sense to first check if the number of lines differ before checking each line for equality? This sounds like a short-cut.

DL6ER: Checking the number of lines means we need to read the entire file and count ` \n`. There is no other way to do this. When we have already read the file lines, we just as well compare them. This is super fast as the disk I/O is the slow work, not the one CPU cycle comparison between two memory regions.

yubiuser (approved):

PR #1804, opened 2023-12-03
Fix CPU utilization calculation on 32 bit systems

DL6ER:

Following [a report on Discourse](https://discourse.pi-hole.net/t/calc-cpu-usage-failed-in-ftl-log/66617), we change how FTL computes its own CPU utilization. The issue is that our current method of using the system-provided function `clock()` can easily overflow on 32bit systems, leading to a permanent failure after little more than two days. This has not been discovered during development as all our machines are 64 bit and the CI tests intentionally running on 32 bit obviously finished quicker than those two-and-a-half days.

This PR changes the way we measure CPU utilization to use resource details obtained from the kernel to compute a ten seconds average of FTL's CPU utilization based on the consumed CPU time (not ticks), instead. This is a very reliable metric and more in-line with what `top` and friends calculate for the processes.

---

---

## Checklist:

### DIFF
--- a/src/api/docs/content/specs/info.yaml
+++ b/src/api/docs/content/specs/info.yaml
@@ -732,7 +732,7 @@ components:
               example: 0.1
             "%cpu":
               type: number
-              description: Percentage of total CPU used by FTL
+              description: Percentage of total CPU used by FTL (ten seconds average)
               example: 1.2
             allow_destructive:
               type: boolean
--- a/src/daemon.c
+++ b/src/daemon.c
@@ -356,26 +356,31 @@ void cleanup(const int ret)
 	log_info("########## FTL terminated after%s (code %i)! ##########", buffer, ret);
 }
 
-static clock_t last_clock = -1;
+static float last_clock = 0.0f;
 static float cpu_usage = 0.0f;
-void calc_cpu_usage(void)
+void calc_cpu_usage(const unsigned int interval)
 {
-	// Get the current CPU usage
-	const clock_t clk = clock();
-	if(clk == (clock_t)-1)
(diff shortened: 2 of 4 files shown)


yubiuser (commented):
Mhh.. I watched it a while and FTL consumed ~1-2% CPU according to htop, but the web interface reports much lower

Screenshot at 2023-12-04 20-44-55

yubiuser:
>  compute a ten minute average

I think you meant "ten seconds average" in the PR description

DL6ER:
Strange, this is the exact same way (`h`)`top` do it, e.g.

https://github.com/mmalecki/procps/blob/fe4c4a7314f32907b9f558ad0d8b8d0ff1cc76be/top.c#L1046-L1050

Could you do one experiment maybe for verification? Note down the shown CPU time of the entire FTL process (2.05 seconds in your screenshot) and wait 60 seconds. Then note down the value shown now (say, 4.50 seconds), subtract them and divide them by the time you waited, e.g.

```
4.5 - 2.05
---------- = 0.040 = 4.0%
    60
```

and see if this number better matches with the averaged CPU load shown by FTL. It will be hard (basically impossible) to the the exact same numbers due to different time averages but you should get at least the same order of magnitude.

Your screenshot already highlights that there is some strange inconsistency in how `htop` computes the numbers as 2% for one thread is already larger than the 1.3% for the entire process (including all its threads)...

yubiuser:
The last commit fixed the discrepancy for me. Both `htop` and `FTL` report similar (+/- 0.1%) usage. Also calculation as you said results in almost the same values.

yubiuser (approved):

PR #1806, opened 2023-12-05
Use WAL, remove (and strip) SQLite3 shared-cache support and improve attached database handling

DL6ER:

This PR seeks to improve the overall database performance by use write-ahead log (WAL) and removing (and actually stripping the entire) shared-cache support from the SQLite3 engine.

While "shared cache" sounds nice at first, this change is actually following [the recommendations from SQLite3 itself](https://www.sqlite.org/sharedcache.html):

> Shared-cache mode is an obsolete feature. **The use of shared-cache mode is discouraged.** Most use cases for shared-cache are better served by [WAL mode](https://www.sqlite.org/wal.html).
>
> Shared-cache mode was invented in 2006 at the request of developers of [Symbian](https://en.wikipedia.org/wiki/Symbian). Their problem was that if the contacts database on the phone was being synced, that would lock the database file. Then if a call came in, the database lock would prevent them from querying the contacts database in order to find the appropriate ring-tone for the incoming call, or a photo of the caller to show on screen, and so forth. **[WAL mode](https://www.sqlite.org/wal.html) (circa 2010) is a better solution to this problem as it permits simultaneous access without breaking transaction isolation.**

WAL itself is not the standard journaling mode in SQLite3 and has a number of [advantages and disadvantages](https://www.sqlite.org/wal.html). Most important for us are - thankfully - only the advantages:
- WAL is significantly faster in most scenarios.
- WAL provides more concurrency as readers do not block writers and a writer does not block readers. Reading and writing can proceed concurrently.
- Disk I/O operations tends to be more sequential using WAL. 

The second commit fixes a bug mentioned a while ago by @yubiuser where he saw that the Query Log on the web interface could trigger `database already attached` errors in "Live" mode. This happened because two independent threads (two web API workers) both tried to attach the disk database when 

---

---

## Checklist:


DL6ER:
It seems we have found an SQLite3 bug.

When using WAL, the attached disk database will work for some time and then start to report "disk I/O error". From this point on, no further interaction with the disk database is possible. When I remove the `PRAGMA journal_mode=WAL`, everything works perfectly fine.

In my tests, I have seen this "disk I/O error" once, but I was never able to reproduce it. The branch continued working for a several days afterwards so I thought this was a one time issue, possibly because I tested some intermediate version of the final commit. Now we see that both x86 builds+tests worked fine (even always on the first try) but SQLite3 is consistently throwing these errors:
```
2023-12-06 13:44:00.008 [3767/T3769] ERR: SQLite3 message: statement aborts at 55: [INSERT INTO disk.query_storage SELECT * FROM query_storage WHERE id > ? AND timestamp < ?] disk I/O error (266)
2023-12-06 13:44:00.009 [3767/T3769] ERR: export_queries_to_disk(): Failed to export queries: disk I/O error
```

DL6ER:
> this is still an SQLite3 bug but we will probably find a way coding around it.

The explanation for what we have seen that SQLite3 while being [thread-safe](https://www.sqlite.org/threadsafe.html) is actually not [fork-safe](https://www.sqlite.org/faq.html#q6):

> Under Unix, you should not carry an open SQLite database across a `fork()` system call into the child process.

This has been wrong in FTL since the beginning, but - so far - we have never witnessed any consequences. This is easily fixed by https://github.com/pi-hole/FTL/pull/1806/commits/3aec9bd1dc8b01156c6b992cbf105064424e6ae3 in this PR.

DL6ER:

src/database/query-table.c
```diff
@@ -123,6 +123,10 @@ bool init_memory_database(void)
 		}
 	}
 
+	// Attach disk database
+	if(!attach_database(memdb, NULL, config.files.database.v.s, "disk"))
+		return false;
```
yubiuser, line 113: Should we through an error here?
___

The commit message says that we detach the database when FTL terminates - does it also detach when it crashes? Or is this an exception we can't handle?

src/database/common.c
```diff
@@ -251,9 +251,15 @@ void SQLite3LogCallback(void *pArg, int iErrCode, const char *zMsg)
...
+		log_warn("SQLite3: %s (%d)", zMsg, iErrCode);
+	else if(iErrCode == SQLITE_NOTICE || iErrCode == SQLITE_SCHEMA)
+		// SQLITE_SCHEMA is returned when the database schema has changed
+		// This is not necessarily an error, as sqlite3_step() will re-prepare
+		// the statement and try again. If it cannot, it will return an error
+		// and this will be handled over there.
+		log_debug(DEBUG_ANY, "SQLite3: %s (%d)", zMsg, iErrCode);
 	else
-		log_err("SQLite3 message: %s (%d)", zMsg, iErrCode);
+		log_err("SQLite3: %s (%d)", zMsg, iErrCode);
```
yubiuser, line 262: This line is exactly the same as 254 so you could omit the first `if` (`if(iErrCode == SQLITE_WARNING)`) as it will default to `else`

src/database/database-thread.c
```diff
@@ -135,16 +160,30 @@ void *DB_thread(void *val)
...
+			analyze_database(db);
+			lastAnalyze = now;
+			DBCLOSE_OR_BREAK();
+		}
+
 		// Intermediate cancellation-point
 		if(killed)
 			break;
 
 		// Update MAC vendor strings once a month (the MAC vendor
```
yubiuser, line 180: Is this necessary at all? We only download the vendor database on install/update which will trigger a FTL restart anyway  reading the database. 

P.S. How often is the vendor database updated at all? Maybe it's time to do it again?

src/database/common.c
rdwebdesign: Actually these lines are almost the same. 
Line 254 is executing `log_warn()`. 
Line 262 is executing `log_err()`.

yubiuser: Good catch.

src/database/query-table.c
DL6ER: This makes `init_memory_database()` returning `false` which, in turn, causes the routine calling it (`FTL_fork_and_bind_sockets()`)  log this even:
``` c
	// Initialize in-memory databases
	if(!init_memory_database())
		log_crit("Cannot initialize in-memory database.");
```
The error will furthermore be logged independently via the SQLite3 logging callback we have installed.

DL6ER: We do try to detach on a crash:

https://github.com/pi-hole/FTL/blob/710e0bb3b7ded854ae06d5bbaeb257cf667f77e6/src/signals.c#L259

https://github.com/pi-hole/FTL/blob/710e0bb3b7ded854ae06d5bbaeb257cf667f77e6/src/daemon.c#L346

https://github.com/pi-hole/FTL/blob/710e0bb3b7ded854ae06d5bbaeb257cf667f77e6/src/database/query-table.c#L141-L143

src/database/database-thread.c
DL6ER: This is *not* done on an FTL restart, it is only done once per month.

> How often is the vendor database updated at all? Maybe it's time to do it again?

There is a `cron` job running on `ftl.pi-hole.net` doing this in frequent intervals, have a look at the last modified timestamp online:
Screenshot from 2024-01-06 09-03-16

I will push another small change that will ensure the steps are run once per week/month (new behavior) and not after FTL has been running continuously for one week/month (current behavior).

DL6ER: Pushed

yubiuser: Where do you save the lastMACVendor between restarts? If it's not stored externally, you can't be sure the update is executed every month if FTL is restarted in-between.

DL6ER: This is addressed by the last commit.

The update is started randomly within the first hour of the next month.

Example: If FTL is (re)started on the last day (31st), the update will still be done in *the next* month (one day after the restart in this example).

Edge-cases are considered: If FTL is offline in the first hour of say February, the first update will be performed on March 1st. When FTL is restarted within the first hour of March 1st (and before the update happens), the update is done in April.

Analyze is similar but within the first hour of each *week*.

Note: FTL's calculation is simplified and doesn't calculate months in a calendar sense but in a time sense, i.e., the first day of a month may actually be the 6th or so. This is because leap years continue to mix things up since the beginning of time (January 1st, 1970). In the end, it doesn't matter...

yubiuser (approved):

PR #1809, opened 2023-12-07
Warn if the system has no resolver configured.

DL6ER:
# What does this implement/fix?

Warn if the system has no resolver configured. The system may not be ready to resolve host names, not even when we insert FTL's IP address manually.

**Related issue or feature (if applicable):** N/A

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** N/A

---
**By submitting this pull request, I confirm the following:** 

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against. 
2. I have commented my proposed changes within the code.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

## Checklist:

- [x] The code change is tested and works locally.
- [x] I based my code and PRs against the repositories `developmental` branch.
- [x] I [signed off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits. Pi-hole enforces the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
- [x] I [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all my commits. Pi-hole requires signatures to verify authorship
- [x] I have read the above and my PR is ready for review.

PR #1826, opened 2023-12-15
Improve and fix host name resolution

DL6ER:

More recent Pi-hole installations may not be using `127.0.0.1` in `/etc/resolv.conf` but something like `192.168.0.1` or even `9.9.9.9`. So far, FTL uses a hack that ensures that - whatever is set in `resolv.conf` - we are always enforcing the use of FTL as DNS resolver when doing internal name lookups.

This is meaningful as external servers would very likely have no idea about the internal network structure and could not provide useful answers. Assuming the router is used in typical at-home deployments, it may still work (partially, locally defined records would not be available, for instance) but if the system is even using an external server such as Quad9, no answer will be available.

When switching from `glibc` to `musl` we lost the hack FTL has been using for years to temporary change the resolver used by the C library as `musl` does not care about the `_res` datastructure *at all* (they only add it for ABI compatibility reasons) and simply always reads `/etc/resolv.conf` itself for each query. As we - quite obviously - do not want to interfere with the system and change `/etc/resolv.conf` for each name lookup, we have to stop using the library-provided `getnameinfo()` function and implement a tiny DNS resolver ourselves. This is done in this PR. The new implementation ensure we are always asking FTL on the configured DNS port without relying on any special library behavior.

This PR is the follow-up on https://github.com/pi-hole/FTL/pull/1809 and is the final fix for this bug reported [on Discourse](https://discourse.pi-hole.net/t/dashboard-nur-noch-ips-unter-clients/66690).

---

---

## Checklist:

### DIFF
--- a/src/resolve.c
+++ b/src/resolve.c
@@ -287,27 +287,60 @@ static char *__attribute__((malloc)) ngethostbyname(const char *host, const char
 }
 
 // Convert hostname from network to host representation
+// This routine supports DNS compression pointers
 // 3www6google3com -> www.google.com
 static u_char * __attribute__((malloc)) __attribute__((nonnull(1,2,3))) name_fromDNS(unsigned char *reader, unsigned char *buffer, uint16_t *count)
 {
 	unsigned char *name = calloc(MAXHOSTNAMELEN, sizeof(char));
-	unsigned int p = 0, jumped = 0, offset;
+	unsigned int p = 0, jumped = 0;
 
 	// Initialize count
 	*count = 1;
 
 	// Parse DNS label string encoding (e.g, 3www6google3com)
-	while(*reader!=  0)
+	//
+	// In its text format, a domain name is a sequence of dot-separated
+	// "labels". The dot separators are not used in binary DNS messages.
+	// Instead, each label is preceded by a byte containing its length, and
+	// the name is terminated by a zero-length label representing the root
+	// zone.
+	while(*reader != 0)
 	{
-		// Check for too long host names
-		if(*reader >= 192)
+		if(*reader >= 0xC0)
 		{
-			offset = (*reader)*256 + *(reader + 1) - 49152; // 49152 = 11000000 00000000
+			// RFC 1035, Section 4.1.4: Message compression
+			//
+			// A label can be up to 63 bytes long; if the length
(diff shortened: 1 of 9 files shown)


DL6ER:
Some tests are failing due to https://community.arm.com/support-forums/f/compilers-and-libraries-forum/54957/arm-none-eabi-gcc-packed-attribute-causes-inefficient-alignment

src/resolve.c
```diff
@@ -75,81 +136,275 @@ static bool valid_hostname(char* name, const char* clientip)
...
+	// In its text format, a domain name is a sequence of dot-separated
+	// "labels". The dot separators are not used in binary DNS messages.
+	// Instead, each label is preceded by a byte containing its length, and
+	// the name is terminated by a zero-length label representing the root
+	// zone.
+	while(*reader != 0)
+	{
+		if(*reader >= 0xC0)
+		{
+			// RFC 1035, Section 4.1.4: Message compression
```
DL6ER, line 311: https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.4

yubiuser (commented):
Is this expected? `hostname` returning something else than `pihole-FTL ptr IP`
IP of this device is `10.0.1.24`

```
nanopi@nanopi:~$ hostname
nanopi
nanopi@nanopi:~$ pihole-FTL ptr 10.0.1.24
pi.hole
nanopi@nanopi:~$ pihole-FTL ptr 127.0.0.1
localhost.lan

```

DL6ER:
@yubiuser What is your `dns.piholePTR` set to? This is affecting the expected result. It should only be identical to `hostname` if you have `dns.piholePTR = "HOSTNAME"`. This should be consistent to the behavior before this change.

DL6ER:
The most recent comment links to a Discourse discussion confirming this works for them:

> Big thanks, issue solved immeditely! :upside_down_face:

PromoFaux (approved):
> The most recent comment links to a Discourse discussion confirming this works for them:

Reading over the description and the issue satisfies me.

rdwebdesign (approved):
I think this is good to go (based on a quick test and the user's comment), _unless you are still waiting for an answer about `dns.piholePTR` from yubiuser_.

PR #1832, opened 2023-12-18
Limit number of clients returned by /api/history/clients

DL6ER:

Limit number of clients returned by the `/api/history/clients` endpoint to 20 (default value).
This can be overwritten at compile-time and at run-time using the query parameter `?N=...`

---

---

## Checklist:

### DIFF
--- a/src/api/api.h
+++ b/src/api/api.h
@@ -24,6 +24,7 @@
 int api_handler(struct mg_connection *conn, void *ignored);
 
 // Statistic methods
+int __attribute__((pure)) cmpdesc(const void *a, const void *b);
 int api_stats_summary(struct ftl_conn *api);
 int api_stats_query_types(struct ftl_conn *api);
 int api_stats_upstreams(struct ftl_conn *api);
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -426,6 +426,8 @@ components:
                         type: string
                     maxHistory:
                       type: integer
+                    maxClients:
+                      type: integer
                     allow_destructive:
                       type: boolean
                     temp:
@@ -700,6 +702,7 @@ components:
               excludeClients: [ '1.2.3.4', 'localhost', 'fe80::345' ]
               excludeDomains: [ 'google.de', 'pi-hole.net' ]
               maxHistory: 86400
+              maxClients: 10
               allow_destructive: true
               temp:
                 limit: 60.0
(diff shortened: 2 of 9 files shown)


yubiuser:
Could you implement a "all-other-clients" client with the sum of all their queries for every client exceeding `N`? Otherwise the two dashboard graphs would not match in terms to queries/timeslot

DL6ER:
Thanks for the suggestion. I added this and also reduced the default `N` down to 10 as I had 18 devices in my network over night and the graph already rendered pretty useless. Now with the "other clients" entry compensating for this, it seemed a wise choice.

yubiuser (commented):
Works as expected, but I have a wish and one point for discussion.

1) Can we make this a config option -> users can set whatever N they want for their web interface to be displayed

2) Now we limit to the most active **overall** clients. However, the issue we want to solve is: to many clients per time slot (rendering the graph useless). Image a situation where we have only 10 clients in the morning and 10 different clients in the afternoon (each within one time slot). Despite both not exceeding N on per-timeslot basis, 10 of them will be removed. How about (I know, much more code) to limit the number of clients returned on per-timeslot basis with a lower N (e.g. 3-5)

DL6ER:
Sorry, I missed your comment before.

> Can we make this a config option -> users can set whatever N they want for their web interface to be displayed

Yes, I did just add this.

> How about (I know, much more code) to limit the number of clients returned on per-timeslot basis with a lower N (e.g. 3-5)

I think this will reintroduce part of the problem - even when N is further reduces. We have 240 timeslots in 24 hours, even with N = 3, this is close to 1000. But even when this is surely worst-case thinking, I guess we will get massive confusion with repeated colors as some of your morning clients (smart coffee maker) may be red until 10AM, then they are replaced by some other component (solar system reporting tool) which got the same red color because if limited colors available.

yubiuser:
> I think this will reintroduce part of the problem - even when N is further reduces. We have 240 timeslots in 24 hours, even with N = 3, this is close to 1000. But even when this is surely worst-case thinking, I guess we will get massive confusion with repeated colors as some of your morning clients (smart coffee maker) may be red until 10AM, then they are replaced by some other component (solar system reporting tool) which got the same red color because if limited colors available.

I see your valid point. It's a tricky decision. The current implementation can lead to situations where "other clients" is the biggest/only client within a certain timeslot, making it a bit sensless. However, this should not stop this PR. Depending on use feedback we could change it later.

Screenshot at 2024-01-16 21-11-13

yubiuser (dismissed):
Small nit pick: the order in which objects in the json are returned is `clients` first, and `history` second. In the documentation it's the other way round.

yubiuser (approved):

PR #1842, opened 2023-12-25
Do not overwrite errors on ROLLBACK

DL6ER:

Do not check errors on `ROLLBACK TRANSACTION` when `gravityDB_delFromTable()` fails. We remove this to avoid overwriting the initial cause of the error. See linked Discourse thread for context.

Additional context from [the SQLite3 documentation](https://www.sqlite.org/lang_transaction.html):

> [!TIP]
> 3. Response To Errors Within A Transaction
>
> If certain kinds of errors occur within a transaction, the transaction may or may not be rolled back automatically. The errors that can cause an automatic rollback include:
>
> *   [SQLITE_FULL](https://www.sqlite.org/rescode.html#full): database or disk full
> *   [SQLITE_IOERR](https://www.sqlite.org/rescode.html#ioerr): disk I/O error
> *   [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy): database in use by another process
> *   [SQLITE_NOMEM](https://www.sqlite.org/rescode.html#nomem): out of memory 
> 
> For all of these errors, SQLite attempts to undo just the one statement it was working on and leave changes from prior statements within the same transaction intact and continue with the transaction. However, depending on the statement being evaluated and the point at which the error occurs, it might be necessary for SQLite to rollback and cancel the entire transaction. An application can tell which course of action SQLite took by using the [sqlite3_get_autocommit()](https://www.sqlite.org/c3ref/get_autocommit.html) C-language interface.
> 
> **It is recommended that applications respond to the errors listed above by explicitly issuing a ROLLBACK command.** If the transaction has already been rolled back automatically by the error response, then the ROLLBACK command will fail with an error, but no harm is caused by this.
>
> Future versions of SQLite may extend the list of errors which might cause automatic transaction rollback. Future versions of SQLite might change the error response. In particular, we may choose to simplify the interface in future versions of SQLite by causing the errors above to force an unconditional rollback.

---

---

## Checklist:

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -1840,15 +1840,11 @@ bool gravityDB_delFromTable(const enum gravity_list_type listtype, const cJSON* 
 		*message = sqlite3_errmsg(gravity_db);
 		log_err("gravityDB_delFromTable(%d) - SQL error prepare(\"%s\"): %s",
 		        listtype, querystr, *message);
+
 		// Rollback transaction
 		querystr = "ROLLBACK TRANSACTION;";
-		rc = sqlite3_exec(gravity_db, querystr, NULL, NULL, NULL);
-		if(rc != SQLITE_OK)
-		{
-			*message = sqlite3_errmsg(gravity_db);
-			log_err("gravityDB_delFromTable(%d): SQL error exec(\"%s\"): %s",
-			        listtype, querystr, *message);
-		}
+		sqlite3_exec(gravity_db, querystr, NULL, NULL, NULL);
+
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
I'm not sure if I understand why we overwrite the error message in the first place with the original code. Is because we have `*message = sqlite3_errmsg(gravity_db);` twice in a row and only the latter gets printed to the user on the web interface?

Because logging to `FTL.log` should happen anyway.

DL6ER:
You are spot on with both. Logging will always happen but the UI reports only the last value in `*message`

yubiuser (approved):

PR #1860, opened 2024-01-17
X.509: Create and export CA certificate

DL6ER:

Currently, Pi-hole's mechanism to create a self-signed certificate, well, creates a self-signed certificate. Recent user questions have, however, revealed that there is software (most worth mentioning is Firefox) not liking (as in: explicitly discouraging, sometimes even preventing) the usage of such self-signed certificates. Adding the certificate to said browsers (so the web interface page is considered "safe") is not possible in this case.

This PR changes the way Pi-hole generates the TLS certificate for itself. We go a more traditional way by first creating a (self-signed) root certificate authority (CA) and then using this CA to sign a second certificate we specifically create for the TLS web server. The generated CA is exported to a file and can then be imported into Firefox, Chrome, etc.

If you are using `curl` or friends, nothing changes - you can still use the certificate PEM file as you were used before.

> [!NOTE]
> Pi-hole will not recreate already existing self-signed certificates. If you want to use this new certificates, you will have to run
> ```
> sudo pihole-FTL --gen-x509 /etc/pihole/tls.pem
> ```
> to force recreation of the certificate. Possibly with a domain if you want to use a domain different from `pi.hole`, e.g.
> ```
> sudo pihole-FTL --gen-x509 /etc/pihole/tls.pem pihole.lan
> ```

---

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** https://github.com/pi-hole/docs/pull/980

---

## Checklist:


src/webserver/x509.c
```diff
@@ -139,48 +209,73 @@ bool generate_certificate(const char* certfile, bool rsa, const char *domain)
...
+	printf("Generating new CA with serial number %s...\n", serial1);
+	mbedtls_x509write_crt_set_version(&ca_cert, MBEDTLS_X509_CRT_VERSION_3);
+
+	mbedtls_x509write_crt_set_serial_raw(&ca_cert, serial1, sizeof(serial1)-1);
+	mbedtls_x509write_crt_set_md_alg(&ca_cert, MBEDTLS_MD_SHA256);
+	mbedtls_x509write_crt_set_subject_key(&ca_cert, &ca_key);
+	mbedtls_x509write_crt_set_subject_key_identifier(&ca_cert);
+	mbedtls_x509write_crt_set_issuer_key(&ca_cert, &ca_key);
+	mbedtls_x509write_crt_set_authority_key_identifier(&ca_cert);
+	mbedtls_x509write_crt_set_issuer_name(&ca_cert, "CN=pi.hole,O=Pi-hole,C=DE");
```
yubiuser, line 243:
```suggestion
	mbedtls_x509write_crt_set_issuer_name(&ca_cert, "CN=pi.hole,O=Pi-hole,C=DE");
```
I'm not sure where Pi-hole is located, but it might no be "DE".

(Same applies to the next line and 266)

yubiuser (commented):
Should we add a log line announcing the creation of a CA here

```
2024-01-21 19:24:47.886 [654233M] INFO: FTL is running as user pihole (UID 999)
2024-01-21 19:24:47.903 [654233M] INFO: Created SSL/TLS certificate for nanopi.lan at /etc/pihole/tls.pem
2024-01-21 19:24:47.903 [654233M] INFO: Reading certificate from /etc/pihole/tls.pem ...
2024-01-21 19:24:47.904 [654233M] INFO: Using SSL/TLS certificate file /etc/pihole/tls.pem
```
___

This is not true anymore

```
TLS X.509 certificate generator:
    Generate a self-signed certificate suitable for SSL/TLS
    and store it in outfile.
``
```
____

Should `pihole-FTL --read-x509-key` read also the `tls_ca.crt`?

src/webserver/x509.c
DL6ER: > Should we add a log line announcing the creation of a CA here

No, it's not even trivial to transport the CA name over there. The vast majority of users will likely add a permanent exception instead of adding the CA certificate. When they want it, the documentation will tell them how - without the need to read any logs.

---

> Should pihole-FTL --read-x509-key read also the tls_ca.crt?

I don't think so. It is a generic tool to reads a certificate. This is still what it does. The CA is independent.

---

> I'm not sure where Pi-hole is located, but it might no be "DE".

In the end it doesn't matter. We can add whatever we want.

DL6ER:
> Should we add a log line announcing the creation of a CA here

No, it's not even trivial to transport the CA name over there. The vast majority of users will likely add a permanent exception instead of adding the CA certificate. When they want it, the documentation will tell them how - without the need to read any logs.

---

> Should pihole-FTL --read-x509-key read also the tls_ca.crt?

I don't think so. It is a generic tool to reads a certificate. This is still what it does. The CA is independent.

---

> I'm not sure where Pi-hole is located, but it might no be "DE".

In the end it doesn't matter. We can add whatever we want.

yubiuser (approved):

PR #1902, opened 2024-03-08
Accelerate CI builds substantially

DL6ER:

**Extra short summary:** Reduced build time for almost all binaries from ~30 min to ~3 min

---

**Longer description:**

Building FTL binaries using `buildx` slowed down our CI to roughly 16 minutes per build. If builds sporadically fail, they are repeated and the build needs half an hour. This is quite some unfortunate situation as we need to wait half an hour after pushing a fix until a user can try it.

This PR changes how Pi-hole builds binaries on the CI. Before, we built (and tested!) all binaries in parallel using GHA runners (`x86_64`). Only once all of them were finished, the binaries were uploaded:

Screenshot-old-CI

The new method has three partially independent branches:

1. We build `riscv64` on GHA runners
   Screenshot from 2024-03-08 09-23-23
2. We build `x86_64` and `i386` in parallel on GHA runners
   Screenshot from 2024-03-08 09-23-26
3. We build all `arm` binaries on **self-hosted ARM64-powered runners**
   Screenshot from 2024-03-08 09-23-29

Once `x86_64`, `i386`, and all the `arm` builds are finished (typically **3 minutes** now), we deploy them. The `riscv64` binaries will take a much longer time (there are self-hosted servers available at this point and building them on `arm64` is even slower than on `x86`) but since they are not needed by the vast majority of users, we accept that there is still some larger delay here.

---

7. I have commented my proposed changes within the code.
8. I am willing to help maintain this change if there are issues with it later.
9. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
10. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,6 +150,7 @@ jobs:
           target_dir: ${{ needs.smoke-tests.outputs.OUTPUT_DIR }}
           event_name: ${{ github.event_name }}
           actor: ${{ github.actor }}
+          upload_docs: true
           SSH_KEY: ${{ secrets.SSH_KEY }}
           KNOWN_HOSTS: ${{ secrets.KNOWN_HOSTS }}
           SSH_USER: ${{ secrets.SSH_USER }}
@@ -171,6 +172,7 @@ jobs:
           target_dir: ${{ needs.smoke-tests.outputs.OUTPUT_DIR }}
           event_name: ${{ github.event_name }}
           actor: ${{ github.actor }}
+          upload_docs: false
           SSH_KEY: ${{ secrets.SSH_KEY }}
           KNOWN_HOSTS: ${{ secrets.KNOWN_HOSTS }}
           SSH_USER: ${{ secrets.SSH_USER }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
           [[ $FAIL == 1 ]] && exit 1 || echo "Branch name depth check passed."
         shell: bash
 
-  build-gha-x86:
+  gha-x86:
     runs-on: ubuntu-latest
     needs: smoke-tests
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -61,132 +61,117 @@ jobs:
...
+          bin_name: ${{ matrix.bin_name }}
+          artifact_name: ${{ matrix.bin_name }}-binary
+          git_branch: ${{ needs.smoke-tests.outputs.GIT_BRANCH }}
+          git_tag: ${{ needs.smoke-tests.outputs.GIT_TAG }}
+          event_name: ${{ github.event_name }}
 
   deploy:
     if: github.event_name != 'pull_request'
-    needs: [smoke-tests, build]
+    needs: [smoke-tests, build-gha-x86, build-self-hosted]
```
PromoFaux, line 139: Just reading over this file again. I'm wondering if `smoke-tests` is `need`ed at the `deploy` stage?

if `build-gha-x86` or `build-self-hosted` are true, then it stands to reason that `smoke-tests` must also be true (else they would never start)

I think I've logicked that correctly? Why isn't logicked a real word?

DL6ER:
Further improvement (direct deployment) on its way... changed back to draft mode and will re-request review later when it's ready.

DL6ER:
We used to have a combined `deploy`ment step as - pre-v6 - we have still been using cross-compilation toolchains and been only been able to test the `x86_64` build but not the `arm` builds. Hence, we wanted to wait for `arm` binary deployment until we had a success message from the CI tests.

With Pi-hole v6, we use `docker buildx` instead of cross-compilation allowing us to test each of the binaries on their own. Hence, we can directly deploy everything which passes the CI tests. If any architecture fails due to build for some arch-specific thing, merging the PR will still be blocked by the failed CI build. The fact that the other binaries have been deployed is harmless as they have been tested okay themselves.

Updated workflow screenshot with immediate deployment:
grafik

resulting in:
grafik

DL6ER:
On its last iteration, it got even simpler:
grafik

PromoFaux (approved):

Issue #1920, opened 2024-03-28
dns.hosts(0): invalid hostname


asdf4w3t5:
https://discourse.pi-hole.net/t/local-dns-settings-x-invalid-value/68985/4

![image](https://github.com/pi-hole/FTL/assets/99051107/2c8709e7-9358-485a-92c8-900fdf10eb92)

seems the same

https://github.com/pi-hole/FTL/blob/development-v6/src/tools/gravity-parseList.c#L54

no hostnames work since the update of the most recent one.

It is unrelated to the forum thread, also this occurs with pre-existing domains and is very annoying since I have a lot of them.
It is validating domains I have entered months ago and saying they are wrong. And not letting add new ones.

I have deleted all of them out of pihole and just added them to my hosts in my unbound server so I don't have to deal with the pihole and can test better

it looks like any domain that is asdasdasd.a <-- single character at TLD fails 
this testing is too greedy, Let people put in whatever they want or at least check a box to input whatever they want.. They already had to check "expert" to get here.

DL6ER:
The forum post was about capital letters being rejected, we fixed this.

> It is validating domains I have entered months ago and saying they are wrong

The issue you are reporting is about a TLD with a length of one whereas the minimum enforced length is 2:

https://github.com/pi-hole/FTL/blob/91d8738a7f32eacabb0e65c760a4d505295e0051/src/tools/gravity-parseList.c#L97-L101

following https://data.iana.org/TLD/tlds-alpha-by-domain.txt which does not list a single one-character TLD.

However, I do agree with you that users may define such TLDs locally so we should relax this as well. Nonetheless, I want to advise against using TLDs that are not defined in the wild for you private use. Imagine they get defined at some point in the future and suddenly some of your internal configurations become problematic. This threat is not hypothetical as we have seen more recently where a malicious party has registered `fritz.box` where this is the host name of a lot of popular routers out in the wild.

DL6ER:
Also, looking at the title of this PR:
```
dns.hosts[0]: invalid hostname
```
will have shown you where to look. It means that the zeroth element was causing the validation issue.

asdf4w3t5:
>The forum post was about capital letters being rejected, we fixed this.

yes that is how I was able to say it is unrelated

This issue did not occur until I updated yesterday. Last time I updated was a month ago or so

>will have shown you where to look. It means that the zeroth element was causing the validation issue.

So I should have fixed it?

DL6ER:
[...]

PR #1922, opened 2024-03-28
Allow one-character TLDs in hostname validation

DL6ER:

See title, fixes https://github.com/pi-hole/FTL/issues/1920

---

**Related issue or feature (if applicable):** https://github.com/pi-hole/FTL/issues/1920

---

## Checklist:

### DIFF
--- a/src/tools/gravity-parseList.c
+++ b/src/tools/gravity-parseList.c
@@ -94,12 +94,6 @@ inline bool __attribute__((pure)) valid_domain(const char *domain, const size_t 
 	if(domain[last_dot + 1] == '-' || domain[len - 1] == '-')
 		return false;
 
-	// TLD length check
-	// The last label must be at least 2 characters long
-	// (len-1) because we start counting from zero
-	if((len - 1) - last_dot < 2)
-		return false;
-
 	return true;
 }
 


src/tools/gravity-parseList.c
```diff
@@ -95,9 +95,9 @@ inline bool __attribute__((pure)) valid_domain(const char *domain, const size_t
 		return false;
 
 	// TLD length check
-	// The last label must be at least 2 characters long
+	// The last label must be at least one character long
```
rdwebdesign, line 98: Maybe we could add the issue number here, to help us understand in the future why this is different then the expected.

```suggestion
	// The last label must be at least one character long (see FTL #1920)
```

DL6ER: We have never done this before and I don't think it's good practice. You can always use `git blame` to get context. Otherwise, it's difficult to maintain such links when stuff gets rewritten/extended in the future and references may become outdated

rdwebdesign: I agree the issue number is not the best idea.

I just think we should have a better comment explaining the text _"at least one character long"_, since this is not what we expect for TLD and we might forget the reason why this was set.

Anyway, it's just a suggestion.

rdwebdesign: The check and comments were removed.

Explanation (in your words):
"We already check for empty labels and every label that is non-empty has at least one character in it".

rdwebdesign (approved):
Tested using the same hostname reported on the issue (1920):

image

Issue #1975, opened 2024-05-25
PTR requests not being generated for local clients


moya2162:
### Versions

Core
    Version is 250a44c (Latest: null)
    Branch is development-v6
    Hash is 250a44ca (Latest: N/A)
Web
    Version is 59b1dcf (Latest: null)
    Branch is development-v6
    Hash is 59b1dcfc (Latest: N/A)
FTL
    Version is vDev-825a14a (Latest: null)
    Branch is development-v6
    Hash is 825a14a6 (Latest: N/A)

### Platform

- OS and version:  Armbian Bookworm

- Platform:  Docker

### Expected behavior

Pi-Hole should be issuing PTR request for clients to be forwarded upstream. 

### Actual behavior / bug

Pi-Hole is not issuing PTR requests. Enabled debug.resolver and following error is repeating for all my local clients, both ipv4 and ipv6:

```
2024-05-25 13:43:23.450 DEBUG_RESOLVER getIfaceFromIP(): "SELECT interface FROM network JOIN network_addresses ON network_addresses.network_id = network.id WHERE network_addresses.ip = ? AND interface != 'N/A' AND interface IS NOT NULL;" with ? = "192.168.X.XX"
2024-05-25 13:44:14.303 DEBUG_RESOLVER  ---> not found
2024-05-25 13:44:14.304 DEBUG_RESOLVER getIfaceFromIP(): "SELECT interface FROM network JOIN network_addresses ON network_addresses.network_id = network.id WHERE network_addresses.ip = ? AND interface != 'N/A' AND interface IS NOT NULL;" with ? = "2601:204:0:XXXX:XXXX:XXXX:XXXX:XXXX"
2024-05-25 13:45:15.840 DEBUG_RESOLVER  ---> not found
2024-05-25 13:45:15.842 DEBUG_RESOLVER getIfaceFromIP(): "SELECT interface FROM network JOIN network_addresses ON network_addresses.network_id = network.id WHERE network_addresses.ip = ? AND interface != 'N/A' AND interface IS NOT NULL;" with ? = "fd71:2002:f23:0:XXXX:XXXX:XXXX:XXXX"
2024-05-25 13:45:15.846 DEBUG_RESOLVER  ---> not found
```

### Additional Info

- dns.bogusPriv = disabled
- dns.EDNS0ECS = enabled
- dns.listeningMode = ALL
- resolver.resolveIPv4 = enabled
- resolver.resolveIPv6 = enabled
[...]

PR #1962, opened 2024-05-17
Add Clang compiler support and update project to C17 standard

DL6ER:
# What does this implement/fix?

Add Clang compiler support, tested with
- Clang 14.0.0 (Ubuntu 22.04.3 LTS),
- Clang 16.0.2 (Alpine 3.18), and
- Clang 17.0.6 (Alpine Edge)

[C17](https://en.wikipedia.org/wiki/C17_(C_standard_revision)) is the informal name for ISO/IEC 9899:2018 and was published in June 2018. It replaced [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision)) and will eventually be superseded by C23 when it is published. C17 fixes numerous minor defects in C11 without introducing new language features. It required a number of subtle code changes without any real changes. We also bump the required minimum CMake version to the lowest version supporting the C17 standard.

---

**Related issue or feature (if applicable):** N/A

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** N/A

---
**By submitting this pull request, I confirm the following:** 

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against. 
2. I have commented my proposed changes within the code.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

## Checklist:

- [x] The code change is tested and works locally.
- [x] I based my code and PRs against the repositories `developmental` branch.
- [x] I [signed off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits. Pi-hole enforces the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
- [x] I [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all my commits. Pi-hole requires signatures to verify authorship
- [x] I have read the above and my PR is ready for review.

PR #1976, opened 2024-05-27
Improve internal DNS resolver

DL6ER:

This PR implements TCP lookups for the internal DNS client used to derive names for clients and upstream servers. The reason for adding this was #1975 during which we have found that existing UDP lookups may fail silently because of UDP packet fragmentation/truncation. Using TCP avoids this issue altogether. It can reliably be triggered in `development-v6` when setting `8.8.8.8` as upstream server **and** enabling DNSSEC validation.

While discussing #1975, we found another glitch that causes names of upstream servers to never be resolved when starting with no/an empty database. Only after the first restart of FTL after database initialization, upstream server names are resolved. This is fixed now as well.

---

**Related issue or feature (if applicable):** #1975 

---

## Checklist:


DL6ER:
Note that the failed spellcheck will be fixed by https://github.com/pi-hole/FTL/pull/1962

src/args.c
```diff
@@ -499,15 +499,26 @@ void parse_args(int argc, char* argv[])
 	}
 
 	// Local reverse name resolver
-	if(argc == 3 && strcasecmp(argv[1], "ptr") == 0)
+	if((argc == 3 || argc == 4) && strcasecmp(argv[1], "ptr") == 0)
```
yubiuser, line 502: This new argument is not documented in the help text

Screenshot at 2024-05-28 21-28-12

yubiuser (changes requested):

src/args.c
DL6ER: grafik

yubiuser (approved):

PR #1994, opened 2024-06-13
Check if the newly set password is the same as the old one

DL6ER:

This is especially worth adding in the context of forced passwords (via env var) as the password is always "set" on each config change. This PR effectively prevents FTL from wiping all sessions and forcing the user to re-login when changing even settings which are 100% unrelated to the API.

---

---

## Checklist:

### DIFF
--- a/src/config/password.c
+++ b/src/config/password.c
@@ -601,6 +601,13 @@ int run_performance_test(void)
 
 bool set_and_check_password(struct conf_item *conf_item, const char *password)
 {
+	// Check if the newly set password is the same as the old one
+	if(verify_password(password, config.webserver.api.pwhash.v.s, false) == PASSWORD_CORRECT)
+	{
+		log_debug(DEBUG_CONFIG, "Password unchanged, not updating");
+		return true;
+	}
+
 	// Get password hash as allocated string (an empty string is hashed to an empty string)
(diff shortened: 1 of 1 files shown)


src/config/password.c
```diff
@@ -601,6 +601,13 @@ int run_performance_test(void)
 
 bool set_and_check_password(struct conf_item *conf_item, const char *password)
 {
+	// Check if the newly set password is the same as the old one
+	if(verify_password(password, config.webserver.api.pwhash.v.s, false) == PASSWORD_CORRECT)
```
yubiuser, line 605: Do we also need to check if no password is set (or empty) to exit the check early?

rdwebdesign:
Do we still need this check?

https://github.com/pi-hole/FTL/blob/2406e1a70ee9ef83b98ed672df2705d02b016da2/src/api/config.c#L289-L294

src/config/password.c
rdwebdesign: This is already checked here:

https://github.com/pi-hole/FTL/blob/2406e1a70ee9ef83b98ed672df2705d02b016da2/src/config/password.c#L611-L621

DL6ER: I agree it's worthwhile to do this before investing any work in trying to validate a hash (it takes a lot of time). I have added and pushed this.

yubiuser (approved):

PR #1995, opened 2024-06-14
 Limit app password permissions

DL6ER:

Limit app password permissions by default. Add new `webserver.api.app_sudo` mode for users to remove this new limitation if they really need to

grafik

----

---

## Checklist:


src/api/config.c
```diff
@@ -1037,6 +1037,16 @@ int api_config(struct ftl_conn *api)
...
 		return api_config_get(api);
 
+	// Check if this is an app session and reject the request if app sudo
+	// mode is disabled
+	if(api->session != NULL && api->session->app && !config.webserver.api.app_sudo.v.b)
+	{
+		return send_json_error(api, 403,
+		                       "forbidden",
+		                       "config read-only",
+		                       "app session but webserver.api.app_sudo is false");
```
rdwebdesign, line 1047: Maybe we should use a more detailed and user friendly message:

```suggestion
		                       "Unable to change configuration (read-only)",
		                       "The current app session is not allowed to modify Pi-hole config settings (webserver.api.app_sudo is false)");
```

rdwebdesign (dismissed):

src/api/config.c
yubiuser: I agree with rd

DL6ER:
grafik

yubiuser:
I think I found a bug here.
Login via app password and try a config change. Won't work. Enabled `app_sudo` > config change will work. Disable `app_sudo` > Config change will still work

(Also the web interface won't show this active session as coming from a app password)

```
chrko@ThinkPad-X230:~$ curl -k -XPOST "https://pi.hole/api/auth" --data '{"password":"nDgnVeJVko5buF1HMA/zfZ7KbTDPAbOpv4zvlDkIq8I="}'
{
	"session":	{
		"valid":	true,
		"totp":	false,
		"sid":	"pNMUGNtG42aiQAIwtYT5ng=",
		"csrf":	"eKihaYo8VP9sNHfJLg6syA=",
		"validity":	300,
		"message":	"app-password correct"
	},
	"took":	1.1221466064453125
}chrko@ThinkPad-X230:~$ curl -k -X UT "https://pi.hole/api/config/dns/upstreams/1.1.1.1" --data '{"sid":"pNMUGNtG42aiQAIwtYT5ng="}'
{
	"error":	{
		"key":	"forbidden",
		"message":	"Unable to change configuration (read-only)",
		"hint":	"The current app session is not allowed to modify Pi-hole config settings (webserver.api.app_sudo is false)"
	},
	"took":	0.00031328201293945312
}chrko@ThinkPad-X230:~$ curl -k -X PUT "https://pi.hole/api/config/dns/upstreams/1.1.1.1" --data '{"sid":"pNMUGNtG42aiQAIwtYT5ng="}'
{
	"took":	0.050773859024047852
}chrko@ThinkPad-X230:~$ curl -k -X PUT "https://pi.hole/api/config/dns/upstreams/2.2.2.2" --data '{"sid":"pNMUGNtG42aiQAIwtYT5ng="}'
{
	"took":	0.057279586791992188

```

DL6ER:
Ah, yes. Your case is special. The reason is that FTL restarted in between (to add the new DNS server) and sessions restoring from the database had a small copy-paste bug causing the `app` status not to be saved to the database and, hence, all `app` sessions backed up during a restart became full-blown sessions after the restart and the setting became ineffective by design for them.

src/config/config.c
```diff
@@ -1018,6 +1018,12 @@ void initConfig(struct config *conf)
 	conf->webserver.api.app_pwhash.d.s = (char*)"";
 	conf->webserver.api.app_pwhash.c = validate_stub; // Only type-based checking
 
+	conf->webserver.api.app_sudo.k = "webserver.api.app_sudo";
+	conf->webserver.api.app_sudo.h = "Should application password API sessions be allowed to modify config settings?\n Setting this to true allows third-party applications using the application password to modify advanced settings, e.g., the upstream DNS servers, DHCP server settings, or changing passwords. This setting should only be enabled if really needed and only if you trust the applications using the application password.";
```
yubiuser, line 1022: I think this affects all settings, not only advanced settings.

```suggestion
	conf->webserver.api.app_sudo.h = "Should application password API sessions be allowed to modify config settings?\n Setting this to true allows third-party applications using the application password to modify settings, e.g., the upstream DNS servers, DHCP server settings, or changing passwords. This setting should only be enabled if really needed and only if you trust the applications using the application password.";
```

yubiuser (commented):
Somewhere in `send_json_error()` must be a type (tab instead of space) between `message:` and the actual message.

```
	"error":	{
		"key":	"forbidden",
		"message":	"Unable to change configuration (read-only)",
		"hint":	"The current app session is not allowed to modify Pi-hole config settings (webserver.api.app_sudo is false)"
	},

```

src/config/config.c
DL6ER: We also need to edit `test/pihole.toml` or CI tests will complain that the code does not match the config file. I will, hence, do the changed locally and push them instead of simply accepting the suggestion here.

DL6ER:
`send_json_error()` uses cJSON to create the output. If you have `webserver.api.prettyJSON = false`, the output will be most compact, if the value is `true`, the expected separator between between keys and values is [`\t`](https://github.com/pi-hole/FTL/commit/6fe4a512935443664b4bfeae4f97d85b8033c193#diff-7a0dcba91f37fdd2ab05bf4cfa99026a109cdb4f9147badfa3383238f1dd8560R1655) so just as in your example.

Have a look at https://github.com/pi-hole/FTL/blob/0367117cad371bb9cbe41dfd0ead69c39e8e5192/src/webserver/http-common.c#L23-L47
where you can also see that the expected (formatted) output has tabs as separators.

yubiuser:
I have `webserver.api.prettyJSON = true` and was only surprised of the different distances between `key` and `value`

yubiuser (approved):

yubiuser (approved):

PR #2001, opened 2024-06-25
Improve Query Log suggestions

DL6ER:

Implement actual TOP suggestions for the Query Log by ensuring the most relevant domains and clients are returned (sorted by the most active being at the top) by the API `/api/queries/suggestions` endpoint. The result of this is used in the dropdown menus on the Query Log Advanced Filtering box:

image

---

**Related issue or feature (if applicable):** [Discourse](https://discourse.pi-hole.net/t/unable-to-select-correct-client-by-name-in-advanced-filter/70863)

---

## Checklist:


.github/Dockerfile
```diff
@@ -5,13 +5,13 @@ WORKDIR /app
...
+ENV CI_ARCH=${CI_ARCH}
 ARG GIT_BRANCH="test"
-ENV GIT_BRANCH ${GIT_BRANCH}
+ENV GIT_BRANCH=${GIT_BRANCH}
 ARG GIT_TAG="test"
-ENV GIT_TAG ${GIT_TAG}
+ENV GIT_TAG=${GIT_TAG}
 ARG BUILD_OPTS=""
-ENV BUILD_OPTS ${BUILD_OPTS}
+ENV BUILD_OPTS=${BUILD_OPTS}
```
DL6ER, line 14: Fixing CI warnings
```
 4 warnings found (use --debug to expand):
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 8)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 10)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 12)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 14)
```

DL6ER:
Confirmed working by OP on Discourse

yubiuser (commented):
First, this PR is a great enhancement for the suggested domains/clients.

But I would like to suggest to return other data than it is returning right now: Currently, it returns the `max_count` items from the resp. SQL queries. Esp. for domains this might be undesired. If I have a lot of high count permitted domains, but only a low count number of blocked domains, it will return only permitted domains as suggestions. But I think it should return both top allowed and top blocked domains. So maybe we could instead re-use the data we have for the top domains already (/stats/top_domain) and use those as suggestions instead?
For dis

DL6ER:
> But I think it should return both top allowed and top blocked domains. 

Yes, we can make this return, e.g. 30 top permitted and 30 top blocked domains just to populate the dropdown.

> maybe we could instead re-use the data we have for the top domains already (/stats/top_domain) and use those as suggestions instead?

This endpoint still uses the internal data structure and does not use the database. On the one hand, this data is already pre-processed and accessing it is fast, on the other hand, accessing this data blocks DNS resolution during the time it is accessed and is not very flexible. I will think about this, maybe we just don't need any flexibility here.

DL6ER:
As you can assume from the prolonged inactivity on this PR, I'm really thorn how to proceed on this matter. True, your point about "simply" reusing the data from the Top Lists is valid and implementing it would actually be trivial.

The Top Lists use the internal data structures prepared and updated by FTL on every incoming query. They have the benefit of being pre-computed and being available in a directly machine-digestible form. Reading from this data takes only picoseconds. So far, so good. The drawback is here that we need to use locking ensure we have consistent data when new queries are incoming. So every API call has to wait until the last incoming query has finished processing, and then the next query has to wait for the API call to finish before it can be processed. Typically, this isn't an issue except for lowest-end hardware getting hammered with millions of queries a day. This can safely be considered unsupported for good reasons, though.

However, my (longer-term) goal is to actually go away from utilizing the internal data to using the database. The Query Log was the first big success on this as it allows us to easily leverage all the powerful SQL features without the aforementioned locking issues possibly preventing queries to be processed in a timely fashion. This has already been done. Now, I tried the same for the suggestions.

The issue is that getting the same numbers from the database requires a full scan (sadly no index can be used immediately) on the respective tables like `SELECT COUNT(*), domain_id GROUP BY domain_id` but with the added complexity of having to use a `JOIN` to limit this computation to the most recent 24 hours. Without this limitation (top entries over the lifetime of the database), the effort would actually be less due to the tree structure.

So, it will either be:

1. Refrain to using the internal data and accept that there may be delays in query processing. The delay will be so super-small so this is actually something worth considering.
2. Add metatables to the database basically exporting FTL's internal knowledge of the Top List composition into the database. As they are momentary data (24h rolling windows), it'd be fine for the to exist in the in-memory database only and don't need to be copied to the on-disk database.

As no. 2 will require a substantial amount of code and will actually have some intrinsic delays (the internal FTL data is always 100% correct, the database only immediately after an update and we do not want to update on *every* single query), we will probably go with no. 1 and this long comment here might remind me in the future why we are not doing it for the Top Lists API endpoints, either.

DL6ER:

yubiuser (changes requested):
It does not make sense to show the same client name multiple times
Screenshot at 2024-07-17 22-01-19

DL6ER:
This can happen when devices appear with different IP addresses on the network, e.g. once on Ethernet, once on WiFi. It may also happen when using a non-deterministic DHCP server

yubiuser (approved):

PR #2004, opened 2024-07-01
Improve NTP implementation

DL6ER:

Start NTP server thread(s) only after the first successful NTP time sync. This ensures the time we offer is correct. As a by-product this also gives some time for alternative NTP servers (e.g., `crony`) to finish their startup so we don't take the port before they can do it. Also, disable DNSSEC time-windows validation until first sync succeeded to avoid DNSSEC-related issues during NTP sync.

If the time is updated by a large amount, restart FTL to ensure we load the correct history window.

---

---

## Checklist:

### DIFF
--- a/src/ntp/client.c
+++ b/src/ntp/client.c
@@ -39,10 +39,9 @@
 #include <sys/timex.h>
 // log_ntp_message()
 #include "database/message-table.h"
-// load_queries_from_disk()
-#include "database/query-table.h"
 // check_capability()
 #include "capabilities.h"
+
 struct ntp_sync
 {
 	bool valid;
@@ -593,20 +592,11 @@ static void *ntp_client_thread(void *arg)
 	prctl(PR_SET_NAME, thread_names[NTP_CLIENT], 0, 0, 0);
 
 	// Run NTP client
-	bool first_run = true;
 	bool ntp_server_started = false;
 	while(!killed)
 	{
 		// Run NTP client
 		const bool success = ntp_client(config.ntp.sync.server.v.s, true, false);
-
-		// Load queries from database after first NTP synchronization
-		if(first_run)
-		{
-			load_queries_from_disk();
-
-			first_run = false;
-		}
 
 		if(success && !ntp_server_started)
 		{
@@ -636,8 +626,7 @@ bool ntp_start_sync_thread(pthread_attr_t *attr)
 	   strlen(config.ntp.sync.server.v.s) == 0 ||
 	   config.ntp.sync.interval.v.ui == 0)
 	{
-		log_info("NTP sync is disabled - NTP server will not be available");
-		load_queries_from_disk();
+		log_info("NTP sync is disabled");
 		ntp_server_start();
 		return false;
 	}
@@ -649,10 +638,6 @@ bool ntp_start_sync_thread(pthread_attr_t *attr)
 	if(!check_capability(CAP_SYS_TIME))
 	{
 		log_warn("Insufficient permissions to set system time (CAP_SYS_TIME required), NTP client not available");
-		// Send SIGUSR7 to embedded dnsmasq instance to signal time is
-		// assumed to be correct
-		kill(main_pid(), SIGUSR7);
-		load_queries_from_disk();
 		ntp_server_start();
 		return false;
 	}
@@ -661,10 +646,6 @@ bool ntp_start_sync_thread(pthread_attr_t *attr)
 	if(pthread_create(&threads[NTP_CLIENT], attr, ntp_client_thread, NULL) != 0)
 	{
 		log_err("Cannot create NTP client thread");
-		// Send SIGUSR7 to embedded dnsmasq instance to signal time is
-		// assumed to be correct
-		kill(main_pid(), SIGUSR7);
-		load_queries_from_disk();
 		ntp_server_start();
 		return false;
 	}
--- a/src/ntp/client.c
+++ b/src/ntp/client.c
@@ -603,6 +603,12 @@ static void *ntp_client_thread(void *arg)
 		{
 			load_queries_from_disk();
 
+			// If this was the first run and NTP time synchronization was
+			// successful, we send SIGUSR7 to the embedded dnsmasq instance
+			// to signal time is now guaranteed to be correct
+			if(success)
+				kill(main_pid(), SIGUSR7);
+
 			first_run = false;
 		}
 
@@ -634,6 +640,9 @@ bool ntp_start_sync_thread(pthread_attr_t *attr)
 	   config.ntp.sync.interval.v.ui == 0)
 	{
 		log_info("NTP sync is disabled - NTP server will not be available");
+		// Send SIGUSR7 to embedded dnsmasq instance to signal time is
+		// assumed to be correct
+		kill(main_pid(), SIGUSR7);
 		load_queries_from_disk();
 		return false;
 	}
--- a/src/ntp/client.c
+++ b/src/ntp/client.c
@@ -657,7 +657,7 @@ bool ntp_start_sync_thread(pthread_attr_t *attr)
 	// in starting the thread.
 	if(!check_capability(CAP_SYS_TIME))
 	{
-		log_warn("Insufficient permissions to set system time, NTP client not available");
+		log_warn("Insufficient permissions to set system time (CAP_SYS_TIME required), NTP client not available");
 		// Send SIGUSR7 to embedded dnsmasq instance to signal time is
 		// assumed to be correct
 		kill(main_pid(), SIGUSR7);
--- a/src/ntp/client.c
+++ b/src/ntp/client.c
@@ -605,13 +605,16 @@ static void *ntp_client_thread(void *arg)
 		// Get time after NTP sync
 		const time_t after = time(NULL);
 
-		// If the time was updated by more than one hour, restart FTL to
-		// import recent data. This is relevant when the system time was
-		// set to an incorrect value (e.g., due to a dead CMOS battery
-		// or overall missing RTC) and the time was off.
-		if(first_run && after - before > 3600)
+		// If the time was updated by more than a certain amount,
+		// restart FTL to import recent data. This is relevant when the
+		// system time was set to an incorrect value (e.g., due to a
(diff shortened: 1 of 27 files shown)


PromoFaux:
Without `SYS_TIME` on a container:

```
pihole  | 2024-07-01 20:32:06.427 BST [243M] WARNING: Insufficient permissions to set system time, NTP client not available
```

👍 

Is there some value in including a hint that `SYS_TIME` is unavailable in this warning message?

<details><summary>Similarly</summary>
<p>
Maybe something here about `SYS_NICE`

```
pihole  | 2024-07-01 20:32:33.023 BST [244M] WARNING: Cannot set process priority to -10: Permission denied. Process priority remains at 0
```


</p>
</details>

yubiuser (commented):
Why does it try to open the database if it errored before?

```
nanopi@nanopi:~$ pihole-FTL ntp de.ntp.org
Using NTP server: de.ntp.org
Error NTP client: Cannot resolve NTP server address: Name has no usable address "de.ntp.org"
Cannot open database in read-write mode
```

DL6ER:
> Why does it try to open the database if it errored before?

Because it wants to store this in the `message` table. But it does make sense to never store in the message table when running interactively. Pushed now.

Gontier-Julien:
@DL6ER With this PR would this mean that it won't throw a warning anymore when there already is an ntp server on the server/system?
Also i don't know/think it already exist so correct me if i'm wrong as always, but there isn't an single option to disable the ntp client & server currently no? As i already have an ntp server that also sync the clock of the server it self there isn't much of a need for pi-hole to act either as a server or client then, so maybe as single option could be added to disable it?

DL6ER:
> With this PR would this mean that it won't throw a warning anymore when there already is an ntp server on the server/system?

Well, it'd warn to the log file that the port is already taken and, hence, Pi-hole won't start its own NTP server. But you can safely ignore this warning if you know you have a local NTP server. There will be no warning added to the Pi-hole diagnosis system.

---

> an single option to disable the ntp client & server

There are already options to disable it, even when it is not a single one:
```
ntp.ipv4.active = false
ntp.ipv6.active = false
ntp.sync.active = false
```
will disable all NTP components inside Pi-hole. I don't think we really need **one** option.

Gontier-Julien:
> Well, it'd warn to the log file that the port is already taken and, hence, Pi-hole won't start its own NTP server. But you can safely ignore this warning if you know you have a local NTP server. There will be no warning added to the Pi-hole diagnosis system.

Well awesome then! ^^

> There are already options to disable it, even when it is not a single one:
> 
> ```
> ntp.ipv4.active = false
> ntp.ipv6.active = false
> ntp.sync.active = false
> ```
> 
> will disable all NTP components inside Pi-hole. I don't think we really need **one** option.

That what i've done currently, but it fine by me too tbh, i was think so it simple for anyone, but i guess that if someone already have a ntp server then it should be easy

Alright then, thank ^^

yubiuser (commented):
What signal are you sending to `dnsmasq` to signal correct time? You created a custom `SIGUSR7` which is interpreted as `SIGUP`

```
2024-07-04 21:06:30.699 CEST [670283/T670448] INFO:  -> DNS cache records: 4020
2024-07-04 21:06:30.699 CEST [670283/T670448] INFO:  -> Known forward destinations: 2
2024-07-04 21:06:30.700 CEST [670283/T670464] INFO: NTP server listening on 0.0.0.0:123 (IPv4)
2024-07-04 21:06:30.700 CEST [670283/T670465] INFO: NTP server listening on :::123 (IPv6)
2024-07-04 21:06:30.703 CEST [670283M] INFO: Received SIGHUP, flushing cache and re-reading config
2024-07-04 21:06:30.704 CEST [670283M] INFO: Blocking status is enabled

```

`dnsmasq` config says, one should use `SIGINT`

____

I noticed that queries that happen after `dnsmasq` start before  starting DNSSEC validation are not logged to the query database. Basically everything between

```
Jul  4 21:18:55 dnsmasq[670676]: DNSSEC validation enabled
Jul  4 21:18:55 dnsmasq[670676]: DNSSEC signature timestamps not checked until receipt of SIGINT
....
Jul  4 21:19:03 dnsmasq[670676]: now checking DNSSEC signature timestamps

```

Is this unavoidable?

DL6ER:
> What signal are you sending to `dnsmasq` to signal correct time? You created a custom `SIGUSR7` which is interpreted as `SIGUP`
>
> [...]
>
> `dnsmasq` config says, one should use SIGINT

Yes. The new signal was introduced because `SIGINT` has a different effect when starting dnsmasq in debug mode (`pihole-FTL -d`): it terminates the program.
This is also mentioned in `man dnsmasq` but it doesn't seem useful here. Hence , I introduced a new signal that does the exact same and reuses the same code *except* going the `exit()` path.

The logged message contains the `SIGHUP` a fixed string which is now outdated. It could already have be triggered with `SIGINT` before as well so it wasn't necessarily accurate anyway - removed with the last commit.

---

> I noticed that queries that happen after dnsmasq start before starting DNSSEC validation are not logged to the query database.
>
> [...]

> Is this unavoidable?

Sorry, I see I have not mentioned this before. It is hard to avoid. The issue is that we can only add queries after the database importing from disk or we will not have the `domain`, `client`, and `forward` IDs as we have not imported these at this point.

**edit** Turns out it helps to step back and look at things later again - it is actually less hard, I just had to arrange some code to disentangle database initialization. We have now two parts, one running during database initialization (importing all the dependencies for *new* queries) and another one importing the actual queries which can be delayed until we know we have a proper time (or will never get one).

---

> How can a DNSSEC validation be secure before the DNSSEC timestamp can be checked?

Because `dnsmasq` still performs all the checks, it will not just consider `signature expired` or `signature not yet valid` as errors. The cache is flushed once time is ready to ensure all domains are re-checked and we don't trust an `SECURE` which was not time-checked.

yubiuser:
I fear it did not work. The query at `23:03:10` has not been logged to the database.

```
nanopi@nanopi:~$ cat /var/log/pihole/pihole.log | grep DNSSEC
Jul  5 23:03:08 dnsmasq[684108]: DNSSEC validation enabled
Jul  5 23:03:08 dnsmasq[684108]: DNSSEC signature timestamps not checked until receipt of SIGINT
...
Jul  5 23:03:15 dnsmasq[684108]: now checking DNSSEC signature timestamps

nanopi@nanopi:~$ cat /var/log/pihole/pihole.log | grep sigok
Jul  5 23:03:10 dnsmasq[684108]: query[A] sigok.ippacket.stream from 10.0.1.202
Jul  5 23:03:10 dnsmasq[684108]: forwarded sigok.ippacket.stream to 8.8.8.8
Jul  5 23:03:10 dnsmasq[684108]: reply sigok.ippacket.stream is <CNAME>
Jul  5 23:03:10 dnsmasq[684108]: reply sigok.rsa2048-sha256.ippacket.stream is 195.201.14.36

nanopi@nanopi:~$ pihole-FTL sqlite3 -h  /etc/pihole/pihole-FTL.db "Select * from queries where domain is 'sigok.ippacket.stream'"
id       timestamp         type  status  domain                 client     forward     additional_info  reply_type  reply_time            dnssec  list_id
-------  ----------------  ----  ------  ---------------------  ---------  ----------  ---------------  ----------  --------------------  ------  -------
2930152  1716923761.61009  1     2       sigok.ippacket.stream  10.0.1.24  8.8.8.8#53  (null)           3           0.220105886459351     1       (null) 
2930156  1716923761.41762  1     2       sigok.ippacket.stream  10.0.1.24  8.8.8.8#53  (null)           12          0.181222915649414     5       (null) 
2930240  1716923869.86778  1     2       sigok.ippacket.stream  10.0.1.24  8.8.8.8#53  (null)           12          0.207485675811768     5       (null) 
2930250  1716923870.08774  1     2       sigok.ippacket.stream  10.0.1.24  8.8.8.8#53  (null)           3           0.401464700698853     1       (null) 
3569511  1720119147.28513  1     2       sigok.ippacket.stream  10.0.1.59  8.8.8.8#53  (null)           3           0.428323030471802     1       (null) 
3569513  1720119148.74269  1     3       sigok.ippacket.stream  10.0.1.59  (null)      (null)           3           0.000569820404052734  1       (null) 
3570094  1720119705.92446  1     2       sigok.ippacket.stream  10.0.1.59  8.8.8.8#53  (null)           3           0.266446113586426     1       (null) 
3570098  1720119710.19207  1     3       sigok.ippacket.stream  10.0.1.59  (null)      (null)           3           0.000896692276000977  1       (null) 

```

Only part of the query has been captured

Screenshot at 2024-07-05 23-11-44

src/ntp/client.c
```diff
@@ -588,52 +589,87 @@ bool ntp_client(const char *server, const bool settime, const bool print)
...
+		// Get time after NTP sync
+		const time_t after = time(NULL);
 
-		// Load queries from database after first NTP synchronization
-		if(first_run)
+		// If the time was updated by more than one hour, restart FTL to
+		// import recent data. This is relevant when the system time was
+		// set to an incorrect value (e.g., due to a dead CMOS battery
+		// or overall missing RTC) and the time was off.
+		if(first_run && after - before > 3600)
```
yubiuser, line 612: `after - before > 3600` should we use `abs()` here. Just in case - for any reason -  `before > after`

rdwebdesign: Another question:

Should we use a hard-coded value of 3600 (1 hour) or the value of `ntp.sync.interval.v.ui`?

DL6ER: `before > after` means the time has been set back, like from year 2030 to 2024. I don't think this is likely to happen as most updates will either pull the device from Jan 1st, 1970 or from some earlier time stored by `fake-hwclock`. Anyway, it also won't hurt err on the side of caution.

> Should we use a hard-coded value of 3600 (1 hour) or the value of `ntp.sync.interval.v.ui`?

I don't think abusing the interval is a useful compromise here. Users may set the interval to a few seconds ... or even one ... you never know. A sensible constant may be `GCinterval` currently defaulting to 10 minutes.

yubiuser (approved):
Maybe you can update the PR text because some if it is not valid anymore after the last commits.

PR #2005, opened 2024-07-01
Show Unauthorized warnings only in debug mode

DL6ER:

Log API Unauthorized warnings caused by `send_json_error()` only if `debug.api` is `true`.

---

---

## Checklist:

### DIFF
--- a/src/api/config.c
+++ b/src/api/config.c
@@ -713,7 +713,7 @@ static int api_config_patch(struct ftl_conn *api)
 			return send_json_error_free(api, 400,
 			                            "bad_request",
 			                            "This config option can only be set in pihole.toml, not via the API",
-			                            key, true);
+			                            key, true, true);
 		}
 
 		// Check if this is a write-only config item with the placeholder value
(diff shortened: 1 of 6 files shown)


src/api/config.c
```diff
@@ -713,7 +713,7 @@ static int api_config_patch(struct ftl_conn *api)
 			return send_json_error_free(api, 400,
 			                            "bad_request",
 			                            "This config option can only be set in pihole.toml, not via the API",
-			                            key, true);
+			                            key, true, false);
```
yubiuser, line 716: Why is this the only one set to `false`?

DL6ER: Because I thought it's not very useful to print this to the log. Given that we also print when a config item is forced through environment, we can probably just keep this as well.

yubiuser:
And now that the value defaults to `true` all the time we could also just hard-code it. Or do you want to keep the flexibility to set it to `false` at some point in the future?

DL6ER:
The key is that it isn't always `true`. Check out `send_json_unauthorized()` where it is indeed conditional.

yubiuser (approved):

PR #2017, opened 2024-07-25
Fix termination in dnsmasq failure mode

DL6ER:

When the embedded `dnsmasq` fails to start because of some incorrect configuration and reversing to the previous configuration isn't possible, FTL remains being active so users can saee the web interface and the critical `dnsmasq` error message shown in the diagnosis system. However, certain misconfigurations (e.g., `files.log.dnsmasq` invalid) can cause the process to get into some strange meta-state where `SIGTERM` is incorrectly translated into a meta-signal on which nobody is listening. As a result, the entire process starts to ignore `SIGTERM` altogether and can only forcefully be killed though `SIGKILL`.

Finding this defect was quite a challenge which is carefully documented on the linked Discourse discussion. This PR contains a rather simple fix: Only translate `SIGTERM` into a meta-state if we know `dnsmasq` is up and running and handling signals as it should. If `dnsmasq` fails to initialize at any point, FTL remains active at reacting itself to `SIGTERM` events.

#### How to test

Set `files.log.dnsmasq = "/var/log/pih/pihole.log"` (or any other invalid path) and observe that `kill $(cat /var/run/pihole-FTL.pid)` does not cause FTL to actually terminate. Repeat on this branch and see that it does work.

---

**Related issue or feature (if applicable):** https://discourse.pi-hole.net/t/suggested-systemd-service-file-to-insure-pihole-db-is-populated-with-data-if-dbinterval-is-non-default/71213

---

## Checklist:

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -201,6 +201,10 @@ void parse_args(int argc, char* argv[])
 	if(strEndsWith(argv[0], "luac"))
 		exit(run_luac(argc, argv));
 
+	// Special (undocumented) mode to test kernel signal handling
+	if(argc == 2 && strcmp(argv[1], "sigtest") == 0)
+		exit(sigtest());
+
 	// If the binary name is "sqlite3"  (e.g., symlink /usr/bin/sqlite3 -> /usr/bin/pihole-FTL),
 	// we operate in drop-in mode and consume all arguments for the embedded SQLite3 engine
 	// Also, we do this if the first argument is a file with ".db" ending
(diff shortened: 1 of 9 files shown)


yubiuser (approved):
Is there any code that should be send upstream? Or does the problem only exists when `dnsmasq` is used as embedded part of Pi-hole?

DL6ER:
The latter. Our modification to `dnsmasq`  caused this unforeseen problem. The issue is that `dnsmasq` typically fails hard when this happens. However, we figured this is not nice behavior as many users wouldn't be able to see what is happening when FTL dies (and took the webserver along!). So we defer the handling of `SIGTERM` but deferring to handling by someone who is not there isn't a clever move. This PR changes this to only defer when we know that there is actually someone else there handling the signal as it should be handled. If not, we do it ourselves.

PR #2022, opened 2024-08-01
Improve network table code

### LINKED ISSUE #2021: Bug: "Network overview" -> “First seen” later than ”Last Query"
### Versions
- Pi-hole: v5.18.3
- AdminLTE: v5.25.2
- FTL: v5.21

### Platform
- OS and version: Raspberry Pi OS Lite (64-bit), Bookworm
- Platform: Raspberry Pi 

### Expected behavior

“First seen” earlier than ”Last Query” (at least according to my understanding)

### Actual behavior / bug

“First seen” later than ”Last Query"

### Steps to reproduce

Imo not necessary

### Debug Token

...

### Screenshots

Screenshot 2024-07-30 124833 - m 2

### Additional context

N/A

DL6ER:

This PR improves the network-table code. It consists of two commits:

1. https://github.com/pi-hole/FTL/commit/1c2e44d519182fe3e3681cd46620407daede1f4d fixes #2021 by using the client's `firstSeen` instead of `now` when adding new devices to the database
2. https://github.com/pi-hole/FTL/commit/73634762ed04e5b6101de465a3baee35dad40d46 removes using `asprintf()`  from the network-table code for both speed (no heap allocations) and safety (against unlikely-but-possible SQL injections). It should come at no functional changes. It also makes the code shorter and easier to read.

Condition for the mentioned SQL injections would be a malformed `macvendor.db` which is downloaded from a trustworthy source but could be modified on user systems (need `root` for this).

---

**Related issue or feature (if applicable):** Fixes #2021 

---

## Checklist:

### DIFF
--- a/src/database/aliasclients.c
+++ b/src/database/aliasclients.c
@@ -109,6 +109,7 @@ bool import_aliasclients(sqlite3 *db)
 
 	// Loop until no further data is available
 	int imported = 0;
+	const double now = double_time();
 	while((rc = sqlite3_step(stmt)) != SQLITE_DONE)
 	{
 		// Check if we ran into an error
@@ -132,7 +133,7 @@ bool import_aliasclients(sqlite3 *db)
 		}
 
 		// Try to open existing client
-		const int clientID = findClientID(aliasclient_str, false, true);
+		const int clientID = findClientID(aliasclient_str, false, true, now);
 
 		clientsData *client = getClient(clientID, true);
 		if(client == NULL)
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -758,6 +758,98 @@ int db_query_int(sqlite3 *db, const char* querystr)
 	return result;
 }
 
+int db_query_int_int(sqlite3 *db, const char* querystr, const int arg)
+{
+	log_debug(DEBUG_DATABASE, "db_query_int_arg: \"%s\"", querystr);
+
+	sqlite3_stmt* stmt;
+	int rc = sqlite3_prepare_v2(db, querystr, -1, &stmt, NULL);
+	if( rc != SQLITE_OK )
+	{
+		if( rc != SQLITE_BUSY )
+			log_err("Encountered prepare error in db_query_int(\"%s\"): %s",
+			        querystr, sqlite3_errstr(rc));
+		return DB_FAILED;
+	}
+
+	// Bind argument to prepared statement
+	if((rc = sqlite3_bind_int(stmt, 1, arg)) != SQLITE_OK)
+	{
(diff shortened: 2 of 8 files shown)


yubiuser (dismissed):
Screenshot at 2024-08-01 13-32-26

yubiuser:
Oh, this one is interesting... Everything after a flush of the network table
Screenshot at 2024-08-01 13-36-18

DL6ER:
Thanks for the thorough testing. This revealed that there was something hidden wrong about `firstSeen` for clients imported from the database. See the last commit's message without reiterating this here.

What I noticed as well is that we have `firstSeen` and `lastQuery` in (sub-)millisecond accuracy but promote it to INTEGER timestamps for the database where we have:
https://github.com/pi-hole/FTL/blob/c9275ace5f5822c79121d8536a0587fd1b6157e7/src/database/network-table.c#L46-L47

We could do three things:
1. Nothing - which is what I chose right now
2. Ignore the column type and still write the `double` value - feels dirty but would very likely "just work"
   > SQLite uses a more general dynamic type system. In SQLite, the datatype of a value is associated with the value itself, not with its container.

   ([source](https://www.sqlite.org/datatype3.html))
4. Change the column type to `REAL` - painful because altering existing column types isn't possible, we'd need to create a new table with the new types, import the existing data, remove the old table and rename the new to the old table. Plus what is needed to handle possible foreign key links.
5. We can modify the schema without creating a new table and doing all the stuff by directly manipulating the table code of the network table via the magic `[main.sqlite_schema`](https://www.sqlite.org/schematab.html) table.

yubiuser (commented):
Timestamps look good now. However, there seems to be some conflict between "Last Query" and "Number of queries". How can the number be zero when the last query is still within the 24h time frame? Aren't we importing the number of queries on a restart - or is this data pure in-memory?
Screenshot at 2024-08-01 21-52-39

yubiuser:
> Nothing - which is what I chose right now

This seems reasonable. I don't think we need sub-second precision here.

DL6ER:
> How can the number be zero when the last query is still within the 24h time frame? Aren't we importing the number of queries on a restart - or is this data pure in-memory?

Yes, there is indeed a special mechanism in place that prevents counts from the database to be added to the total number of queries. This is useful to not add, say 1,000 queries a client has made five times to the database when you restart `pihole-FTL` five times. We could add an exception in case a client is added completely newly to the database, I'm not sure right now how I feel about it. What do you think?

yubiuser:
Looks better now. Only `127.0.0.1` is still off (last row)
Screenshot at 2024-08-03 13-49-33

yubiuser (approved):

PR #2030, opened 2024-08-12
Implement DNS caching for queries blocked upstream

DL6ER:

Implement DNS caching for queries blocked upstream (`NXDOMAIN + no RA bit, `0.0.0.0/::`, and known Umbrella IP blocking pages) accompanied with suitable CI testing. This change improves handling of externally blocked queries in two ways:

1. Once we know the domain is externally blocked, we don't forward it again for the same client (it is currently forwarded each time)
2. We can recognize cache content with the specified addresses/flags as being upstream blocked and don't return them as "OK (cached)" through the API/in the database

---

---

5. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
6. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

## Checklist:

### DIFF
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -239,7 +239,7 @@ components:
                       type: integer
                     optimizer:
                       type: integer
-                    upstreamTTL:
+                    upstreamBlockedTTL:
                       type: integer
                 revServers:
                   type: array
@@ -665,7 +665,7 @@ components:
             cache:
               size: 10000
               optimizer: 3600
-              upstreamTTL: 86400
+              upstreamBlockedTTL: 86400
             revServers:
               - "true,192.168.0.0/24,192.168.0.1,lan"
             blocking:
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -239,6 +239,8 @@ components:
                       type: integer
                     optimizer:
                       type: integer
+                    upstreamTTL:
+                      type: integer
                 revServers:
                   type: array
                   items:
@@ -663,6 +665,7 @@ components:
             cache:
               size: 10000
               optimizer: 3600
+              upstreamTTL: 86400
             revServers:
               - "true,192.168.0.0/24,192.168.0.1,lan"
             blocking:
--- a/src/enums.h
+++ b/src/enums.h
@@ -130,6 +130,9 @@ enum domain_client_status {
 	REGEX_BLOCKED,
 	ALLOWED,
 	SPECIAL_DOMAIN,
+	UPSTREAM_BLOCKED_NXRA,
+	UPSTREAM_BLOCKED_NULL,
+	UPSTREAM_BLOCKED_IP,
 	NOT_BLOCKED
 } __attribute__ ((packed));
 
--- a/.github/.codespellignore
+++ b/.github/.codespellignore
@@ -12,3 +12,4 @@ dnsmasq
 iif
 prefered
 padd
+rabit
--- a/src/api/padd.c
+++ b/src/api/padd.c
@@ -52,7 +52,7 @@ int api_padd(struct ftl_conn *api)
 		// Find most recently blocked query
 		for(int queryID = counters->queries - 1; queryID > 0 ; queryID--)
 		{
-			const queriesData* query = getQuery(queryID, true);
+			const queriesData *query = getQuery(queryID, true);
 			if(query == NULL)
 				continue;
 
--- a/src/api/stats.c
+++ b/src/api/stats.c
@@ -748,7 +748,7 @@ int api_stats_recentblocked(struct ftl_conn *api)
 	cJSON *blocked = JSON_NEW_ARRAY();
 	for(int queryID = counters->queries - 1; queryID > 0 ; queryID--)
 	{
-		const queriesData* query = getQuery(queryID, true);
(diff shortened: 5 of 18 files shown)


Anuskuss:
I set up my upstream to respond with `NXDOMAIN` + no `ra` (+ [EDE15](https://www.rfc-editor.org/rfc/rfc8914#name-extended-dns-error-code-15-)) and this works for me.
```
$ dig @127.0.0.1 -p53 +noall +answer +comments one.one.one.one 
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 56340
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; EDE: 15 (Blocked): (1.1.1.1)
;; ANSWER SECTION:
one.one.one.one. 2	IN	A	0.0.0.0
```
```
$ dig @127.0.0.1 -p53 +noall +answer +comments one.one.one.one
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 27011
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; ANSWER SECTION:
one.one.one.one. 2	IN	A	0.0.0.0
```

Unfortunately the EDE message doesn't get cached yet (notice how it disappears in the second request) but this is way better than before. Now it actually responds with my `BLOCKINGMODE` (`NULL` in my case) and shows `Blocked (external, NXRA)` both times in the log instead of `OK (cache)` the second time. Good job :+1:

DL6ER:
We have a follow-up already in the pipeline (branch `new/edns_reasoning`) but I will open the PR only after this has been merged to ease reviewing. It will offer EDE to the clients why the Pi-hole has decided to block queries. This feature can be controlled by a new setting `dns.blocking.edns`

Examples:
``` bash
$ dig doubleclick.net | grep EDE
; EDE: 29: (synthesized)

$ dig doubleclick.net | grep EDE
; EDE: 15 (Blocked): (gravity)

$ dig denied.ftl | grep EDE
; EDE: 15 (Blocked): (denylist)

$ dig some.regex.example.com | grep EDE
; EDE: 15 (Blocked): (regex)

$ dig null.ftl | grep EDE
; EDE: 15 (Blocked): (upstream NULL)
```

yubiuser (commented):
> Once we know the domain is externally blocked, we don't forward it again for the same client 

Not forwarding for how long? After a certain time we should allow forwarding again - the domain might not be blocked anymore

DL6ER:
> Not forwarding for how long?

This is a good question - something I have not thought about. So far, the cache is kept valid until something changes *locally* (gravity update, changes to the domainlists, etc.). But with this PR we may need to make DNS records actually expire... I wonder: is "it is cleared on the next gravity run" too long?

src/enums.h
```diff
@@ -130,6 +130,9 @@ enum domain_client_status {
 	REGEX_BLOCKED,
 	ALLOWED,
 	SPECIAL_DOMAIN,
+	UPSTREAM_BLOCKED_NXRA,
```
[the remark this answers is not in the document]
DL6ER: No, they are only internally used meta-states. The cache is never exported to the database.

yubiuser:
>  I wonder: is "it is cleared on the next gravity run" too long?

Personally, I think yes. Such a feature reminds me to `unbound`'s `serve-expired` with a default limit of 1 day. 
https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-serve-expired-ttl

rdwebdesign:
>  I wonder: is "it is cleared on the next gravity run" too long?

This raises another question: Will we add a new config option to change this expiration time?

I agree with yubiuser. `1 day` seems to be a good default value.

Anuskuss:
Personally I don't think that's necessary. I think that if upstream decides to outright block a domain it has do be something serious (e.g. illegal). So I think once a week is plenty.

DL6ER:
New config option added, defaulting to one day - as suggested. Note that `pihole -g` will still flush the entire cache *including* the queries blocked upstream.

DL6ER:
A commit concerning EDE sneaked into this PR which I wanted to keep separate to ease reviewing. Force-pushed away.

src/api/docs/content/specs/config.yaml
```diff
@@ -239,6 +239,8 @@ components:
                       type: integer
                     optimizer:
                       type: integer
+                    upstreamTTL:
```
yubiuser, line 242:
```suggestion
                    upstreamBlockedTTL:
```

yubiuser (commented):
Can you squeeze in `Blocked` in the config option name? Just to make it clear that this option does not 'overwrite' all TTLs form upstream responses, but just set them for blocked queries.

src/api/docs/content/specs/config.yaml
DL6ER: done in the latest commit, needed to be done in 12 locations

yubiuser (approved):

PR #2033, opened 2024-08-16
Update embedded CivetWeb server

DL6ER:

This PR updates the embedded webserver to its latest `master` HEAD.

Hereby, the following changes/fixes directly applicable to Pi-hole are added:

- https://github.com/civetweb/civetweb/pull/1208
- https://github.com/civetweb/civetweb/pull/1209
- https://github.com/civetweb/civetweb/pull/1277
- https://github.com/civetweb/civetweb/pull/1278
- https://github.com/civetweb/civetweb/pull/1279

Furthermore, this PR also already includes:
- https://github.com/civetweb/civetweb/pull/1281
- https://github.com/civetweb/civetweb/pull/1286

---

---

## Checklist:


src/webserver/civetweb/civetweb.h
```diff
@@ -23,9 +23,9 @@
 #ifndef CIVETWEB_HEADER_INCLUDED
 #define CIVETWEB_HEADER_INCLUDED
 
-#define CIVETWEB_VERSION "1.16"
+#define CIVETWEB_VERSION "1.17"
```
timtjtim, line 26: would it be better to use 1.16.1 or something in case there are further changes before 1.17 is signed off?

DL6ER: This string is directly sourced from the unmodified upstream code. It's anyway wrong as we're applying Pi-hole specific patches into CivetWeb. 

Setting this to `1.16.1` feels even more wrong to me as such a version had never existed. I could probably go with some final patch which appends something like `-pihole` to the version string defined here. Maybe even a specific commit but that's another manual change needed we may have forgotten about when the next Civetweb update is some next year or even later

rdwebdesign: Complementing the answer above, this is the upstream code (master branch):

https://github.com/civetweb/civetweb/blob/3649a5fe2e4ebed42b93dbdd4662cd1894ae95fe/include/civetweb.h#L26-L29

timtjtim: Hmm - really not a fan of civetweb doing it that way. I like the idea of the specific commit hash as it gives traceability, within the code but I'm not fussy - your call!

DL6ER: As said, all the minor things we have up manually take care about that cannot properly be automated tend to be forgotten in the future (this had happened often enough in the past) to such an extend that I'd like to avoid it as much as possible. 

We are not exposing the server as a library or in any other way offer any interface to it (I know it is *possible* but it's not supported to use the binary as library). I think it's fine to not correspond to the version 100% perfectly, it shouldn't really matter.

DL6ER: To be clear: I'll either add a static string `1.17-pihole` or something like `1.17 (modified)` to the output of `pihole -vv`

yubiuser (approved):

PR #2047, opened 2024-08-31
Fix CPU % calculation

DL6ER:

Use total number of *configured* (= existing) cores rather then the number of available (= online and assigned) cores when computing the CPU utilization of the system. This fixes the displayed CPU % in case a container got fewer cores assigned to it than are available on the host system.

---

**Related issue or feature (if applicable):** https://discourse.pi-hole.net/t/cpu-usage-is-wrong-shown-calculated-on-lxc-debian-container/71999

---

## Checklist:

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -262,8 +262,8 @@ static void check_load(void)
 	if (getloadavg(load, 3) == -1)
 		return;
 
-	// Get number of CPU cores
-	const int nprocs = get_nprocs();
+	// Get total number of CPU cores
+	const int nprocs = get_nprocs_conf();
 
 	// Warn if 15 minute average of load exceeds number of available
 	// processors
--- a/src/api/info.c
+++ b/src/api/info.c
@@ -12,7 +12,7 @@
 #include "webserver/http-common.h"
 #include "webserver/json_macros.h"
 #include "api/api.h"
-// sysinfo()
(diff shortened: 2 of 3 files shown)


src/gc.c
```diff
@@ -262,8 +262,8 @@ static void check_load(void)
 	if (getloadavg(load, 3) == -1)
 		return;
 
-	// Get number of CPU cores
-	const int nprocs = get_nprocs();
+	// Get number of totally existing CPU cores
```
yubiuser, line 265:
```suggestion
	// Get number of configured CPU cores
```

Just to use the same name as in the location

DL6ER: As you see, I'm a bit struggling with the vocabulary here

- Configured means "totally physically existing", but
- Available (sometimes called "online") means something different. It is the number of cores the VM configuration allows to be used

...

yubiuser: I see what you mean.  If the function would be called `get_nprocs_exist` it would all be much clearer.
Configured and available are to similar in meaning I think.
___

Maybe we leave "totally existing" in `gc.c` but change L162 in `info.c` to also "totally existing"?

DL6ER: Done that and force-pushed over the existing commit

yubiuser (approved):

PR #2086, opened 2024-10-16
API Documentation fixes

irsheep:

Fix some typos in the API documentation and add a new section

Fixed an incorrect description in the "until" parameter, it was using the same description as "from"
Fixed capitalization of "Pi-hole configuration", which was creating 2 sections in API documentation, "Pi-hole Configuration" and "Pi-hole configuration"
Added a section for "Local DNS Records"  with examples how to add and remove "A" and "CNAME" records.
Removed "$ref: 'common.yaml#/components/schemas/took'" from requestBody

N/A

---

---

### DIFF
--- a/src/api/docs/content/specs/localdns.yaml
+++ b/src/api/docs/content/specs/localdns.yaml
@@ -8,7 +8,6 @@ components:
         summary: Get local DNS records
         tags:
           - "Local DNS Records"
-        security: []
         operationId: "get_records"
         description: |
           Get local DNS records (A or CNAME).
--- a/src/api/docs/content/specs/localdns.yaml
+++ b/src/api/docs/content/specs/localdns.yaml
@@ -166,7 +166,7 @@ components:
       schema:
         type: string
       required: true
-      description: The IP address (IPv4 or IPv6) and domain seperated by a 'SPACE'
+      description: The IP address (IPv4 or IPv6) and domain separated by a 'SPACE'
       example: 8.8.8.8 googledns.com
     cname_value:
       in: path
--- a/src/api/docs/content/specs/localdns.yaml
+++ b/src/api/docs/content/specs/localdns.yaml
@@ -1,0 +1,199 @@
+openapi: 3.0.2
+components:
+  paths:
+    record_type:
+      parameters:
+        - $ref: 'localdns.yaml#/components/parameters/record_type'
+      get:
+        summary: Get local DNS records
+        tags:
+          - "Local DNS Records"
+        security: []
+        operationId: "get_records"
+        description: |
+          Get local DNS records (A or CNAME).
+        responses:
+          '200':
+            description: OK
+            content:
+              application/json:
+                schema:
+                  allOf:
+                    - $ref: 'localdns.yaml#/components/schemas/record_type'
+                    - $ref: 'common.yaml#/components/schemas/took'
+          '401':
+            description: Unauthorized
+            content:
+              application/json:
+                schema:
+                  allOf:
+                    - $ref: 'common.yaml#/components/errors/unauthorized'
+                    - $ref: 'common.yaml#/components/schemas/took'
+
+    host_value:
+      parameters:
+        - $ref: 'localdns.yaml#/components/parameters/host_value'
+      put:
+        summary: Add a host record
+        tags:
+          - "Local DNS Records"
+        operationId: "add_host_record"
+        description: |
+          Add a 'host' (A or AAAA) DNS record to the Pi-hole hosts file.
+        responses:
+          '201':
+            description: Created item
+            content:
+              application/json:
+                schema:
+                  allOf:
+                    - $ref: 'common.yaml#/components/schemas/took'
+          '401':
+            description: Unauthorized
+            content:
+              application/json:
+                schema:
+                  allOf:
+                    - $ref: 'common.yaml#/components/errors/unauthorized'
+                    - $ref: 'common.yaml#/components/schemas/took'
+      delete:
+        summary: Delete a host record
+        tags:
+          - "Local DNS Records"
+        operationId: "delete_host_record"
+        description: |
+          Remove a host DNS record from the hosts file.
+        responses:
+          '204':
+            description: No Content (deleted)
+          '400':
+            description: Bad request
+            content:
+              application/json:
+                schema:
+                  allOf:
+                    - $ref: 'common.yaml#/components/errors/bad_request'
+                    - $ref: 'common.yaml#/components/schemas/took'
+          '401':
+            description: Unauthorized
+            content:
+              application/json:
+                schema:
+                  allOf:
+                    - $ref: 'common.yaml#/components/errors/unauthorized'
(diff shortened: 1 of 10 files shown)


src/api/docs/content/specs/queries.yaml
```diff
@@ -114,6 +114,7 @@ components:
             description: Load queries from on-disk database rather than from in-memory
             required: false
             schema:
+              type: boolean
               enum:
                 - true
                 - false
```
rdwebdesign, line 120: Here the type is `enum` (enumerated type or enumerated set), containing 2 items: `true` and `false`.

rdwebdesign (dismissed):

src/api/docs/docs.h
```diff
@@ -136,6 +136,11 @@
 #include "hex/specs/padd.yaml"
 };
 
+static const unsigned char specs_localdns_yaml[] = {
+#include "hex/specs/localdns.yaml"
```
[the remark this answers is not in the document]
irsheep: I have looked at this, but this seems to be normal, as docs.h does contain other includes to yaml files.

src/api/docs/content/specs/localdns.yaml
```diff
@@ -0,0 +1,199 @@
...
+      required: true
+      description: The type of DNS record hosts (A) or cnameRecord (CNAME)
+      example: hosts
+    host_value:
+      in: path
+      name: record
+      schema:
+        type: string
+      required: true
+      description: The IP address (IPv4 or IPv6) and domain seperated by a 'SPACE'
```
DL6ER, line 169: There is a small typo here: `seperated ==> separated`

```suggestion
      description: The IP address (IPv4 or IPv6) and domain separated by a 'SPACE'
```

src/api/docs/content/specs/localdns.yaml
```diff
@@ -0,0 +1,199 @@
...
+components:
+  paths:
+    record_type:
+      parameters:
+        - $ref: 'localdns.yaml#/components/parameters/record_type'
+      get:
+        summary: Get local DNS records
+        tags:
+          - "Local DNS Records"
+        security: []
```
DL6ER, line 11: Why are you defining that `security` is not required?

image


This is definitely wrong. Just leave this away and the endpoint should correctly receive standard security restrictions:

```suggestion
```

DL6ER (changes requested):
Thank you for your submission. The spell check found a small glitch, see below.

To be honest, I am not sure how I feel about adding dedicated documentation for local DNS records where `/config` is already documented. Yes, I see how you are getting here (making it more obvious) but one might then come and ask why we are doing it for local DNS records and local CNAMEs but not for other array-like settings (e.g., DNS upstream servers) where the same `PUT`/`DELETE` can be used.

Also, there is some glitch when you leave the field empty, you get the **full config** instead:
Screenshot

irsheep:
> Thank you for your submission. The spell check found a small glitch, see below.
> 
> To be honest, I am not sure how I feel about adding dedicated documentation for local DNS records where `/config` is already documented. Yes, I see how you are getting here (making it more obvious) but one might then come and ask why we are doing it for local DNS records and local CNAMEs but not for other array-like settings (e.g., DNS upstream servers) where the same `PUT`/`DELETE` can be used.
> 
> Also, there is some glitch when you leave the field empty, you get the **full config** instead: Screenshot

Sorry didn't intended to close your comment without addressing this.

When I was implementing this change I thought about the same, but decided to propose it anyway for a couple of reasons
- In the previous version of Pi-hole A and CNAME records had their own API (and only realized that this is part of `/config` after it was almost done)
- The path to update or delete the DNS records requires 4 elements in the path `/config/dns/_type_/_record_` , but that is not clear in the **Pi-hole configuration** section
- to document that format when specifying the _record_. An A record uses a space "IP fqdn" and CNAME uses a comma. At first I was thinking that it was a mistake, but then realized that this is because how it written to the hosts and dnsmask files.

This one might be cheeky because just found now, the same could be said with **DNS control**

As for the glitch that is because as you mention proposed **Local DNS records** section part of the `/config` so by leaving the `type` blank in that request you're in fact requesting `/config/dns`.

DL6ER:
> As for the glitch that is because as you mention proposed **Local DNS records** section part of the `/config` so by leaving the `type` blank in that request you're in fact requesting `/config/dns`.

Yes, I know :-) My comment was more meant into the direction "is there any way to prevent this from happening?". I could have been clearer about my intention. As you have dedicated `PUT/DELETE` descriptions, you may do the same for `GET` and avoid the parameter `{type]`. Otherwise, I do seem to recall that there was once a mentioning of the `parameter` property `allowEmptyValue` as in
``` yaml
  parameters:
    record_type:
      in: path
      name: type
      schema:
        type: string
        enum:
          - hosts
          - cnameRecords
      required: true
      description: The type of DNS record hosts (A) or cnameRecord (CNAME)
      example: hosts
      allowEmptyValue: false                                 <-----------------------
```
but I haven't tested it myself.

> An A record uses a space "IP fqdn" and CNAME uses a comma.

Ah yes, right. That is in the description of `pihole.toml` and can also be queried over the API but it isn't obvious at all from the very generic/abstract OpenAPI format how the particular formatting of the individual endpoints would be.

image
image

Or, of course, you don't need to know it at all if you are using https://pi.hole/admin/settings/dnsrecords which does everything for you (if configuring via the web interface).

> document that format

Where did you add something describing this? I have only see the examples.

irsheep:
> > As for the glitch that is because as you mention proposed **Local DNS records** section part of the `/config` so by leaving the `type` blank in that request you're in fact requesting `/config/dns`.
> 
> Yes, I know :-) My comment was more meant into the direction "is there any way to prevent this from happening?". I could have been clearer about my intention. As you have dedicated `PUT/DELETE` descriptions, you may do the same for `GET` and avoid the parameter `{type]`. Otherwise, I do seem to recall that there was once a mentioning of the `parameter` property `allowEmptyValue` as in
> 
> ```yaml
>   parameters:
>     record_type:
>       in: path
>       name: type
>       schema:
>         type: string
>         enum:
>           - hosts
>           - cnameRecords
>       required: true
>       description: The type of DNS record hosts (A) or cnameRecord (CNAME)
>       example: hosts
>       allowEmptyValue: false                                 <-----------------------
> ```
> 
> but I haven't tested it myself.
> 
> > An A record uses a space "IP fqdn" and CNAME uses a comma.
> 
> Ah yes, right. That is in the description of `pihole.toml` and can also be queried over the API but it isn't obvious at all from the very generic/abstract OpenAPI format how the particular formatting of the individual endpoints would be.
> 
> image image
> 
> Or, of course, you don't need to know it at all if you are using https://pi.hole/admin/settings/dnsrecords which does everything for you (if configuring via the web interface).
> 
> > document that format
> 
> Where did you add something describing this? I have only see the examples.

I looked at the `allowEmptyValue` and it doesn't seem to work, I do agree with your suggestion to have the `GET` method added for each of the record types, it might make is less ambiguous.

It is in the description/example of the `record` parameter, don't think that makes sense adding it anywhere else.

PR #2094, opened 2024-10-22
Update embedded SQLite to 3.47.0

DL6ER:

As you can see from the commits below, I first wanted to make the SQLite3 `percentile` extension part of Pi-hole's `sqlite3`. I wanted to use it for future statistical tools tailored towards debugging possibly slow upstream responses. Anyway, after I have done this, the SQLite3 team decided themselves to make the very same module an integral part of the `sqlite3` shell (see 9.2 below). Instead of doing all kinds to `git` magic to undo my previous changes, I decided to (partially) revert and, instead, keep them here as they clearly indicate how to statically bundle official SQLite3 extensions in case we ever need this again.

Of interest may be the new (still experimental) `sqlite3_rsync` (9.1 below) feature allowing live cross-system-boundary database syncs (via `ssh`). The new `.www` output format (9.3 below) generates HTML output for each run SQL statement. This involves rendering a HTML page and calling something like `xdg-open /tmp/abcdef123456.htm`. This will only work if you run this locally (not over SSH).

## SQLite3 CHANGELOG 2024-10-21 (3.47.0)

  1. Allow arbitrary expressions in the second argument to the [`RAISE` function](https://sqlite.org/lang_createtrigger.html#raise). 
  2. If the RHS of [the `->>` operator](https://sqlite.org/json1.html#jptr) is negative, then access array elements counting from the right. 
  3. Fix a problem with rolling back hot journal files in the seldom-used [`unix-dotfile` VFS](https://sqlite.org/vfs.html#standard_unix_vfses). 
  4. FTS5 tables can now be dropped even if they use a non-standard tokenizer that has not been registered. 
  5. Fix the [`group_concat()` aggregate function](https://sqlite.org/lang_aggfunc.html#group_concat) so that it returns an empty string, not a `NULL`, if it receives a single input value which is an empty string. 
  6. Enhance the [`generate_series()`](https://sqlite.org/series.html) table-valued function so that it is able to recognize and use constraints on its output value. 
  7. Preupdate hooks now recognize when a column added by `ALTER TABLE ADD COLUMN` has a non-null default value. 
  8. Performance optimizations: 
    1. Improved reuse of subqueries associated with the [`IN` operator](https://sqlite.org/lang_expr.html#in_op), especially when the IN operator has been duplicated due to [predicate push-down](https://sqlite.org/optoverview.html#pushdown). 
    2. Use a Bloom filter on subqueries on the right-hand side of the [`IN` operator](https://sqlite.org/lang_expr.html#in_op), in cases where that seems likely to improve performance. 
    3. Ensure that queries like "`SELECT func(a) FROM tab GROUP BY 1`" only invoke the `func()` function once per row. 
    4. No attempt is made to create [automatic indexes](https://sqlite.org/optoverview.html#autoindex) on a column that is known to be non-selective because of its use in other indexes that have been analyzed. 
    5. Adjustments to the query planner so that it produces better plans for star queries with a large number of dimension tables. 
    6. Add the "order-by-subquery" optimization, that seeks to disable sort operations in outer queries if the desired order is obtained naturally due to ORDER BY clauses in subqueries. 
    7. The "indexed-subtype-expr" optimization strives to use expressions that are part of an index rather than recomputing the expression based on table values, as long as the query planner can prove that the [subtype](https://sqlite.org/c3ref/value_subtype.html) of the expression will never be used. 
    8. Miscellaneous coding tweaks for faster runtimes. 
  9. Enhancements to SQLite-related command-line programs: 
    1. Add the experimental [`sqlite3_rsync`](https://sqlite.org/rsync.html) program. 
    2. Add extension functions [`median()`](https://sqlite.org/percentile.html#*medianfunc), [`percentile()`](https://sqlite.org/percentile.html#*percentilefunc), [`percentile_cont()`](https://sqlite.org/percentile.html#*percentilecontfunc), and [`percentile_disc()`](https://sqlite.org/percentile.html#*percentilediscfunc) to the [CLI](https://sqlite.org/cli.html). 
    3. Add the [`.www` dot-command](https://sqlite.org/cli.html#dotwww) to the [CLI](https://sqlite.org/cli.html). 
    4. The [`sqlite3_analyzer`](https://sqlite.org/sqlanalyze.html) utility now provides a break-out of statistics for WITHOUT ROWID tables. 
    5. The [`sqldiff`](https://sqlite.org/sqldiff.html) utility avoids creating an empty database if its second argument does not exist. 
  10. Enhance the [`sqlite_dbpage`](https://sqlite.org/dbpage.html) table-valued function such that [`INSERT`](https://sqlite.org/lang_insert.html) can be used to increase or decrease the size of the database file. 
  11. SQLite no longer makes any use of the "`long double`" data type, as hardware support for long double is becoming less common and long double creates challenges for some compiler tool chains. Instead, SQLite uses [Dekker's algorithm](https://csclub.uwaterloo.ca/~pbarfuss/dekker1971.pdf) when extended precision is needed. 
  12. The [TCL Interface](https://sqlite.org/tclsqlite.html) for SQLite supports TCL9. Everything probably still works for TCL 8.5 and later, though this is not guaranteed. Users are encouraged to upgrade to TCL9. 
  13. JavaScript/WASM: 
    1. Fix a corruption-causing bug in the JavaScript "`opfs`" VFS. 
    2. Correct "`mode=ro`" handling for the "`opfs`" VFS. 
    3. Work around a couple of browser-specific OPFS quirks. 
  14. FTS5 Changes: 
    1. Add the `fts5_tokenizer_v2` API and the `locale=1` option, for creating custom locale-aware tokenizers and fts5 tables that may take advantage of them. 
    2. Add the `contentless_unindexed=1` option, for creating contentless fts5 tables that store the values of any UNINDEXED columns persistently in the database. 
    3. Allow an FTS5 table to be dropped even if it uses a custom tokenizer whose implementation is not available. 

---

---

## Checklist:


DL6ER:
Now returning the expected results, e.g.
``` sql
> SELECT isIPv6('::1');
1

> SELECT isIPv6('127.0.0.1');
0
```

src/database/shell.c
```diff
@@ -15076,119 +15918,1209 @@ void sqlite3_intck_close(sqlite3_intck *p){
...
+static int vfstraceCurrentTimeInt64(sqlite3_vfs *pVfs, sqlite3_int64 *pTimeOut){
+  vfstrace_info *pInfo = (vfstrace_info*)pVfs->pAppData;
+  sqlite3_vfs *pRoot = pInfo->pRootVfs;
+  return pRoot->xCurrentTimeInt64(pRoot, pTimeOut);
 }
 
 /*
-** Close any read transaction the integrity-check object is holding open
-** on the database.
+** Return th3 most recent error code and message
```
rdwebdesign, line 17007: Typo:

```suggestion
** Return the most recent error code and message
```

src/database/shell.c
```diff
@@ -1644,37 +973,212 @@ static void SQLITE_CDECL iotracePrintf(const char *zFormat, ...){
...
+      n++;
+      i++;
+    }else{
+      i++;
+    }
+  }
+  return n;
+}
+#endif
+
```
rdwebdesign, line 1141: Do we need this?

```suggestion
```

rdwebdesign (changes requested):

DL6ER:
@rdwebdesign This file is copied 1:1 from the SQLite3 project - we don't change it ourselves as this would either cause conflicts with future updates or the next update would simply overwrite it. These changes would be to happen here: https://sqlite.org/src/file/src/shell.c.in

rdwebdesign:
I didn't pay attention... I thought this file was one of our files. Ignore my comments.

yubiuser (approved):

PR #2128, opened 2024-12-05
Add DHCPv6 probing to dhcp-discover tool

DL6ER:

This PR extends the embedded `dhcp-discover` to also scan for IPv6 routers possibly broadcasting DNS configuration.

---

---

## Checklist:


DL6ER:
1. You are not using the latest version as negative displays are not possible with the latest commit (87a2ad52).
2. Not possible without much overhead. All the scanning threads are running parallel and don't sync, but I will try something.
3. Hmm, interesting, this probably has to do with the interface error "network unreachable". For me, probing `docker` works.

yubiuser:
> You are not using the latest version as negative displays are not possible with the latest commit

Interesting, because it was a fresh checkout. Maybe one of the builders failed at some point?
___

Looks good now. Code review will follow.

```
Error while sending Router Solicitation on docker0: Network unreachable
No answer on docker0
Received 1 DHCP (IPv4) and 0 RA (IPv6) answers on end0
```

src/tools/dhcpv6-discover.c
```diff
@@ -0,0 +1,823 @@
...
+/**
+ * @brief Resolves an IPv6 address by hostname and interface name.
+ *
+ * This function takes a hostname and an interface name, resolves the hostname
+ * to an IPv6 address, and fills the provided sockaddr_in6 structure with the
+ * resolved address.
+ *
+ * @param name The hostname to resolve.
+ * @param ifname The name of the network interface.
+ * @param numeric A flag indicating whether the hostname is a numeric IPv6 address.
```
yubiuser, line 43: This param does not exist in the function call. Only three arguments are taken.

DL6ER: Ah, right. This parameter *did exist* but I removed it later because I made this the default as it is often not useful to try to reverse-resolve an IPv6 address in internal networks and it was causing some long delays because (at least my ISP's) DNS servers time can run into timeouts when trying to resolve GUA IPv6 addresses.

DL6ER: Let's make this a suggestions here we can later collect in case you have further suggestions (avoiding tiny commits):
```suggestion
```

src/tools/dhcpv6-discover.c
```diff
@@ -0,0 +1,822 @@
...
+/**
+ * @brief Sends a Router Solicitation message and waits for a Router Advertisement response.
+ *
+ * This function performs Neighbor Discovery (ND) by sending a Router Solicitation (RS)
+ * message to the specified target and waits for a Router Advertisement (RA) response.
+ *
+ * @param fd The file descriptor of the socket to use for sending and receiving messages.
+ * @param ifname The name of the network interface to use for sending the message.
+ *
+ * @return 0 on success, -1 on error, -2 if no response is received after retries.
```
yubiuser, line 729: I don't see where it returns `-2`

DL6ER: I should probably write documentation only when the code is finished...

```suggestion
 * @return 0 on success, -1 on error.
```

yubiuser (commented):
Can you please rebase/merge development - I get SQL database errors due to changed database column number.

____

There is a small indention error after an error

```

   broadcast: 10.0.1.255
   ntp-server: 10.0.1.1
   Error while sending Router Solicitation on docker0: Network unreachable
Port Control Protocol (PCP) server: 10.0.1.1
```

DL6ER:
Both should be addressed now.

yubiuser (approved):

PR #2130, opened 2024-12-06
Delay IPv6 client name lookups

### LINKED ISSUE #2124: IPv6 hosts are not resolved to hostnames in Query Log
I can see several entries like this in Query Log:

Image

The IPv6 host fd2b:43d7:e9a:b2:c022:bd26:228f:3e04 is known in network table:

Image

So I would expect that the hostname is shown in query log, not the IPv6 address. I can see some more addresses that should be known in network table, but are not resolved. 99% are resolved, but 1% not.

Docker Tag nightly
Core vDev (development, 03932e8c)
FTL vDev (development, 84780798)
Web interface vDev (development, 96f9ee6a)

DL6ER:

This PR is an attempt to fix #2124 

The described issue is that some IPv6 clients don't show a host name in the Query Log even when the network table has one.

The reason for this is timing. When a client issues a query over IPv6 quickly after it got this address, the first query arrived
the Pi-hole *before* the client is known to the network table. When FTL now tries to resolve this address's name, it cannot find a name. A few moments later, when the ARP cache is parsed for the next time, FTL may be able to correlate this new address through an identical MAC address to an existing address having a host name. However, as the default value of `resolver.refreshNames` is `IPV4_ONLY`, FTL actually never tries to resolve the address so it never gets to know the IPv6 client's host name.

The attempt we are doing here is *postponing* name resolution of IPv6 clients by two times the ARP processing interval - this should give plenty time for the network table to catch up so that - when the client name is tried to be resolved for the first (and only!) time - we can get it right away.

By default, ARP cache interpretation is run at the `database.DBinterval` interval.

---

**Related issue or feature (if applicable):** #2124 

---

## Checklist:

### DIFF
--- a/src/FTL.h
+++ b/src/FTL.h
@@ -138,6 +138,12 @@
 // Default: 30 (seconds)
 #define QPS_AVGLEN 30
 
+// How long should IPv6 client host name resolution be postponed?
+// This is done to ensure that the network table had time to catch up on new
+// clients in the network
+// Default: 2 x database.DBinterval (seconds) = 120 s
+#define DELAY_V6_RESOLUTION 2*config.database.DBinterval.v.ui
+
 // Use out own syscalls handling functions that will detect possible errors
 // and report accordingly in the log. This will make debugging FTL crash
 // caused by insufficient memory or by code bugs (not properly dealing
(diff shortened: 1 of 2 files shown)


DL6ER:
TBF, I think this is ready to go in. Even without a confirmation that this fixes the underlying issue, I'm pretty sure it is the right thing to try to solve the issue at hand

rdwebdesign (approved):

yubiuser (approved):

PR #2136, opened 2024-12-19
Update embedded dnsmasq to latest tag v2.91test2

DL6ER:

See title.

CHANGELOG:

- Fix spurious "resource limit exceeded messages". Thanks to Dominik Derigs for the bug report.
- Fix out-of-bounds heap read in order_qsort(). We only need to order two server records on the ->serial field. Literal address records are smaller and don't have this field and don't need to be ordered on it. To actually provoke this bug seems to need the same server-literal to be repeated twice, eg --address=/a/1.1.1.1 --address-/a/1.1.1.1 which is clearly rare in the wild, but if it did exist it could provoke a SIGSEV. Thanks to Daniel Rhea for fuzzing this one.
- Fix buffer overflow when configured lease-change script name is too long. Thanks to Daniel Rhea for finding this one.
- Improve behaviour in the face of non-responsive upstream TCP DNS servers. Without shorter timeouts, clients are blocked for too long and fail wuth their own timeouts.
- Set --fast-dns-retries by default when doing DNSSEC. A single downstream query can trigger many upstream queries. On an unreliable network, there may not be enough downstream retries to ensure that all these queries complete.
- Improve behaviour in the face of truncated answers to queries for DNSSEC records. Getting these answers by TCP doesn't now involve a faked truncated answer to the downstream client to force it to move to TCP. This improves performance and robustness in the face of broken clients which can't fall back to TCP.
- No longer remove data from truncated upstream answers. If an upstream replies with a truncated answer, but the answer has some RRs included, return those RRs, rather than returning and empty answer.
-Fix handling of EDNS0 UDP packet sizes. When talking upstream we always add a pseudoheader, and set the UDP packet size to --edns-packet-max. Answering queries from downstream, we get the answer (either from upstream or local data) If local data won't fit the advertised size (or 512 if there's not an EDNS0 header) return truncated. If upstream returns truncated, do likewise. If upstream is OK, but the answer is too big for downstream, truncate the answer.
- Modify the behaviour of --synth-domain for IPv6. When deriving a domain name from an IPv6 address, an address such as 1234:: would become 1234--.example.com, which is not legal in IDNA2008. Stop using the :: compression method, so 1234:: becomes 1234-0000-0000-0000-0000-0000-0000-0000.example.com
- Fix broken dhcp-relay on *BSD. Thanks to Harold for finding this problem.
- Add --dhcp-option-pxe config. This acts almost exactly like --dhcp-option except that the defined option is only sent when replying to PXE clients. More importantly, these options are sent in reply PXE clients when dnsmasq in acting in PXE proxy mode. In PXE proxy mode, the set of options sent is defined by the PXE standard and the normal set of options is not sent. This config allows arbitrary options in PXE-proxy replies. A typical use-case is to send option 175 to iPXE. Thanks to Jason Berry for finding the requirement for this.
- Support PXE proxy-DHCP and DHCP-relay at the same time. When using PXE proxy-DHCP, dnsmasq supplies PXE information to the client, which also talks to another "normal" DHCP server for address allocation and similar. The normal DHCP server may be on the local network, but it may also be remote, and accessed via a DHCP relay. This change allows dnsmasq to act as both a PXE proxy-DHCP server AND a DHCP relay for the same network.
- Fix erroneous "DNSSEC validated" state with non-DNSSEC upstream servers.  Thanks to Dominik Derigs for the bug report.

---

## Checklist:


DL6ER:
Opening this as draft for now as there is an issue with https://github.com/pi-hole/FTL/pull/2136/commits/7572e592bebc558cff0e8d01f022b5f9229d9926 which is currently under investigation in private communication.

src/config/dnsmasq_config.c
```diff
@@ -345,7 +345,7 @@ bool __attribute__((const)) write_dnsmasq_config(struct config *conf, bool test_
 	{
 		fputs("# Enable query logging\n", pihole_conf);
 		if(conf->misc.extraLogging.v.b)
-			fputs("log-queries=extra\n", pihole_conf);
+			fputs("log-queries=proto\n", pihole_conf);
```
yubiuser, line 348: What does this change?

DL6ER: It adds the protocol used to perform a query. This is a new feature as `dnsmasq` is now able to perform cross-protocol queries when needed, e.g., UDP queries that were truncated due to the packets size. This is particularly an issue with DNSSEC being enabled and extra key information is attached. So far, `dnsmasq` refused to do such queries and sent `TRUNCATED` downstream to the client. Now, `dnsmasq` is able to do a query itself using TCP (no truncation can happen because of allowed packet fragmentation) and later sent the result down to the client again using UDP.

Example log output:
```
                                   |||
                                   vvv
Dec  21 16:31:31 dnsmasq[2281585]: UDP 436 127.0.0.1/36363 query[PTR] 1.2.168.192.in-addr.arpa from 127.0.0.1
```

yubiuser (approved):

PR #2145, opened 2024-12-30
Fix file ownership after pihole-FTL --config as root

catch-error:

When calling _pihole-FTL --config key value_ as root, the resulting pihole.toml and dnsmasq.conf files in /etc/pihole are owned by root:root. It should be pihole:pihole.

**pihole.toml**: Root cause is that when openFTLtoml() is called with mode "w", it'll not open _pihole.toml_, but _pihole.toml.tmp_ instead. closeFTLtoml(), however, always calls chown_pihole() on _pihole.toml_. As in the next step _pihole.toml.tmp_ gets renamed to _pihole.toml_, its root:root ownership persists.

This patch evaluates the open mode in closeFTLtoml(), changing the ownership of _pihole.toml.tmp_ if the mode isn't O_RDONLY.

**dnsmasq.conf**: Add a call to chown_pihole() for the temporary file, if called as root.

N/A

---

---

### DIFF
--- a/src/config/toml_helper.c
+++ b/src/config/toml_helper.c
@@ -89,16 +89,22 @@ FILE * __attribute((malloc)) __attribute((nonnull(1))) openFTLtoml(const char *m
 void closeFTLtoml(FILE *fp)
 {
 	// Release file lock
-	if(flock(fileno(fp), LOCK_UN) != 0)
+	int fn = fileno(fp);
+	if(flock(fn, LOCK_UN) != 0)
 		log_err("Cannot release lock on FTL's config file: %s", strerror(errno));
+
+	// Get access mode
+	int mode = fcntl(fn, F_GETFL);
+	if (mode == -1)
+		log_err("Cannot get access mode for FTL's config file: %s", strerror(errno));
 
 	// Close file
 	if(fclose(fp) != 0)
 		log_err("Cannot close FTL's config file: %s", strerror(errno));
 
 	// Chown file if we are root
-	if(geteuid() == 0)
-		chown_pihole(GLOBALTOMLPATH, NULL);
+	if(geteuid() == 0 && mode != -1)
+		chown_pihole(((mode & O_ACCMODE) == O_RDONLY) ? GLOBALTOMLPATH : GLOBALTOMLPATH".tmp", NULL);
 
 	return;
 }
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -749,6 +749,10 @@ bool __attribute__((const)) write_dnsmasq_config(struct config *conf, bool test_
 		return false;
 	}
 
+	// Chown file if we are root
+	if(geteuid() == 0)
+		chown_pihole(DNSMASQ_TEMP_CONF, NULL);
+
 	log_debug(DEBUG_CONFIG, "Testing "DNSMASQ_TEMP_CONF);
 	if(test_config && !test_dnsmasq_config(errbuf))
 	{


src/config/toml_helper.c
```diff
@@ -89,16 +89,22 @@ FILE * __attribute((malloc)) __attribute((nonnull(1))) openFTLtoml(const char *m
 void closeFTLtoml(FILE *fp)
 {
 	// Release file lock
-	if(flock(fileno(fp), LOCK_UN) != 0)
+	int fn = fileno(fp);
```
DL6ER, line 92:
```suggestion
	const int fn = fileno(fp);
```

src/config/toml_helper.c
```diff
@@ -89,16 +89,22 @@ FILE * __attribute((malloc)) __attribute((nonnull(1))) openFTLtoml(const char *m
...
+	if(flock(fn, LOCK_UN) != 0)
 		log_err("Cannot release lock on FTL's config file: %s", strerror(errno));
 
+	// Get access mode
+	int mode = fcntl(fn, F_GETFL);
```
DL6ER, line 97:
```suggestion
	const int mode = fcntl(fn, F_GETFL);
```

src/config/toml_helper.c
```diff
@@ -89,16 +89,22 @@ FILE * __attribute((malloc)) __attribute((nonnull(1))) openFTLtoml(const char *m
...
+
 	// Close file
 	if(fclose(fp) != 0)
 		log_err("Cannot close FTL's config file: %s", strerror(errno));
 
 	// Chown file if we are root
-	if(geteuid() == 0)
-		chown_pihole(GLOBALTOMLPATH, NULL);
+	if(geteuid() == 0 && mode != -1)
+		chown_pihole(((mode & O_ACCMODE) == O_RDONLY) ? GLOBALTOMLPATH : GLOBALTOMLPATH".tmp", NULL);
```
DL6ER, line 107:
```suggestion
	if(geteuid() == 0 && mode != -1)
	{
		// If we are in read-only mode, we are closing the global TOML file. If, however,
		// we are in write mode, we have actually been writing to the temporary file
		// that will subsequently be moved into place. In either case, ensure that the
		// permissions of the file we have touched here are correct.
		const bool read_only = (mode & O_ACCMODE) == O_RDONLY;
		chown_pihole(read_only ? GLOBALTOMLPATH : GLOBALTOMLPATH".tmp", NULL);
	}
```

catch-error:
Thx for your suggestions @DL6ER. They're done now.

DL6ER:
We're currently seeing issues with the CI, seems ARM runner are currently unavailable. I'll approve and merge once this is resolved

DL6ER:
Actually,

Screenshot_2024-12-30-17-06-17-55_3aea4af51f236e4932235fdada7d1643

catch-error:
Strange... I've now run builds on Raspi OS 12 (aarch64), Debian 12 (x86_64), Ubuntu 24.04 (aarch64 & x86_64) and Fedora 40 (aarch64 & x86_64). All builds go through w/o issues.

toml_helper.c includes sys/file.h, which in turn includes fcntl.h. I can add an explicit include to fcntl.h as well, looks like some environments need this.

DL6ER:
The CI builds against libmusl on Alpine Linux - from what I can see, all your test systems use GNU libc. Alpine Linux ships with the rather new gcc v14.2.0.

catch-error:
Ok, I see. I need to enhance my build environments. It now also compiles on Alpine Linux.

DL6ER (approved):

PR #2166, opened 2025-01-24
Update dnsmasq to v2.91test9 

DL6ER:

Change compared to last tagged version:

### mplement "DNS-0x20 encoding"

This provides extra protection against reply-spoof attacks.

Since DNS queries are case-insensitive, it's possible to randomly flip the case of letters in a query and still get the correct answer back. This adds an extra dimension for a cache-poisoning attacker to guess when sending replies in-the-blind since it's expected that the legitimate answer will have the same pattern of upper and lower case as the query, so any replies which don't can be ignored as malicious.

The amount of extra entropy clearly depends on the number
of a-z and A-Z characters in the query, and this implementation puts a hard limit of 32 bits to make rescource allocation easy. This about doubles entropy over the standard random ID and random port combination.

---

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,6 @@ set(CMAKE_C_STANDARD 17)
 
 project(PIHOLE_FTL C)
 
-set(DNSMASQ_VERSION pi-hole-v2.91test8)
+set(DNSMASQ_VERSION pi-hole-v2.91test9)
 
 add_subdirectory(src)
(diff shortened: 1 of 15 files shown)


yubiuser (commented):
Should we have a test for MiXed cASe queries?

DL6ER:
Again ready for review after adding tests. Initially, I planed to create a new building container tag including `py3-dnspython` but creating new build containers is currently failing on GHA for `armv8` for all recent stable Alpine releases. It is unclear how long this issue persists.

yubiuser (approved):

PR #2182, opened 2025-02-08
Update dnsmasq to 2.91rc3

DL6ER:

This PR contains only one commit to resolve a bug reported through our Discourse forum:

A retry to upstream DNS servers triggered by the following conditions

1) A query asking for the same data as a previous query which has not yet been answered.
2) The second query arrives more than two seconds after the first.
3) Either the source of the second query or the id field differs from the first.

fails to set the case of the retry to the same pattern as the first attempt.

However `dnsmasq` expects the reply from upstream to have the case pattern of the first attempt.

If the answer to the retry arrives before the answer to the first query, dnsmasq will notice the case mismatch, log an error, and ignore the answer.

The worst case scenario would be the first upstream query or reply is lost and there would follow a short period where all queries for that particular domain would fail.

This is a `development` issue, it doesn't apply to previous stable releases.

---

**Related issue or feature (if applicable):** see comment below

---

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,6 @@ set(CMAKE_C_STANDARD 17)
 
 project(PIHOLE_FTL C)
 
-set(DNSMASQ_VERSION pi-hole-v2.91rc2)
+set(DNSMASQ_VERSION pi-hole-v2.91rc3)
 
 add_subdirectory(src)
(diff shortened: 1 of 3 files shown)


DL6ER:
The last commit fixes the missing newline before `Architecture:`

Screenshot from 2025-02-08 10-27-18

like

Screenshot from 2025-02-08 10-27-36

yubiuser (approved):

PR #2258, opened 2025-02-22
Rate-limit internal NTP server

DL6ER:

Do not allow more than 10 NTP requests per second to be served.

---

---

## Checklist:

### DIFF
--- a/src/ntp/server.c
+++ b/src/ntp/server.c
@@ -42,6 +42,8 @@
 #include "enums.h"
 // threads
 #include "signals.h"
+// sleepms()
+#include "timers.h"
 
 uint64_t ntp_last_sync = 0u;
 uint32_t ntp_root_delay = 0u;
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
What happens when FTL slepps and new request arrive? We simply ignore them? Are they queued?

DL6ER:
They are queued in the kernel's receive buffer until it is full. When the buffer is full, any more packets are dropped. Quick testing showed at least 2000 NTP requests would fit into the buffer of my locally running `pihole-FTL` on `x86_64` (the default buffer may differ from system to system). Due to the problem of easy sender spoofing, a per-client rate-limiting wouldn't work either.

I checked the behavior with my Fritzbox router and found the same: When you flood it with requests, it basically becomes unusable and recovers only after a given time. I'd propose we implement the same for our NTP server by artificially lowering the receive buffer to 1 KB, effectively allowing at most 10 - 20 packets (depending on their actual size) to be buffered, i.e., the buffer would be cleared latest every full second after such an attack ended.

yubiuser (approved):

Issue #2274, opened 2025-02-24
Config Validation?


PromoFaux:
Was just testing the `new/print_invalid_env` branch after testing another PR - and figured I would use `webserver.paths.webhome` to see if it would complain... it did not, but obviously it is broken without the secondary slash.

Should FTL check first that it begins and ends with a slash?

```
    # Sub-directory of the root containing the web interface
    #
    # Possible values are:
    #     <valid subpath>, both slashes are needed!
    webhome = "/pihole" ### CHANGED (env), default = "/admin/"
```

Feel free to close this off if massively irrelevant!

PR #2276, opened 2025-02-24
Add webhome validator

DL6ER:

Add specific validator for `webserver.paths.webhome` ensuring there are slashes both at the beginning and the end of the string

---

**Related issue or feature (if applicable):** Fixes https://github.com/pi-hole/FTL/issues/2274

---

## Checklist:

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -1061,7 +1061,7 @@ static void initConfig(struct config *conf)
 	conf->webserver.paths.webhome.t = CONF_STRING;
 	conf->webserver.paths.webhome.f = FLAG_RESTART_FTL;
 	conf->webserver.paths.webhome.d.s = (char*)"/admin/";
-	conf->webserver.paths.webhome.c = validate_filepath;
+	conf->webserver.paths.webhome.c = validate_filepath_two_slash;
 
 	// sub-struct interface
 	conf->webserver.interface.boxed.k = "webserver.interface.boxed";
--- a/src/config/validator.c
+++ b/src/config/validator.c
@@ -270,6 +270,21 @@ bool validate_filepath(union conf_value *val, const char *key, char err[VALIDATO
 	return true;
 }
 
+// Validate a file path that needs to have both a slash at the beginning and at
+// the end
+bool validate_filepath_two_slash(union conf_value *val, const char *key, char err[VALIDATOR_ERRBUF_LEN])
+{
+	// Check if the path starts and ends with a slash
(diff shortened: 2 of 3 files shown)


PromoFaux:
Computer says no:

image

rdwebdesign:
The check fails and the value is used.

I tried with:
```
  FTLCONF_webserver_paths_webroot: '/home/root/www'
  FTLCONF_webserver_paths_webhome: '/adm'   # (missing the last slash)
```
and the web interface is redirected to:
`http://192.168.0.208/admlogin`

rdwebdesign:
The validator is not used when the env vars are used.

If you try to add an invalid `webhome` via web interface or using `pihole-FTL --config` it will correctly fail:
image

```
2830143614c3:/# pihole-FTL --config webserver.paths.webhome '/admin'
Invalid value: webserver.paths.webhome: file path does not start and end with a slash ("/admin")
```

DL6ER:
Yes, but that is because the validators cannot be used when using either the config file directly (or via env vars) because there is no way to communicate the warning/error to the user. What should we do in this case? Revert the config and overwrite what users have done? Ignore then env variable and use the default? It was unclear to me but rethinking this yesterday seems to make this a clear strategy.

Note: This is a problem of general scope and not limited to the `webhome` setting. The validators are never used for config file and env var values so far. If we want to change this, we should do it in a separate PR for cleanness. (#2294)

I think enforcing the validators on env vars (and falling back to ignoring the env var) is still somewhat doable as this will be in the container logs, I do find this more of an issue for direct config file changes. Your feelings may change, hence, I would like to talk about this.

rdwebdesign:
> The validators are never used for config file and env var values so far.

Yeah... I noticed that when I was checking the code.

I agree with your PR suggestion, to validate env vars and send error messages to the log.

Note:

Not sure how to do it when the config file is directly edited.

We can enforce the validation even in this case. If invalid, we set the value to default and send the errors to `FTL.log` (not sure if this is very user friendly). Or we simply don't validate in this case and leave the "user input" there. At least, the wrong value will be on the debug log to help us understand the issue.

yubiuser (approved):
For the env we have a PR, for the direct file edit we have a warning in the toml already. I think this is fine. IF you really want to add the validator to direct file edit my suggestion would be to keep the old value and log the error to `FTL.log`

```

    # Sub-directory of the root containing the web interface
    #
    # Possible values are:
    #     <valid subpath>, both slashes are needed!
    webhome = "/admin/"
```

PR #2278, opened 2025-02-25
fix: make `get_domains` parameters optional

tien:
I'm trying to generate a client from this OpenAPI spec. And it appears that the parameter type for `get_domains` is incorrect.

### DIFF
--- a/src/api/docs/content/specs/domains.yaml
+++ b/src/api/docs/content/specs/domains.yaml
@@ -1,3 +1,50 @@
+.get_path: &get_path
+  summary: Get domain
+  tags:
+    - "Domain management"
+  description: |
+    `{type}`, `{kind}`, and `{domain}` are optional. Specifying any of these may result in only a subset of the available data being returned.
+
+    Valid combinations are:
+    - `/api/domains` (all domains)
+    - `/api/domains/abc.com` (all domains identical to `abc.com`)
+    - `/api/domains/allow` (only allowed domains)
+    - `/api/domains/allow/abc.com` (only allowed domains identical to `abc.com`)
+    - `/api/domains/allow/exact` (only exactly allowed domains)
+    - `/api/domains/allow/exact/abc.com` (only exactly allowed domain identical to `abc.com`)
+    - `/api/domains/allow/regex` (only allowed regex domains)
+    - `/api/domains/allow/regex/abc.com` (only allowed regex domains identical to `abc.com`)
+    - `/api/domains/deny` (only denied domains)
+    - `/api/domains/deny/abc.com` (only denied domains identical to `abc.com`)
+    - `/api/domains/deny/exact` (only exactly denied domains)
+    - `/api/domains/deny/exact/abc.com` (only exactly denied domain identical to `abc.com`)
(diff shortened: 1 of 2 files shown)


yubiuser:
Please base your PR on and target to `development` branch.

src/api/docs/content/specs/domains.yaml
```diff
@@ -34,6 +34,16 @@ components:
...
+        parameters:
+          - allOf:
+            - $ref: '#/components/parameters/type'
+            - required: false
+          - allOf:
+            - $ref: '#/components/parameters/kind'
+            - required: false
+          - allOf:
+            - $ref: '#/components/parameters/domain'
+            - required: false
```
DL6ER, line 46:
```suggestion
        parameters:
          - allOf:
            - $ref: 'domains.yaml#/components/parameters/type'
            - required: false
          - allOf:
            - $ref: 'domains.yaml#/components/parameters/kind'
            - required: false
          - allOf:
            - $ref: 'domains.yaml#/components/parameters/domain'
            - required: false
```

Unfortunately, the tool we use for parsing and validating the specs against the real API in real tests on the CI does not work all that well on relative parts. My suggested changes above should make the CI tests pass.

DL6ER (dismissed):

tien:
Looks like the validator is just buggy and won't accept the `allOf` syntax :v

So I've replaced it with another one, also with added bonus of this one working with relative paths.

DL6ER:
This is too much of a change, looking at various documentation, e.g. [Swagger](https://swagger.io/docs/specification/v3_0/describing-parameters/#path-parameters) it seems what you want to add is not valid/possible *plus* invalid syntax.

Instead of
``` yaml
        parameters:
          - allOf:
            - $ref: 'domains.yaml#/components/parameters/type'
            - required: false
          - allOf:
            - $ref: 'domains.yaml#/components/parameters/kind'
            - required: false
          - allOf:
            - $ref: 'domains.yaml#/components/parameters/domain'
            - required: false
```
it should rather be
``` yaml
        parameters:
            - $ref: 'domains.yaml#/components/parameters/type'
              required: false
            - $ref: 'domains.yaml#/components/parameters/kind'
              required: false
            - $ref: 'domains.yaml#/components/parameters/domain'
              required: false
```
but even this is invalid because of
grafik
(screenshot from my link above) and would need to be added further up in this endpoint definition (there is already a `parameters` object).

The validator change you suggest does not solve the issue, it just doesn't check for this requirement of the OpenAPI specs and, hence, is incomplete.

----

What we'd need instead is to define new endpoints and use referencing to keep copy-pasting minimal:
``` diff
+GET /domains
+GET /domains/{type}
+GET /domains/{type}/{kind}
 GET /domains/{type}/{kind}/{domain}
```
All other (POST, DELETE, PUT) require the parameters. And this then also for
```diff
+GET /groups
 GET /groups/{name}
+GET /clients
 GET /clients/{client}
+GET /lists
 GET /lists/{list}
```

tien:
@DL6ER I've undo the unnecessary changes & split the endpoints into multiple ones as requested.

DL6ER (approved):

PR #2299, opened 2025-02-27
Enable link-time optimizations

DL6ER:

Link Time Optimization (LTO) gives GCC the capability of dumping its internal representation (GIMPLE) to disk, so that all the different compilation units that make up a single executable can be optimized as a single module.

It expands the scope of inter-procedural optimizations to encompass the whole program as optimizations are not performed to individual files but the entire byte-code as a whole massive unit. It can much improve the interplay between the functions defined in the individual objects (in out case: individual source code files).

Potential benefit is a possibly significant performance boost (but commonly 10%) at least in computation-intense parts (that is mostly crypto for us, so DNSSEC, TLS and password hashing). There have also been reports about reduction of RAM usage in the final applications because of better balancing of the stack. [This](https://github.com/CppCon/CppCon2017/blob/master/Presentations/ThinLTO%20-%20Building%20C%2B%2B%20Applications%20with%20Scalable%20Whole%20Program%20Optimization/ThinLTO%20-%20Building%20C%2B%2B%20Applications%20with%20Scalable%20Whole%20Program%20Optimization%20-%20Teresa%20Johnson%20-%20CppCon%202017.pdf) is an interesting short presentation about the topic.

---

---

## Checklist:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -276,6 +276,25 @@ add_custom_target(
         BYPRODUCTS ${CMAKE_CURRENT_SOURCE_DIR}/version.c
         COMMENT "Generating version.c using gen_version.cmake")
 
+# Add link-time optimization (LTO)
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
+    # -flto=<NUM_CORES> means that we use the number of available cores for LTO
+    #
+    # -fuse-linker-plugin: Use the linker plugin for LTO. This generally
+    # improves the LTO performance
+    #
+    # -fno-fat-lto-objects: Do not generate fat LTO objects. This means the
+    # generated object files are smaller but linking without LTO is not possible
+    #
+    # -fwhole-program: Assume that the current compilation unit is the whole
+    # program. This allows the compiler to perform more aggressive optimizations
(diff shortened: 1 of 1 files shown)


DL6ER:
I don't think we will soon make any real progress here...

yubiuser:
Did you figure out, why it failed?

XhmikosR:
I don't have the knowledge to try it myself, but I would try if LTO works with clang so that we can get some numbers with and without LTO.

DL6ER:
With LTO activated, some parts of the generated binary "simply don't work". I don't have a really good explanation for this. I tried to pin this down to a single part which may be failing but a lot of components simply seem to misbehave (sqlite3, regex, ...). At this point it is not possible continuing on this for me given the unclear actual improvement and the lack of time for this feature.

XhmikosR:
@DL6ER is there any chance this works now that the images were updated or is the GCC version still the same as before?

DL6ER:
Let's find out: https://github.com/pi-hole/FTL/actions/runs/15518659316

XhmikosR:
Thanks! It seems the same issue is still present. :/

PR #2318, opened 2025-03-02
Add "application/pkix-cert" and "application/pkix-crl" MIME types.

abra-mixabra:
### Add registered MIME types for serving certificate and CRL files from FTL.

Currently built-in FTL webserver recognizes only a very limited number of MIME types.
While I'm sure the list may be extended significantly I'd like to contribute the MIME types that are commonly used to serve certificate (.crt, .cer) and CRL files (.crl).

See: [RFC-2585](https://datatracker.ietf.org/doc/html/rfc2585#section-4)

### DIFF
--- a/src/webserver/civetweb/civetweb.c
+++ b/src/webserver/civetweb/civetweb.c
@@ -8565,6 +8565,9 @@ static const struct {
      * (http://www.iana.org/assignments/media-types)
      * application types */
     {".bin", 4, "application/octet-stream"},
+    {".cer", 4, "application/pkix-cert"},
+    {".crl", 4, "application/pkix-crl"},
+    {".crt", 4, "application/pkix-cert"},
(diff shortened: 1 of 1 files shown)


DL6ER:
Thank you for you submission. Please note that you are modifying an external component we are not maintaining ourselves here. Your change would simply be overwritten on the next update if this component. Have, I'd encourage you to open a PR towards https://github.com/civetweb/civetweb instead. Once your change has been accepted (their review times are usually a few days up to two weeks max), will import it into Pi-hole during the next update of CivetWeb. You are invited to ping us in an issue when it's time to do so in order to speed up the process of implementing this in Pi-hole.

yubiuser:
Upstream PR: https://github.com/civetweb/civetweb/pull/1329

PR #2357, opened 2025-03-08
Use keep-alive

DL6ER:

Use HTTP/1.1 keep-alive to allows clients to reuse TCP connection for subsequent HTTP requests, which **can improve performance dramatically** as no further TLS handshakes and connection opening/closing are required. This reduces pressure especially on low-end hardware lacking hardware-crypto for TLS.

This is disabled by default in CivetWeb as it puts additional compliance constraints on the request handlers. However, Pi-hole's API already fully complies with these additional constraints so no other changes than enabling this are needed.

We set a default timeout of 5 seconds after which the server closes the ready-to-be-used-again connections to free them for other requests. We intentionally keep such a long timeout so that requests being fired on the background (e.g. the regular update of `GET /api/summary`) can benefit here as well without needing additional handshakes.

Note that, while this reduces the overhead for opening and closing connections when loading several resources from one server, it also blocks one port and one thread at the server during the lifetime of this connection. Unfortunately, most browsers do not seem to close the keep-alive connection after loading all resources required to show a website. The server closes a keep-alive connection, if there is no additional request from the client during this timeout. If there are really many clients using the webserver at the same time, this may require you to increase `webserver.threads` when you see timeouts.

> [!NOTE]
> Don't get me wrong here. This does *not* reduce the number of TLS handshakes from, say, 35 to 1. It just allows already established connections to be reused. If the web interface makes five requests *exactly* parallel, there will be five handshakes. 
> However, if requests happens sequentially, the total number of handshakes is drastically reduced (see TL;DR below). Even on my fairly fast `x86_64` microserver running Pi-hole the speed enhancement was noticeable.
>
> TL;DR: On `development`, we see roughly 70 individual TLS handshakes but **only six** on this branch.

## Before this PR
Each connection needed a connection-establishment incl. a TLS handshake *on its own*.

Screenshot from 2025-03-08 09-51-30
Screenshot from 2025-03-08 09-50-59

## After this PR
Only the first connection needs the full handshake, following connections can reuse the existing connection *without* having to do any additional handshaking *on their own*.

Screenshot from 2025-03-08 09-52-09
Screenshot from 2025-03-08 09-51-27

---

---

## Checklist:

### DIFF
--- a/src/webserver/webserver.c
+++ b/src/webserver/webserver.c
@@ -475,6 +475,8 @@ void http_init(void)
 		"authentication_domain", config.webserver.domain.v.s,
 		"additional_header", webheaders,
 		"index_files", "index.html,index.htm,index.lp",
+		"enable_keep_alive", "yes",
+		"keep_alive_timeout_ms", "5000",
 		NULL, NULL,
 		NULL, NULL, // Leave slots for access control list (ACL) and TLS configuration at the end
 		NULL


PromoFaux (approved):
I played with this option when I was trying to hunt down slowness issues after initial release - I never got around to investigating it fully, though. LGTM

yubiuser (approved):

XhmikosR:
Confirmed! I was right to think something was missing :)

XhmikosR (approved):

PR #2425, opened 2025-04-14
Feature/fix yocto cross build

rehsack:

This PR tries to solve some issues when building FTL*DNS from source from within Yocto.

These are only configure/compile stage fixes according to chosen target platform (beaglebone-black) using yocto.

There is no behavior change, the fixes should work in-place and I expect they still work in existing environment.
`lua_scripts` is a bit hacky ...

---

---

### DIFF
--- a/src/ntp/client.c
+++ b/src/ntp/client.c
@@ -240,7 +240,7 @@ static bool settime_skew(const double offset)
 	tx.offset = 1000000 * offset;
 	tx.modes = ADJ_OFFSET_SINGLESHOT;
 
-	log_debug(DEBUG_NTP, "Gradually adjusting system time by %ld us", tx.offset);
+	log_debug(DEBUG_NTP, "Gradually adjusting system time by %"PRId64" us", (int64_t)tx.offset);
 
 	if(adjtimex(&tx) < 0)
 	{
--- a/src/lua/scripts/CMakeLists.txt
+++ b/src/lua/scripts/CMakeLists.txt
@@ -8,28 +8,40 @@
 # This file is copyright under the latest version of the EUPL.
 # Please see LICENSE file for your rights under this license.
 
-set(sources
-        inspect.lua.hex
-        scripts.h
+set(resources
+        inspect.lua
         )
 
-# Compile files from raw/ into hex/
+set(sources
+        ${resources}
+	scripts.h
+        )
+
+# Ensure target lua_scripts is build before target ftl_lua depending on it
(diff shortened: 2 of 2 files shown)


src/ntp/client.c
```diff
@@ -240,7 +240,7 @@ static bool settime_skew(const double offset)
 	tx.offset = 1000000 * offset;
 	tx.modes = ADJ_OFFSET_SINGLESHOT;
 
-	log_debug(DEBUG_NTP, "Gradually adjusting system time by %ld us", tx.offset);
+	log_debug(DEBUG_NTP, "Gradually adjusting system time by %"PRId64" us", tx.offset);
```
DL6ER, line 243: This causes the 32 bit builds (both `x86` and `arm`) to fail with
```
#9 0.798 /app/src/ntp/client.c: In function 'settime_skew':
#9 0.799 /app/src/ntp/client.c:243:30: error: format '%lld' expects argument of type 'long long int', but argument 4 has type 'long int' [-Werror=format=]
#9 0.799   243 |         log_debug(DEBUG_NTP, "Gradually adjusting system time by %"PRId64" us", tx.offset);
#9 0.799       |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~             ~~~~~~~~~
#9 0.799       |                                                                                   |
#9 0.799       |                                                                                   long int
#9 0.799 /app/src/log.h:71:43: note: in definition of macro 'log_debug'
#9 0.799    71 |                 _FTL_log(LOG_DEBUG, flag, format, ## __VA_ARGS__)
#9 0.799       |                                           ^~~~~~
```

rehsack: The difference seems to be the kernel build, not the architecture. I received the opposite when building for armv7at2hf-neon - which is also 32-bit, but not Debian. I know in libstatgrab we simply test for the type - but without autoconf ... I do a bit research tomorrow if there is something in kernel/libc which suits better.

DL6ER: I'd say a pretty obviousl solution would be casting this to whatever we deem appropriate, e.g.
``` c
log_debug(DEBUG_NTP, "Gradually adjusting system time by %"PRId64" us", (int64_t)tx.offset);
```

rehsack: That's why I wanted to postpone it to today ;)
Sometimes I am searching a too much sophisticated way because of a situation in late 80s with a broken compiler ... let's go forward with your proposal.

DL6ER (dismissed):

PR #2448, opened 2025-05-10
Set dns.interface to eth0 by default

yubiuser:

Set `dns.interface` to `eth0?` by default. We already use `eth0` if no interface is set, but this PR makes this more obvious to the users.

https://github.com/pi-hole/FTL/blob/e0f96fee53413ce32cc0c95878d14de85cddda38/src/config/dnsmasq_config.c#L445-L448

Accompanies: https://github.com/pi-hole/pi-hole/pull/6216 and https://github.com/pi-hole/web/pull/3436

---

---


DL6ER (approved):
While I don't actually think this is the right thing to do, it is certainly better than what we have right now and, hence, a step into the right direction. A proper fix would have to suggest the primary interface of the computer which is - due to predictable network interface names - now rather something like `enp2s0` or similar.

PR #2456, opened 2025-05-15
Automatically detect DNS interface when empty in pihole.toml

DL6ER:

Follow-up on similar PRs across Pi-hole's repositories, see https://github.com/pi-hole/FTL/pull/2448#pullrequestreview-2837769640

You can use
```
pihole-FTL --default-gateway
```
to get which interface would be used on your particular machine

---

**Related issue or feature (if applicable):** https://github.com/pi-hole/FTL/pull/2448#pullrequestreview-2837769640

---

## Checklist:

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -70,6 +70,8 @@
 #include "ntp/ntp.h"
 // check_capability()
 #include "capabilities.h"
+// get_gateway_name()
+#include "tools/netlink.h"
 
 // defined in dnsmasq.c
 extern void print_dnsmasq_version(const char *yellow, const char *green, const char *bold, const char *normal);
@@ -600,13 +602,21 @@ void parse_args(int argc, char *argv[])
 		exit(EXIT_SUCCESS);
 	}
 
-
 	// Set config option through CLI
 	if(argc == 3 && strcmp(argv[1], "migrate") == 0 && strcmp(argv[2], "v6") == 0)
 	{
 		cli_mode = true;
 		log_ctrl(false, true);
 		exit(migrate_config_v6() ? EXIT_SUCCESS : EXIT_FAILURE);
+	}
+
+	// Get name of the default gateway
+	if(argc == 2 && strcmp(argv[1], "--default-gateway") == 0)
+	{
+		cli_mode = true;
+		char *name = get_gateway_name();
+		printf("%s\n", name);
+		exit(EXIT_SUCCESS);
 	}
 
 	// start from 1, as argv[0] is the executable name
@@ -1155,6 +1165,7 @@ void parse_args(int argc, char *argv[])
 			printf("\t                    authentication (if enabled)\n");
 			printf("\t%s--perf%s              Run performance-tests based on the\n", green, normal);
 			printf("\t                    BALLOON password-hashing algorithm\n");
+			printf("\t%s--default-gateway%s   Get default network interface's name\n", green, normal);
 			printf("\t%s--%s [OPTIONS]%s        Pass OPTIONS to internal dnsmasq resolver\n", green, cyan, normal);
 			printf("\t%s-h%s, %shelp%s            Display this help and exit\n\n", green, normal, green, normal);
 			exit(EXIT_SUCCESS);
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -28,8 +28,11 @@
 #include <unistd.h>
 // wait
 #include <sys/wait.h>
+// get_gateway_name()
+#include "tools/netlink.h"
 
 #define HEADER_WIDTH 80
+
 
 static bool test_dnsmasq_config(char errbuf[ERRBUF_SIZE])
 {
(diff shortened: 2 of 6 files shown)


yubiuser (commented):
> Leave empty for auto-detection.

I'm not sure if I like that. That makes it unpredictable. I think we should have this auto detecting code only for the migration and then write it into `pihole.toml`. If the user wants to change it, they happily can do this afterwards.

DL6ER:
I thought about this as well. In the end, it is not unpredictable. You have the Ethernet cable only in one of the sockets so the automatic detection should always return the same. If you happen to use another one (for whatever reason), it may not be that bad that the automatism automatically changes the interface for you.

Writing it statically into the file would also make the tests harder because this would make the `toml` be different based on the machine the CI tests are running one. This wouldn't be a show-stopper and could be coded around, but still, on second and third thought, I still think the "empty is auto" is the best approach here. And I *think* we do this elsewhere, too.

rdwebdesign:
Will the API still return an empty string (just like the `pihole.toml` file)?

In this case, the Web Interface will show `eth0`.

DL6ER:
Yes. And this is a very good point. So we need to add an additional AJAX call to get the interface in this case from the API (instead of blindly substituting with "eth0")

DL6ER:
> We don't need another AJAX call.

Yes, we do need one. This field you are mentioning has to stay empty when it is empty in `pihole.toml`. Otherwise, when the user hits "save & apply", the empty value would be replaced by the substituted one.

We can use `network/gateway` here and just take `gateway.0.interface`. This is guaranteed to be the same FTL will by pick by default for an empty sting.

rdwebdesign:
> We can use `network/gateway` here and just take `gateway.0.interface`. This is guaranteed to be the same FTL will by pick by default for an empty sting.

OK.
Then we need a new API call to get `gateway.0.interface` and adjust the code in `scripts/js/settings-dns.js`.

I will try to open a new PR later.

rdwebdesign (dismissed):
From my point of view this PR is approved, unless yubiuser has something else to add.

DL6ER:
@yubiuser any objections?

src/args.c
```diff
@@ -1155,6 +1165,7 @@ void parse_args(int argc, char *argv[])
 			printf("\t                    authentication (if enabled)\n");
 			printf("\t%s--perf%s              Run performance-tests based on the\n", green, normal);
 			printf("\t                    BALLOON password-hashing algorithm\n");
+			printf("\t%s--default-gateway%s   Get default network interface's name\n", green, normal);
```
yubiuser, line 1168: Do we need a CLI option here?

src/config/dnsmasq_config.c
```diff
@@ -442,10 +445,20 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
...
+	char *interface = conf->dns.interface.v.s;
 	// Use eth0 as fallback interface if the interface is missing
 	if(strlen(interface) == 0)
-		interface = "eth0";
+	{
+		interface = get_gateway_name();
+		if(interface == NULL)
+		{
+			log_err("Unable to get default interface name");
+			interface = (char*)"eth0";
```
yubiuser, line 457: We already fallback to `eth0` within `get_gateway_name`. Do we need a double-net?

src/config/dnsmasq_config.c
```diff
@@ -442,10 +445,20 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 		fputs("\n", pihole_conf);
 	}
 
-	const char *interface = conf->dns.interface.v.s;
+	bool interface_allocated = false;
+	char *interface = conf->dns.interface.v.s;
 	// Use eth0 as fallback interface if the interface is missing
```
yubiuser, line 450:
```suggestion
         // Auto-detect the gateway interface if empty
	// Use eth0 as fallback if the interface name could not be retrieved
```

yubiuser (commented):
No objections to the proposed idea. 
Just for the implementation: 

> We can use network/gateway here and just take gateway.0.interface.

If we have code already at an other place, why not use this directly, but instead create a new function here?

src/args.c
DL6ER: No, I thought this is something we may want to use in the `debug` log

src/config/dnsmasq_config.c
DL6ER: No, we can remove this one

rdwebdesign (approved):

yubiuser:
I was just asking because this also what `get_gateway()` does

https://github.com/pi-hole/FTL/blob/632a90de717b8662e8a00a5a53111f85a4912aa7/src/api/network.c#L47-L70

But it might be more code to extract part of the information from `get_gateway()` instead of just doing what we really need in `get_gateway_name()`

yubiuser (approved):

PR #2405, opened 2025-04-01
Prevent home.arpa queries from being upstreamed. RFC 8375

rrobgill:
RFC8375 defines the home.arpa special use domain for local dns in a home network.
These are not globally resolvable and should not be sent upstream.

**What does this PR aim to accomplish?:**

Prevent queries for home.arpa domains being sent to upstream services.

**How does this PR accomplish the above?:**

Checks if the user has specified home.arpa as their local domain, and that we have configured dnsmasq to use a suitable local server for these queries.

Otherwise, blocks queries to this domain.

**Link documentation PRs if any are needed to support this PR:**
n/a

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [X] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #2474, opened 2025-05-31
[RFC] Prevent .internal queries from being upstreamed. Draft draft-davies-internal-tld-03

Tooa:
Hi :wave:,

after seeing https://github.com/pi-hole/FTL/pull/2405 in the latest release notes, I figured I hand this change in for RFC.

The draft [draft-davies-internal-tld-03](https://datatracker.ietf.org/doc/html/draft-davies-internal-tld-03) defines  `.internal` as  special use domain for local DNS in a home network.
These are not globally resolvable and should not be sent upstream.

Unsure, if you want to upstream this right now as it is still in draft and no RFC published yet. Let me know what you think about this.

I also held back on refactoring the two similar cases. First, because of the rule of three and second because of the RFC character of this PR.

Prevent queries for .internal domains being sent to upstream services.

Checks if the user has specified .internal as a reverse server domain. In that case, we will forward the query to this server.

Otherwise, blocks queries to this domain.

n/a

---

---

### DIFF
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -476,6 +476,7 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 	// Add upstream DNS servers for reverse lookups
 	bool domain_revServer = false;
 	bool domain_homearpa = false;
+	bool domain_internal = false;
 	const unsigned int revServers = cJSON_GetArraySize(conf->dns.revServers.v.json);
 	for(unsigned int i = 0; i < revServers; i++)
 	{
@@ -530,6 +531,10 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 			// Flag if configured a server for queries for home.arpa domains
 			if(strcmp(domain, "home.arpa") == 0)
 				domain_homearpa = true;
+			
+			// Flag if configured a server for queries for .internal domains
+			if(strcmp(domain, "internal") == 0)
+				domain_internal = true;
 		}
 
 		// Forward unqualified names to the target only when the "never forward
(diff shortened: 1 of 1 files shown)


dschaper:
I don't see an issue with it in concept. It looks like ICANN has blocked `.internal` from being a TLD so it should be safe to block it from upstream _if_ that upstream is remote. It could be that the upstream is another internal DNS server that is authoritative for that zone locally. Think something like Active Directory or other manage DNS like `freeIPA`.

DL6ER:
We have a similar discussion going on for `home.arpa` right now in https://github.com/pi-hole/FTL/issues/2506. However, as I stated there as well, I do think that one can safely assume that whatever is defined as "general" upstream for Pi-hole should never receive such internal queries. May it be local or not.

What's exactly what we have the "conditional forwarding" (also called `rev-server`) for. Here you define that, e.g., `129.168.0.1` is to be asked for `1291.68.0.1/24` with domain `myhome`. When you do with with `home.arpa`, this blocking will not be in effect but we - instead - make sure that such queries are always sent to this specific upstream, but never to the "general" ones. This PR adds the exact same logic for `internal`: When you have a `rev-server` defined with domain `internal`, it will continue to work as before. The only thing ensured herein is that `internal` is not handed to those "general" upstream server (may it by `8.8.8.8` or some local `unbound` at `10.1.4.23` or whatever else).

I think this change warrants the forthcoming release to be a 6.x (not 6.2.x) as this *may* effect at least a few isolated cases where users are expecting this to be forwarded to the general upstream (as right now).

DL6ER:
@Tooa Could you please rebase on or merge the current `development` branch into your branch? This will fix the `riscv64` job as well.

DL6ER:
Relevant part from `man dnsmasq`:

> Note that the behaviour for queries which don't match the specified address literal changed in version 2.86. Previous versions, configured with (eg) --address=/example.com/1.2.3.4 and then queried for a RR type other than A would return a NoData answer. From 2.86, the query is sent upstream. To restore the pre-2.86 behaviour, use the configuration --address=/example.com/1.2.3.4 --local=/example.com/ 

Your PR adds this `local=/internal/` so when they have the `address=/internal/1.2.3.4` everything should still work as expected.

DL6ER:
Ready to approve + merge from my side, waiting on @dschaper's comment about if we want to merge it now for 6.2.3 or with for 6.3 (see my comment above)

dschaper:
It's not a bugfix so it should go out with a minor bump, i.e. 6.3.

dschaper (approved):

Issue #2504, opened 2025-06-07
NTP logs WARNING Received NTP request has unsupported version, ignoring


depuhitv:
### Versions

- Pi-hole:  Core version is v6.1.2 (Latest: v6.1.2)
- Web:  Web version is v6.2.1 (Latest: v6.2.1)
- FTL:  FTL version is v6.2.2 (Latest: v6.2.2)
Docker Tag [2025.06.0](https://github.com/pi-hole/docker-pi-hole/releases/2025.06.0)

### Platform

- OS and version:  "Ubuntu 24.04.2 LTS

- Platform:  VM
and Docker

### Expected behavior
Lower NTP client versions 1,2,3 and 4 are supported

### Actual behavior / bug
NTP client version 3 is rejected and debug logs have a message
"WARNING Received NTP request has unsupported version, ignoring"

### Steps to reproduce

1. Enable Debug logs at
Settings -> All Settings --> Debug --> debug.ntp

2. Query pihole NTP using command 
sudo ntpdate -o 4 piholeIP
3. Time adjusted as expected.
5. Query pithole NTP using command sudo ntpdate -o 3 piholeIP
4. Message seen "no server suitable for synchronization found"

## Debug Token

- URL:  

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Additional context

This seems to have due to change merged recently in this release.
https://github.com/pi-hole/FTL/pull/2431 where you have a check to only respond if the NTP version received is 4
```
	// Check if the request is NTP version 4
	}
	if (((recv_buf[0] >> 3) & 0x07) != 0x4) {
		log_warn("Received NTP request has unsupported version, ignoring");
		return false;
	}
```

rrobgill:
>NTP client version 3 is rejected and debug logs have a message "WARNING Received NTP request has unsupported version, ignoring"
>This seems to have due to change merged recently in this release [..] where you have a check to only respond if the NTP version received is 4

Yes it corrected a bug where the version had been determined incorrectly, it may have permitted incorrect responses to be sent to earlier versions.

Did you have version 3 software connecting to Pi-hole previously, and if so was the time synchronization returned accurate ?

A look at the earlier RFC1305 for version 3, seems that it may well work.

Can you point me to some current software that _requires_ the earlier versions that I can test it properly?

depuhitv:
> Did you have version 3 software connecting to Pi-hole previously, and if so was the time synchronization returned accurate ?
[...]

PR #2505, opened 2025-06-07
ntp : ignore client version, always return a v4 packet

rrobgill:

Send an ntp response to any client, no matter the version the client is identifying as.

Some clients (such as win32tm shipping with Microsoft Windows 11, and apparently some IOT devices described in #2504) send their client packet with an unspecified version number, or set to version 1. 

I have confirmed that in the case of Microsoft Windows, it is able to accept and correctly interpret a valid ntp v4 packet in response. 

The RFCs for the earlier versions are from 1992 (v3), 1989 (v2) and 1985 (v1), so it seems unlikely that these clients are actually expecting data to be returned in any of these older formats, but this PR does return the onus of correct interpretation to the clients.

Delete the version check for incoming client packets.

N/A

---

---

### DIFF
--- a/src/ntp/server.c
+++ b/src/ntp/server.c
@@ -82,12 +82,11 @@ static bool ntp_reply(const int socket_fd, const struct sockaddr *saddr_p, const
 	if ((recv_buf[0] & 0x07) != 0x3) {
 		log_warn("Received invalid NTP request: not from an NTP client, ignoring");
 		return false;
-	// Check if the request is NTP version 4
 	}
-	if (((recv_buf[0] >> 3) & 0x07) != 0x4) {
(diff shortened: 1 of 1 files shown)


rrobgill:
If it was just some poorly coded IOT devices, I'd be inclined to leave well enough alone. But with Microsoft Windows sending such requests I believe that on balance this is probably the best approach.

yubiuser (approved):

DL6ER:
Yes, and no. We should remove the enforcement but we should keep logging this in debug mode in case such a cheap IoT device once relly expects such an old version

yubiuser (approved):

PR #2460, opened 2025-05-20
Add option to enable/disable dnsmasq localise-queries option

Manakuremati:
## Thank you for your contribution to the Pi-hole Community!

Please read the comments below to help us consider your Pull Request.

We are all volunteers and completing the process outlined will help us review your commits quicker.

**Please make sure you**

 1. Base your code and PRs against the repositories developmental branch.
 2. [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits as we enforce the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
 3. [Sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all your commits as they must have verified signatures
 4. File a pull request for any change that requires changes to [our documentation](https://docs.pi-hole.net/) at our [documentation repo](https://github.com/pi-hole/docs) 

---

**What does this PR aim to accomplish?:**

Add a config option to enable/disable the dnsmasq localise-queries option.

**How does this PR accomplish the above?:**

Added a new part for this in /src/config/config.c and an IF statement in /src/config/dnsmasq_config.c. By default the option is enabled and writes it into the dnsmasq.conf file. If it is disabled it is not written to dnsmasq.conf and therefore not activated.
I added it to the miscellaneous section as I'm the first person who wishes for this and 99.9% of people probably will not use this option.
Please correct my code. I normally don't work with code so this fairly out of my scope.

Forum Links:
https://discourse.pi-hole.net/t/localise-queries-option-now-always-set/79934?u=asfp
https://discourse.pi-hole.net/t/dnsmasq-localise-queries-checkbox-in-dns-settings/79975?u=asfp

**Link documentation PRs if any are needed to support this PR:**

No links

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
[...]

PR #2524, opened 2025-06-14
Add dns.localise

Manakuremati:
Another one for (https://github.com/pi-hole/FTL/pull/2460)
I think I managed to rebase it correctly this time

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -579,6 +579,13 @@ static void initConfig(struct config *conf)
 	conf->dns.port.d.u16 = 53u;
 	conf->dns.port.c = validate_stub; // Only type-based checking
 
+	conf->dns.localise.k = "dns.localise";
+	conf->dns.localise.h = "Enable/Disable the localise-queries option of dnsmasq. When this setting is disabled dnsmasq will return all possible values for local DNS Records. Enabled by default";
+	conf->dns.localise.t = CONF_BOOL;
+	conf->dns.localise.f = FLAG_RESTART_FTL;
+	conf->dns.localise.d.b = true;
+	conf->dns.localise.c = validate_stub; // Only type-based checking
+		
 	// sub-struct dns.cache
 	conf->dns.cache.size.k = "dns.cache.size";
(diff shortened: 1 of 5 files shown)


src/config/config.c
```diff
@@ -579,6 +579,13 @@ static void initConfig(struct config *conf)
 	conf->dns.port.d.u16 = 53u;
 	conf->dns.port.c = validate_stub; // Only type-based checking
 
+	conf->dns.localise.k = "dns.localise";
+	conf->dns.localise.h = "Enable/Disable the localise-queries option of dnsmasq. When this setting is disabled dnsmasq will return all possbile values for local DNS Records. Enabled by default";
```
rdwebdesign:
```suggestion
	conf->dns.localise.h = "Enable/Disable the localise-queries option of dnsmasq. When this setting is disabled dnsmasq will return all possible values for local DNS Records. Enabled by default";
```

test/pihole.toml
```diff
@@ -209,6 +209,10 @@
   # Port used by the DNS server
   port = 53
 
+  # Enable/Disable the localise-queries option of dnsmasq. When this setting is disabled 
+  # dnsmasq will return all possbile values for local DNS Records. Enabled by default
```
rdwebdesign:
```suggestion
  # dnsmasq will return all possible values for local DNS Records. Enabled by default
```

rdwebdesign (changes requested):

yubiuser:
The file is  self-generated on FTL start in the test environment. It can be very picky when the help text of the option does not match `test/pihole.toml`
You also need to change this line


https://github.com/pi-hole/FTL/blob/8659f1fb48fafc3f1be80c0af6487692a7407948/test/pihole.toml#L1205

Manakuremati:
Changed it to "# 156 total entries out of which 99 entries are default"

test/pihole.toml
```diff
@@ -209,6 +209,10 @@
   # Port used by the DNS server
   port = 53
 
+  # Enable/Disable the localise-queries option of dnsmasq. When this setting is disabled
```
yubiuser:
```suggestion
  # Enable/Disable the localise-queries option of dnsmasq. When this setting is disabled
```

I think there is an additional space....

Issue #2497, opened 2025-06-06
Local addresses defined in dnsmasq config return NXDOMAIN


DragonQ:
### Versions

- Docker Tag: 2025.06.0
- Core: v6.1.2
- FTL: v6.2.2
- Web interface: v6.2.1]

### Platform
- OS and version:  Docker
- Platform: Docker

### Expected behavior

Local addresses defined in dnsmasq config should return correct IPs when `misc.etc_dnsmasq_d` is enabled, e.g.:

```
server=/lan/192.168.1.254
server=/1.168.192.in-addr.arpa/192.168.1.254
server=/iot.lan/192.168.2.254
server=/2.168.192.in-addr.arpa/192.168.2.254
server=/not.lan/192.168.3.254
server=/3.168.192.in-addr.arpa/192.168.3.254
```

### Actual behavior / bug

All local addresses return `NXDOMAIN`.

### Steps to reproduce

`nslookup device.lan`

## Debug Token

- URL: https://tricorder.pi-hole.net/JkPBVQ1R

### Additional context

Downgrading to image 2025.4.0 (FTL 6.1) fixes the issue.

```
services:
  pihole:
    container_name: pihole
    image: 'pihole/pihole:latest'
    restart: unless-stopped
    networks:
      pihole-net:
        ipv4_address: ${PIHOLE_IP}
    volumes:
      - ${PIHOLE_ROOT_PATH}/etc/:/etc/pihole/
      - ${PIHOLE_ROOT_PATH}/log/:/var/log/
      - ${PIHOLE_ROOT_PATH}/dnsmasq.d/:/etc/dnsmasq.d/
    ports:
      - '${HOST_IP}:53:53/tcp'
      - '${HOST_IP}:53:53/udp'
      - '${HOST_IP}:${PIHOLE_GUI_PORT}:80/tcp'
    environment:
      - TZ='Europe/London'
      - WEBPASSWORD=${PIHOLE_GUI_PASSWORD}
      - DNS1=${PIHOLE_UPSTREAM_DNS}
      - DNS2=no
      - MAXDBDAYS=60
    dns:
      - 127.0.0.1
      - 208.67.222.222
  unbound:
    container_name: unbound
    image: 'klutchell/unbound:latest'
    restart: unless-stopped
    cap_add:
      - NET_ADMIN # Needed to fix so-rcvbuf & so-sndbuf errors
    networks:
      pihole-net:
        ipv4_address: ${UNBOUND_IP}
    volumes:
[...]

PR #2531, opened 2025-06-19
Add new dns.domainLocal config option

DL6ER:

Add new dns.domainLocal` config option defaulting to true.

If set, the domain is considered local and queries for this domain are never forwarded upstream unless a `dns.revServer` is configured for this domain.\n If unset, queries for this domain are forwarded upstream to (possibly public) server which is probably not what you want *unless* you have added extra configuration for this domain *or* your upstream servers are able to handle local domains (e.g., router).

---

**Related issue or feature (if applicable):** Fixes https://github.com/pi-hole/FTL/issues/2497

---

## Checklist:

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -494,6 +494,13 @@ static void initConfig(struct config *conf)
 	conf->dns.domainNeeded.d.b = false;
 	conf->dns.domainNeeded.c = validate_stub; // Only type-based checking
 
+	conf->dns.domainLocal.k = "dns.domainLocal";
+	conf->dns.domainLocal.h = "If set, the domain is considered local and queries for this domain are never forwarded upstream unless a dns.revServer is configured for this domain.\n If unset, queries for this domain are forwarded upstream to (possibly public) server which is probably not what you want *unless* you have added extra configuration for this domain *or* your upstream servers are able to handle local domains (e.g., router).";
+	conf->dns.domainLocal.t = CONF_BOOL;
+	conf->dns.domainLocal.f = FLAG_RESTART_FTL;
+	conf->dns.domainLocal.d.b = true;
+	conf->dns.domainLocal.c = validate_stub; // Only type-based checking
+
 	conf->dns.expandHosts.k = "dns.expandHosts";
 	conf->dns.expandHosts.h = "If set, the domain is added to simple names (without a period) in /etc/hosts in the same way as for DHCP-derived names";
 	conf->dns.expandHosts.t = CONF_BOOL;
--- a/test/pihole.toml
+++ b/test/pihole.toml
@@ -1,7 +1,7 @@
-# Pi-hole configuration file (v6.0.4-75-gc53cf00b)
+# Pi-hole configuration file (v6.2.3-16-g7eb5c0bb-dirty)
 # Encoding: UTF-8
 # This file is managed by pihole-FTL
-# Last updated on 2025-03-27 11:04:20 UTC
+# Last updated on 2025-06-19 18:31:11 UTC
 
 [dns]
   # Array of upstream DNS servers used by Pi-hole
@@ -107,6 +107,14 @@
   # If set, A and AAAA queries for plain names, without dots or domain parts, are never
   # forwarded to upstream nameservers
   domainNeeded = false
+
+  # If set, the domain is considered local and queries for this domain are never
+  # forwarded upstream unless a dns.revServer is configured for this domain.
+  # If unset, queries for this domain are forwarded upstream to (possibly public) server
+  # which is probably not what you want *unless* you have added extra configuration for
+  # this domain *or* your upstream servers are able to handle local domains (e.g.,
+  # router).
+  domainLocal = true
 
   # If set, the domain is added to simple names (without a period) in /etc/hosts in the
   # same way as for DHCP-derived names
@@ -1206,7 +1214,7 @@
   all = true ### CHANGED, default = false
 
 # Configuration statistics:
-# 156 total entries out of which 99 entries are default
+# 157 total entries out of which 100 entries are default
 # --> 57 entries are modified
 # 3 entries are forced through environment:
 #   - misc.nice
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -210,6 +210,8 @@ components:
                     type: string
                 domainNeeded:
                   type: boolean
+                domainLocal:
+                  type: boolean
                 expandHosts:
                   type: boolean
                 domain:
@@ -666,6 +668,7 @@ components:
             hosts:
               - "192.168.2.123 mymusicbox"
             domainNeeded: true
+            domainLocal: true
             expandHosts: true
             domain: "lan"
             bogusPriv: true
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -141,6 +141,7 @@ struct config {
 		struct conf_item blockTTL;
 		struct conf_item hosts;
 		struct conf_item domainNeeded;
+		struct conf_item domainLocal;
 		struct conf_item expandHosts;
 		struct conf_item domain;
 		struct conf_item bogusPriv;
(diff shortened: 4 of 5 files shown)


dschaper:
Probably some language confusion for me, let me see if I am clear:

`dns.domain=bender`

Then set `dns.domainLocal=true`? Any query for `*.bender` will never be forwarded, unless a specific revserver entry for `bender` is set. `true,192.168.88.0/24,192.168.88.1,bender`

`dns.domainLocal=false` allows queries to be resolved by upstream DNS servers.

I think it would be better to have the flag as `dns.domain.local` to make it clear that the `local` boolean is specifically for `dns.domain` as it only modifies the behavior of queries for `dns.domain`. camelCasing the flag makes it seem like the flag should be used with the domain to be considered local `dns.domainLocal=bender`.

rdwebdesign:
If there is a technical reason to keep the flag under `dns.`, we can try a different name to make more clear the flag job, like `dns.domainIsLocal` or something like that.

src/config/config.c
```diff
@@ -494,6 +494,13 @@ static void initConfig(struct config *conf)
 	conf->dns.domainNeeded.d.b = false;
 	conf->dns.domainNeeded.c = validate_stub; // Only type-based checking
 
+	conf->dns.domainLocal.k = "dns.domainLocal";
+	conf->dns.domainLocal.h = "If set, the domain is considered local and queries for this domain are never forwarded upstream unless a dns.revServer is configured for this domain.\n If unset, queries for this domain are forwarded upstream to (possibly public) server which is probably not what you want *unless* you have added extra configuration for this domain *or* your upstream servers are able to handle local domains (e.g., router).";
```
PromoFaux: Just a nitpick for consistency with changes here: https://github.com/pi-hole/FTL/pull/2558

```suggestion
	conf->dns.domainLocal.h = "If set, the domain is considered local and queries for this domain are never forwarded upstream unless a dns.revServer is configured for this domain.\n\n If unset, queries for this domain are forwarded upstream to (possibly public) server which is probably not what you want *unless* you have added extra configuration for this domain *or* your upstream servers are able to handle local domains (e.g., router).";
```

src/config/config.c
```diff
@@ -494,6 +494,13 @@ static void initConfig(struct config *conf)
...
+	conf->dns.domainLocal.t = CONF_BOOL;
```
PromoFaux: Just a nitpick for consistency with changes here: https://github.com/pi-hole/FTL/pull/2558

```suggestion
	conf->dns.domainLocal.a = cJSON_CreateStringReference("<true|false>");
	conf->dns.domainLocal.t = CONF_BOOL;
```

test/pihole.toml
```diff
@@ -108,6 +108,14 @@
   # forwarded to upstream nameservers
   domainNeeded = false
 
+  # If set, the domain is considered local and queries for this domain are never
+  # forwarded upstream unless a dns.revServer is configured for this domain.
+  # If unset, queries for this domain are forwarded upstream to (possibly public) server
```
PromoFaux:
```suggestion
  #
  # If unset, queries for this domain are forwarded upstream to (possibly public) server
```

test/pihole.toml
```diff
@@ -108,6 +108,14 @@
...
+  # which is probably not what you want *unless* you have added extra configuration for
+  # this domain *or* your upstream servers are able to handle local domains (e.g.,
+  # router).
```
PromoFaux:
```suggestion
  # router).
  #
  # Allowed values are:
  #     <true|false>
```

src/config/config.c
DL6ER: Not sure we actually need this, but I will comment on the other PR about this

PromoFaux: It's about verbosity and consistentcy. Even if it is obvious to us, we should treat documentation as catering to the lowest common denominator

src/config/config.c
```diff
@@ -495,6 +495,13 @@ static void initConfig(struct config *conf)
 	conf->dns.domainNeeded.d.b = false;
 	conf->dns.domainNeeded.c = validate_stub; // Only type-based checking
 
+	conf->dns.domainLocal.k = "dns.domainLocal";
+	conf->dns.domainLocal.h = "If set, the domain is considered local and queries for this domain are never forwarded upstream unless a dns.revServer is configured for this domain.\n\n If unset, queries for this domain are forwarded upstream to (possibly public) server which is probably not what you want *unless* you have added extra configuration for this domain *or* your upstream servers are able to handle local domains (e.g., router).";
```
yubiuser:
```suggestion
	conf->dns.domainLocal.h = "If set, the domain set in [dns.domain] is considered local and queries for this domain are never forwarded upstream unless a dns.revServer is configured for this domain.\n\n If unset, queries for this domain are forwarded upstream to (possibly public) server which is probably not what you want *unless* you have added extra configuration for this domain *or* your upstream servers are able to handle local domains (e.g., router).";
```

test/pihole.toml
```diff
@@ -137,6 +137,18 @@
   #     true or false
   domainNeeded = false
 
+  # If set, the domain is considered local and queries for this domain are never
```
yubiuser:
```suggestion
  # If set, the domain set in [dns.domain] is considered local and queries for this domain are never
```

yubiuser:
The syntax can be different, but we should add a reference anyways.

yubiuser:
> If we move dns.domain to dns.domain.name (or similar) to make dns.domain be a branch rather than a leaf, this is bordering on a breaking change IMO. External tools working with the API may not understand this and we'd need to write some migration code ourselves, too. Not impossible but certainly some work with potential to break things somewhere.

I'm going back and forth about this. While the current solution is fast, simple and non-breaking, `dns.domain.name` and `dns.domain.local` is probably the **right** solution. No one said we aren't allowed to break things - we only need to tell users. Looking at the current release draft, we already plan a minor bump to 6.3. IMO this change would fit in a minor version bump, no need to wait for v7....

DL6ER:
Yes, but I think we should write a blog post about this upcoming change so users are prepared. I will also work on some migration logic so that users won't even notice (unless external tools are involved).

DL6ER:
Migration works:

Screenshot From 2025-09-17 09-29-57

automatically migrates to

Screenshot From 2025-09-17 09-45-11

DL6ER:
Note that this was only visible on the first start of `pihole-FTL` after migration happens. On any subsequent start, the issues wasn't there any longer.

src/config/config.c
```diff
@@ -599,6 +591,22 @@ void initConfig(struct config *conf)
...
...eam *unless* you have configured a dns.revServer exactly for this domain. In the latter case, all queries for this domain are sent exclusively to this server (including reverse lookups).\n\n For DHCP, this has two effects; firstly it causes the DHCP server to return the domain to any hosts which request it, and secondly it sets the domain which it is legal for DHCP-configured hosts to claim. The intention is to constrain hostnames so that an untrusted host on the LAN cannot advertise its name via DHCP as e.g. \"google.com\" and capture traffic not meant for it. If no domain suffix is specified, then any DHCP hostname with a domain part (ie with a period) will be disallowed and logged. If a domain is specified, then hostnames with a domain part are allowed, provided the domain part matches the suffix. In addition, when a suffix is set then hostnames without a domain part have the suffix added as an optional domain part. For instance, we can set domain=mylab.com and have a machine whose DHCP hostname is \"laptop\". The IP address for that machine is available both as \"laptop\" and \"laptop.mylab.com\".\n\n You can disable setting a domain by setting this option to an empty string.";
```
yubiuser, line 600: Maybe 'This DNS domain is purely local' should be followed by ', unless you either set dns.revServer or dns.domain.local'.

Because here we claim it is purely local, but we offer two options to forward it anyway: revServer (which is mentioned and explained) and the new option dns.domain.local

DL6ER: We already have description that this may not be the case with `revServer`. I modified the description in the latest commit.

yubiuser: > I modified the description in the latest commit. 

Did you pusht his commit? The last one I see is the development merge commit.

DL6ER: The change is therein, it seems I silently squashed my commit into it...

image

yubiuser (commented):
Perfect. Let's squash this down a bit ,then it's good to go.

DL6ER:
The `rebase` messed up the `pihole.toml` footer so the config file did not meet the expectations, causing a failure. The solution was to regenerate the `test/pihole.toml` file on this branch. I squashed and force-pushed the change into the last commit.

yubiuser (approved):

PR #2539, opened 2025-06-27
Ensure cache size is sufficient for DNSSEC

rrobgill:

Ensures that when DNSSEC is enabled, the cache size is sufficient for DNSSEC to operate correctly.

Dnsmasq has a hardcoded lower cache size of 150 for DNSSEC to operate. Docs mentions the 150 limit but it is currently possible for a user to use the webUI or edit pihole.toml to set it below this size while DNSSEC is enabled, resulting in pihole-FTL being unable to start:

```
2025-06-27 01:09:44.822 CRIT Error in dnsmasq configuration: cannot reduce cache size from default when DNSSEC enabled
```

Adds a test when writing dnsmasq.conf, so that if dnssec is enabled any cache size below 150 is overridden with the cache is set to 150. This catches changes whether made in the UI or manual edits to pihole.toml.

Logs a note that the change was made.

Does not make any changes to cache size unless DNSSEC is enabled. (So that the cache may still be set to a small size or disabled entirely when the user does not have DNSSEC enabled).

Adds text to the settings page noting the minimum size that applies if DNSSEC is enabled, along with corresponding text in the tests.

N/A

---

---

### DIFF
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -348,6 +348,12 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 	}
 	fputs("# Set the size of dnsmasq's cache. The default is 150 names. Setting the cache\n", pihole_conf);
 	fputs("# size to zero disables caching. Note: huge cache size impacts performance\n", pihole_conf);
+	// Ensure cache is large enough for DNSSEC operation if DNSSEC is enabled
+	if (conf->dns.dnssec.v.b && conf->dns.cache.size.v.ui < 150)
+	{
+		log_warn("Requested cache size insufficient for DNSSEC operation, overriding to 150");
+		conf->dns.cache.size.v.ui = 150;
+	}
 	fprintf(pihole_conf, "cache-size=%u\n", conf->dns.cache.size.v.ui);
 	fputs("\n", pihole_conf);
 
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -588,7 +588,7 @@ static void initConfig(struct config *conf)
 		
 	// sub-struct dns.cache
 	conf->dns.cache.size.k = "dns.cache.size";
-	conf->dns.cache.size.h = "Cache size of the DNS server. Note that expiring cache entries naturally make room for new insertions over time. Setting this number too high will have an adverse effect as not only more space is needed, but also lookup speed gets degraded in the 10,000+ range. dnsmasq may issue a warning when you go beyond 10,000+ cache entries.";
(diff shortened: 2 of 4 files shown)


src/config/dnsmasq_config.c
```diff
@@ -348,6 +348,12 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 	}
 	fputs("# Set the size of dnsmasq's cache. The default is 150 names. Setting the cache\n", pihole_conf);
 	fputs("# size to zero disables caching. Note: huge cache size impacts performance\n", pihole_conf);
+	// Ensure cache is large enough for DNSSEC operation if DNSSEC is enabled
+	if (conf->dns.dnssec.v.b && conf->dns.cache.size.v.ui < 150)
+	{
+		log_info("Requested cache size insufficient for DNSSEC operation, overriding to 150");
```
yubiuser:
```suggestion
		log_warn("Requested cache size insufficient for DNSSEC operation, overriding to 150");
```

I would make this a warning.

rrobgill:
That's an interesting problem the FTL database isn't open, neither on disk nor in memory, while dnsmasq_config.c is running.

DL6ER:
I don't think we need a message table entry about this. However, I do agree with @yubiuser's concern about users wanting to intentionally disable the cache but still using DNSSEC. In case of `dns.cache.size = 0` but DNSSEC enabled, there are three options:

1. Silently disable DNSSEC *although* user doesn't want this,
2. Silently re-enable DNS caching *although* user doesn't want this, or
3. Fail hard (current behavior) so user has no other way than noticing something they want isn't possible.

I have to say that I'm in favor of option 3. I don't really think this PR is really needed (no hard feelings intended) but I do see it being useful when users want to set it to, say, 50, which does not make much sense in the first place. The DNS cache is by far not the largest memory consumption part of FTL.

rrobgill:
> Fail hard (current behavior) so user has no other way than noticing something they want isn't possible.

All good.

DL6ER:
We could also amend option 3 by adding some custom config validation functions for `dns.cache.size` and `dns.dnssec`. They can cross-check each other options, rejecting enabling DNSSEC or lowering the DNS cache size when the other option would result in an incompatible configuration. Thereby, we should be able to ensure such configuration is rejected early enough to never* get into the state of such incompatible setups.

*) We can not really "protect" against users directly modifying `pihole.toml` on-disk, because there is not really something we can revert to. Otherwise, we'd again end up in no. 1 or 2 above which is (I think) undesirable. Nonetheless, such custom validators would guard changing those settings (even at the same time) through the API or the CLI with the possibility to show a clear human-readable error **directly** to the user (and not only bury it in the log files).

rrobgill:
Yeah, that gets out of hand fairly quickly if I think about every possible interaction between different settings.

PR #2540, opened 2025-06-27
Add `pihole-FTL create-default-config` option and use it to upload `pihole.toml` to `ftl.pi-hole.net`

yubiuser:

Add new undocumented flag `pihole-FTL create-default-config <file.toml>` which will generate a default config file without starting FTL. This `pihole.toml` is then uploaded during the build process to ftl.pi-hole.net for later use (e.g. in https://github.com/pi-hole/docs/pull/1215)

Supersedes https://github.com/pi-hole/FTL/pull/2537 

---

---

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -610,6 +610,18 @@ void parse_args(int argc, char *argv[])
 		exit(migrate_config_v6() ? EXIT_SUCCESS : EXIT_FAILURE);
 	}
 
+	// Undocumented option to create an all-default dummy config file
+	if(argc == 3 && strcmp(argv[1], "create-default-config") == 0)
+	{
+		// Enable stdout printing
+		cli_mode = true;
+		log_ctrl(false, true);
+		if(create_default_config(argv[2]))
+			exit(EXIT_SUCCESS);
+		else
+			exit(EXIT_FAILURE);
+	}
+
 	// start from 1, as argv[0] is the executable name
 	for(int i = 1; i < argc; i++)
 	{
--- a/src/args.c
+++ b/src/args.c
@@ -616,6 +616,15 @@ void parse_args(int argc, char *argv[])
 		// Enable stdout printing
 		cli_mode = true;
 		log_ctrl(false, true);
+
+		// Validate the output filename
+		if(strstr(argv[2], "..") || strchr(argv[2], '/') || strchr(argv[2], '\\'))
+		{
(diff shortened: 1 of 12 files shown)


yubiuser:
Successfully uploaded to https://ftl.pi-hole.net/new/gen_config/

DL6ER (dismissed):

DL6ER:
Just thinking out aloud here: Now that we upload the `pihole.toml` file from various branches, it may be useful to include the branch (`if != master`) in the header of the file as well

DL6ER (approved):

src/args.c
```diff
@@ -616,6 +616,15 @@ void parse_args(int argc, char *argv[])
 		// Enable stdout printing
 		cli_mode = true;
 		log_ctrl(false, true);
+
+		// Validate the output filename
+		if(strstr(argv[2], "..") || strchr(argv[2], '/') || strchr(argv[2], '\\'))
```
yubiuser, line 621: Does this mean, the `/` character is not allowed?

DL6ER: Yes. Whenever I remove this, Code QL starts grumbling about (positive) path-traversal. If you think we need this, I can re-allow it and we silence the CodeQL "issue report".

yubiuser: It depends. When we really only ever want to use this internally during CI it's fine. But I fear users will find this and try to do `pihole-FTL create-default-config /home/pi/subdir/toml.conf`

DL6ER: I think we should relax this only when someone actually comes and requests this.

rdwebdesign:
@yubiuser 

The file upload is not needed anymore. FTL#2558 already handles the docs update.

I'm not sure if we should keep the [file generation and upload](https://github.com/pi-hole/FTL/pull/2540/commits/cd9ee5b07e9549eaf7d99a312d0e9d1bb83a633d) or if we remove them.
Will this improve the build speed?

rdwebdesign:
I have the same doubt. 
I'm not sure the file upload is a problem.
I think we should remove it only if we will improve the build process. If not, leave it as it is.

DL6ER:
It is also "archived" like here: https://ftl.pi-hole.net/v6.3/pihole.toml

This may turn out useful when users want to check in which version what change happened without wanting to navigate through `git blame` walking down prior versions and checking the related tag.

yubiuser:
Then we should keep it for now

PR #2553, opened 2025-07-03
Add python script to convert parse pihole.toml and produce a markdown…

PromoFaux:

Utilises some python code to convert the generated default `pihole.toml` file to a markdown formatted page to PR directly to the docs repo.

Keeping draft for now as it is currently configured to always create the PR while we iron out the wrinkles. Once we are satisfied I can add in the conditon that it only runs on master and then open the PR fully

---

---
- [x] I have read the above and my PR is ready for review. _Check this box to confirm_… formatted page to add to docs.pi-hole.net

Signed-off-by: Adam Warner <me@adamwarner.co.uk>

### DIFF
--- a/tools/pihole_toml_to_markdown.py
+++ b/tools/pihole_toml_to_markdown.py
@@ -1,0 +1,285 @@
+import re
+from pathlib import Path
+
+
+def parse_toml_with_comments(filepath):
+    """
+    Parse a TOML file with comments and generate a Markdown documentation string.
+
+    Args:
+        filepath (str or Path): Path to the TOML file.
+
+    Returns:
+        str: Markdown-formatted documentation.
+    """
+    with open(filepath, encoding="utf-8") as f:
+        lines = f.readlines()
+
+    # Start with a Markdown header and introductory documentation block
+    documentation = [
+        """<!-- markdownlint-disable MD033 -->
+# Pi-hole FTL Configuration Reference
+
+This page documents the available options of `pihole-FTL`. They are typically managed via the [TOML](https://toml.io/)-formatted configuration file `/etc/pihole/pihole.toml`. This file may be edited directly or you can use the command line (CLI) option, the web interface, the application programming interface (API) or environment variables.
+
+Using the web interface, the API or the CLI is preferred as they can do error checking for you, trying to prevent any incompatible options which could prevent FTL from starting on a severely broken configuration.
+
+To edit with the command line, use the format `key.name=value`, e.g:
+
+```text
+sudo pihole-FTL --config dns.dnssec=true
+```
+
+!!! note "Environment Variables"
+    **⚙️ Configuration Precedence**
+    Every Pi-hole setting in this file can be overridden using an environment variable.
+    This is especially common in Docker deployments.
+    Environment variable names follow the format:
+    ```text
+    FTLCONF_<section>_<key>
+    ```
+    For example:
+    ```text
+    FTLCONF_dns_upstreams
+    FTLCONF_database_DBimport
+    ```
+    ⚠️ **If a setting is defined via an environment variable, it becomes read-only.**
+    You will not be able to override it through the TOML file, the command line, or the web interface until the variable is removed from the environment.
+
+---
+<!-- markdownlint-disable-file MD034 -->
+"""
+    ]
+    section_stack = []
+    comment_buffer = []
+    in_config = False  # <-- New flag to skip file header comments
+
+    lines_iter = iter(lines)
+    for line in lines_iter:
+        stripped = line.strip()
+
+        # Handle section headers
+        if re.match(r"^\[.*\]$", stripped):
+            in_config = True  # <-- Start processing comments now
+            section_stack = [stripped.strip("[]")]
+            documentation.append(f"\n## `[{'.'.join(section_stack)}]`\n")
+            continue
+        
+        # If we are in a config section, start buffering comments
+        elif stripped.startswith("#"):
+            if in_config:
+                comment_buffer.append(stripped.lstrip("#").strip())
+            continue
+
+        # Handle key-value pairs
+        elif "=" in stripped and in_config:
+            key, value = map(str.strip, stripped.split("=", 1))
+            value_lines = [value]
+
+            # Check if value is a multi-line array
+            if value.startswith("[") and not value.endswith("]"):
+                # Multi-line array
+                while not value_lines[-1].strip().endswith("]"):
+                    next_line = next(lines_iter).rstrip("\n")
+                    value_lines.append(next_line)
+                value = "\n".join(value_lines)
+
+            else:
+                value = value_lines[0]
+                
+            documentation.append(f"### `{key}`\n")
+
+            # Process the comments collected for this key
+            if comment_buffer:
+                adjusted_comments = []
+
+                i = 0
+                while i < len(comment_buffer):
+                    line = comment_buffer[i]
+                    is_bullet = line.lstrip().startswith("-")
+                    prev_is_bullet = i > 0 and comment_buffer[i - 1].lstrip().startswith("-")
+                    prev_is_blank = i > 0 and comment_buffer[i - 1].strip() == ""
+
+                    # Fix malformed emphasis due to underscores or asterisks
+                    line = re.sub(r'\b(_[a-zA-Z0-9.-]+)', r'`\1`', line)
+                    line = re.sub(r'\*\.[a-zA-Z0-9]+', lambda m: f"`{m.group(0)}`", line)
+
+                    # Bold "Allowed values are:"
+                    line = re.sub(
+                        r'(^|\s)(Allowed values are:)', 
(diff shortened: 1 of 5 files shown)


PromoFaux:
This is now ready for review, and the additional protection of only running the pull request code on push to master has been added

An example of the PR that will be produced by this additonal action can be found here: https://github.com/pi-hole/docs/pull/1258 (Note: I think we should merge that one ahead of this code making it into `master` anyway - as it is useful)

tools/pihole_toml_to_markdown.py
```diff
@@ -0,0 +1,174 @@
+import re
+from pathlib import Path
+
+def parse_toml_with_comments(filepath):
+    with open(filepath, encoding='utf-8') as f:
+        lines = f.readlines()
+
+    documentation = ["""---
+description: This file is automatically generated when new code is pushed to the master branch of the FTL Repo. Any proposed changes should be made against https://github.com/pi-hole/ftl/tools/pihole_toml_to_markdown.py
+---
```
DL6ER: This string will be used by link-preview generators, e.g.
image

I don't think this is a good fit and should instead just say that we always show the config options of the most recent FTL release here. This sentence here could go towards the end of the page, it is not important and probably nobody would file a PR against the documentation *first*.

tools/pihole_toml_to_markdown.py
```diff
@@ -0,0 +1,174 @@
...
+
+# Pi-hole FTL Configuration Reference
+
+This page documents the available options in the `pihole-FTL.conf` file, which is typically managed via the TOML-formatted configuration file:
+
+```text\n/etc/pihole/pihole.toml\n```
+
+The file can be edited directly, however you can also use the command line option or the web interface.
```
DL6ER:
```suggestion
This page documents the available options of `pihole-FTL`. They are typically managed via the [TOML](https://toml.io/)-formatted configuration file `/etc/pihole/pihole.toml`. This file may be edited directly or you can use the command line (CLI) option, the web interface, the application programming interface (API) or environment variables.

Using the web interface, the API or the CLI is preferred as they can do error checking for you, trying to prevent any incompatible options which could prevent FTL from starting on a severely broken configuration.
```

tools/pihole_toml_to_markdown.py
```diff
@@ -0,0 +1,174 @@
...
+
+To edit with the command line, use the format `key.name=value`, e.g:
+
+```text\npihole-FTL --config dns.dnssec=true\n```
```
DL6ER:
```suggestion
```text\nsudo pihole-FTL --config dns.dnssec=true\n```
```
`sudo` will be needed unless you are either user `pihole` (or `root`)

tools/pihole_toml_to_markdown.py
```diff
@@ -0,0 +1,174 @@
...
+
+!!! note
```
DL6ER:
```suggestion
!!! note "Environment variables"
```

DL6ER (changes requested):
This looks good overall, I added a few review comments. In general, I think handling of examples is still missing, e.g.

Image

where I think `Example: ....` should be on their own line and bold-face/code-face the same way as the default value.

---

"Possible values" should stay on the same line when there is just one option (not an list), an example of this not happening can be seen in the screenshot just above.

---

And last nitpick: `[dns]` should be in ``` as well.

tools/pihole_toml_to_markdown.py
PromoFaux: Cool, I'll do this one in a separate commit, because there is another place

tools/pihole_toml_to_markdown.py
rdwebdesign: This can be changed to a Note or just a paragraph, at the end of the file.

PromoFaux:
> I think handling of examples is still missing, e.g.

The difficulty here is that the examples and possible values formatting is not always consistent in the source material. Ideally I don't want to add in special handling for each section - but maybe that's what we have to do

rdwebdesign:
I noticed another thing...

When "possible value" are a list, the option could be on a separate line. The source code is on a different line, but markdown is putting everything on the same line:

image

I think we could add a `<br>` to look like this:

image

DL6ER:
> The difficulty here is that the examples and possible values formatting is not always consistent in the source material. Ideally I don't want to add in special handling for each section - but maybe that's what we have to do

We can also make it more consistent in the FTL config source code generating this. I just don't know what/where you'd like to see changes

rdwebdesign:
I really like the table format for these options.

I can't suggest a good solution for the text wrapping and column width issue because you didn't commit the changes and I can't see the issue. 
I think we can force line breaks inside the table cells to look better (the text generated by FTL already have line breaks, but markdown/HTML won't show them unless we replace them if `<br>`).

PromoFaux:
For the wrapping issue, we can do the tables as HTML and just stick a thing in to tell the linter to ignore the "no embedded HTML" rule

rdwebdesign:
I think we can try to replace `\n` inside the cells with `<br>`, even using markdown tables. 
Most markdown flavors allow `<br>` inside markdown tables (I think this will work here).

PromoFaux:
We can squash a bunch of these commits down once we've finished - otherwise it's going to get silly.

rdwebdesign:
Where did the comments go?

PromoFaux:
See the commit message...

PromoFaux:
The angled brackets bit needs to be a little smarter

image

PromoFaux:
Likewise these sections are not correct 

image


from:

```toml
    # Array of clients to be excluded from certain API responses (regex):
    # - Query Log (/api/queries)
    # - Top Clients (/api/stats/top_clients)
    # This setting accepts both IP addresses (IPv4 and IPv6) as well as hostnames.
    # Note that backslashes "\" need to be escaped, i.e. "\\" in this setting
    #
    # Example: [ "^192\\.168\\.2\\.56$", "^fe80::341:[0-9a-f]*$", "^localhost$" ]
    #
    # Possible values are:
    #     array of regular expressions describing clients
    excludeClients = []

    # Array of domains to be excluded from certain API responses (regex):
    # - Query Log (/api/queries)
    # - Top Clients (/api/stats/top_domains)
    # Note that backslashes "\" need to be escaped, i.e. "\\" in this setting
    #
    # Example: [ "(^|\\.)\\.google\\.de$", "\\.pi-hole\\.net$" ]
    #
    # Possible values are:
    #     array of regular expressions describing domains
    excludeDomains = []
```

rdwebdesign:
Hmm... It was late and I didn't see all that mess. We need to revert my commits.

We can try to convert bullet points to tables only after the text "Possible values are:", but I'm not sure how to do it.

Also, we need to avoid replacing `<` and `>` if they are inside code blocks, but replace all others.

tools/pihole_toml_to_markdown.py
```diff
@@ -0,0 +1,192 @@
...
+            section_stack = [stripped.strip('[]')]
+            documentation.append(f"## [{'.'.join(section_stack)}]\n")
+            continue
+
+        # Array of tables: [[section]]
+        elif re.match(r'^\[\[.*\]\]$', stripped):
+            in_config = True
+            section_stack = [stripped.strip('[]')]
+            documentation.append(f"## [{'.'.join(section_stack)}]\n")
+            continue
```
rdwebdesign: This block is never used.

Our TOML file doesn't use sections with double brackets (`[[section]]`).

dschaper:
Thank you!!

tools/pihole_toml_to_markdown.py
DL6ER: Double square brackets isn't a thing in TOML at all for *sections*

tools/pihole_toml_to_markdown.py
```diff
@@ -0,0 +1,229 @@
...
+                    i += 1
+
+                documentation.append(escape_angle_brackets_outside_backticks("\n".join(adjusted_comments)))
+                documentation.append("")  # spacer after comment block
+            # Format default value for Markdown
+            if "\n" in value:
+                documentation.append("**Default value:**")
+                documentation.append("```toml")
+                documentation.append(f"{value}")
+                documentation.append("```")
```
rdwebdesign:
```suggestion
            if "\n" in value:
                documentation.append("**Default value:**")
                documentation.append("")
                documentation.append("```toml")
                documentation.append(f"{value}")
                documentation.append("```")
                documentation.append("")
```

rdwebdesign (changes requested):

tools/pihole_toml_to_markdown.py
PromoFaux: Changed locally, will be reflected in the next FP

tools/pihole_toml_to_markdown.py
PromoFaux: I removed the block - but I did see on stackexchange somewhre (lost the link now) that apparently it is valid syntax.

PR #2569, opened 2025-07-15
NTP ipv6 crash fix - ntp reply & logging

rrobgill:

Avoid the buffer overflow when repying to ipv6 ntp packets.

This was observed to cause crashes on 32 bit systems #2567.

Use SOCKADDR_STORAGE struct to hold address (guaranteed sufficient for ipv4 or ipv6 address)

Uses pointer casting to avoid potential buffer overflow copying socket addresses into socket structures.

Together this should be safer against buffer overflows and alignment issues on 32 bit systems

Adds more detailed ntp debug logging.

N/A

---

---

### DIFF
--- a/src/ntp/server.c
+++ b/src/ntp/server.c
@@ -238,9 +238,16 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
 	while (true)
 	{
 		unsigned char buf[48];
-		struct sockaddr src_addr;
+		struct sockaddr_storage src_addr;
 		socklen_t src_addrlen = sizeof(src_addr);
-		while(recvfrom(fd, buf, sizeof(buf), 0, &src_addr, &src_addrlen) < 48);  // ignore invalid requests
+
+		// Receive packet
+		ssize_t received = recvfrom(fd, buf, sizeof(buf), 0, (struct sockaddr *)&src_addr, &src_addrlen);
+		if(received != 48)
+		{
+			log_debug(DEBUG_NTP, "Received invalid NTP packet size: %zd", received);
+			continue;
+		}
 
 		// Get the current time in NTP format directly after receiving
 		// the request
@@ -249,31 +256,36 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
 		// Print the request
 		if(config.debug.ntp.v.b)
 		{
-			if(protocol == AF_INET6)
+			if (src_addr.ss_family == AF_INET6)
 			{
-				struct sockaddr_in6 sin6;
-				memcpy(&sin6, &src_addr, sizeof(sin6));
-
-				char ip[INET6_ADDRSTRLEN];
-				const in_port_t port = ntohs(sin6.sin6_port);
-				inet_ntop(protocol, &sin6.sin6_addr, ip, sizeof(ip));
+				char ip[INET6_ADDRSTRLEN] = { 0 };
+				struct sockaddr_in6 *sin6 = (struct sockaddr_in6 *)&src_addr;
+				inet_ntop(protocol, &sin6->sin6_addr, ip, sizeof(ip));
+				const in_port_t port = ntohs(sin6->sin6_port);
 				log_debug(DEBUG_NTP, "Received NTP request from [%s]:%u", ip, port);
+			}
+			else if (src_addr.ss_family == AF_INET)
+			{
+				char ip[INET_ADDRSTRLEN] = { 0 };
+				struct sockaddr_in *sin = (struct sockaddr_in *)&src_addr;
+				inet_ntop(protocol, &sin->sin_addr, ip, sizeof(ip));
+				const in_port_t port = ntohs(sin->sin_port);
+				log_debug(DEBUG_NTP, "Received NTP request from %s:%u", ip, port);
 			}
 			else
 			{
-				struct sockaddr_in sin;
-				memcpy(&sin, &src_addr, sizeof(sin));
-
-				char ip[INET6_ADDRSTRLEN];
-				const in_port_t port = ntohs(sin.sin_port);
-				inet_ntop(protocol, &sin.sin_addr, ip, sizeof(ip));
-				log_debug(DEBUG_NTP, "Received NTP request from %s:%u", ip, port);
+				// I don't think this should ever happen
+				log_debug(DEBUG_NTP, "Received NTP request with malformed address family");
+				continue;
 			}
 		}
 
 		// Handle the request
-		ntp_reply(fd, &src_addr, src_addrlen, buf, &recv_time);
-		log_debug(DEBUG_NTP, "NTP reply sent");
+		if (!ntp_reply(fd, (struct sockaddr *)&src_addr, src_addrlen, buf, &recv_time)) {
+			log_debug(DEBUG_NTP, "Failed to send NTP reply");
+		} else {
+			log_debug(DEBUG_NTP, "NTP reply sent");
+		}
 
 		// Sleep for 100 msec, this allows no more than 10 requests per
 		// second


rrobgill:
There is a one-liner from [https://seriot.ch/projects/tiny_ntp_client.html](https://seriot.ch/projects/tiny_ntp_client.html) that doesn't rely on any ntp package being installed.

To test on localhost,
ipv6:
date -d@$((0x`printf c%47s|nc -6 -uw1 ::1 123|xxd -s40 -l4 -p`-64#23GDW0))
ipv4:
date -d@$((0x`printf c%47s|nc -uw1 127.0.0.1 123|xxd -s40 -l4 -p`-64#23GDW0))

If it gets a response it prints the time and date reported by the ntp server.

I was thinking something like this may be possible to add to the tests, a quick sanity check that the returned time is within 2 seconds either way of the sent time (it should be the same on localhost, but this allows for second rollover and a little extra padding.

Is Pi-hole able to sync its own time within the test environment to even be able to send a response?

```
@test "NTP IPv4" {
SEND=$(date +%Y%m%d%H%M%S)
RECV=$(date -d@$((0x`printf c%47s|nc -6 -uw1 ::1 123|xxd -s40 -l4 -p`-64#23GDW0)) +%Y%m%d%H%M%S)
DIFF="$(($RECV - SEND))"
[[ "$DIFF" -ge -2 && "$DIFF" -le 2 ]]
}

@test "NTP IPv4" {
SEND=$(date +%Y%m%d%H%M%S)
RECV=$(date -d@$((0x`printf c%47s|nc -4 -uw1 192.168.1.1 123|xxd -s40 -l4 -p`-64#23GDW0)) +%Y%m%d%H%M%S)
DIFF="$(($RECV - SEND))"
[[ "$DIFF" -ge -2 && "$DIFF" -le 2 ]]
}

**Comment by rrobgill** [comment]
> does not return anything for me on neither my v86_64 nor the riscv64 Pi-hole so the one-liner doesn't seem to work.

Sorry I hadn't notices but the one-liners got mangled by github's formatting, it clobbered the backticks. 

```
date -d@$((0x`printf c%47s|nc -6 -uw1 ::1 123|xxd -s40 -l4 -p`-64#23GDW0))
date -d@$((0x`printf c%47s|nc -uw1 127.0.0.1 123|xxd -s40 -l4 -p`-64#23GDW0))
```

**Comment by DL6ER** [comment]
Thank you. With
```
date -d@$((0x`printf c%47s|nc -6 -uw1 ::1 123|xxd -s40 -l4 -p`-64#23GDW0))
```
I don't see any crashes, also not on RISCV, but, as said, this is a 64 bit system so the crash may not actually be expected.

File: src/ntp/server.c [hunk 1]
```diff
@@ -238,9 +238,15 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
...
-		struct sockaddr src_addr;
+		struct sockaddr_storage src_addr;
 		socklen_t src_addrlen = sizeof(src_addr);
-		while(recvfrom(fd, buf, sizeof(buf), 0, &src_addr, &src_addrlen) < 48);  // ignore invalid requests
+
+		// Receive packet
+		ssize_t received = recvfrom(fd, buf, sizeof(buf), 0, (struct sockaddr *)&src_addr, &src_addrlen);
+		if (received != 48) {
+			log_debug(DEBUG_NTP, "Received invalid NTP packet size: %zd", received);
+		continue;
```
DL6ER:
```suggestion
			continue;
```

src/ntp/server.c
```diff
@@ -238,9 +238,15 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
...
 	{
 		unsigned char buf[48];
-		struct sockaddr src_addr;
+		struct sockaddr_storage src_addr;
 		socklen_t src_addrlen = sizeof(src_addr);
-		while(recvfrom(fd, buf, sizeof(buf), 0, &src_addr, &src_addrlen) < 48);  // ignore invalid requests
+
+		// Receive packet
+		ssize_t received = recvfrom(fd, buf, sizeof(buf), 0, (struct sockaddr *)&src_addr, &src_addrlen);
+		if (received != 48) {
```
DL6ER:
```suggestion
		if(received != 48)
		{
```

src/ntp/server.c
```diff
@@ -238,9 +238,15 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
 	while (true)
 	{
 		unsigned char buf[48];
-		struct sockaddr src_addr;
+		struct sockaddr_storage src_addr;
```
DL6ER, line 241: Using `struct sockaddr_storage` seems to be the right thing here. Interesting that the `man recvfrom` didn't mention this when I looked at it while implementing this. Anyway, [socket.h description](https://pubs.opengroup.org/onlinepubs/009696799/basedefs/sys/socket.h.html) says:

> The header shall define the `sockaddr_storage` structure. This structure shall be:
>
> - Large enough to accommodate all supported protocol-specific address structures
> - Aligned at an appropriate boundary so that pointers to it can be cast as pointers to protocol-specific address structures and used to access the fields of those structures without alignment problems

src/ntp/server.c
```diff
@@ -249,31 +255,37 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
 		// Print the request
 		if(config.debug.ntp.v.b)
 		{
-			if(protocol == AF_INET6)
-			{
-				struct sockaddr_in6 sin6;
-				memcpy(&sin6, &src_addr, sizeof(sin6));
+			in_port_t port = 0;
 
+			if (protocol == AF_INET6 && src_addrlen >= sizeof(struct sockaddr_in6))
```
DL6ER: As an alternative, it seems cleaner to - instead checking if the address length is `>=` - if the family matches, i.e.,

```suggestion
			if (protocol == AF_INET6 && src_addr.ss_family == AF_INET_6)
```
I am not even sure if we need the `protocol` check here, in the end what matters is what is received on the socket, not what the socket is "meant to be" (I can put only one suggestion here:

``` c
if (src_addr.ss_family == AF_INET_6)
```

src/ntp/server.c
```diff
@@ -249,31 +255,37 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
 		// Print the request
 		if(config.debug.ntp.v.b)
 		{
-			if(protocol == AF_INET6)
-			{
-				struct sockaddr_in6 sin6;
-				memcpy(&sin6, &src_addr, sizeof(sin6));
+			in_port_t port = 0;
```
DL6ER: Why did you reduce the scope of `port` defining it deeper nested with `const` prefix seems superior.

src/ntp/server.c
```diff
@@ -249,31 +255,37 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
...
 				char ip[INET6_ADDRSTRLEN];
-				const in_port_t port = ntohs(sin6.sin6_port);
-				inet_ntop(protocol, &sin6.sin6_addr, ip, sizeof(ip));
+				struct sockaddr_in6 *sin6 = (struct sockaddr_in6 *)&src_addr;
+				inet_ntop(protocol, &sin6->sin6_addr, ip, sizeof(ip));
+				port = ntohs(sin6->sin6_port);
 				log_debug(DEBUG_NTP, "Received NTP request from [%s]:%u", ip, port);
 			}
-			else
+			else if (protocol == AF_INET && src_addrlen >= sizeof(struct sockaddr_in))
```
DL6ER: See comment above re using `src_addr.ss_family` instead

src/ntp/server.c
```diff
@@ -249,31 +255,37 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
...
 		{
-			if(protocol == AF_INET6)
-			{
-				struct sockaddr_in6 sin6;
-				memcpy(&sin6, &src_addr, sizeof(sin6));
+			in_port_t port = 0;
 
+			if (protocol == AF_INET6 && src_addrlen >= sizeof(struct sockaddr_in6))
+			{
 				char ip[INET6_ADDRSTRLEN];
```
DL6ER:
```suggestion
				char ip[INET6_ADDRSTRLEN] = { 0 };
```
just for safety in case `inet_ntop` fails below (I don't think it ever will at this point, but anyway)

src/ntp/server.c
```diff
@@ -249,31 +255,37 @@ static void request_process_loop(const int fd, const char *ipstr, const int prot
...
-			else
+			else if (protocol == AF_INET && src_addrlen >= sizeof(struct sockaddr_in))
 			{
-				struct sockaddr_in sin;
-				memcpy(&sin, &src_addr, sizeof(sin));
-
-				char ip[INET6_ADDRSTRLEN];
-				const in_port_t port = ntohs(sin.sin_port);
-				inet_ntop(protocol, &sin.sin_addr, ip, sizeof(ip));
+				char ip[INET_ADDRSTRLEN];
```
DL6ER:
```suggestion
				char ip[INET_ADDRSTRLEN] = { 0 };
```

DL6ER (approved):

PR #2542, opened 2025-06-28
Remove `'unsafe-inline'` from default-src

XhmikosR:
It's only needed in style-src for chart.js

## Thank you for your contribution to the Pi-hole Community!

Please read the comments below to help us consider your Pull Request.

We are all volunteers and completing the process outlined will help us review your commits quicker.

**Please make sure you**

 1. Base your code and PRs against the repositories developmental branch.
 2. [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits as we enforce the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
 3. [Sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all your commits as they must have verified signatures
 4. File a pull request for any change that requires changes to [our documentation](https://docs.pi-hole.net/) at our [documentation repo](https://github.com/pi-hole/docs) 

---

**What does this PR aim to accomplish?:**

**How does this PR accomplish the above?:**

**Link documentation PRs if any are needed to support this PR:**

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [ ] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #2575, opened 2025-07-18
Improve default CSP headers

DL6ER:

This PR picks up where #2542 left. I imported the branch and checked everything is still working as expected. As usual, a second pair of eyes won't harm.

Most importantly:Modern CSP best practices recommend minimizing the use of `unsafe-inline` and only allowing it where strictly necessary (e.g., for styles). Accordingly, `unsafe-inline` is now only allowed for **styles** (because Chart.js needs it), **no longer for scripts**. This reduces the risk of cross-site scripting (XSS) attacks. As we have seen in the past, inline scripts are a common attack vector.

Other CSP parts:
- `style-src 'self' 'unsafe-inline'`: allows styles both from files and inline in the code/on the HTML pages.
- `img-src 'self' data:` allows images from the same origin and embedded images via data URIs, but blocks external image sources, reducing exposure to malicious content. This is merely a precaution while we are adding more granular control, I don't think we have any external content right now.

A few other things, such as removing `minimum-scale=1,user-scalable=yes` is done as these two are their respective defaults, anyway.

---

---

## Checklist:

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,13 +9,15 @@
         "ms-vscode.cpptools-extension-pack",
         "eamodio.gitlens",
         "github.copilot",
-        "ms-python.python"
+        "ms-python.python",
+        "ms-python.black-formatter"
       ]
     }
   },
   "mounts": [
     "type=bind,source=${localEnv:HOME}/.ssh,target=/root/.ssh,readonly",
-    "type=bind,source=/var/www/html/admin,target=/var/www/html/admin,readonly"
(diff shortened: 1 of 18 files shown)


rdwebdesign:
@DL6ER 

The 3rd commit mentioned above is `https://github.com/pi-hole/FTL/pull/2542/commits/7377e08fe9b94c5d6d1379ca72666c560d563e8e` and removes the request for the external font files.

The current CSP is already blocking the font request:

image

but there are errors: 

image

What do you want to do?
Remove the font completely _(using the code form the other commit)_, or fix the previous CSP and allow to load the external font?

DL6ER:
I don't want external resources to be loaded

PR #2593, opened 2025-08-06
Add autocomplete feature

DL6ER:

This PR adds bash-like completion support for most relevant subcommands of `pihole-FTL`. The completion algorithm can suggest a lot of things including completing along the entire `--config` path, e.g.

```
   pihole-FTL --co <Tab>
-> pihole-FTL --config

   pihole-FTL --config de <Tab>
-> pihole-FTL --config debug.

   another <Tab>
   debug.aliasclients  debug.clients       debug.events        debug.inotify       debug.overtime      debug.shmem         
   debug.all           debug.config        debug.extra         debug.locks         debug.queries       debug.status        
   debug.api           debug.database      debug.flags         debug.netlink       debug.regex         debug.tls           
   debug.arp           debug.dnssec        debug.gc            debug.networking    debug.reserved      debug.vectors       
   debug.caps          debug.edns0         debug.helper        debug.ntp           debug.resolver      debug.webserver

   pihole-FTL --config debug.net <Tab>
-> pihole-FTL --config debug.network

   pihole-FTL --config debug.network t <Tab>
-> pihole-FTL --config debug.network true
```
---

It can also suggest the correct `enum` values, e.g.
```
   pihole-FTL --config dns.listeningMode <Tab>
   all     bind    local   none    single

   pihole-FTL --config dns.listeningMode a<Tab>
-> pihole-FTL --config dns.listeningMode all
```

For review and testing:  Check out `pihole-FTL` on branch `new/autocomplete` and run in your terminal:
```
_complete_FTL() { COMPREPLY=($(pihole-FTL --complete "${COMP_WORDS[@]}")); }
complete -F _complete_FTL pihole-FTL
```
Thereafter, completion should work.
---

---

## Checklist:

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -87,6 +87,9 @@ bool daemonmode = true, cli_mode = false;
 int argc_dnsmasq = 0;
 const char** argv_dnsmasq = NULL;
 
+// Prototypes
+static void suggest_complete(const int argc, char *argv[]);
+
 static bool __attribute__ ((pure)) is_term(void)
 {
 	// test whether STDOUT refers to a terminal or if env variable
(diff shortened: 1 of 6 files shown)


sawsanders:
I think having the default value there is the way to go. You can always see what the current value is by leaving the value field blank and FTL will display the current setting.

BUT... there is value in quickly editing large lists, like `dns.hosts` and `misc.dnsmasq_lines` where you could `tab` and the current values would appear, allowing you to quickly edit or add to the string...  I suppose you could accomplish the same thing by viewing the current value from the CLI, then copy and pasting. 

There's no way to make everyone happy, but I'm happy with how you have it!

src/args.c
```diff
@@ -1216,3 +1267,286 @@ void test_dnsmasq_options(int argc, const char *argv[])
...
+			"--teleporter", "--gen-x509", "--read-x509",
+			"--read-x509-key", "--list-dhcp4", "--list-dhcp6",
+			"gravity", "ntp", "gzip", "dhcp-discover",
+			"arp-scan", "idn2", "sha256sum", "verify",
+			"--default-gateway", "--totp", "--perf",
+			"verify", "ptr", "sqlite3_rsync",
+			"sha256sum", "lua", "--lua",
+			"luac", "--luac", "debug", "test",
+			"-f", "no-daemon", "--gzip"
+		};
```
rdwebdesign: I think we should include these options: `-v`, `-vv` and `-h`.

```suggestion
		const char *options[] = {
			"version", "-v", "-vv", "tag", "branch", "help", "-h", "dnsmasq-test",
			"regex-test", "lua", "sqlite3", "--config",
			"--teleporter", "--gen-x509", "--read-x509",
			"--read-x509-key", "--list-dhcp4", "--list-dhcp6",
			"gravity", "ntp", "gzip", "dhcp-discover",
			"arp-scan", "idn2", "sha256sum", "verify",
			"--default-gateway", "--totp", "--perf",
			"verify", "ptr", "sqlite3_rsync",
			"sha256sum", "lua", "--lua",
			"luac", "--luac", "debug", "test",
			"-f", "no-daemon", "--gzip"
		};
```

Also, maybe we should sort the entries. 
It will make a lot easier to find out which options we already added.

DL6ER: Done

src/args.c
```diff
@@ -1216,3 +1267,281 @@ void test_dnsmasq_options(int argc, const char *argv[])
...
+
+		// Provide matching suggestions
+		list_matches(last_word, options, ArraySize(options), true);
+	}
+	else if(argc == 5 && strEndsWith(argv[3], "sqlite3"))
+	{
+		// pihole-FTL sqlite3 ...
+		const char *options[] = {
+			"-h", "-ni"
```
yubiuser, line 1358: Nitpicking here, but both options could be used simultaneous.

src/args.c
```diff
@@ -1216,3 +1267,281 @@ void test_dnsmasq_options(int argc, const char *argv[])
...
+						};
+
+						// Provide matching suggestions
+						list_matches(last_word, options, ArraySize(options), false);
+					}
+					else if(conf_item->t == CONF_ENUM_BLOCKING_MODE)
+					{
+						// pihole-FTL --config dns.blocking.mode ...
+						const char *options[] = {
+							"IP", "NX", "NULL", "IP_NODATA_AAAA", "NODATA"
```
yubiuser: Does it make sense to list all value options here (as well in the other config keys) instead of getting them from the api source code where they are already defined? Like you did this with the theme?

src/args.c
DL6ER: Done, I prepared the code to complete the remainder if any of the two is already used.

src/args.c
DL6ER: This was not done because in most parts the corresponding `..._MAX` value was not defined. I added them now.

PR #2604, opened 2025-08-13
Update CI branch

yubiuser:

Updates the `CI` branch to latest development to fix the RISCV tests at the `docker-base` repo 

P.s. Why don't we use the `development` branch as target for the docker-base tests but this special branch?

---

---


DL6ER:
The `special/CI_development` is used by the containers to build the binary. It may be necessary to install additional packages or modify the build process when upgrading the base-images. The issue here is that, most often, these changes do *not* work in the old containers but are required in the new so, if we'd be using `development` here, the containers would not build but we also couldn't commit the required changes as this would cause the mandatory tests for a `development` build to fail.

DL6ER (approved):

yubiuser:
I see chicken-egg-issue

PR #2609, opened 2025-08-17
Fix default value autocomplete suggestions

DL6ER:

Currently, `pihole-FTL --config dns.domain <Tab>` leads to `pihole-FTL --config dns.domain '"lan"'` which is incorrect due to the extra single-quotes. This PR fixes that by ensuring single-quotes are not added for ordinary strings (as double-quotes are already added by `cJSON_PrintUnformatted()`), but only for JSON arrays (like `dns.upstreams`) where they are still needed, e.g., `--config webserver.headers`.

Hence, on this branch, the autocomplete suggestion above: `pihole-FTL --config dns.domain <Tab>` now correctly leads to `pihole-FTL --config dns.domain "lan"` (mind the missing extra `''`).

The second commit furthermore adds the special escaping needed for single quotes `'"'"'` as (`ba`)`sh` do not support escaping single quotes in the C-like style `\'`.

---

---

## Checklist:

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -1445,20 +1445,15 @@ void suggest_complete(const int argc, char *argv[])
 						if(val != NULL && (value = cJSON_PrintUnformatted(val)) != NULL)
 						{
 							// Add '' to the output if it is a string
-							if(conf_item->t == CONF_STRING ||
-							   conf_item->t == CONF_STRING_ALLOCATED ||
-							   conf_item->t == CONF_JSON_STRING_ARRAY)
+							if(conf_item->t == CONF_JSON_STRING_ARRAY &&
+							   value[0] != '\'')
 							{
-								// If the value is a string, we need to add quotes
-								if(value[0] != '\'')
+								char *tmp = calloc(strlen(value) + 3, sizeof(char));
+								if(tmp != NULL)
 								{
-									char *tmp = malloc(strlen(value) + 3);
-									if(tmp != NULL)
-									{
-										sprintf(tmp, "'%s'", value);
-										free(value);
-										value = tmp;
-									}
+									sprintf(tmp, "'%s'", value);
+									free(value);
+									value = tmp;
 								}
 							}
 
--- a/src/args.c
+++ b/src/args.c
@@ -1419,164 +1419,191 @@ void suggest_complete(const int argc, char *argv[])
 				if(strcmp(conf_item->k, argv[4]) == 0)
 				{
 					// See if we can suggest a value
-					if(conf_item->t == CONF_BOOL || conf_item->t == CONF_ALL_DEBUG_BOOL)
+					switch(conf_item->t)
 					{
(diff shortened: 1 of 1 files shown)


rdwebdesign:
Not sure if this is intended, but:

```
pihole-FTL --config misc.de <TAB>
pihole-FTL --config misc.delay_startup <TAB>
pihole-FTL --config misc.delay_startup 0    # the default value is correctly shown
```
But:
```
pihole-FTL --config misc.p <TAB>
pihole-FTL --config misc.privacylevel <TAB>
pihole-FTL --config misc.privacylevel        # no value is shown
```

DL6ER:
You can verify that the second commit is working as expected by running
```
sudo pihole-FTL --config webserver.headers <Tab>
```
getting
```
sudo pihole-FTL --config webserver.headers '["X-DNS-Prefetch-Control: off","Content-Security-Policy: default-src '"'"'self'"'"'; style-src '"'"'self'"'"' '"'"'unsafe-inline'"'"'; img-src '"'"'self'"'"' data:;","X-Frame-Options: DENY","X-XSS-Protection: 0","X-Content-Type-Options: nosniff","Referrer-Policy: strict-origin-when-cross-origin"]'
```
and submitting this. The fact that `/etc/pihole/pihole.toml` (line 943ff.) does **not** say "CHANGED" means that our escaping worked correctly.

rdwebdesign:
I just saw your new commit (you pushed it while I was typing). I will test again.

rdwebdesign:
I will wait for the build process, after the third commit... 🕙

rdwebdesign (dismissed):
```
piholetest:/# pihole-FTL -vv
****************************** FTL **********************************
Version:         vDev-a933e2d
...
```

- Quotes are fixed. 
- `misc.privacylevel` correctly returns the numeric values.

Approved, unless you want to accept the Copilot recommendations.

DL6ER:
This is actually an interesting problem. And one I have not thought about before. What you discovered seemed to me the right behavior when implementing this: Suggest the default value wherever it is (more or less) clear what needs to be added (a string, a number, ...). The ENUMs are just special. They do only allow discrete values which is why I made FTL suggest those. But we cannot auto-complete with the default setting at the same time.

Ideally, I would love see something like
```
pihole-FTL --config dns.listeningMode <TAB>
pihole-FTL --config dns.listeningMode <TAB> <TAB>
ALL     BIND    LOCAL (default)   NONE    SINGLE  
pihole-FTL --config dns.listeningMode
```
but when I return this to `complete`, it'd either suggest `(default)` as an option on its own or suggest the entire `LOCAL (default)` when you type `LO <TAB>`. Both options are not what we want.

Reading https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html and https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html doesn't immediately spark any new ideas here.

Any further ideas?

PR #2521, opened 2025-06-13
Webserver: Allow webhome to be root

rrobgill:
**What does this PR aim to accomplish?:**

Allow users to set webhome to root "/", to serve the admin interface on the bare domain.

Prevents the redirect loop noted by @averyvigolo in #2518.

**How does this PR accomplish the above?:**

Tests if webhome is root and if so does not redirect from root.

Logs a warning if the blockingmode is IP and webhome is root on webserver startup.

**Link documentation PRs if any are needed to support this PR:**

N/A

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [X] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #2610, opened 2025-08-24
Only use redirect_root_handler if webhome is set (fixes #2518)

### LINKED ISSUE #2518: Setting `webserver.paths.webhome` to `/` causes an infinite redirect when accessing pi.hole
### Versions

- Pi-hole: 6.1.2
- Web: 6.2.1
- FTL:  6.2.2

### Platform

- Platform:  x86-64 GNU/Linux

### Expected behavior

The pihole dashboard should be served correctly when accessing pi.hole (or whichever configured domain) if `webserver.paths.webhome` is set to `/`.

Or if this is not a supported configuration, setting this value should cause some error or fallback to the default value.

### Actual behavior / bug

Accessing `http://pi.hole:PORT/` causes an infinite redirect to `/`.

### Steps to reproduce

Set `webroot` directly to a checkout of the Pihole Web repo https://github.com/pi-hole/web, set `webhome` to `/`

```
[webserver.paths]
webhome = "/"
webroot = "/path/to/pihole-web-repo/"
```

Attempt to access `http://pi.hole:PORT/`. In a browser, you'll see an error like "this page is not behaving properly", and perhaps a hint about too many redirects.

This is obvious with `curl -Lv http://pi.hole:PORT/`, where you'll see an infinite (or until cURL gives up) redirect to `/`.

## Debug Token

https://tricorder.pi-hole.net/4W6Xbd3n/

[tricorder.log](https://github.com/user-attachments/files/20714207/tricorder.log)

### Additional context

This appears to be due to https://github.com/pi-hole/FTL/blob/7605481fb12184154069dcd0cb08e3fa2f83efa1/src/webserver/webserver.c#L157-L170

There's no check for whether `webpath` is `/` (or empty?), so there's no safeguard against an infinite redirect.
Is there a good reason to not support this behaviour? 

I'm packaging pihole for [Nixpkgs](https://github.com/NixOS/nixpkgs), where software is preferably not installed into the standard paths on the filesystem, but instead installed into independent paths in `/nix/store/`. In my initial configuration, I simply pointed `webroot` to the path to the Pihole web files and set `webpath` to `/` so it would be served from the root of the webserver.

However, a user mentioned they encountered this infinite redirect when accessing the web page via `pi.hole` only.

averyvigolo:

This fixes an infinite redirect on the home page when authentication is enabled. The redirects are caused by XHR calls to the API, which receive a 401 response, and the error handlers simply reload the page.

If webhome is not set, the default request handler should be used to properly
handle authentication. So, conditionally enable redirect_root_handler, if
webhome is not empty or `/`. This fixes the problem, as there's an immediate
redirect to /login, before any XHR calls.

Remove the initial fix in https://github.com/pi-hole/FTL/pull/2521, as it is no
longer necessary. That fix involved checking in redirect_root_handler, if the
redirect destination is the same as the request URI.

---

---


src/webserver/webserver.c
```diff
@@ -720,7 +714,14 @@ void http_init(void)
 	// prefix should be stripped away by the reverse proxy
 	mg_set_request_handler(ctx, "/api", api_handler, NULL);
 
-	mg_set_request_handler(ctx, "/$", redirect_root_handler, NULL);
+	if(strcmp(prefix_webhome, "/") == 0 || strlen(prefix_webhome) == 0)
```
DL6ER: Sorry for the review having taken so long... 

```suggestion
	if(strcmp(prefix_webhome, "/") == 0)
```

`prefix_webhome` should never be empty as `webserver.paths.webhome` has to go through `validate_filepath_two_slash` which rejects the setting if `strlen(val->s) < 1`:

image

Sure, this can be bypassed b via the TOML file, but the "acceptable values" section is pretty specific what is needed here:

> A valid subpath, **both slashes are needed!**

I'd rather go through and ensure `prefix_webhome` is not empty to avoid unintended side-effect in other locations. As said, this would anyway be considered an invalid configuration.

averyvigolo: Pushed your suggested change

> I'd rather go through and ensure prefix_webhome is not empty to avoid unintended side-effect in other locations. As said, this would anyway be considered an invalid configuration.

Just to clarify, does that mean you don't want to change `validate_filepath_two_slash`? That's fine by me, even if it's not obvious that just `/` is an acceptable value

DL6ER:
Sorry for this having fallen between the cracks, could you bring your commit onto the latest `development` so we can merge it? Sorry again for the extra trouble...

averyvigolo:
> Sorry for this having fallen between the cracks, could you bring your commit onto the latest `development` so we can merge it? Sorry again for the extra trouble...

Should be up to date now, thanks!
Not sure why it didn't let me do it through the web UI.

PR #2611, opened 2025-08-24
Support local API sockets for webserver requests

averyvigolo:

Support local socket files for webserver requests, and allow using it for API requests e.g. via the `pihole` command. Authentication for requests on a socket file is not necessary, as we can trust that basic file permissions will ensure only authorised users can write to the socket. 

This greatly simplifies the authentication flow for local API requests, and no temporary password is needed.

Use of Unix sockets is also very common when using a reverse proxy like nginx. Instead of binding to a regular port, you can use redirect to the socket file.

* enable Unix socket support in civetweb
* patch civetweb to pass through the socket path in request data, when the request comes in through a socket
* add a new option `webserver.api.private_socket`, to create and use the unauthenticated socket, and combine this path alongside `webserver.port` when initialising civetweb

- [ ] TODO: docs

---

---

### DIFF
--- a/test/pihole.toml
+++ b/test/pihole.toml
@@ -1107,6 +1107,15 @@
     #     true or false
     cli_pw = true
 
+    # Path to create a UNIX socket path, used by the pihole command for local API requests.
+    # This socket will be owned by the user running FTL (typically pihole). Note, if you
+    # change from the default value, give the path as an environment variable
+    # PIHOLE_API_SOCKET when running the pihole command.
+    #
+    # Possible values are:
+    #     <any writeable path without a comma>
+    private_socket = "/run/pihole-FTL.sock"
+
     # Array of clients to be excluded from certain API responses (regex):
     # - Query Log (/api/queries)
     # - Top Clients (/api/stats/top_clients)
@@ -1593,9 +1602,5 @@
   all = true ### CHANGED, default = false
 
 # Configuration statistics:
-# 156 total entries out of which 103 entries are default
-# --> 53 entries are modified
-# 3 entries are forced through environment:
-#   - misc.nice
-#   - misc.check.shmem
-#   - debug.api
+# 157 total entries out of which 103 entries are default
+# --> 57 entries are modified
--- a/test/run.sh
+++ b/test/run.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -xe
+
 # Skip tests on targets not supporting them
 if [[ ${TEST} == "false" ]]; then
   echo "Skipping tests (CI_ARCH: ${CI_ARCH})!"
@@ -56,7 +58,7 @@ cp test/pihole.toml /etc/pihole/pihole.toml
 chown pihole:pihole /etc/pihole/pihole.toml
 
 # Prepare 01-pihole-tests.conf
-mkdir /etc/dnsmasq.d
+mkdir -p /etc/dnsmasq.d
 cp test/01-pihole-tests.conf /etc/dnsmasq.d/01-pihole-tests.conf
 
 # Prepare versions file (read by /api/version)


test/pihole.toml
```diff
@@ -1593,9 +1602,5 @@
...
 
 # Configuration statistics:
-# 156 total entries out of which 103 entries are default
-# --> 53 entries are modified
-# 3 entries are forced through environment:
-#   - misc.nice
-#   - misc.check.shmem
-#   - debug.api
+# 157 total entries out of which 103 entries are default
+# --> 57 entries are modified
```
PromoFaux, line 1606:
```suggestion
# --> 53 entries are modified
# 3 entries are forced through environment:
#   - misc.nice
#   - misc.check.shmem
#   - debug.api
```

test/pihole.toml
```diff
@@ -1593,9 +1602,5 @@
   all = true ### CHANGED, default = false
 
 # Configuration statistics:
-# 156 total entries out of which 103 entries are default
-# --> 53 entries are modified
-# 3 entries are forced through environment:
-#   - misc.nice
-#   - misc.check.shmem
-#   - debug.api
+# 157 total entries out of which 103 entries are default
```
PromoFaux, line 1605:
```suggestion
# 157 total entries out of which 104 entries are default
```

test/pihole.toml
```diff
@@ -1107,6 +1107,15 @@
     #     true or false
     cli_pw = true
 
+    # Path to create a UNIX socket path, used by the pihole command for local API requests.
+    # This socket will be owned by the user running FTL (typically pihole). Note, if you
+    # change from the default value, give the path as an environment variable
+    # PIHOLE_API_SOCKET when running the pihole command.
+    #
+    # Possible values are:
```
PromoFaux, line 1115:
```suggestion
    # Allowed values are:
```

PromoFaux (changes requested):
Can't speak to any of the more technical changes, can at least help fix _one_ of the tests :)

```diff
```diff
diff --git a/test/pihole.toml b/test/pihole.toml
index c727ccf1..83fe8ff0 100644
--- a/test/pihole.toml
+++ b/test/pihole.toml
@@ -1112,7 +1112,7 @@
     # change from the default value, give the path as an environment variable
     # PIHOLE_API_SOCKET when running the pihole command.
     #
-    # Possible values are:
+    # Allowed values are:
     #     <any writeable path without a comma>
     private_socket = "/run/pihole-FTL.sock"
 
@@ -1602,5 +1602,9 @@
   all = true ### CHANGED, default = false
 
 # Configuration statistics:
-# 157 total entries out of which 103 entries are default
-# --> 57 entries are modified
+# 157 total entries out of which 104 entries are default
+# --> 53 entries are modified
+# 3 entries are forced through environment:
+#   - misc.nice
+#   - misc.check.shmem
+#   - debug.api
```
```

averyvigolo:
Thanks @PromoFaux. I just did a quick rebase on dev to push and open this to gauge interest, not surprised the results are slightly different from what I expected.

I did have a question on how you typically iterate on the tests locally. I tried out the Docker image in .github, but from what I recall, it clean rebuilds pihole-FTL each time from scratch, which makes the feedback loop pretty slow. I've hacked the dockerfile to build in the current directory so it can keep the build directory in between runs, but currently it fails at binding to port 53. Might be a podman issue tho, I'll try to keep hacking on it.

PromoFaux:
I like the sound the idea - but it will require an in-depth review from DL6ER.

> I did have a question on how you typically iterate on the tests locally.

If you use vscode, there is a `.devcontainer` configuration that gives you everything you need to build and test the binary without installing the dependencies. Then to test I simply run `CI_ARCH=linux/amd64 ./build.sh test` within that container

DL6ER:
Thank you for your contribution. I like the overall idea. However, I wonder why you consider it necessary patching CivetWeb to along pass the socket path in the request data. I see you are comparing it to check if the socket is the right one. However, I'd like to argue that there is vanishing likeliness that there is another. And, even, if, does it harm?

I'd rather add a real validator for `webserver.port` ensuring that the string does not contain any of `x` or `/` which are both not allowed in the specified format:

> A comma-separated list of `<[ip_address:]port>`

We can then be sure to have only one socket file.

PR #2615, opened 2025-08-27
Pin github actions to SHA

yubiuser:

Pins the actions used in our workflow by commit. This is the recommended way to prevent supply chain attacks

https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/

The human-friendly comment tag should be auto-updated by dependabot as well

https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/

____

This also updates some outdated actions to their latest version, no idea why dependabot did not pick them up

---

---


DL6ER:
Don't get me wrong, I am not against this in any way but - what does it give us in reality? I don't think the blog post told us this very important detail. Hypothetical scenario: Say some code is compromised. Okay, they can replace the tag and all workflows around the world will start using it. This is bad. Now, with SHA pinning, they cannot do this. Better. However, they can very well create a new release, `dependabot` will come and open update PRs throughout the field, but, yeah, we are slow picking them up and - if the compromise is detected (and communicated!) quickly enough - this may suffice to get the warning *before* we merge the change. If we are "too fast" (difficult to say, though) about this and are already on a malicious SHA, the pinning will keep us there, no way for the maintainers of the pirated repository to just delete the tag (I don't think you can really delete the SHA hashes, can you?). I'd give or take, this change can be positive and negative depending on your particular point of view.

yubiuser:
I see your point. When a bad actor releases a new compromised version and dependabot picks it up it's a kind of race condition - is the misbehavior noticed before we update the action? This is still a possible attack vector.
But as you said, another vector - re-tagging of a now compromised version - is not possible anymore.

So tagging by SHA will prevent 1 of 2 possible attack ways. Better than nothing I'd say.

PR #2620, opened 2025-09-02
Do not set domainname when the kernel replies with "(none)"

DL6ER:

See title.

---

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** https://discourse.pi-hole.net/t/host-name-of-client-203-0-113-41-dns-vm-1-none-contains-at-least-one-invalid-character-hex-28-at-position-10/82042

---

## Checklist:

### DIFF
--- a/src/daemon.c
+++ b/src/daemon.c
@@ -229,7 +229,13 @@ const char *hostname(void)
 		if(uname(&buf) == 0)
 		{
 			strncpy(nodename, buf.nodename, HOSTNAMESIZE);
-			strncpy(dname, buf.domainname, HOSTNAMESIZE);
+
+			// Only set domain name if node name is not empty: the
+			// kernel replies with '(none)' in this case. When the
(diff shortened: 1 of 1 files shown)


rdwebdesign:
I know that any domain name starting with or containing `(` is invalid, but normally FTL doesn't remove/ignore/sanitize any other invalid domains and simply generates a warning.

Maybe I'm being too "nitpicky", but I think we should check if the domain name is exactly equal to `(none)` and ignore the value only in that case. I consider `(none)` a special case, as it is an invalid value returned by the kernel.

In other cases like `(my_custom_domain)` or any other name starting with `(`, I think we should just warn that the domain name is invalid.

What do you think?

DL6ER:
But can you even set a domain with `()` in Linux?

rdwebdesign:
OK. 
I agree this is enough.

PR #2656, opened 2025-10-21
Disable flush_logs endpoint inside docker containers

rdwebdesign:

The flush process needs a FTL restart, but this is not possible inside docker containers (not enough permissions). Currently, if the API endpoint is called inside a container, it returns "success" and FTL logs success messages, but in reality FTL is never restarted and the process is not correctly completed.

This PR effectively disables the `flush_logs` endpoint when FTL is inside a container. Inside containers the API will respond with `403 forbidden`.

This is related to: https://github.com/pi-hole/pi-hole/pull/6391

FTL checks if it is running inside docker and, if so, returns a 403 error message.

---

---

### DIFF
--- a/src/api/action.c
+++ b/src/api/action.c
@@ -137,6 +137,30 @@ int api_action_restartDNS(struct ftl_conn *api)
 	return send_json_success(api);
 }
 
+// This function checks if a given PID is running inside a docker container
+static bool is_in_docker(const pid_t pid)
+{
+	char filename[sizeof("/proc/%u/cgroup") + sizeof(int)*3];
+	snprintf(filename, sizeof(filename), "/proc/%d/cgroup", pid);
+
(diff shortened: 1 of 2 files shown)


DL6ER:
FTL can restart just fine in `docker` as the binary is able to restart within itself without the need for an external helper like `systemd`. It doesn't exit, either, so the `docker` container should also not think it's dead. When we disable the ability to restart we also remove the ability to all the settings that do need a restart because `dnsmasq` configuration is changed.

rdwebdesign:
> (including if this change is needed at all)

Yes. I'm not sure if this is needed at all.

Apparently we need to change how Web handles some things and not try to use `pihole flush` command from web interface.

rdwebdesign:
Closing. 
Apparently this is unnecessary. It can be revisited, if needed.

PR #2704, opened 2025-11-09
Update embedded SQLite to 3.51.0

DL6ER:

We mostly benefit from no. 6 (general performance improvements) but also acknowledge 3 which offers us the ability to use `carrays` in the future. I have checked the existing source code if we can already benefit now, but this isn't the case.

For the curious (why "carrays"?): Currently, the `IN` operator can only use string inputs (`... IN (1,2,3)`), we cannot bind like (`... IN (?)`). With carrays, this is now possible like:

``` sql
SELECT * FROM table WHERE col IN carray(?123);
``` 
and then bind the values like
``` c
const int32_t data[] = {1, 2, 3};
sqlite3_carray_bind(stmt, 123, data, sizeof(data)/sizeof(data[0]), CARRAY_INT32, SQLITE_TRANSIENT);
```

---

Full changelog:

## SQLite Release 3.51.0 On 2025-11-04

1.  New macros in sqlite3.h:
    1.  SQLITE\_SCM\_BRANCH → the name of the branch from which the source code is taken.
    2.  SQLITE\_SCM\_TAGS → space-separated list of tags on the source code check-in.
    3.  SQLITE\_SCM\_DATETIME → ISO-8601 date and time of the source code check-in.
2.  Two new JSON functions, [jsonb\_each()](../json1.html#jbeach) and [jsonb\_tree()](../json1.html#jbtree) work the same as the existing json\_each() and json\_tree() functions except that they return [JSONB](../json1.html#jsonbx) for the "value" column when the "type" is 'array' or 'object'.
3.  The [carray](../carray.html) and [percentile](../percentile.html) extensions are now built into [the amalgamation](../amalgamation.html), though they are disabled by default and must be activated at compile-time using the [\-DSQLITE\_ENABLE\_CARRAY](../compile.html#enable_carray) and/or [\-DSQLITE\_ENABLE\_PERCENTILE](../compile.html#enable_percentile) options, respectively.
4.  Enhancements to [TCL Interface](../tclsqlite.html):
    1.  Add the `-asdict` flag to the `eval` command to have it set the row data as a dict instead of an array.
    2.  User-defined functions may now `break` to return an SQL NULL.
5.  [CLI](../cli.html) enhancements:
    1.  Increase the precision of ".timer" to microseconds.
    2.  Enhance the "box" and "column" formatting modes to deal with double-wide characters.
    3.  The ".imposter" command provides read-only [imposter tables](../imposter.html) that work with [VACUUM](../lang_vacuum.html) and do not require the --unsafe-testing option.
    4.  Add the --ifexists option to the CLI command-line option and to the [.open command](../cli.html#dotopen).
    5.  Limit columns widths set by the ".width" command to 30,000 or less, as there is not good reason to have wider columns, but supporting wider columns provides opportunity to malefactors.
6.  Performance enhancements:
    1.  Use fewer CPU cycles to commit a read transaction.
    2.  Early detection of joins that return no rows due to one or more of the tables containing no rows.
    3.  Avoid evaluation of scalar subqueries if the result of the subquery does not change the result of the overall expression.
    4.  Faster window function queries when using "BETWEEN :x FOLLOWING AND :y FOLLOWING" with a large :y.
7.  Add the [PRAGMA wal\_checkpoint=NOOP;](../pragma.html#pragma_wal_checkpoint) command and the [SQLITE\_CHECKPOINT\_NOOP](../c3ref/c_checkpoint_full.html) argument for [sqlite3\_wal\_checkpoint\_v2()](../c3ref/wal_checkpoint_v2.html).
8.  Add the [sqlite3\_set\_errmsg()](../c3ref/set_errmsg.html) API for use by extensions.
9.  Add the [sqlite3\_db\_status64()](../c3ref/db_status.html) API, which works just like the existing [sqlite3\_db\_status()](../c3ref/db_status.html) API except that it returns 64-bit results.
10.  Add the [SQLITE\_DBSTATUS\_TEMPBUF\_SPILL](../c3ref/c_dbstatus_options.html) option to the [sqlite3\_db\_status()](../c3ref/db_status.html) and [sqlite3\_db\_status64()](../c3ref/db_status.html) interfaces.
11.  In the [session extension](../sessionintro.html) add the [sqlite3changeset\_apply\_v3()](../session/sqlite3changeset_apply.html) interface.
12.  For the [built-in printf()](../printf.html) and the [format() SQL function](../lang_corefunc.html#format), omit the leading '-' from negative floating point numbers if the '+' flag is omitted and the "#" flag is present and all displayed digits are '0'. Use '%#f' or similar to avoid outputs like '-0.00' and instead show just '0.00'.
13.  Improved error messages generated by [FTS5](../fts5.html).
14.  Enforce STRICT typing on computed columns.
15.  Improved support for VxWorks
16.  JavaScript/WASM now supports 64-bit WASM. The canonical builds continue to be 32-bit but creating one's own 64-bit build is now as simple as running "make".
17.  Improved resistance to database corruption caused by an application [breaking Posix advisory locks using close()](../howtocorrupt.html#posix_close_bug).

---

---

## Checklist:


patch/sqlite3/0001-print-FTL-version-in-interactive-shell.patch
```diff
@@ -20,10 +20,10 @@ index 6280ebf6..a5e82f70 100644
  #else
  int SQLITE_CDECL wmain(int argc, wchar_t **wargv){
    char **argv;
-@@ -33467,6 +33469,7 @@ int SQLITE_CDECL wmain(int argc, wchar_t **wargv){
+@@ -33656,6 +33658,7 @@ int SQLITE_CDECL wmain(int argc, wchar_t **wargv){
+     if( stdin_is_interactive ){
        char *zHome;
        char *zHistory;
-       int nHistory;
 +      print_FTL_version();
```
yubiuser, line 27: Why did you add this here?

DL6ER: The highlighting of Github tricked you. We always added this, what I removed is the `int nHistory;`, check out the background color of the line numbers:

image

yubiuser (approved):

patch/sqlite3/0001-print-FTL-version-in-interactive-shell.patch
yubiuser: Ah yes. I checked an older version of the patch file and it is already there.

PromoFaux (approved):

PR #2705, opened 2025-11-09
Implement simple partial matching for regex in /api/search/{domain}

DL6ER:

Implement partial matching for regular expressions. Note that this will only work for literal parts of regular expressions and will not find complicated regular expressions.

This change allows to find *literal* parts in regular expressions like `gofile` in a regex like
``` regex
^store[0-9]+\.gofile\.io$
```

Complex regular expressions such as
``` regex
^junk\.[a-z]*\.com$
```
will still not be found by `example` even though `junk.example.com` would be blocked by the regex above. This is expected and actually both by design (many regular expressions have `[a-z0-9]*` parts) and by concept (partial matching in complex regular expressions is a difficult problem).

To bring this to the attention of the user, https://github.com/pi-hole/web/pull/3654 amends the description of the partial matching checkbox:

image

---

---

## Checklist:

### DIFF
--- a/src/api/docs/content/specs/search.yaml
+++ b/src/api/docs/content/specs/search.yaml
@@ -162,7 +162,7 @@ components:
               properties:
                 partial:
                   type: boolean
-                  description: Whether partial matching was requested
+                  description: Whether partial matching was requested. Note that partial matching may not find complex regex entries.
                   example: false
                 N:
                   type: integer
--- a/src/api/search.c
+++ b/src/api/search.c
@@ -29,6 +29,7 @@ static int search_table(struct ftl_conn *api, const char *item,
 	if(ids != NULL)
 	{
 		// Set item to NULL to indicate that we are searching for IDs
+		// instead of domains
 		item = NULL;
 		// Strip "[" and "]" from ids
 		ids[strlen(ids)-1] = '\0';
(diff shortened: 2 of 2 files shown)


PromoFaux:
Just thinking out loud here, but assuming that it's computationaly too much to "just" take the list of each regex a user may have in their list and then test them against the query that is being passed into -q?

rdwebdesign:
I'm still unsure about adding partial regex search.

I think users expect an "impossible" result from the partial match.

Apparently some users expect that the "partial regex match" would return results if **any possible match** of a regex contains the search string. 
In other words, they expect a regex would be transformed into a list of all possible domains that will match the regex and then the search would return all list items containing the _search string_.

Example:
Imagine the search string is `abc`.

- The list of all possible matches for `^domain.abctelecom.net$` is only the domain "domain.abctelecom.net", so this regex matches the search.
- The list of all possible matches for `^domain.xyztelecom.net$` is only the domain "domain.xyztelecom.net", so this regex doesn't match the search.
- The list of all possible matches for `^(a|b)bc\.com$` has only 2 domains: "abc.com" and "bbc.com", so this regex also matches the search for `abc`.

But that logic is impractical. 
We can't list all possible matches for `^[a-z]*\.com$` and then search for the string on the list.

PromoFaux (approved):

PR #2706, opened 2025-11-09
Get earliest query timestamp from database

PromoFaux:

Retrieves the the earliest timestamp found in the database for both in memory and on disk databses. Will be utilised on the web interface for determining the "all time" range in the datetimepicker of the query page.

We reuse the existing `get_number_of_queries_in_DB` function and add an optional pointer to determine if we actually need the earliest timestamp or not. Maybe the function should be renamed, now. Not sure - I'll leave that up to @DL6ER 

---

---

### DIFF
--- a/src/api/docs/content/specs/info.yaml
+++ b/src/api/docs/content/specs/info.yaml
@@ -994,8 +994,22 @@ components:
                   example: "pihole"
         queries:
           type: integer
-          description: Number of queries in long-term database
+          description: Number of queries in in-memory database
           example: 536956
+        earliest_timestamp:
+          type: number
+          description: Unix timestamp of the earliest query in the in-memory database
+          example: 1611742120.5
+          nullable: true
+        queries_disk:
+          type: integer
+          description: Number of queries in on-disk database
+          example: 1234567
+        earliest_timestamp_disk:
+          type: number
+          description: Unix timestamp of the earliest query in the on-disk database
+          example: 1608450520.3
+          nullable: true
         sqlite_version:
           type: string
           description: Version of embedded SQLite3 engine
--- a/src/api/docs/content/specs/info.yaml
+++ b/src/api/docs/content/specs/info.yaml
@@ -998,18 +998,16 @@ components:
           example: 536956
         earliest_timestamp:
           type: number
-          description: Unix timestamp of the earliest query in the in-memory database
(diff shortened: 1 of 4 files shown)


src/database/query-table.c
```diff
@@ -532,7 +540,13 @@ int get_number_of_queries_in_DB(sqlite3 *db, const char *tablename)
 	}
 	rc = sqlite3_step(stmt);
 	if( rc == SQLITE_ROW )
+	{
+		// Get count from first column
 		num = sqlite3_column_int(stmt, 0);
+		// Get timestamp from second column if requested
+		if(earliest_timestamp != NULL && sqlite3_column_type(stmt, 1) != SQLITE_NULL)
```
DL6ER, line 547: So when there is no minimum timestamp, the column is not read and the value stays how it was (initialized to `0.0`). I think this should be said in the API specs or the frontend may fail / show Jan 1st, 1970 in this case.

src/api/docs/content/specs/info.yaml
```diff
@@ -994,8 +994,22 @@ components:
...
         queries:
           type: integer
-          description: Number of queries in long-term database
+          description: Number of queries in in-memory database
           example: 536956
+        earliest_timestamp:
+          type: number
+          description: Unix timestamp of the earliest query in the in-memory database
+          example: 1611742120.5
+          nullable: true
```
DL6ER:
```suggestion
```
`nullable: true` is not actually possible. Your code always produces a number. Yes, it may be `0.0` but it can never be `null`.

src/api/docs/content/specs/info.yaml
```diff
@@ -994,8 +994,22 @@ components:
...
+        queries_disk:
+          type: integer
+          description: Number of queries in on-disk database
+          example: 1234567
+        earliest_timestamp_disk:
+          type: number
+          description: Unix timestamp of the earliest query in the on-disk database
+          example: 1608450520.3
+          nullable: true
```
DL6ER:
```suggestion
```

DL6ER (changes requested):

src/database/query-table.c
PromoFaux: I will add a comment in the specs

DL6ER (approved):

PR #2710, opened 2025-11-13
update gravity - improve domain validation processing speed

rrobgill:

Speed up operation of Update Gravity / `pihole -g`

Incorporates several small efficiency gains within the main domain validation loop. The individual gains per execution are comparatively minor but as they applied to every entry in blocklists and allowlists during the update process, their effects are cumulative.

- Use a lookup table for valid characters, with single comparison per character instead of repeated comparisons and branches per character

- Short circuit the ip address testing rather than testing every string as an IPv4 and IPv6 address:
   Only perform full ipv4 address test if first character of token is numeric
   Only perform full ipv6 address test if the line contains a colon, 
   and a colon is found within the first 5 characters of the token
 
- Limits check for unicode BOM to once per file

**Test results**

On test system with the following lists:

```
https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts	
https://media.githubusercontent.com/media/zachlagden/Pi-hole-Optimized-Blocklists/refs/heads/main/lists/all_domains.txt	
https://cdn.jsdelivr.net/gh/hagezi/dns-blocklists@latest/adblock/ultimate.txt (ABP style, added as blocklist)
https://raw.githubusercontent.com/arapurayil/aBL/refs/heads/main/filters/main.txt (ABP style - added as allowlist)
https://raw.githubusercontent.com/DandelionSprout/adfilt/refs/heads/master/Alternate%20versions%20Anti-Malware%20List/AntiMalwareDomains.txt 
```

Results of `time pihole -g`

Before (current dev):
real	0m27.348s
user	0m20.069s
sys	0m3.631s

After (this patch):
real	0m22.964s
user	0m15.607s
sys	0m3.598s

N/A

---

---

### DIFF
--- a/src/tools/gravity-parseList.c
+++ b/src/tools/gravity-parseList.c
@@ -31,6 +31,23 @@ static const char *false_positives[] = {
 	"ip6-allhosts"
 };
 
+// Lookup table containing characters that are valid in domain names
+// Domain must not contain any character other than [a-zA-Z0-9.-_]
+static const unsigned char valid_domain_char[256] = {
+	['a' ... 'z'] = 1, ['A' ... 'Z'] = 1, ['0' ... '9'] = 1,
+	['-'] = 1, ['.'] = 1, ['_'] = 1,
+};
+
+// Fast test for string containing specified character in 
+// selected range from start of string
+static inline bool string_has_within(const char *s, const char character, const int maxlen)
+{
+    for (int i = 0; i < maxlen && s[i] != '\0'; ++i)
+        if (s[i] == character)
+            return true;
+    return false;
+}
+
 // Print progress for files larger than 10 MB
 // This is to avoid printing progress for small files
 // which would be printed too often as affect performance
@@ -47,19 +64,17 @@ inline bool __attribute__((pure)) valid_domain(const char *domain, const size_t 
 	if(domain == NULL || len == 0 || len > 255)
 		return false;
 
-	// Loop over line and check for invalid characters
+	// Loop over line
 	int last_dot = -1;
 	for(unsigned int i = 0; i < len; i++)
 	{
-		// Domain must not contain any character other than [a-zA-Z0-9.-_]
-		if(domain[i] != '-' && domain[i] != '.' && domain[i] != '_' &&
-		   (domain[i] < 'a' || domain[i] > 'z') &&
-		   (domain[i] < 'A' || domain[i] > 'Z') &&
-		   (domain[i] < '0' || domain[i] > '9'))
+		// Check for invalid characters
+		unsigned char c = (unsigned char)domain[i];
+		if(!valid_domain_char[c])
 			return false;
 
 		// Individual label length check
-		if(domain[i] == '.')
+		if(c == '.')
 		{
 			// Label must be longer than 0 characters, i.e., two consecutive
 			// dots are not allowed
@@ -270,18 +285,23 @@ int gravity_parseList(const char *infile, const char *outfile, const char *adlis
 	ssize_t invalid_domains_list_lengths[MAX_INVALID_DOMAINS] = { -1 };
 	unsigned int invalid_domains_list_len = 0;
 	unsigned int exact_domains = 0, abp_domains = 0, invalid_domains = 0;
+	bool first_line = true; // Flag to test UTF-8 Bom only on first line
 	while((read = getline(&line, &len, fpin)) != -1)
 	{
-
 		// Handle UTF-8 BOM (Byte Order Mark) if present at start of file
-		if (read >= 3 &&
-			(unsigned char)line[0] == 0xEF &&
-			(unsigned char)line[1] == 0xBB &&
-			(unsigned char)line[2] == 0xBF)
+		if (first_line)
 		{
-			// Shift line contents left by 3 bytes to remove BOM
-			memmove(line, line + 3, read - 3);
-			read -= 3;
+			// Clear first_line flag immediately after check
+			first_line = false;
+			if (read >= 3 &&
+				(unsigned char)line[0] == 0xEF &&
+				(unsigned char)line[1] == 0xBB &&
+				(unsigned char)line[2] == 0xBF)
+			{
+				// Shift line contents left by 3 bytes to remove BOM
+				memmove(line, line + 3, read - 3);
+				read -= 3;
+			}
 		}
 
 		// Update total read bytes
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
I see im improvements around 1-2 secs for my particular lists. I guess it depends on your hardware and list content how much improvement one sees.

src/tools/gravity-parseList.c
```diff
@@ -47,19 +54,17 @@ inline bool __attribute__((pure)) valid_domain(const char *domain, const size_t
...
-		   (domain[i] < 'A' || domain[i] > 'Z') &&
-		   (domain[i] < '0' || domain[i] > '9'))
+		// Check for invalid characters
+		unsigned char c = (unsigned char)domain[i];
+		if(!valid_domain_char[c])
 			return false;
 
 		// Individual label length check
-		if(domain[i] == '.')
+		if(c == '.')
```
DL6ER, line 77: I don't think this changes anything in the actually generated machine code. I - sometimes - look at FTL code in Assembly, just to see what clever tricks the compiler does. This change does not seem to cause any change in Assembly as the content of `domain[i]` is in a register already. However, it also doesn't hurt so we can add this here - mainly to avoid someone else suggesting this again in the future or because other compilers than `gcc` may be less clever in this regard.

src/tools/gravity-parseList.c
```diff
@@ -108,7 +113,7 @@ static inline bool __attribute__((pure)) valid_abp_domain(const char *line, cons
 			return false;
 
 		// First four characters must be "@@||"
-		if(line[0] != '@' || line[1] != '@' || line[2] != '|' || line[3] != '|')
+		if (memcmp(line, "@@||", 4) != 0)
```
DL6ER: Is calling a function with all the related overhead really faster than doing a direct comparison in code that can be highly efficiently implemented by the compiler?

Mind that there is no magic behind the standard library functions, and `memcmp()` is implemented like: 
``` c
int memcmp(const void *vl, const void *vr, size_t n)
{
	const unsigned char *l=vl, *r=vr;
	for (; n && *l == *r; n--, l++, r++);
	return n ? *l-*r : 0;
}
```
so actually a loop ([source code](https://git.musl-libc.org/cgit/musl/tree/src/string/memcmp.c)). As the compiler is usually not doing loop-unrolling over into linked a standard library, running this will *definitely* cost more than doing the chain of `== || == || ...` here. Moreover, because the first mismatch will short-circuit the entire `if`, skipping the rest of the function.

src/tools/gravity-parseList.c
```diff
@@ -47,19 +54,17 @@ inline bool __attribute__((pure)) valid_domain(const char *domain, const size_t
...
 	for(unsigned int i = 0; i < len; i++)
 	{
-		// Domain must not contain any character other than [a-zA-Z0-9.-_]
-		if(domain[i] != '-' && domain[i] != '.' && domain[i] != '_' &&
-		   (domain[i] < 'a' || domain[i] > 'z') &&
-		   (domain[i] < 'A' || domain[i] > 'Z') &&
-		   (domain[i] < '0' || domain[i] > '9'))
+		// Check for invalid characters
+		unsigned char c = (unsigned char)domain[i];
+		if(!valid_domain_char[c])
```
DL6ER, line 73: This is definitely a good improvement.

src/tools/gravity-parseList.c
```diff
@@ -270,18 +275,20 @@ int gravity_parseList(const char *infile, const char *outfile, const char *adlis
...
-			(unsigned char)line[1] == 0xBB &&
-			(unsigned char)line[2] == 0xBF)
+		if (first_line)
 		{
-			// Shift line contents left by 3 bytes to remove BOM
-			memmove(line, line + 3, read - 3);
-			read -= 3;
+			// Clear first_line flag immediately after check
+			first_line = false;
+			if(read >= 3 && memcmp(line, "\xEF\xBB\xBF", 3) == 0)
```
DL6ER: Similar here: I am pretty sure using `memcmp()` actually increases the overhead.

DL6ER (changes requested):
Thank you for this pull request. I have made a few comments. The main speed increase likely comes from calling `inet_pton()` which is definitely a worthwhile change. Maybe also small bits from the lookup table (but definitely less). However, the two locations where direct comparisons are replaced with `memcmp()` calls should be undone due to the reasons I outlined.

src/tools/gravity-parseList.c
rrobgill: reverted

src/tools/gravity-parseList.c
rrobgill: reverted

src/tools/gravity-parseList.c
```diff
@@ -342,6 +352,9 @@ int gravity_parseList(const char *infile, const char *outfile, const char *adlis
 		if(read < 1)
 			continue;
 
+		// Flag lines containing colons (potential IPv6 addresses)
+		bool line_has_colon = (strchr(line, ':') != NULL);
+
```
DL6ER: Sorry ... I have one more suggestion but then this is definitely good to go in :-)
```suggestion
```

src/tools/gravity-parseList.c
```diff
@@ -352,14 +365,24 @@ int gravity_parseList(const char *infile, const char *outfile, const char *adlis
...
+			}
 
-			// IPv6 addresses
-			struct in6_addr buffer6 = { 0 };
-			if (inet_pton(AF_INET6, token, &buffer6) == 1)
-				goto next_domain;
+			// Ipv6 address
+			// Only test for IPv6 address if line contained a colon
+			// and it's in this token's first 5 characters
+			if (line_has_colon && memchr(token, ':', 5) != NULL)
```
DL6ER:
```suggestion
			if (strchr(line, ':') != NULL && memchr(token, ':', 5) != NULL)
```
Avoid `strchr()` - which walks the entire line - to be called for IPv4 address lines at all.

DL6ER: I'm also wondering it this may be a possible place for the code to crash. `memchr` accesses up to 5 bytes inside `token` but couldn't the latter be shorter than that? So `memchr` possibly looks into uninitialized/unallocated memory?

Something like
``` c
memchr(token, ':', min(strlen(token), 5))
```
would be safer, however, `strlen()` alone will already walk the entire line so you could even use a tight loop here:
``` c
static inline bool string_has_within(const char *s, const char character, const int maxlen)
{
    for (int i = 0; i < maxlen && s[i] != '\0'; ++i)
        if (s[i] == character)
            return true;
    return false;
}
```
and then
```suggestion
			if (string_has_within(token, ':', 5))
```
(note: not tested)

DL6ER (changes requested):

rrobgill:
Tested now with your latest changes.

The strchr() was definitely more expensive than I realised. 

`pihole -g` is an additional 300ms faster on test machine using set of lists above with your changes applied.

DL6ER (approved):

PR #2741, opened 2025-12-12
Enhancements to the documentation markdown generator

PromoFaux:

Enhances the markdown generation script to flesh out the examples. If there is an "Example:" line detected then we use that in the practical examples instead of the default:

Before:
image

After:

image

This also corrects the docker compose example to use multiline strings rather than invalid array syntax. This is nicer to read than a semicolon delimited string (although this also would work)

Docs PR as example: https://github.com/pi-hole/docs/pull/1327

---

---


tools/pihole_toml_to_markdown.py
```diff
@@ -158,37 +216,65 @@ def parse_toml_with_comments(filepath):
...
+            # Convert TOML-style array to proper YAML format
+            if example_display_value.strip().startswith("["):
+                # This is an array - convert to YAML format with pipe syntax
+                # Extract content between [ and ]
+                array_content = example_display_value.strip()[1:-1].strip()
+                if array_content:
+                    # Format as YAML multiline string with pipe syntax
+                    # Use smart splitting that respects quoted strings
+                    items = parse_array_items(array_content)
+                    documentation.append(f"      {env_var}: |")
```
DL6ER:
```suggestion
                    documentation.append(f"      {env_var}: |-")
```

Shouldn't we use `|-`  for removing the final newline?

PromoFaux: Good question well asked. I didn't know about this option

PromoFaux: Both ways ultimately produce the same output in the toml file

```
  dnsmasq_lines = [
    "address=/example.com/",
    "address=/example.org/",
    "address=/example.net/"
  ] ### CHANGED (env), default = []
```

If you would prefer we didn't send the final new line then that's also fine and I can update my other docker docs PR to follow suit

DL6ER (approved):

PR #2772, opened 2026-02-01
home.arpa and internal TLDs may be non-local without revServer

### LINKED ISSUE #2769: Cannot forward home.arpa unless revServer configured, even when pihole domain is home.arpa
### Versions
Core version is v6.3 (Latest: v6.3)
Web version is v6.4 (Latest: v6.4)
FTL version is v6.4.1 (Latest: v6.4.1)

### Platform

- OS and version:  Debian 12 

- Platform:  Proxmox LXC

### Expected behavior

My chosen domain for internal servers is home.arpa.  My goal is to have pihole handle only DNS blocking, and forward all other requests to opnsense for local name resolution and forwarding to unbound.  So I simply set my upstream to opnsense and that should be all.

### Actual behavior / bug

However, it's hard coded that home.arpa is set to local in dnsmasq.conf unless there's a revServer with home.arpa as the domain, so my desired configuration cannot be done - any lookups for X.home.arpa never get forwarded past pihole and there is no way to change it other than setting a revServer, which I do not want to do.

### Steps to reproduce

1. set upstream dns to a router that can resolve X.home.arpa
2. clear all revServers
3. nslookup X.home.arp
4. You'll get NXdomain and the query is never forwarded to upstream

## Debug Token

- URL:  https://tricorder.pi-hole.net/yGothn8m/

### Additional context

I see two possible solutions here.

When flagging domain_homearpa, make it so that if pihole's domain is home.arpa, and dhcp is not in use, it will forward home.arpa upstream. This does make some assumptions though and may not be the best.

A simple solution would just be an option flag "Allow forwarding of home.arpa to upstream" so it's a conscious decision by the user to allow this for their specific setup.

DL6ER:

Fix that `dns.domain = "home.arpa"` ( or `"internal"`) **and** `dns.domain.local = false` should result in these domains being sent to regular upstreams - even in the absence of matching `revServer` settings. So far, a valid `revServer` line for this configuration is needed to function as expected.

This is a direct fix for #2769 

---

**Related issue or feature (if applicable):**  Closes #2769

---

## Checklist:

### DIFF
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -569,15 +569,30 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 		fputs("local=//\n\n", pihole_conf);
 	}
 
-	// Ensure that home.arpa domains (RFC 8375) are not sent upstream,
-	// unless explicitly configured by user as local domain with either:
-	//  - revServer entry for "home.arpa" (revServer_homearpa), or
-	//  - dns.domain == "home.arpa" AND dns.domain.local == false
+	// Ensure that home.arpa domains (RFC 8375) are not forwarded to
+	// upstream servers by default. However, we skip adding the protection
+	// when the user has explicitly configured an exception. The exceptions
+	// are:
+	//  - a reverse server has been configured for the "home.arpa" TLD, OR
+	//  - the configured DNS domain equals "home.arpa" and that domain is
+	//    explicitly marked non-local.
 	const bool domain_homearpa = strlen(conf->dns.domain.name.v.s) > 0 &&
 	                             strcasecmp(conf->dns.domain.name.v.s, "home.arpa") == 0;
-	if (!revServer_homearpa && !domain_homearpa && !config.dns.domain.local.v.b)
+	if(revServer_homearpa)
 	{
-		fputs("# Do not forward home.arpa domains to upstream servers\n",pihole_conf);
+		fputs("# A reverse server is configured for \"home.arpa\".\n", pihole_conf);
+		fputs("# All queries for this domain will be forwarded to this\n", pihole_conf);
+		fputs("# upstream server\n\n", pihole_conf);
+	}
+	else if(domain_homearpa && !config.dns.domain.local.v.b)
+	{
+		fputs("# The configured DNS domain is \"home.arpa\" and is explicitly\n", pihole_conf);
+		fputs("# marked non-local. Pi-hole will be forwarding queries for this\n", pihole_conf);
+		fputs("# domain to upstream servers.\n\n", pihole_conf);
+	}
+	else
+	{
+		fputs("# Do not forward .home.arpa domains to upstream servers\n",pihole_conf);
 		fputs("local=/home.arpa/\n\n",pihole_conf);
 	}
 
@@ -585,14 +600,24 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 	// draft-davies-internal-tld-05) are not forwarded to upstream servers
 	// by default. However, we skip adding the protection when the user has
 	// explicitly configured an exception. The exceptions are:
-	//  - a reverse server has been configured for the "internal" TLD
-	//    (revServer_internal == true), OR
+	//  - a reverse server has been configured for the "internal" TLD, OR
 	//  - the configured DNS domain equals "internal" and that domain is
-	//    explicitly marked non-local (domain_internal == true &&
-	//    config.dns.domain.local.v.b == false).
+	//    explicitly marked non-local.
 	const bool domain_internal = strlen(conf->dns.domain.name.v.s) > 0 &&
 	                             strcasecmp(conf->dns.domain.name.v.s, "internal") == 0;
-	if (!revServer_internal && !(domain_internal && config.dns.domain.local.v.b == false))
+	if(revServer_internal)
+	{
+		fputs("# A reverse server is configured for \"internal\".\n", pihole_conf);
+		fputs("# All queries for this domain will be forwarded to this\n", pihole_conf);
+		fputs("# upstream server\n\n", pihole_conf);
+	}
+	else if(domain_internal && !config.dns.domain.local.v.b)
+	{
+		fputs("# The configured DNS domain is \"internal\" and is explicitly\n", pihole_conf);
+		fputs("# marked non-local. Pi-hole will be forwarding queries for this\n", pihole_conf);
+		fputs("# domain to upstream servers.\n\n", pihole_conf);
+	}
+	else
 	{
 		fputs("# Do not forward .internal domains to upstream servers\n",pihole_conf);
 		fputs("local=/internal/\n\n",pihole_conf);
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -582,13 +582,17 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 	}
 
 	// Ensure that internal domains (Internet-Draft
-	// draft-davies-internal-tld-05) are not sent upstream, unless
-	// explicitly configured by user as local domain with either:
-	//  - revServer entry for "internal" (revServer_internal), or
-	//  - dns.domain == "internal" AND dns.domain.local == false
+	// draft-davies-internal-tld-05) are not forwarded to upstream servers
+	// by default. However, we skip adding the protection when the user has
+	// explicitly configured an exception. The exceptions are:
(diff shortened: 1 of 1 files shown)


Aerankas (approved):
My vote doesn't count, but looks good to me and thanks for the fix.

yubiuser:
@Aerankas did you actually test this brach in your setup?

You can switch to this implementation by runing

`pihole checkout ftl fix/dns_domain_local`

Aerankas:
I did attempt to verify this, thanks for the command to switch to dev branch.  I left revServers blank, set dns.domain.name to home.arpa and dns.domain.local to disabled, as I wanted.  I can see the dnsmasq.conf line  (local=/home.arpa/) is gone.  However I can't seem to resolve a home.arpa address (NXDOMAIN).  This could be a problem on my end since I did change to a different domain and tried to just create a test - In opnsense unbound I set query forwarding for home.arpa to point to dnsmasq. In dnsmasq I added a hosts record with name piholeverify, domain home.arpa and a valid IP on my network.  I'm not yet sure where the request is failing.
Edit: nslookup direct to opnsense also returns NXDOMAIN, so I think it's possible the request is being forwarded correctly and my test setup is poor.

Aerankas:
I was able to set up a proper test and unfortunately it's not working as expected.

Direct to opnsense works:

nslookup piholeverify.home.arpa 192.168.0.1
Server:  opnsense.lan
Address:  192.168.0.1

Name:    piholeverify.home.arpa
Address:  192.168.0.10

Via pihole fails (it's forwarding to 192.168.0.1):
nslookup piholeverify.home.arpa
Server:  pi.hole
Address:  192.168.1.10

*** pi.hole can't find piholeverify.home.arpa: Non-existent domain

Aerankas:
Ah. It's appending .lan which is why it fails.

root@pihole:/etc/pihole# pihole -t | grep home.arpa
Feb  3 11:01:55: query[A] piholeverify.home.arpa.lan from 192.168.0.50
Feb  3 11:01:55: forwarded piholeverify.home.arpa.lan to 192.168.1.1
Feb  3 11:01:55: reply piholeverify.home.arpa.lan is NXDOMAIN

That's probably my connection-specific suffix? I tried setting is manually to prevent this (I'm committed to keeping my internal domain as .lan now so I don't want to reconfigure it back to home.arpa, so the testing is a little finicky).

Now I can see both the .lan request and the proper request, but getting a cached nxdomain

root@pihole:/etc/pihole# pihole -t | grep home.arpa
Feb  3 11:19:13: query[A] piholeverify.home.arpa.lan from 192.168.0.50
Feb  3 11:19:13: forwarded piholeverify.home.arpa.lan to 192.168.1.1
Feb  3 11:19:13: reply piholeverify.home.arpa.lan is NXDOMAIN
Feb  3 11:19:13: query[AAAA] piholeverify.home.arpa.lan from 192.168.0.50
Feb  3 11:19:13: forwarded piholeverify.home.arpa.lan to 192.168.1.1
Feb  3 11:19:13: reply piholeverify.home.arpa.lan is NXDOMAIN
Feb  3 11:19:13: query[A] piholeverify.home.arpa from 192.168.0.50
Feb  3 11:19:13: cached piholeverify.home.arpa is NXDOMAIN
Feb  3 11:19:13: query[AAAA] piholeverify.home.arpa from 192.168.0.50
Feb  3 11:19:13: cached piholeverify.home.arpa is NXDOMAIN

Aerankas:
And it works!

root@pihole:/etc/pihole# pihole reloaddns
  [✓] Flushing DNS cache
root@pihole:/etc/pihole# pihole -t | grep home.arpa
Feb  3 11:22:02: query[A] piholeverify.home.arpa.lan from 192.168.0.50
Feb  3 11:22:02: forwarded piholeverify.home.arpa.lan to 192.168.1.1
Feb  3 11:22:02: reply piholeverify.home.arpa.lan is NXDOMAIN
Feb  3 11:22:02: query[AAAA] piholeverify.home.arpa.lan from 192.168.0.50
Feb  3 11:22:02: forwarded piholeverify.home.arpa.lan to 192.168.1.1
Feb  3 11:22:02: reply piholeverify.home.arpa.lan is NXDOMAIN
Feb  3 11:22:02: query[A] piholeverify.home.arpa from 192.168.0.50
Feb  3 11:22:02: forwarded piholeverify.home.arpa to 192.168.1.1
Feb  3 11:22:02: reply piholeverify.home.arpa is 192.168.0.10
Feb  3 11:22:02: query[AAAA] piholeverify.home.arpa from 192.168.0.50
Feb  3 11:22:02: forwarded piholeverify.home.arpa to 192.168.1.1
Feb  3 11:22:02: reply piholeverify.home.arpa is NODATA-IPv6

yubiuser:
Thanks for testing!

src/config/dnsmasq_config.c
```diff
@@ -571,17 +569,26 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
...
-	// Ensure that home.arpa domains (RFC 8375) are not sent upstream, unless configured by
-	// user as local domain, with server setting applied above
-	if (!domain_homearpa)
+	// Ensure that home.arpa domains (RFC 8375) are not sent upstream,
+	// unless explicitly configured by user as local domain with either:
+	//  - revServer entry for "home.arpa" (revServer_homearpa), or
+	//  - dns.domain == "home.arpa" AND dns.domain.local == false
+	const bool domain_homearpa = strlen(conf->dns.domain.name.v.s) > 0 &&
+	                             strcasecmp(conf->dns.domain.name.v.s, "home.arpa") == 0;
+	if (!revServer_homearpa && !domain_homearpa && !config.dns.domain.local.v.b)
```
yubiuser: Shouldn't this be something like


```suggestion
	if ((!revServer_homearpa) ||(!domain_homearpa && !config.dns.domain.local.v.b))
```

This would reflect what is written in the comment above.

rdwebdesign: Or maybe the comment should be fixed to reflect what the code does.

andreasbehnke:
I had the exact same problem (using upstream DNS for resolving local hosts to FQN xyz.home.arpa) and setting domain name to home.arpa and local to false plus checking out this branch fixed it.

src/config/dnsmasq_config.c
DL6ER: Yes, the code logic is broken. It should look like
``` c
if (!revServer_internal && !(domain_internal && !config.dns.domain.local.v.b))
```
to add `local=/internal/` unless a matching `rev-server` exists or the domain is `internal` with `domain.local == false`

The suggestion above would always add the protection when there is no rev-server, regardless of the `domain_homearpa` and `config.dns.domain.local.v.b` values.

However, we only want the protection if
- no rev server is configured AND
- the domain is `internal` AND `domain.local == false`

What makes the comment a bit harder to understand is that its intent is negated by itself (add the lines unless ...) but the `if` is formulated positively. I will clarify the comment as well.

src/config/dnsmasq_config.c
```diff
@@ -571,17 +569,30 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
...
```
yubiuser:
```suggestion
	if (!revServer_homearpa && !(domain_homearpa && config.dns.domain.local.v.b == false))
```

yubiuser (commented):
I still find it confusing that the comments talk about two conditions connected with OR but the code only uses AND. But maybe need a fresh look tomorrow.

src/config/dnsmasq_config.c
DL6ER: `!revServer_homearpa && !(domain_homearpa && !config.dns.domain.local.v.b)`

For brevity, let's use `R=revServer`, `D=domain`, `L=domain.local`:

- If the user explicitly set the domain to `home.arpa` and marked it non-local (`D=true`, `L=false`) we must NOT add the protection.
  - Your initial suggestion `(!R) || (!D && !L)` evaluates to `true` when `R=false` and so would still add the protection — incorrect.
  - My `!R && !(D && !L)` evaluates to `false` (doesn't add protection) — correct.

- If a `revServer` exists (`R=true`) we must NOT add the protection.
  - Your initial variant can still be `true` when `R=true` if `(!D && !L)` is `true`, so it may add protection despite a `revServer` — incorrect.
  - My variant is false when `R=true` — correct.

The comment yields the exception condition `(R || (D && !L))` the protection should be **the negation of that**: `!R && !(D && !L)`. We can write `!(R || (D && !L))` which is the same expression just needing more steps for evaluation. However, if it makes things more readable, this is justified.

DL6ER: I pushed a major simplification. I hope this is now clear.

yubiuser (approved):
Now a simple mind can understand this as well :-)

PR #2788, opened 2026-02-26
Fix overTime data when database.DBimport = false

### LINKED ISSUE #2785: Total queries timeline not advancing when logging is disabled
### Versions

- Pi-hole:  v6.4
- Web:  v6.4.1
- FTL:  v6.5

### Platform

- OS and version:  Debian 13 Trixie / Linux pi 6.12.62+rpt-rpi-v8
- Platform:  Raspberry Pi 4 Model B Rev 1.1

### Expected behavior

The total queries timeline should advance as usual.

### Actual behavior / bug

The total queries timeline collects all queries at the 10min when pihole-FTL was restarted at the right of the timeline. Enabling logging resolves this issue and everything works as expected. Re-disabling query logging and setting anonymous mode brings the error back.

### Steps to reproduce

1. Go to 'Settings/Privacy'
2. Deselect 'Log DNS queries and replies' then enable 'Anonymous mode'
3. Watch the total queries timeline in the dashboard

## Debug Token

- URL:  https://tricorder.pi-hole.net/2JySHl8x/

### Screenshots

Image

DL6ER:

Always mark database import as done even when `database.DBimport = false` or on database errors.

Otherwise, garbage collection will never run, causing the memory allocation to always grow and overtime information to be incorrect (handled in GC code as well)

---

**Related issue or feature (if applicable):** Fixes #2785 

---

## Checklist:

### DIFF
--- a/src/database/database-thread.c
+++ b/src/database/database-thread.c
@@ -36,6 +36,8 @@
 #include "database/sqlite3-ext.h"
 // PRId64
 #include <inttypes.h>
+// db_import_done
+#include "gc.h"
 
 static bool analyze_database(sqlite3 *db)
 {
@@ -152,6 +154,9 @@ void *DB_thread(void *val)
 	// Asynchronously import queries from the on-disk database
 	if(config.database.DBimport.v.b)
 		DB_read_queries();
+
+	// Signify that the import is done, so garbage collection will run
+	db_import_done = true;
 
 	// Log some information about the imported queries (if any)
 	log_counter_info();
--- a/src/database/query-table.c
+++ b/src/database/query-table.c
@@ -1535,14 +1535,9 @@ void DB_read_queries(void)
 	}
 
 	if( rc == SQLITE_DONE )
-	{
-		db_import_done = true;
 		log_info("Imported %zu queries from the long-term database", imported_queries);
-	}
 	else
-	{
 		log_err("DB_read_queries() - SQL error step: %s", sqlite3_errstr(rc));
-	}
 
 	if((int)imported_queries < counted_queries)
 	{


src/database/database-thread.c
```diff
@@ -153,6 +155,9 @@ void *DB_thread(void *val)
 	if(config.database.DBimport.v.b)
 		DB_read_queries();
 
+	// Signify that the import is done, so garbage collection will run
+	db_import_done = true;
+
```
rdwebdesign, line 160: This is a question:

The new code always sets `db_import_done = true`.

Before the change, did the `DB_read_queries()` function always modify the value of `db_import_done` to `true`, or was there a possibility that this variable would still be `false`?

In other words, should we use the original `query-table.c` and only add an `else` here?

```suggestion
	else
	{
		// Signify that the import is done, so garbage collection will run
		db_import_done = true;
	}

```

DL6ER: > Before the change, did the DB_read_queries() function always modify the value of db_import_done to true, or was there a possibility that this variable would still be false?

This is the perfect place to respond with the German word: Jein (a blend of Yes and No).

---

The function did set `db_import_done = true`, however, not always. When there were database importing issues (like a corrupted database), it would *not* have set it to `true`, i.e., in this case the garbage collection (GC) would never run, the in-memory database would grow ever larger and the "everything in the last slot issue" would persist, too.

The only effect of `db_import_done` is postponing any possible GC run until we are sure database import isn't running any longer. Even when it failed, it won't happen again, hence, it is fine to set it unconditionally as I did here.

rdwebdesign: OK.

DL6ER:
Confirmed working by OP in https://github.com/pi-hole/FTL/issues/2785#issuecomment-3977307090

rdwebdesign (approved):

PR #2816, opened 2026-03-20
Performance optimizations and bug fixes

DL6ER:

## Summary

Broad sweep across the FTL engine covering performance work, correctness fixes, database schema improvements, signal handler hardening, and crash diagnostics.

**145 commits** across 56 files (+2851/−1395 lines): 78 bug fixes, 45 performance optimizations, and supporting refactors.

Testing showed approximately 15% query processing speedup: ~92 µs → ~80 µs for each new query.

---

### Performance: Gravity Database

- **Shared prepared statements with `carray()` binding** — replace per-client statement vectors with one shared statement per process, rebound via `sqlite3_carray_bind()` before each lookup. Query plan goes from 4 SEARCH operations (through 3 LEFT JOINs) to a single covering index seek for gravity/antigravity.
- **SHM intarray storage for client group IDs** — store group ID arrays in shared memory with deduplication via `addintarray()`. TCP forks get group data for free without recomputation.
- **Junction table `(group_id, ...)` indexes** — enable index seeks for `carray()`-filtered queries on `adlist_by_group` and `domainlist_by_group` instead of full scans.
- **`WITHOUT ROWID` on junction tables** — `adlist_by_group`, `domainlist_by_group`, `client_by_group`, and `info` tables use the composite PK directly as the B-tree storage key, eliminating one level of indirection per JOIN.
- **carray rebind deduplication** — `bind_client_groups()` helper tracks the last-bound SHM `groupspos` per statement. Since `addintarray()` deduplicates, clients sharing the same group set skip the rebind entirely.
- **Memory-mapped I/O** — `PRAGMA mmap_size = 256 MiB` on gravity.db and the on-disk query database. Eliminates `pread()` syscall overhead for all domain lookups.
- **`posix_fadvise(POSIX_FADV_WILLNEED)`** on gravity.db at startup — pre-warms the kernel page cache asynchronously, eliminating cold-start page faults.
- **ANTIGRAVITY_TABLE support** — new enum, count query, `getTable()` case, SHM counter, and API exposure for antigravity domain counts.
- **Correctness: reverted adlist ID caching** — pre-computed per-client adlist ID arrays bypassed view JOINs for ~4x speedup but became stale when users modified groups/adlists at runtime. Reverted to live view queries that check `adlist.enabled` and `group.enabled` on every lookup.

### Performance: Query Processing Hot Path

- **Eliminate redundant SHM lookups in `FTL_check_blocking()`** — pass pre-fetched domain/client/cache pointers instead of re-looking up by ID.
- **Stack-based ABP pattern generation** — replace cJSON heap allocations with a stack struct containing pre-computed suffix offsets. Zero heap allocation in the blocking check path.
- **Lazy ABP offset computation** — defer suffix offset calculation until after the exact-match check fails, avoiding all ABP work on exact matches (the common case).
- **Skip empty list checks** — `gravity_has_exact_allowlist`, `gravity_has_exact_denylist`, `gravity_has_antigravity` flags skip entire functions when the corresponding lists are empty.
- **Regex query-type pre-filtering** — check the query-type bitmask before running `regexec()`, skipping the expensive regex engine when the type doesn't match.
- **Replace `tolower()` with branchless ASCII bit-trick** in `strcpy_tolower()`.
- **Replace `time(NULL)` calls** with the query's existing timestamp where applicable.
- **Move `find_mac()` outside SHM lock** — netlink kernel I/O no longer blocks all other threads.
- **Interface caching** in `_FTL_iface()` — skip repeated interface list iteration when the pointer hasn't changed.

### Performance: Garbage Collection

- **Deferred compaction** — instead of `memmove()`-ing all surviving queries on every GC run, advance `queries_offset` and compact lazily only when the physical array runs out of room.
- **O(1) query ID lookup** — direct-mapped 4096-entry cache for dnsmasq ID → FTL query index, replacing O(N) linear scan.
- **Incremental reference counting** — `domain->count`, `client->count`, `cache->refcount`, `domain->cname_refcount` are maintained per-query instead of recomputed during GC, eliminating O(queries) scan in `recycle()`.
- **Single-pass lookup table compaction** — GC recycles clients/domains/upstreams and compacts lookup tables in one pass instead of per-entry removal.

### Performance: Database Writes

- **Three-phase `queries_to_database()`** — Phase 1 (under SHM lock): snapshot query data. Phase 2 (no lock): bulk SQLite writes. Phase 3 (brief lock): write-back of DB indices. Minimizes lock contention.
- **Linking table ID caching** — hash table caches domain/client/forward/addinfo row IDs, eliminating 4 REPLACE subqueries per query.
- **Batched dedup linking table INSERTs** — `INSERT OR IGNORE` with `sqlite3_changes()` check instead of SELECT-then-INSERT.
- **`PRAGMA synchronous=NORMAL`** on WAL-mode disk database — skip per-commit fsync, relying on WAL checksums for crash safety.

### Performance: Data Structures

- **Shrink `DNSCacheData` from 48 to 36 bytes** on all architectures.
- **Reorder `queriesData` and `clientsData`** for cache-line alignment — hot fields in the first 64-byte line, cold `overTime[]` at the end.
- **`O(1)` string deduplication** — replace `memmem()` linear scan in `addstr()` with a process-local open-addressing hash table (FNV-1a).
- **FNV-1a hash** replaces Jenkins' One-at-a-Time in `hashStr()`.
- **Heap-based top-K selection** for `/api/stats/top_domains` and `/api/stats/top_clients` — O(N log K) instead of full sort.

### Database Schema & Query Improvements

- **`queries` VIEW migration (v22)** — replace 4 correlated subqueries with LEFT JOINs. Dramatically faster for users querying `pihole-FTL.db` via sqlite3 CLI or third-party tools.
- **`ORDER BY ... DESC LIMIT` on database top items** — `/api/stats/database/top_domains` and `top_clients` now return correctly sorted results (was returning arbitrary GROUP BY order) and push the LIMIT into SQL instead of scanning all rows.
- **Fix `NOL NULL` → `NOT NULL`** typo on `client.ip` in gravity schema — SQLite silently ignored the misspelled constraint, allowing NULL IPs.
- **Update triggers use PK lookup** — `tr_adlist_update`, `tr_client_update`, `tr_domainlist_update` now use `WHERE id = NEW.id` instead of UNIQUE text column lookup.
- **Fix missing comma in `index_creation` array** — C string literal concatenation merged two index definitions into one entry.
- **MEMDB_VERSION bump to 22** and expected schema test updates.

### Signal Handler Hardening

- **`SA_ONSTACK`** — crash handlers run on a dedicated 16 KiB alternate signal stack, enabling backtrace generation even on stack overflow faults.
- **`SA_RESETHAND`** — crash signal disposition resets to `SIG_DFL` after first invocation, preventing infinite re-entry if the handler itself faults.
- **Async-signal-safe SIGTERM handler** — save `si_pid`/`si_uid` in volatile globals, defer the expensive sender lookup (`/proc/<pid>/cmdline`, `getpwuid()`, logging) to `log_sigterm_info()` called from `cleanup()` outside signal context. Eliminates potential deadlock if SIGTERM arrives while malloc lock is held.
- **Async-signal-safe RT signal handler** — remove `log_info(strsignal())` call; events are logged when processed in the main loop.
- **Guard `delete_shm()` against uninitialized SHM** — early return when `name == NULL`, preventing garbage munmap/close/shm_unlink during `pihole-FTL crash` test.
- **Fix leaked `FILE*`** in original SIGTERM handler (missing `fclose` on success path).

### Crash Diagnostics

- **GDB-style backtrace output** — frames show `#N  0xADDR in func () at file:line` matching GDB's `bt` convention. Library frames show `from libname`. Unresolved frames show `in ?? ()`.
- **`_Unwind_Backtrace`** via GCC's libgcc — works on glibc AND musl, static AND dynamic, all architectures. Falls back to `dladdr()` for library symbols, then `/proc/self/maps` for mapping names.
- **PIE/ASLR-aware addr2line** — subtracts the `__ehdr_start` load base from instruction pointers for correct source line resolution.
- **Backtrace end marker** — `--- end of backtrace (N frames) ---` for clear visual termination.

### Correctness Fixes (78 commits)

Critical:
- `gravity_updated()` closed global `gravity_db` instead of local handle — broke all DNS queries on gravity check error
- `new_sqlite3_stmt_vec()` allocated n² instead of n items — massive memory waste
- Session cookie used compile-time default timeout instead of configured value
- Zombie detection in `process_alive()` used `strcmp` on a substring — never matched
- SHM deadlock: `parse_neighbor_cache()` held lock across `continue` on NULL client
- Heap buffer overflows: `add_to_fifo_buffer()` OOB write, `process_arp_reply()` OOB device access
- Swapped bind parameters in `db_update_disk_counter` — on-disk counters never updated
- `% 0xFF` vs `& 0xFF` for EDNS VERSION byte — corrupted version 0 detection
- EDE info-code byte-order: double-swap on little-endian, broken on big-endian
- Stale SHM pointers after `find_mac()` unlock/relock — use-after-free risk
- Unchecked `realloc` in teleporter upload — NULL deref + memory leak on failure

Resource leaks (30+ fixes):
- Missing `dbclose()` on error paths across all API database endpoints
- Missing `sqlite3_finalize()` on prepare/bind failures
- Missing `ROLLBACK` before early returns inside open transactions
- Missing `free()` for cJSON objects, punycode buffers, query strings
- Missing `fclose()` for file handles

Y2038 fixes:
- `sqlite3_column_int64()` for all `time_t` database fields
- `sqlite3_bind_int64()` for time fields in network table inserts

### Test Infrastructure

- `test-fast` build option — skips performance tests for faster CI iteration
- Updated test expectations for removed log messages, new schema version, new VIEW definition

---

---

## Checklist:


DL6ER:
Real-world experiments suggest a speed improvement from ~ 92 µs -> ~ 80 µs for each new query. This is not a lot, but my testing machine is also a very fast one. In relative terms it is roughly 15 %.

yubiuser (commented):
I think there is no need to do the performance calculation on a fresh FTL start.

```
2026-03-22 20:03:09.987 UTC [240677/T240710] INFO:  -> DNS cache records: 47
2026-03-22 20:03:09.987 UTC [240677/T240710] INFO:  -> Known forward destinations: 3
2026-03-22 20:03:09.987 UTC [240677/T240710] DEBUG_PERFORMANCE: FTL cache stats: no queries in last 5 minutes
2026-03-22 20:03:09.987 UTC [240677/T240710] DEBUG_PERFORMANCE: Query perf [new_query (total under lock)]: no calls in last 5 minutes
2026-03-22 20:03:09.987 UTC [240677/T240710] DEBUG_PERFORMANCE: Query perf [findClientID]: no calls in last 5 minutes
2026-03-22 20:03:09.987 UTC [240677/T240710] DEBUG_PERFORMANCE: Query perf [findDomainID]: no calls in last 5 minutes
2026-03-22 20:03:09.987 UTC [240677/T240710] DEBUG_PERFORMANCE: Query perf [check_blocking]: no calls in last 5 minutes
2026-03-22 20:03:09.987 UTC [240677/T240710] DEBUG_PERFORMANCE: Query perf [reply (total under lock)]: no calls in last 5 minutes
2026-03-22 20:03:09.988 UTC [240677/T240710] DEBUG_PERFORMANCE: Gravity lookup stats [gravity (exact)]: no calls in last 5 minutes
2026-03-22 20:03:09.988 UTC [240677/T240710] DEBUG_PERFORMANCE: Gravity lookup stats [antigravity (exact)]: no calls in last 5 minutes
2026-03-22 20:03:09.988 UTC [240677/T240710] DEBUG_PERFORMANCE: Gravity lookup stats [gravity (ABP)]: no calls in last 5 minutes
2026-03-22 20:03:09.988 UTC [240677/T240710] DEBUG_PERFORMANCE: Gravity lookup stats [antigravity (ABP)]: no calls in last 5 minutes
2026-03-22 20:03:09.988 UTC [240677/T240710] DEBUG_PERFORMANCE: Gravity lookup stats [denylist]: no calls in last 5 minutes
2026-03-22 20:03:09.988 UTC [240677/T240710] DEBUG_PERFORMANCE: Gravity lookup stats [allowlist]: no calls in last 5 minutes
```
_____

Why does it print the FTL PID message twice?

```
2026-03-22 20:03:09.158 UTC [240677M] INFO: Parsed config file /etc/pihole/pihole.toml successfully
2026-03-22 20:03:09.159 UTC [240677M] INFO: PID file does not exist or not readable
2026-03-22 20:03:09.159 UTC [240677M] INFO: No other running FTL process found.
2026-03-22 20:03:09.167 UTC [240677M] INFO: PID of FTL process: 240677
2026-03-22 20:03:09.171 UTC [240677M] INFO: listening on 0.0.0.0 port 53
2026-03-22 20:03:09.172 UTC [240677M] INFO: listening on :: port 53
2026-03-22 20:03:09.178 UTC [240677M] INFO: PID of FTL process: 240677
2026-03-22 20:03:09.183 UTC [240677M] INFO: Database version is 21
```

DL6ER:
Resolved the findings from @copilot and also addressed @yubiuser's comments above.

yubiuser:
This can now be rebases - let's see if it passes the tests

DL6ER:
Depends on https://github.com/pi-hole/FTL/pull/2845

src/api/stats.c
```diff
@@ -78,6 +78,27 @@ static int __attribute__((pure)) cmpdesc_te(const void *a, const void *b)
 		return 0;
 }
 
+// Min-heap sift-down for top-K selection: maintain the smallest element at the
+// root so it can be replaced when a larger one is found
```
yubiuser, line 82: For sift-down should not the smallest element stay at the root?
```suggestion
// root so it can be replaced when a smaller one is found
```

DL6ER: I think the current wording is accurate - but I can see how it reads ambiguously, so let me unpack it.

The structure is a min-heap of size `K` used to track the top-`K` largest counts:
- The root holds the smallest of the `K` elements currently in the heap (yes, smallest at the root, as you'd expect from a min-heap).
- When we see a new candidate that is larger than the root (i.e. larger than the current weakest member of our top-`K`), it kicks the root out and takes its place; then we sift down to restore the invariant. That's the if(`entry_count > top_domains[0].count`) block right below.

So "maintain the smallest at the root so it can be replaced when a larger one is found" is describing exactly that: the root is the smallest member of the heap, and it gets replaced by anything larger we encounter later.

Replacing only on a smaller element would do the opposite: it would gradually collect the smallest values from the input, which is not what we want for a "top-`K` largest" selection. Remember that we want to sort the largest value first, not the other way around.

yubiuser: Thanks for the through explanation.

yubiuser:
If been running this for several days now without issues. I think after the next patch release this is good to go into `development` for an extended testing period.

yubiuser:
I think now is a good moment to merge this.

rdwebdesign:
The PR text should be updated by DL6ER:

> **145 commits** across 56 files (+2851/−1395 lines): <mark>78 bug fixes, 45 performance optimizations, and supporting refactors.</mark>
> Testing showed approximately <mark>15%</mark> query processing speedup: ~92 µs → <mark>~80 µs</mark> for each new query.

New commits were added. The values are outdated.

src/shmem.c
```diff
@@ -148,6 +148,143 @@ static pid_t shmem_pid = 0;
...
+	if(str_hash_table == NULL || str_hash_cap == 0)
+		return SIZE_MAX;
+
+	uint32_t idx = hash & (str_hash_cap - 1);
+	while(str_hash_table[idx].hash != 0)
+	{
+		if(str_hash_table[idx].hash == hash)
+		{
+			const char *candidate = &((const char*)shm_strings.ptr)[str_hash_table[idx].offset];
+			if(memcmp(candidate, input, len) == 0)
```
PromoFaux, line 237: Full disclosure, I'm reviewing this with some black box assistance, but this is far too big a PR to just YOLO-merge and see what happens...

But what if `input` is a _prefix_ of `candidate`? e.g `example.co.` -> `example.co.uk`, or `192.168.0.1` -> `192.168.0.100` - would the code as is not return a false match? Maybe a +1 to include the terminator... (or `strcmp` which already accounts for that)


```txt
candidate string : "example.co.uk"
input string : "example.co"  (len = 10)

memcmp(candidate , input, len)   = 0
memcmp(candidate , input, len+1) = 46
strcmp(candidate, input) = 46

----------------------------------------------

candidate string : "192.168.0.100"
input string : "192.168.0.1"  (len = 11)

memcmp(candidate , input, len)   = 0
memcmp(candidate , input, len+1) = 48
strcmp(candidate, input) = 48
```

PromoFaux: Ping @DL6ER in case you missed it

rdwebdesign:
Tests are failing.
The number of entries in `pihole.toml` test is wrong.

rdwebdesign:
@DL6ER 

I started a new container using this branch + Core `development` + Web `development`.

I saw a few unexpected warnings in the end of `FTL.log`:

```
2026-04-29 14:39:37.008 WARNING Trying to free NULL pointer in get_top_domains() (/app/src/api/stats.c:357)
2026-04-29 14:39:37.008 WARNING Trying to free NULL pointer in get_top_domains() (/app/src/api/stats.c:357)
2026-04-29 14:39:37.023 WARNING Trying to free NULL pointer in get_top_clients() (/app/src/api/stats.c:600)
2026-04-29 14:39:37.023 WARNING Trying to free NULL pointer in get_top_clients() (/app/src/api/stats.c:600)
```

**NOTE:**
I recreated the container and the messages are gone. Not sure what was the difference.

src/shmem.c
DL6ER: Here `len` is `strlen(input)+1` (see `addstr()` above), so the `memcmp()` includes the trailing `NUL` and rejects prefix-only matches (`example.co` vs `example.co.uk`, `192.168.0.1` vs `192.168.0.100`). I added a short comment make that explicit in the code as I understand that `len` can look ambiguous at first glance.

rdwebdesign (approved):
Approved. 

I think we need an extended testing period (like suggested by *yubiuser*).
Let's merge it and test `development` branch for a while.

yubiuser (approved):

src/dnsmasq_interface.c
```diff
@@ -947,7 +1063,7 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
 	// Process interface information of client (if available)
 	// Skip interface name length 1 to skip "-". No real interface should
 	// have a name with a length of 1...
-	if(!internal_query && strlen(interface) > 1)
+	if(!internal_query && interface[0] != '\0' && interface[1] != '\0')
```
PromoFaux, line 1066: Clever...

test/gravity.db.sql
```diff
@@ -106,6 +106,15 @@ CREATE TRIGGER tr_domainlist_update AFTER UPDATE ON domainlist
       UPDATE domainlist SET date_modified = (cast(strftime('%s', 'now') as int)) WHERE id = NEW.id;
     END;
 
+-- Views use LEFT JOINs deliberately: groups can be removed from entries at
```
PromoFaux, line 109: Just a nitpick - we've used `/*` ... `*/` for much smaller comments further down the file, can we either:
 - Do the same here?
 - Use `--` below

PromoFaux: Forget it, I'm not going to lose any sleep over it

PromoFaux (approved):
I can't say I understand 100% of the changes but I can say that I've read 100% of the changes, and there are no show stoppers for me. My only remaining query is around a change 

My only remaining question is around the `info.c` change from getgrgid(st.st_uid) to getgrgid(st.st_gid) which looks correct (**u**id -> **g**id when referring to group), but I don’t see it described in this commit message. The message talks only about stats_database.c. Was the message accidentally carried over during a rebase/amend?

Not  a blocker -just curious

DL6ER:
Another possibility is that I saw it while I was reading the code and fixed it and, because I was doing something else at that moment, then forgot about having done this when I finished my changes.

dschaper:
Damn!

PR #2820, opened 2026-03-25
Extend CI tests

DL6ER:

 - Port and create new (in total now ~120) API integration tests, existing ones are ported from the difficult to maintain inline BATS `curl | jq` assertions to native `pytest` with requests, covering HTTP errors, config validation, domain search, history, lists, queries, Lua pages, auth workflow, stats, network, logs, PADD, and OpenAPI spec validation
 - Change all test from simple assertions (like `.total > 0`) to always use exact numbers everywhere (like `.total == 137`) to make the tests much tighter and precise. Any change that may even cause only a digit to change, will now be caught by the tests.
 - Fix a double-free bug in `printFTLenv()` which was however, covered and cannot lead to a crash (so mainly a cosmetic issue)
 - Fix our `ResponseVerifyer` to respect OpenAPI required fields instead of reporting all absent optional properties as errors
 - Restructure test runner to execute BATS (DNS/regex/CLI/config) -> `pytest` (API) -> final BATS (log validation + FTL termination) in sequence

### Test infrastructure changes

 - `test/run.sh`: Reordered to run BATS core tests first, then `pytest` API tests, then a final BATS pass for log validation.
    GDB and perf tests are now opt-in.
 - `build.sh`: Added `test-fast` target that sets skips the final time-consuming performance test
 - `test/test_suite.bats`: Removed ~300 lines of API tests (now in `pytest`), improve polling for async DB export in query ID 0 test, consolidated CLI password set/remove into a single idempotent test.

Overall, this makes the testing infrastructure much more maintainable and the test at the same time much more tight.
---

---

## Checklist:

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -89,13 +89,6 @@ if ! su pihole -s /bin/sh -c /home/pihole/pihole-FTL; then
   exit 1
 fi
 
-# Prepare BATS
-if [ -z "$BATS" ]; then
-  mkdir -p test/libs
-  git clone --depth=1 --quiet https://github.com/bats-core/bats-core test/libs/bats > /dev/null
-  BATS=test/libs/bats/bin/bats
-fi
-
 # Give FTL some time for startup preparations
 sleep 2
 
@@ -111,12 +104,43 @@ echo "FTL verbose version (CLI): "
 echo -n "Contained dnsmasq version (DNS): "
 dig TXT CHAOS version.bind @127.0.0.1 +short
 
-# Install py3-dnspython
-apk add --no-cache py3-dnspython
+# Install Python test dependencies
+apk add --no-cache py3-dnspython py3-requests py3-pytest py3-yaml
 
-# Run tests
-$BATS -p "test/test_suite.bats"
+# Pre-warm DNSSEC root key cache. dnsmasq's DNSSEC validation can
+# trigger internal DNSKEY queries for the root zone at unpredictable
+# times. By explicitly querying DNSKEY for "." first, we force the
+# root key into cache so all subsequent DNSSEC validation uses the
+# cached key. This makes the total query count deterministic.
+dig DNSKEY . @127.0.0.1 +dnssec > /dev/null 2>&1
+sleep 1
+
+RET=0
+
+# Run pytest API tests first (OpenAPI validation, rate limiting).
+# These must run before BATS because the BATS suite terminates FTL
+# as its final test.
+echo "Running pytest API tests..."
+python3 -m pytest test/api/ -v
 RET=$?
+
+# Run BATS test suite (skipped when SKIP_BATS=1, e.g. via "test-api")
+if [[ "${SKIP_BATS}" != "1" ]]; then
+  if [ -z "$BATS" ]; then
+    mkdir -p test/libs
+    git clone --depth=1 --quiet https://github.com/bats-core/bats-core test/libs/bats > /dev/null
+    BATS=test/libs/bats/bin/bats
+  fi
+
+  echo "Running BATS test suite..."
+  $BATS -p "test/test_suite.bats"
+  BATS_RET=$?
+
+  # Combine exit codes: fail if either suite failed
+  if [[ $BATS_RET != 0 ]]; then
+    RET=$BATS_RET
+  fi
+fi
 
 curl_to_tricorder() {
   curl --silent --upload-file "${1}" https://tricorder.pi-hole.net
@@ -146,9 +170,11 @@ fi
 # Restore umask
 umask "$OLDUMASK"
 
-# Run performance tests
-if ! su pihole -s /bin/sh -c "/home/pihole/pihole-FTL --perf"; then
-  echo "pihole-FTL --perf failed to start"
+# Run performance tests (skipped when SKIP_PERF_TEST=1, e.g. via "test-fast")
+if [[ "${SKIP_PERF_TEST}" != "1" ]]; then
+  if ! su pihole -s /bin/sh -c "/home/pihole/pihole-FTL --perf"; then
+    echo "pihole-FTL --perf failed to start"
+  fi
 fi
 
 # Remove copied file
--- a/test/run.sh
+++ b/test/run.sh
@@ -129,6 +129,12 @@ echo "Running BATS test suite..."
 $BATS -p "test/test_suite.bats"
 RET=$?
 
+# Trigger network table update (PARSE_NEIGHBOR_CACHE) so mock-hwaddr
+# devices like ip-127.0.0.1 exist before pytest checks them.
+# RT signal offset 5 maps to PARSE_NEIGHBOR_CACHE in signals.c.
+kill -SIGRTMIN+5 "$(cat /run/pihole-FTL.pid)" 2>/dev/null
+sleep 2
+
 # Run pytest API tests (FTL is still running — BATS no longer terminates it)
 # Skip on riscv64 — the emulated runner is too slow for the full API suite
 if [[ "${CI_ARCH}" != "linux/riscv64" ]]; then
--- a/test/run.sh
+++ b/test/run.sh
@@ -124,7 +124,7 @@ if [ -z "$BATS" ]; then
 fi
 
 # Run BATS test suite (includes DNS, regex, CLI, config tests;
-# terminates FTL as its final test)
+# FTL remains running for the pytest API tests afterwards)
 echo "Running BATS test suite..."
 $BATS -p "test/test_suite.bats"
 RET=$?
--- a/test/run.sh
+++ b/test/run.sh
@@ -187,8 +187,8 @@ fi
 # Restore umask
 umask "$OLDUMASK"
 
-# Run performance tests (skipped when SKIP_PERF_TEST=1, e.g. via "test-fast")
-if [[ "${SKIP_PERF_TEST}" != "1" ]]; then
+# Run performance tests (opt-in via RUN_PERF_TEST=1)
+if [[ "${RUN_PERF_TEST}" == "1" ]]; then
   if ! su pihole -s /bin/sh -c "/home/pihole/pihole-FTL --perf"; then
     echo "pihole-FTL --perf failed to start"
   fi
--- a/test/run.sh
+++ b/test/run.sh
@@ -130,11 +130,16 @@ $BATS -p "test/test_suite.bats"
 RET=$?
 
 # Run pytest API tests (FTL is still running — BATS no longer terminates it)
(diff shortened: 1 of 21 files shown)


PromoFaux:
Risc is never fun.

```
  5/6 ✗ Expected number of config file rotations
     (in test file test/test_final.bats, line 38)
       `[[ ${lines[0]} == "6" ]]' failed
     pihole.toml write count: 1
```

test/test_final.bats
```diff
@@ -0,0 +1,66 @@
...
+  [[ ${lines[0]} == "0" ]]
+}
+
+@test "Expected number of config file rotations" {
+  # BATS:   1x pihole.toml write (dns.reply.host API PATCH)
+  # BATS:   2x pihole.toml writes (CLI password set/remove processes)
+  # pytest: 3x pihole.toml writes (password, app_pwhash, serve_all via API)
+  run bash -c 'grep -c "INFO: Config file written to /etc/pihole/pihole.toml" /var/log/pihole/FTL.log'
+  printf "pihole.toml write count: %s\n" "${lines[0]}"
+  [[ ${lines[0]} == "6" ]]
```
PromoFaux:
```suggestion
  # On RISCV64, pytest is skipped (too slow), so only BATS writes occur
  if [[ "${CI_ARCH}" == "linux/riscv64" ]]; then
    [[ ${lines[0]} == "1" ]]
  else
    [[ ${lines[0]} == "6" ]]
  fi
```

DL6ER: But ... the comment suggests it should be 3.

yubiuser (commented):
Should we have a test that checks deleting an item returns 204?

test/run.sh
```diff
@@ -146,9 +187,11 @@ fi
 # Restore umask
 umask "$OLDUMASK"
 
-# Run performance tests
-if ! su pihole -s /bin/sh -c "/home/pihole/pihole-FTL --perf"; then
-  echo "pihole-FTL --perf failed to start"
+# Run performance tests (skipped when SKIP_PERF_TEST=1, e.g. via "test-fast")
+if [[ "${SKIP_PERF_TEST}" != "1" ]]; then
+  if ! su pihole -s /bin/sh -c "/home/pihole/pihole-FTL --perf"; then
```
yubiuser, line 189: Do we need those tests at all? All other tests are true/false, but the performance test return performance numbers  - we don't judge on them in any way here.

DL6ER: Yes, let me make them opt-in

DL6ER:
Thank you for the suggestion, I decided to go beyond it and add much more additional test coverage, it became four commits:

`fix: sync test gravity.db.sql schema with production database`
While adding PUT/DELETE tests, I discovered the test `gravity.db.sql` had drifted significantly from the real schema: missing composite UNIQUE constraints (`UNIQUE(domain,type)` / `UNIQUE(address,type)`), missing `ON DELETE CASCADE`, a `NOL NULL` typo, incorrect views, and missing indexes. This caused `INSERT ... ON CONFLICT` to silently fail. The schema now matches `/etc/pihole/gravity.db` exactly.

`chore: make performance tests opt-in, increase CI timeouts`
`./build.sh test` no longer runs `--perf` by default. Use `./build.sh test-perf` to include it. Also bumped CI QEMU timeouts (build 15→30 min, test 10→20 min) for RISC-V.

`fix: return HTTP 400 when all gravity DB inserts fail`
Found that `api_list_write()` returned HTTP 200/201 even when every `gravityDB_addToTable()` call failed — the error was captured in `processed.errors` but the status code indicated success. Now returns 400 with `database_error` when no items succeed. Also added the `database_error` example to the PUT 400 specs for groups, clients, and lists (domains already had it).

`test: add PUT/DELETE, batch delete, query filter, and auth tests`
~80 new tests bringing the total to 130, covering:
- DELETE 204/404 for all major resources (groups, domains, clients, lists, config array items)
- PUT create/replace round-trips with cleanup
- Batch delete via `POST :batchDelete`
- DNS blocking toggle (disable + re-enable)
- Auth logout (`DELETE /api/auth`) and session delete by ID
- Missing GET endpoints (clients, config, gateway, routes, sensors, metrics, host)
- Query filters (domain, client_ip, upstream address, `blocklist` pseudo-upstream)
- Cursor pagination, query suggestions, partial search
- Stats database endpoints with valid time ranges
- TOTP credential suggestion
- Config PATCH round-trip (bool + integer, change → verify → restore)
- NTP server protocol-level test (UDP NTPv4 request/response with ≤2s accuracy check)

test/run.sh
```diff
@@ -111,13 +103,62 @@ echo "FTL verbose version (CLI): "
...
+# devices like ip-127.0.0.1 exist before pytest checks them.
+# RT signal offset 5 maps to PARSE_NEIGHBOR_CACHE in signals.c.
+kill -SIGRTMIN+5 "$(cat /run/pihole-FTL.pid)" 2>/dev/null
+sleep 2
+
+# Run pytest API tests (FTL is still running — BATS no longer terminates it)
+# Skip on riscv64 — the emulated runner is too slow for the full API suite
+if [[ "${CI_ARCH}" != "linux/riscv64" ]]; then
+  echo "Running pytest API tests..."
+  python3 -m pytest test/api/ -v
```
yubiuser, line 139:
```suggestion
  python3 -m pytest -n auto test/api/ -v
```

test/run.sh
```diff
@@ -111,13 +103,62 @@ echo "FTL verbose version (CLI): "
 echo -n "Contained dnsmasq version (DNS): "
 dig TXT CHAOS version.bind @127.0.0.1 +short
 
-# Install py3-dnspython
-apk add --no-cache py3-dnspython
+# Install Python test dependencies
+apk add --no-cache py3-dnspython py3-requests py3-pytest py3-yaml
```
yubiuser:
```suggestion
apk add --no-cache py3-dnspython py3-requests py3-pytest py3-yaml py3-pytest-xdist
```

yubiuser (commented):
Might run a bit faster with more cores.

Should the python dependencies moved to the docker-base-image?

DL6ER:
Unfortunately, parallel test execution (`-n auto`) would break the current suite — the tests mutate shared state on a single running FTL instance and rely on execution order:

- **`test_z_auth.py`** is explicitly order-dependent (tests 01->10): create app password -> set it -> login -> set regular password -> verify wrong/correct -> rate-limit -> remove password. The tests share state via class variables, so splitting them across workers would break the chain.
- **`test_m_mutations.py`** does PUT -> verify -> DELETE -> verify sequences against the live server. Concurrent mutations would cause nondeterministic failures.
- **`test_api.py`** asserts exact counts (e.g., `len(groups) == 6`). A parallel mutation test temporarily adding a group would break these assertions.
- All tests hit the same FTL instance at `127.0.0.1` - there's no per-worker isolation for DNS blocking toggles, config patches, or auth state changes.

The full pytest suite runs in ~15 seconds on my machine, so parallelism doesn't buy much here. Making the tests xdist-safe would require either significant redesign (fully independent tests with per-test setup/teardown of the entire server state) or careful xdist grouping that would add significant amounts of work and needs reconsideration for each new test we add in the future. Also, parallelism can cause falky tests as their random order cannot reliably be tested in all flavors.

Just think about running the auth test (setting a password) while another test API test is running (that is not aware of a password): Test would fail as API would require a login. Such flaky tests would then lead to an entire of the entire test step (including setting up the test harnmess + 2x BATS) so this will for sure be slower in sum than non-parallel python tests.

And re the dependencies: Yes, for sure. Let's see if someone beats me to it or if I will forget (whatever comes first :smirk: )

test/run.sh
yubiuser: Valid reason not to add parallelism written below

test/run.sh
yubiuser: Valid reason not to add parallelism written below

yubiuser:
> And re the dependencies: Yes, for sure. Let's see if someone beats me to it or if I will forget (whatever comes first 😏 )

https://github.com/pi-hole/docker-base-images/pull/153

test/run.sh
```diff
@@ -80,28 +80,20 @@ export FTLCONF_debug_api="not_a_bool"
...
-fi
-
 # Give FTL some time for startup preparations
 sleep 2
 
-# Attach debugger and immediately continue running the binary
-# In case a non-ignored signal occurs (a crash), create a full backtrace
-gdb -p $(cat /run/pihole-FTL.pid) --ex continue --ex "bt full" &
+# Optionally attach gdb for crash backtraces (opt-in via GDB=1)
+if [[ "${GDB}" == "1" ]]; then
```
yubiuser, line 93: Why did you make this opt-in? Because we have automated back-trace now?

PromoFaux:
> https://github.com/pi-hole/docker-base-images/pull/153

Merged - should be available as `nightly` if you want to test it against this PR

edit: eventually, anyway... bloody risc

test/run.sh
DL6ER: Because it does not run on the CI anyway (missing capabilities in the containers) and because we have the built-in backtraces now

yubiuser (commented):
`test/libs` should be added to `.gitignore`

DL6ER:
Done. I also added a session concurrency test (which is slowing down the tests notable ~ 10 seconds to circumvent rate-limiting due to many logins). Unfortunately, I was not able to trigger the crash once (this branch does not have the fix) but I'm pretty sure I *could* have done it this way. Do we want to keep them (despite making the CI run slower) or not (as they are already passing right now).

yubiuser (dismissed):
I think we should keep it. ~10 secs is not much but the tests are valuable.

rdwebdesign:
10 seconds increase in tests time is acceptable.
We won't even notice it when compared to riscv64 times... 😄

DL6ER:
@yubiuser sorry for voiding the approval but I managed to redesign the tests to actually FAIL here (more or less) reliably while running through after manually running `git merge 6fc8deed` sourcing the bugfix commit :tada:

yubiuser:
No problem. If this fails at the right spot now I'll still going to approve and looking forward for `6fc8deed` to fix it again.

yubiuser (approved):
```
  FAILED test/api/test_s_auth_stress.py::TestSetCookieRace::test_cookie_echoes_correct_sid
  FAILED test/api/test_s_auth_stress.py::TestSessionListRace::test_listed_sessions_not_corrupted_during_deletes
  FAILED test/api/test_s_auth_stress.py::TestValidateDeleteRace::test_validate_while_deleting_same_session
```

yubiuser (approved):

PR #2831, opened 2026-04-03
Resolve empty backtraces when addr2line is not installed

DL6ER:

When addr2line was absent from the runtime system, popen() succeeded but produced no output, leaving func as an empty string. The check `strcmp(func, "??")` did not match "", so the code fell through to the success branch and logged only bare frame numbers (#0, #1, ...) with no addresses or symbols.

Fix the condition to also check for empty func, so the dladdr() and /proc/self/maps fallbacks execute correctly. Additionally, when any frame cannot be resolved, print the addr2line commands the user can run manually after installing binutils in case the tool was missing.

---

---

## Checklist:

### DIFF
--- a/src/signals.c
+++ b/src/signals.c
@@ -147,17 +147,18 @@ static void find_mapping_name(const void *addr, char *buf, const size_t buflen)
 // Log one backtrace frame as a single line.
 // Resolved:   "  #N  func_name                    src/file.c:line"
 // Unresolved: "  #N  0xADDRESS  (reason)"
-static void log_frame(const int idx, const void *addr, const void *rel_addr)
+// Returns true if addr2line resolved the frame, false otherwise.
+static bool log_frame(const int idx, const void *addr, const void *rel_addr)
 {
 	if(!config.misc.addr2line.v.b)
 	{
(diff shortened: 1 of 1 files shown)


yubiuser:
> binutils in case the tool was missing.

Have you seen this package being missed in the wild? I just checked and it's part of all supported distos `pihole-meta` package and we install in in docker as well.

DL6ER:
I have no better explanation as to why the backtrace was completely empty in the referenced issue.

yubiuser (commented):
There is one CodeQL failure: https://github.com/pi-hole/FTL/runs/70673223990?pr=2831

DL6ER:
@yubiuser It is a false-positive. The CodeQL complaint is that we build a command using
```
addr2line -f -e \"%s\" %p
```
where `%s` is the binary's name (`pihole-FTL` in our case). We get it from `argv[0]`. It is the same you get for
``` bash
echo $0
```
or similar. This is *guaranteed* to be the binary name where dangerous stuff like `;` or similar is impossible in a binary's name. Hence, I flagged this false-positive.

yubiuser (approved):

PR #2834, opened 2026-04-05
Preserve log file path config when fopen fails

### LINKED ISSUE #2827: Missing CAP_CHOWN results in files.log.ftl to be empty
### Actual behavior / bug

When `CAP_CHOWN` is missing, FTL will set `files.log.ftl` empty. It stays empty even after re-adding the cap.
This can be reproduced easily in docker with adding

```
    cap_drop:
      - CAP_CHOWN
```

### Expected behavior

`files.log.ftl` should stay untouched

DL6ER:

When `CAP_CHOWN` is missing (e.g. in Docker with `cap_drop`), `init_FTL_log()` could fail to open the log file and set `config.files.log.ftl.v.s = NULL`. This `NULL` value was then persisted as an empty string when the TOML config was rewritten shortly after during `readFTLconf()`. The empty path remained even after restoring `CAP_CHOWN`, permanently breaking file logging.

Use a separate static flag to track log file availability instead of clobbering the config value, so the configured path survives fopen failures and is correctly written back to `pihole.toml`

---

**Related issue or feature (if applicable):** Fixes: #2827

---

## Checklist:

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -59,6 +59,8 @@ void init_FTL_log(const char *name)
 			syslog(LOG_ERR, "Opening of FTL\'s log file failed, using syslog instead!");
 			ftl_log_available = false;
 		}
+		else
+			ftl_log_available = true;
 
 		// Close log file
 		if(logfile != NULL)
--- a/src/log.c
+++ b/src/log.c
@@ -30,6 +30,7 @@
 #include "gc.h"
 
 static bool print_log = true, print_stdout = true;
+static bool ftl_log_available = true;
 static const char *process = "";
 bool debug_flags[DEBUG_MAX] = { false };
 
@@ -56,7 +57,7 @@ void init_FTL_log(const char *name)
 			printf("ERROR: Opening of FTL log (%s) failed: %s\nUsing syslog instead!\n",
 			       config.files.log.ftl.v.s, strerror(errno));
 			syslog(LOG_ERR, "Opening of FTL\'s log file failed, using syslog instead!");
-			config.files.log.ftl.v.s = NULL;
+			ftl_log_available = false;
 		}
 
 		// Close log file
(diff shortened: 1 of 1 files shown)


src/log.c
```diff
@@ -56,7 +57,7 @@ void init_FTL_log(const char *name)
 			printf("ERROR: Opening of FTL log (%s) failed: %s\nUsing syslog instead!\n",
 			       config.files.log.ftl.v.s, strerror(errno));
 			syslog(LOG_ERR, "Opening of FTL\'s log file failed, using syslog instead!");
-			config.files.log.ftl.v.s = NULL;
+			ftl_log_available = false;
 		}
```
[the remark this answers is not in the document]
DL6ER: Resolved via [32e3d66](https://github.com/pi-hole/FTL/pull/2834/commits/32e3d663883c7bd77abab04247e35051a30b1662)

src/log.c
```diff
@@ -56,8 +57,10 @@ void init_FTL_log(const char *name)
 			printf("ERROR: Opening of FTL log (%s) failed: %s\nUsing syslog instead!\n",
```
yubiuser, line 57: Where is this logged to when the log file is not available?

rdwebdesign: I guess this line prints to _stdout_ and the next line uses syslog:

https://github.com/pi-hole/FTL/blob/32e3d663883c7bd77abab04247e35051a30b1662/src/log.c#L56-L61

yubiuser: I tested the branch with our docker container and it was not printed. But maybe it would on bare metal

rdwebdesign: I started my previous answer with "I guess"... maybe I'm wrong.

DL6ER: @rdwebdesign is right: line 57 goes to stdout and line 59 to syslog. init_FTL_log() runs from main() before we daemonize and before log_ctrl() silences stdout, so both streams are actually open at that point.

The reason you may not have seen anything in the container:
- The Pi-hole image doesn't run a syslog daemon (and `/dev/log` isn't bind-mounted), so `syslog()` likely silently dropped the message.
- `stdout` is captured by s6 and should show up in `docker logs pihole`, but it won't appear in `pihole -t` or the web log viewer - those tail `pihole-FTL.log`, which is exactly the file we failed to open.

On bare metal you'd see it on the console and in `journald`, which is why it looks different there.

Actually, this PR doesn't change where the error surfaces - it only stops us from clobbering the configured path on `fopen()` failure.

yubiuser (approved):

PR #2865, opened 2026-04-25
dnsmasq: check address= rules before following cached CNAME chain

pwnflakes:

Fixes `address=` rules being bypassed when a subdomain has a cached CNAME entry from a prior AAAA query.

When `address=/example.com/192.168.0.1` is configured, subdomains that resolve upstream via a CNAME chain intermittently return the real upstream IP instead of the configured address. The bypass is triggered by this sequence:

1. **A query** for `api.example.com` — `lookup_domain` finds the `address=` rule, returns the configured IP correctly ✓
2. **AAAA query** for `api.example.com` — no IPv6 `address=` rule exists, query goes upstream, `api.example.com CNAME → something.cdn.net` gets written into the cache ✗
3. **Subsequent A query** — `answer_request` finds the cached CNAME at the top of its lookup loop and follows the chain, returning the CDN's real IP. `lookup_domain` is never reached ✗

The same bypass can occur when another network client has already queried the subdomain and populated the shared cache with the CNAME.

This is distinct from the existing deep CNAME inspection for blocklists — `FTL_CNAME` calls `FTL_check_blocking` for each CNAME hop but never checks `address=` rules. The same issue has been reported twice to the upstream dnsmasq mailing list but I don't believe anything has changed from this being raised:
- [April 2024](https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2024q2/017559.html)
- [September 2025](https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2025q3/018320.html)

In the CNAME-following loop of `answer_request` (`rfc1035.c`), before following a cached CNAME chain for `A` or `AAAA` queries, `lookup_domain` is called for the queried name. If an `address=` rule matches, the configured local address response is returned directly and the loop breaks. This mirrors the existing `forward_query` path: `lookup_domain` finds the matching configured entries, `is_local_answer` confirms they apply to the requested address family, and all matching configured addresses for that family are emitted.

Verified on a Pi 3B+ running Pi-hole v6.6.1:

1. Configure `address=/imgur.com/192.168.8.3`, restart FTL to flush cache
2. `dig AAAA apidocs.imgur.com` — upstream returns `apidocs.imgur.com CNAME phs.getpostman.com`, CNAME written to cache
3. `dig A apidocs.imgur.com` — returns `192.168.8.3` (0ms, served from cache) ✓

Without this fix, step 3 returns the real upstream IP.

No documentation changes required.

---

---

### DIFF
--- a/src/dnsmasq/rfc1035.c
+++ b/src/dnsmasq/rfc1035.c
@@ -1729,6 +1729,56 @@ size_t answer_request(struct dns_header *header, char *limit, size_t qlen,
 	  }  
 	
 	cname_target = cache_get_cname_target(crecp);
+
+	/*********** Pi-hole modification ***********
+	   address= rules must take priority over cached CNAME chains. A CNAME
+	   cached via an AAAA query (where no IPv6 address= match exists) would
+	   otherwise cause subsequent A queries to bypass lookup_domain entirely.
+	   Check for a local address= match before following the chain. */
+	if (qtype == T_A || qtype == T_AAAA)
+	  {
+	    int addr_flags = (qtype == T_AAAA) ? F_IPV6 : F_IPV4;
+	    int first_addr, last_addr, addr_index;
+	    if (lookup_domain(name, addr_flags, &first_addr, &last_addr) &&
+		(is_local_answer(now, first_addr, name) & addr_flags))
+	      {
+		union all_addr local_addr;
+		ans = 1;
+		sec_data = 0;
+		auth = 0;
+		if (addr_flags & F_IPV4)
+		  {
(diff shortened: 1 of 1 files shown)


DL6ER:
We have a strict rule to never modify upstream code when this can be avoided. This is mainly to ensure we remain upstream-compatible and do not end in conflict-hell when trying to import future patches. Please submit your change to dnsmasq-discuss@lists.thekelleys.org.uk, instead. We have a special fast-track for `dnsmasq` patches following their `master`. Your patch will land here, too, after it has been integrated upstream.

Issue #2849, opened 2026-04-12
`pihole-FTL --config` sometimes returns the wrong value if `sudo` is not used


rdwebdesign:
### Expected behavior

`pihole-FTL --config <option>` should return the current value of `<option>`.
If FTL can't read the current value, an error message or warning should be returned.

### Actual behavior / bug

In some cases, when `sudo` is not used, the command `pihole-FTL --config <option>` returns the default value for `<option>`, instead of the current selected value. 
No error or warning messages are shown. Just the default value is returned.

Repeating the same command with `sudo` (`sudo pihole-FTL --config <option>`), returns the expected value.

This is a problem when Pi-hole user expects to see the current selected value, but is using an OS user without enough permissions and forgets to add `sudo`.

**Note:** the issue depends on the user permissions and it happens only in some cases. Some tests returned the expected value even without `sudo`.

DL6ER:
> Note: the issue depends on the user permissions and it happens only in some cases. Some tests returned the expected value even without sudo.

Yes. I guess the cases it is not working is when `/etc/pihole/pihole.toml` is correctly owned by user `pihole` with proper permissions and the user they are using is not able to read the file. `sudo` trivially bypasses any permission checks.

What is the expected behavior? An error? Falling back to the default when the config file cannot be read is some sensible way, too, but I agree that this can also cause confusion.

rdwebdesign:
> What is the expected behavior? An error? 

Personally, I think the intend of `pihole-FTL --config <option>` command is to return the **current** value of an option.

In this case, we must return an error message about lack of permissions, when `pihole.toml` can't be read.

> Falling back to the default when the config file cannot be read is some sensible way

I agree this is useful in some cases, but with the current behavior FTL is just giving a `returned_value` and users can't differentiate between the current value or the default value.
It's impossible to know the permissions failed and the default value was returned.

In this case, we need to at least show a warning saying something like:
```
returned_value
Warning: pihole.toml can't be read due to missing permissions. Default value was used.
```

rdwebdesign:
I was able to replicate the issue with the following steps:

1. install Pi-hole on a fresh OS (I used a VM with Debian 13)

2. change an option (example: `sudo pihole-FTL --config webserver.port '80'`, or edit the file manually).  
   Everything works as expected:
   - `pihole-FTL --config webserver.port` returns `80`.
   - `sudo pihole-FTL --config webserver.port` returns `80`.
[...]

PR #2873, opened 2026-04-29
Revert "Clarify when --config cannot read the config file"

PromoFaux:
I think we should revert this change...

Reasoning:

 - Tests in core and docker are now broken
 - Fresh installs are broken (i'm even running as root!)
 
image

Examining the `readFTLConf` function...

 - There is only one path that returns `true` which is path where the file is readable
 - If we don't explicitly call it with `rewrite=true` then we will return false (And FTL will log that we are using defaults)
 - If we call with `rewrite=true` then the file is created but for some reason we then return `false`.... this does not make sense to me. If we have created the file, why not return `true`?

IMO the true fix to @rdwebdesign's [original issue](https://github.com/pi-hole/FTL/issues/2849) thread would be to explicitly test whether the file is unreadable because either:
  1) It does not exist (maybe it does not exist + neither do backups)
		- In which case return `true`, use defaults as expected
  2) There are permissions issues
		- In which case return `false` - the file exists but there are issues reading it

Function as exists is below so you don't have to go searching for it:

```c
bool readFTLconf(struct config *conf, const bool rewrite)
{
	// Initialize config with default values
	initConfig(conf);

	// First, read the environment
	getEnvVars();

	// Try to read TOML config file
	// If we cannot parse /etc/pihole.toml (due to missing or invalid syntax),
	// we try to read the rotated files in /etc/pihole/config_backup starting at
	// the most recent one and going back in time until we find a valid config
	for(unsigned int i = 0; i < MAX_ROTATIONS; i++)
	{
		toml_datum_t toml = { 0 };
		if(readFTLtoml(NULL, conf, toml, rewrite, NULL, i, false))
		{
			// If successful, we write the config file back to disk
			// to ensure that all options are present and comments
			// about options deviating from the default are present
			if(rewrite)
			{
				writeFTLtoml(true, NULL);
				char errbuf[ERRBUF_SIZE] = { 0 };
				write_dnsmasq_config(conf, false, errbuf);
				write_custom_list();
			}
			return true;
		}
	}

	log_info("No config file nor backup available, using defaults");

	// If we reach this point, we could not read the TOML config file When
	// this functions is invoked to run without rewriting, we are likely
	// running interactively and do not want to migrate settings (yet):
	// using defaults is fine in this case
	if(!rewrite)
		return false;

	// When we reach this point but the FTL TOML config file exists, it may
	// contain errors such as syntax errors, etc. We move it into a
	// ".broken" location so it can be revisited later
	if(file_exists(GLOBALTOMLPATH))
	{
		const char new_name[] = GLOBALTOMLPATH ".broken";
		rotate_files(new_name, NULL);
		rename(GLOBALTOMLPATH, new_name);
	}

	// Determine and set default webserver ports if not imported from
	// setupVars.conf
	get_web_port(&config);

	// Initialize the TOML config file
	writeFTLtoml(true, NULL);
	char errbuf[ERRBUF_SIZE] = { 0 };
	write_dnsmasq_config(conf, false, errbuf);
	write_custom_list();

	return false;
}
```

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -364,15 +364,7 @@ void parse_args(int argc, char *argv[])
 		// Enable stdout printing
 		cli_mode = true;
 		log_ctrl(false, false);
-		if(!readFTLconf(&config, false))
-		{
-			fprintf(stderr,
-			        "Error: Could not read or parse config file \"%s\" or its backup files. "
-			        "Please check that the file is accessible and that permissions are correct, "
-			        "or try running with sudo.\n",
-			        GLOBALTOMLPATH);
-			exit(EXIT_FAILURE);
-		}
+		readFTLconf(&config, false);
 		log_ctrl(false, true);
 		clear_debug_flags(); // No debug printing wanted
 		if(argc == 2)
--- a/test/test_suite.bats
+++ b/test/test_suite.bats
@@ -778,28 +778,6 @@ setup() {
   [[ "${lines[0]}" == "80o,443os,[::]:80o,[::]:443os" ]]
 }
 
-@test "CLI --config fails when config and backups are unavailable" {
-  run bash -c '
-    config_backup=/etc/pihole/pihole.toml.batsbak
-    backup_dir_backup=/etc/pihole/config_backups.batsbak
-    cleanup() {
-      if [[ -e "$config_backup" ]]; then
-        mv "$config_backup" /etc/pihole/pihole.toml || { echo "Failed to restore /etc/pihole/pihole.toml" >&2; return 1; }
-      fi
-      if [[ -e "$backup_dir_backup" ]]; then
-        mv "$backup_dir_backup" /etc/pihole/config_backups || { echo "Failed to restore /etc/pihole/config_backups" >&2; return 1; }
-      fi
-    }
-    trap cleanup EXIT
-    mv /etc/pihole/pihole.toml "$config_backup" || { echo "Failed to move /etc/pihole/pihole.toml for test setup" >&2; exit 1; }
-    mv /etc/pihole/config_backups "$backup_dir_backup" || { echo "Failed to move /etc/pihole/config_backups for test setup" >&2; exit 1; }
-    ./pihole-FTL --config debug.regex 2>&1
-  '
-  printf "%s\n" "${lines[@]}"
-  [[ $status -eq 1 ]]
-  [[ ${lines[0]} == *'Could not read or parse config file "/etc/pihole/pihole.toml" or its backup files.'* ]]
-}
-
 # NOTE: Log validation (WARNING/ERROR/CRIT/DB checks) moved to the final
 # log scan in run.sh, which runs after both BATS and pytest complete.
 


PromoFaux:
I don't really understand why we wouldn't want --config to return default values...

rdwebdesign:
> I don't really understand why we wouldn't want --config to return default values...

I understand the intention of using `pihole-FTL --config` command to get default values, but maybe we're "abusing" the command when we use it like this.

We need to define what is the real job of this command.

Users expect to see the value set on `pihole.toml`.
`pihole-FTL --help` says this command should return the **current** value of the option:

image

I agree that returning the default value is useful.
I'm just not sure if this should be done by this command. Maybe we need another option.

In my opinion:
 - if the intention is to *"return the current value set on the file"*, then the fix in **PR2855** is correct. 
   In this case, maybe we need a different option to return the default value if the file is not readable.
 - if the intention of this command is to *"always return some value, even if there is no file"*, then we need to fix the command description.

rdwebdesign:
@DL6ER 

Do you have any suggestions?

Should `pihole-FTL --config` return default values without `sudo`? 
Or do you have a different suggestion?

PromoFaux:
If the current value _is_ the default value, then it's still the current value.

> Users expect to see the value set on pihole.toml.

The description of the command is _already correct_.

If no file exists, then FTL will be creating one and using it's default values and so returning the default values is not only the correct thing to do but it also in line with the --help output.

The confusion you are describing comes from the (rare?) case where the permissions of the toml file are mismatched and so `piholeFTL --config` is returning the default values because it is not allowed (?!) to read it's own config file and so is returning the default _in place of_ the custom configured value.

The proper fix, as stated above, is to explicitly check if we can't read because of:
1) Permissions
2) non-existence of the file.

If 1 then return with an error message
if 2 then the file need to be created anyway... and so on initialisation the configured value will be the default value

I stand by it, #2855 has broken more than it fixes and should be reverted until a proper solution is found.

DL6ER:
I fully agree with @PromoFaux. The blast radius is just too wide. Printing an error in case the file has permissions will make others sad, too, as this will deviate from the current output-is-single-line but we can simply re-route it to `stderr` so any script already doing `2>/dev/null` will not actually see any change.

If you don't agree or think we should do something else, feel free to discuss but I concur that the cleanest strategy for now is reverting and then rethinking.

DL6ER (approved):

rdwebdesign:
>  I concur that the cleanest strategy for now is reverting and then rethinking.

I agree we need a solution to the broken tests. We need to rethink how to solve the issue that happens when the permissions are wrong.

rdwebdesign:
> but we can simply re-route it to `stderr` so any script already doing `2>/dev/null` will not actually see any change.

This is a good solution.

Maybe my point wasn't clear during the discussion. 

I don't have a preference for the output showing an error message, instead of a single value.
I just want to fix the issue that happens when the permissions doesn't match, without breaking other stuff, of course.

PR #2043, opened 2024-08-26
Fix possible double-free and avoid memory leaking

DL6ER:
# What does this implement/fix?

This PR fixes a possible double `free`s that can crash FTL. It can happen when a client hostname contains invalid characters (e.g., spaces), reported and confirmed [here](https://discourse.pi-hole.net/t/pihole-6-0-beta-and-sqlite-issue/71586/26).

The issue we are dealing with here is ensuring that the `allocated = yes/no` for config strings has been checked at every location where memory is allocated and free'd. In some places, we had to change things to avoid memory being leaked.

---

**Related issue or feature (if applicable):** see above

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** N/A

---
**By submitting this pull request, I confirm the following:** 

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against. 
2. I have commented my proposed changes within the code.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

## Checklist:

- [x] The code change is tested and works locally.
- [x] I based my code and PRs against the repositories `developmental` branch.
- [x] I [signed off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits. Pi-hole enforces the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
- [x] I [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all my commits. Pi-hole requires signatures to verify authorship
- [x] I have read the above and my PR is ready for review.

PR #2875, opened 2026-05-02
fix: check NULL returns from strdup/calloc in rotate_files()

jluzzi123:

**Please make sure you**

---

`rotate_files()` in `src/files.c` calls `strdup()` and `calloc()` four times in its loop body without checking the return values for NULL. On a system under memory pressure, which is more realistic on Pi hardware than on a typical server, any of those failures leads to either a NULL pointer dereference (crash) or a NULL being silently passed back to the caller through `*first_file`.

This is proactive hardening rather than a fix for a specific crash report. The same NULL-check pattern was already applied in #2043 for similar allocations in `src/config/` and `src/resolve.c`, so this brings `rotate_files()` in line with that existing approach.

Four NULL guards are added in the order the allocations occur:

- `strdup(path)` to `fname`: log error and return
- first `calloc()` to `old_path`: log error, free `fname`, return
- second `calloc()` to `new_path`: log error, free `fname` and `old_path`, return
- `strdup(new_path)` to `*first_file`: log error and continue. The rotation still proceeds, the caller just does not receive the first file path back

The cleanup order is intentional: `filename` points into `fname`'s buffer via `basename()`, so `fname` is freed only after `filename` is no longer referenced.

No documentation changes needed.

---

---

### DIFF
--- a/src/files.c
+++ b/src/files.c
@@ -487,22 +487,44 @@ void rotate_files(const char *path, char **first_file)
 	{
 		// Construct old and new paths
 		char *fname = strdup(path);
+		if(fname == NULL)
+		{
+			log_err("rotate_files(): strdup() failed: %s", strerror(errno));
+			return;
+		}
 		const char *filename = basename(fname);
 		// extra 6 bytes is enough space for up to 999 rotations ("/", ".", "\0", "999")
 		const size_t buflen = strlen(filename) + MAX(strlen(BACKUP_DIR), strlen(path)) + 6;
 		char *old_path = calloc(buflen, sizeof(char));
+		if(old_path == NULL)
+		{
+			log_err("rotate_files(): calloc() failed: %s", strerror(errno));
+			free(fname);
+			return;
+		}
 		if(i == 1)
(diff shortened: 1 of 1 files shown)


PromoFaux:
@jluzzi123 - no issues with generative AI assistance as such, but all PR's follow the default pull request template.

Also, does this change relate to an existing issue report, or is it purely speculative?

jluzzi123:
Hi @PromoFaux, sorry about that. I've updated the description to follow the template.

To answer your question: no crash report behind this. I was reading through `src/files.c` and noticed those four allocations didn't have NULL checks on the returns. Given that FTL runs on Pi hardware where memory pressure is a more realistic concern than on a typical server, and that the same pattern was already applied in #2043 for other files in the codebase, it seemed like a reasonable gap to close. The change only touches the failure paths. The success path is untouched.

Understand if proactive hardening isn't something you want to pull in right now.

DL6ER (approved):
If we are under memory pressure, more serious trouble is ahead but having this extra defensiveness isn't bad.

PR #2878, opened 2026-05-03
Make error messages more human friendly

yubiuser:
**What does this PR aim to accomplish?:**

Makes error messages more human friendly, as not everyone understands `UNIQUE constraint failed`
Raised in https://github.com/pi-hole/web/issues/3773#issuecomment-4365942542

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.2)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #2879, opened 2026-05-05
Use custom message for UNIQUE constraint error message

rdwebdesign:

Currently, when an "UNIQUE constraint" error is found while adding items to group management tables, the returned message is not very user friendly.

This PR uses a custom error message.

image

Redo of https://github.com/pi-hole/FTL/pull/2878
Suggested in https://github.com/pi-hole/web/issues/3773#issuecomment-4365942542

Testing the returned error code.

If the error code is `SQLITE_CONSTRAINT` (19), we show a custom message.
We still show the original message for other error codes.

---

---

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -2020,7 +2020,10 @@ bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
 	}
 	else
 	{
-		*message = sqlite3_errmsg(gravity_db);
+		if(rc == SQLITE_CONSTRAINT)
+			*message = "The item is already present";
+		else
+			*message = sqlite3_errmsg(gravity_db);
 	}
 
 	// Finalize statement and close database handle
--- a/src/api/docs/content/specs/clients.yaml
+++ b/src/api/docs/content/specs/clients.yaml
@@ -484,7 +484,7 @@ components:
             error:
               key: "database_error"
               message: "Could not add to gravity database"
-              hint: "Client already exists (UNIQUE constraint failed)"
+              hint: "The item is already present"
             took: 0.003
   parameters:
     client:
--- a/src/api/docs/content/specs/clients.yaml
+++ b/src/api/docs/content/specs/clients.yaml
@@ -164,7 +164,7 @@ components:
 
           On success, a new resource is created at `/clients/{client}`.
 
-          The `database_error` with message `UNIQUE constraint failed` error indicates that this client already exists.
+          The `database_error` with message `The item is already present` error indicates that this client already exists.
         requestBody:
           description: Callback payload
           content:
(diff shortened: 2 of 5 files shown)


src/database/gravity-db.c
```diff
@@ -2020,7 +2020,10 @@ bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
 	}
 	else
 	{
-		*message = sqlite3_errmsg(gravity_db);
+		if(rc == SQLITE_CONSTRAINT)
+			*message = "Item already in the database (UNIQUE constraint failed)";
```
DL6ER:
```suggestion
			*message = "Record already present";
```
But maybe native speakers have an even better idea. Would need to be adapted in the other changes, too.

rdwebdesign: OK. 
The word "database" appears twice. Once on the title and once on the message. We can remove it here.

Also, the "(UNIQUE constraint failed)" can be removed. If was left here because there is [code that (in some cases) expects this text](https://github.com/pi-hole/web/blob/db6da0645a3094afd148e804e0ad9fc66ab0b551/scripts/js/utils.js#L581-L583), but I don't remember to ever see messages using that code.

---

I would like to suggest a different text, with "item" instead of "record".

The intention of this change is to return user friendly messages, specially for users without any programming or database background.

What do you think about "The item is already present"?
```suggestion
			*message = "The item is already present";
```

rdwebdesign:
@DL6ER 

I changed the messages. I force pushed to avoid unnecessary commits.

**EDIT:**

I noticed there are many other places showing _"UNIQUE constraint failed"_ on the same 4 yaml files, like these:
https://github.com/pi-hole/FTL/blob/144f0449ea848bf07c1207fd04dab19278019d0a/src/api/docs/content/specs/clients.yaml#L167
https://github.com/pi-hole/FTL/blob/144f0449ea848bf07c1207fd04dab19278019d0a/src/api/docs/content/specs/clients.yaml#L437
https://github.com/pi-hole/FTL/blob/144f0449ea848bf07c1207fd04dab19278019d0a/src/api/docs/content/specs/clients.yaml#L460

---

Should we change all of them?

DL6ER:
Yes, I think if we want to change this, we should fix it consistently.

rdwebdesign:
Done.

DL6ER (approved):

DL6ER:
Waiting for https://github.com/pi-hole/FTL/pull/2919 to be merged first or, otherwise, CI tests on `development` would fail

src/database/gravity-db.c
rdwebdesign: Resolved.

rdwebdesign:
> Waiting for *2919* to be merged first

The other PR was merged.
Do we need to wait some other CI tests fix? Or can we merge this one?

PR #2872, opened 2026-04-27
Bats

yubiuser:
**What does this PR aim to accomplish?:**

- adds/uses BATS libraries to the test (https://github.com/bats-core/bats-assert, https://github.com/bats-core/bats-file, https://github.com/bats-core/bats-support) - the functions give a much nicer output if the assertion fails
- removes all ` printf "count: %s\n" "${lines[@]}"` calls which are used for debugging in case the test fails - replaced with the proper `bats::on_failure()` hook

This will need https://github.com/pi-hole/docker-base-images/pull/156 and a new tag of the base image.

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.2)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #2880, opened 2026-05-05
Improve crash backtraces for non-reproducible faults

DL6ER:

Prefer signal-context frame walking in crash handlers, with `_Unwind_Backtrace` fallback. Add mapping safety checks and log the unwind source so field crash reports show whether frames came from signal context or fallback.

---

---

## Checklist:

### DIFF
--- a/src/signals.c
+++ b/src/signals.c
@@ -9,17 +9,22 @@
 *  Please see LICENSE file for your rights under this license. */
 
 #include "FTL.h"
-// Universal unwind backtrace via GCC's libgcc — works on glibc AND musl, static AND dynamic
+// Universal unwind backtrace via GCC's libgcc - works on glibc AND musl, static AND dynamic
 #if defined(USE_UNWIND)
 #  include <unwind.h>
 #  include <limits.h>    // PATH_MAX
-#  include <inttypes.h>  // SCNxPTR — portable sscanf format for uintptr_t
-#  include <sys/mman.h>  // mmap() — used for the intentional crash test subcommand
-#  include <dlfcn.h>     // dladdr() — dynamic symbol lookup from .dynsym
+#  include <sys/mman.h>  // mmap() - used for the intentional crash test subcommand
+#  include <dlfcn.h>     // dladdr() - dynamic symbol lookup from .dynsym
+#  include <ucontext.h>  // ucontext_t - register snapshot from SA_SIGINFO handlers
+#  include <fcntl.h>     // open() - raw, signal-safe /proc/self/maps reading
+#  include <poll.h>      // poll() - bound the addr2line subprocess wall-clock
+#  include <sys/wait.h>  // waitpid()/WIFEXITED() - reap the addr2line child
 #endif
 #include "signals.h"
 // logging routines
 #include "log.h"
+// cli_color() - ANSI colors only when stdout is an interactive terminal
+#include "args.h"
 // ls_dir()
 #include "files.h"
 // gettid()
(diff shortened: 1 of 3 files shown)


DL6ER:
After this round of reviews (last round was only about comment accuracy, no functional requests), I squashed this PR down from 8 into one commit and rebased on `development`.

DL6ER:
A short summary of what changed here and how to check it, now that the review rounds have settled.

The goal of this branch is that a single crash, which in the field is almost always non-reproducible, gives us the most useful data it can, and that trying to symbolize the backtrace can never hang or silence the crash handler.

What changed compared to development:

- Frames are now taken from the signal context (frame-pointer walk) when possible, and we fall back to _Unwind_Backtrace only if that fails. The header notes when frames came "from signal context" (e.g. `Backtrace (5 frames from signal context):`), so a field report tells us how they were obtained.
- Frame #0 is now the actual faulting instruction. We no longer subtract 1 from the top/signal frames, and the fallback path uses _Unwind_GetIPInfo so it only adjusts real return addresses.
- When the frame-pointer walk stops early on a corrupt or not-yet-set-up frame pointer (the kind of damaged stack that often causes the crash), we cross-check it against the libgcc/_Unwind DWARF-CFI unwinder, which can cross frame-pointer-less boundaries the walk cannot. The cross-check is aligned to the faulting PC so our own handler and the signal trampoline are skipped, and it is printed as a second "Cross-check backtrace" section only when it actually recovers more frames than the truncated walk. A clean crash prints just the one backtrace.
- /proc/self/maps is read with raw open/read instead of fopen/fgets/sscanf, so the crash path stays async-signal-safe.
- addr2line is run once per object (batched) instead of once per frame, and the call is bounded by a 5 second watchdog using poll(). A hung addr2line is killed and reaped, so it cannot wedge the handler and it leaves no zombie.
- If addr2line is missing, hangs, or dies, we still print a raw "addr2line -f -e <obj> <addr>" line for every frame that stayed unresolved, so the return addresses are always in the log and can be symbolized offline. A short note says why symbolization did not run (not installed / unavailable / timed out). Fully resolved backtraces print none of this.
- The work buffers (frame table, addr2line argv/io, maps snapshot) are now _Thread_local, so two threads faulting at the same time do not corrupt each other's output.
- The output is rendered in gdb's style and, on an interactive terminal, is colored (function names yellow, source locations green) through the existing cli_color() helper. The daemon writes no color codes, so FTL.log stays plain text.

Why this is worth it: previously a single inline popen(addr2line) per frame with no timeout meant one slow or broken addr2line could lose every later frame and the summary, which is the worst case for a crash you cannot reproduce. Now the raw addresses are always logged, symbolization is cheaper (one process per object instead of one per frame) and bounded in time, the whole crash path is async-signal-safe and thread-safe, and a damaged frame-pointer chain gets a second opinion from libgcc. The build already ships -fno-omit-frame-pointer, -g3 and -rdynamic and is not stripped, so the frame-pointer walk and the dladdr/addr2line resolution all have what they need.

How to test (build with `bash build.sh`):

1. `./pihole-FTL backtrace`
   You should see a "Backtrace (N frames):" header, every frame resolved to "func file:line", and a "--- end of backtrace ---" line. No "not installed" or reproducer lines.

2. `PATH="" ./pihole-FTL backtrace`
   You should see "addr2line is not installed" plus an `addr2line -f -e "<obj>" <addr>` line for each unresolved frame. This shows the raw addresses survive even with no symbolizer present.

3. Put a slow shim called addr2line first on PATH (for example a script that runs `sleep 30`) and run `./pihole-FTL backtrace`. It should return within about 5 seconds, report the timeout, fall back to the reproducer lines, and leave no leftover process.

4. `./pihole-FTL crash`
   The header should read "Backtrace (N frames from signal context):" and frame #0 should be the faulting instruction (the `*ptr = 'x';` line in src/args.c). A clean crash unwinds all the way to _start, so no cross-check section is printed.

5. `FORCE_COLOR=1 ./pihole-FTL backtrace` (or run it in a terminal) shows the function names in yellow and the source paths in green; piping the same command to a file or `cat -v` shows no escape codes, confirming the log file stays plain.

6. The new tests in test/test_suite.bats cover the three backtrace variants, and test/arch_test.sh adds check_backtrace and asserts "from signal context" on the crash test, run per architecture in CI under QEMU.

PromoFaux (dismissed):

yubiuser:
Depending on what gets merged first, this PR or https://github.com/pi-hole/FTL/pull/2872 needs adjustment to the tests to use the BATS helper scripts.

DL6ER:
Merged `development` and migrated tests to the new bats helper functions.

DL6ER:
Assuming review still holds, no changes other than merging `development` and adjusting the test code.

PR #2882, opened 2026-05-06
Fix mbedTLS 100% CPU busy-spin on non-blocking TLS sockets

DL6ER:

Fixes a 100% CPU busy-spin in the embedded CivetWeb web server when it is built with the mbedTLS backend. A single misbehaving or slow TLS client can peg one CPU core per connection - e.g. opening a TCP connection to the HTTPS port and never completing the handshake spins a worker thread at 100% indefinitely. With a handful of such connections FTL saturates every core. This is trivially reachable from the network, so it is effectively a low-effort DoS as well as the source of the high-CPU reports against idle HTTPS keep-alive connections. This replaces the temporary backoff-sleep mitigation (the `MG_MBEDTLS_WANT_RETRY_DELAY_MS` workaround) with a proper event-driven fix that is suitable for upstreaming to CivetWeb.

Non-blocking mbedTLS calls can repeatedly return `MBEDTLS_ERR_SSL_WANT_READ` / `WANT_WRITE` without making application-visible progress. Three code paths reacted by retrying immediately (or by waiting on the wrong socket event), which is what spun the CPU. Each now waits for the operation mbedTLS actually asked for, mirroring what the OpenSSL path (`sslize()`) already does:

- **Handshake** (`mbed_ssl_handshake`, `mod_mbedtls.inl`): the loop had no wait at all. It now waits with `mbedtls_net_poll()` for read- or write-readiness (matching `WANT_READ` / `WANT_WRITE`), with a finite timeout so it stays responsive to shutdown; `ASYNC_IN_PROGRESS` yields briefly. The accepted socket fd is passed in via the existing bio context.
- **Reads** (`pull_inner`, `civetweb.c`): instead of mapping `WANT_*` to "0 bytes" and returning immediately, it now waits with the existing `mg_poll()` on `POLLIN` / `POLLOUT`, bounded by the request timeout and the server stop flag.
- **Writes** (`push_inner`, `civetweb.c`): it already polled before retrying, but always for `POLLOUT`. When `mbedtls_ssl_write()` returns `WANT_READ` (e.g. during a renegotiation) the socket is almost always writable, so it spun; it now polls for the event mbedTLS requested.

No new configuration and no magic constants - the change is behaviour-only and confined to the mbedTLS paths. The same change is being submitted upstream to CivetWeb so the vendored copy can eventually drop the local delta.

Changes:
- `src/webserver/civetweb/mod_mbedtls.inl`: event-driven handshake wait via `mbedtls_net_poll()`; `mbed_ssl_handshake()` now takes the socket context.
- `src/webserver/civetweb/civetweb.c`: `pull_inner()` and `push_inner()` mbedTLS branches wait on the correct socket event via `mg_poll()`; removes the `MG_MBEDTLS_WANT_RETRY_DELAY_MS` backoff.
- `test/repro_idle_https_keepalive.py`: regression probe for the spin.

## How can this actually be tested?

Built with the mbedTLS backend and exercised against a running FTL, sampling process CPU. The probe is sensitive - built against the pre-fix code it shows the spin, against this PR it does not:

| Scenario | Before (busy-loop) | After (this PR) |
|----------------------------------------|--------------------|-----------------|
| 4 stalled TLS handshakes (8 s) | ~390% (4 cores) | <1% |
| 4 idle HTTPS keep-alive conns (8 s) | idle | idle |
| 20 concurrent HTTPS requests | 200 OK | 200 OK |
| Large HTTPS responses (write path) | 200 OK | 200 OK |
| Handshake that stalls 3 s then resumes | n/a | 200 OK |

The "resumes after stalling" case confirms the poll wait wakes and the handshake completes - it is an event wait, not a fixed delay. The handshake and read spins are directly reproducible; the write spin requires a renegotiation that makes `mbedtls_ssl_write()` return `WANT_READ`, so it is addressed by inspection (wait for the requested event) and verified not to regress normal writes.

The handshake spin needs no special tooling - **any plain-TCP client that connects to the HTTPS port and never sends a ClientHello makes the pre-fix server spin**. Open a few stalled connections with bash's `/dev/tcp` and sample the FTL process' CPU straight from `/proc` (cores summed).

Check out `test/civetweb_handshake_cpu.sh` for a simple demonstration. The regression test is provided on my laptop:
- this branch: 0% CPU - PASS (exit 0),
- `development` branch 577% CPU - FAIL (exit 1).

---

**Related issue or feature (if applicable):** https://github.com/civetweb/civetweb/pull/1405

---

## Checklist:

### DIFF
--- a/src/webserver/civetweb/civetweb.c
+++ b/src/webserver/civetweb/civetweb.c
@@ -6219,6 +6219,10 @@ push_inner(struct mg_context *ctx,
 	/* Try to read until it succeeds, fails, times out, or the server
 	 * shuts down. */
 	for (;;) {
+		/* Which socket event to wait for before retrying a write that made
+		 * no progress. Default to write-readiness; a non-blocking TLS write
+		 * can instead need read-readiness (see the mbedTLS branch below). */
+		int want_events = POLLOUT;
 
 #if defined(USE_MBEDTLS)
 		if (ssl != NULL) {
@@ -6227,6 +6231,14 @@ push_inner(struct mg_context *ctx,
 				if ((n == MBEDTLS_ERR_SSL_WANT_READ)
 				    || (n == MBEDTLS_ERR_SSL_WANT_WRITE)
 				    || n == MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS) {
+					/* mbedTLS may need to read (e.g. during a
+					 * renegotiation) before it can write. Wait for the
+					 * operation it actually asked for: polling for write-
+					 * readiness while it wants to read busy-spins at 100%
+					 * CPU because the socket is almost always writable. */
+					if (n == MBEDTLS_ERR_SSL_WANT_READ) {
+						want_events = POLLIN;
+					}
 					n = 0;
 				} else {
 					fprintf(stderr, "SSL write failed, error %d\n", n);
@@ -6328,7 +6340,7 @@ push_inner(struct mg_context *ctx,
 			unsigned int num_sock = 1;
 
 			pfd[0].fd = sock;
-			pfd[0].events = POLLOUT;
+			pfd[0].events = (short)want_events;
 
 			if (ctx->context_type == CONTEXT_SERVER) {
 				pfd[num_sock].fd = ctx->thread_shutdown_notification_socket;
@@ -6488,6 +6500,30 @@ pull_inner(FILE *fp,
 				if ((nread == MBEDTLS_ERR_SSL_WANT_READ)
 				    || (nread == MBEDTLS_ERR_SSL_WANT_WRITE)
 				    || nread == MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS) {
+					/* mbedTLS consumed data from the socket (e.g. a non-
+					 * application TLS record) or holds buffered data that
+					 * does not yield application bytes yet, so the socket can
+					 * stay poll-readable while no progress is visible here.
+					 * Wait for the socket to become ready for the operation
+					 * mbedTLS asked for (read vs. write), bounded by the
+					 * request timeout, instead of returning immediately and
+					 * letting the request loop busy-spin at 100% CPU. */
+					num_sock = 1;
+					pfd[0].fd = conn->client.sock;
+					pfd[0].events =
+					    (nread == MBEDTLS_ERR_SSL_WANT_WRITE) ? POLLOUT : POLLIN;
+
+					if (conn->phys_ctx->context_type == CONTEXT_SERVER) {
+						pfd[num_sock].fd =
+						    conn->phys_ctx->thread_shutdown_notification_socket;
+						pfd[num_sock].events = POLLIN;
+						num_sock++;
+					}
+
+					mg_poll(pfd,
+					        num_sock,
+					        (int)(timeout * 1000.0),
+					        &(conn->phys_ctx->stop_flag));
 					nread = 0;
 				} else {
 					fprintf(stderr, "SSL read failed, error %d\n", nread);
--- a/src/webserver/civetweb/civetweb.c
+++ b/src/webserver/civetweb/civetweb.c
@@ -479,6 +479,13 @@ _civet_safe_clock_gettime(int clk_id, struct timespec *t)
 #define SOCKET_TIMEOUT_QUANTUM (2000) /* in ms */
 #endif
 
+/* Non-blocking mbedTLS operations can repeatedly return WANT_READ/WRITE
+ * without making application-visible progress. Add a tiny backoff to avoid
+ * tight retry loops on idle HTTPS keep-alive connections. */
+#if !defined(MG_MBEDTLS_WANT_RETRY_DELAY_MS)
+#define MG_MBEDTLS_WANT_RETRY_DELAY_MS (5)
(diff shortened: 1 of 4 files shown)


DL6ER:
TODO: Need to submit a PR upstream

yubiuser (commented):
Should such a change not better be a patch at https://github.com/pi-hole/FTL/tree/master/patch/civetweb

rdwebdesign:
I think this is the intention:

> TODO: Need to submit a PR upstream

gkuchta:
FWIW, I think I maybe ran into this after issuing a request to the admin UI (clean session; was just going to add a blocklist entry). I saw a single pihole-FTL thread spike to, and stay at, 100% cpu use. From `strace`:


> 3185 20:44:39.629235 <... select resumed>) = 0 (Timeout) <0.048528> 3185 20:44:39.629253 select(0, NULL, NULL, NULL, {tv_sec=0, tv_usec=100000} <unfinished ...> 3191 20:44:39.629265 poll([{fd=35, events=POLLIN}, {fd=36, events=POLLIN}, {fd=37, events=POLLIN}, {fd=38, events=POLLIN}, {fd=34, events=POLLIN}], 5, 2000) = 4 ([{fd=35, revents=POLLIN}, {fd=36, revents=POLLIN}, { fd=37, revents=POLLIN}, {fd=38, revents=POLLIN}]) <0.000007> 3191 20:44:39.629292 poll([{fd=35, events=POLLIN}, {fd=36, events=POLLIN}, {fd=37, events=POLLIN}, {fd=38, events=POLLIN}, {fd=34, events=POLLIN}], 5, 2000) = 4 ([{fd=35, revents=POLLIN}, {fd=36, revents=POLLIN}, { fd=37, revents=POLLIN}, {fd=38, revents=POLLIN}]) <0.000007> 3191 20:44:39.629319 poll([{fd=35, events=POLLIN}, {fd=36, events=POLLIN}, {fd=37, events=POLLIN}, {fd=38, events=POLLIN}, {fd=34, events=POLLIN}], 5, 2000) = 4 ([{fd=35, revents=POLLIN}, {fd=36, revents=POLLIN}, { fd=37, revents=POLLIN}, {fd=38, revents=POLLIN}]) <0.000007>


fd35 = 0.0.0.0:80
fd36 = 0.0.0.0:443
fd37 = [::]:80
fd38 = [::]:443
poll() returns POLLIN for all four listener fds in ~7us
FTL performs no accept/read/write (or any other calls) between polls; basically just an infinite poll() loop.
Recv-Q remains nonzero (Recv-Q was at 2) on listeners
no inbound 80/443 traffic observed via tcpdump
dns resolution continued without interruption

If I run into it again I can try to get some more useful info via gdb or something, but it's just my home network so I just dumped what info I could and HUP'd the process

DL6ER:
Sorry for the long delay in replying - real life has been really busy lately.

Thanks for the detailed dump - that's genuinely useful.

One thing stands out though: the strace shows the spinning thread sitting in a tight `poll()` loop over the **listener** sockets (fd 35–38 = your `:80`/`:443` listeners), returning `POLLIN` on all of them but never calling `accept()`, with `Recv-Q=2` pending. That's the **master/accept thread**, whereas this PR fixes a busy-spin in the mbedTLS read/handshake path on connections that have *already* been accepted (`pull_inner` / `mbed_ssl_handshake`).

So as it stands this looks like it may be a *related but distinct* loop rather than the exact one this PR targets — possibly the accept loop or the worker queue getting wedged. It's plausible the two are connected (e.g. workers stuck spinning never return to the pool and starve the accept side), but the trace only shows one hot thread and it's the master on the listeners, so I can't tie it to the TLS path from this alone.

If you hit it again, the one thing that would nail it down is a backtrace of the 100%-CPU thread under gdb:

```
gdb -p <pid>
(gdb) info threads
(gdb) thread apply all bt
```

(or at least `bt` for the hot LWP). That'll tell us immediately whether it's the mbedTLS loop this PR addresses or the accept side. The `Recv-Q` nonzero + no `accept()` symptom in particular makes me suspect the latter.

Appreciate you grabbing what you could before HUP'ing it!

gkuchta:
Absolutely no worries, @DL6ER; we're all just out here existing on a spinning rock; real life is way more important than all this. I appreciate that you'd left a signpost for me to find, and I further appreciate your clarification about what that `poll()` loop was actually up to! 

I haven't seen this spin loop condition again, but if I do, I'll snag the back traces and report back here. Thanks for putting your eyes on this, and for lending some of your brain juice.

Cheers and be well!

DL6ER:
Instead of the original band-aid fix (adding the `sleep`s), I rewrote this entire PR to fix the underlying actual issue. While doing this, I found that CivetWeb already has a fix for this implemented but only for the OpenSSL path. Nobody ever ported it to MbedTLS which is what we are doing here. Upstream PR to follow (this time I will actually do it :crossed_fingers:  ).

DL6ER:
Currently rebasing on latest `development`

DL6ER:
Upstream PR: https://github.com/civetweb/civetweb/pull/1405
(also added to PR description)

DL6ER:
Superseded by replacing mbedTLS by OpenSSL which doesn't have this issue in CivetWeb: https://github.com/pi-hole/FTL/pull/2941

PR #2897, opened 2026-05-21
[DESIGN]: structured logging: JSON and journald

darkexplosiveqwx:
See https://github.com/pi-hole/docker-pi-hole/pull/2036 for more information.

~~Major issue:
`dnsmasq` closes the file descriptors of STDIN, STDOUT and STDERR, this prevents us from logging after dnsmasq has started. Removing the lines is only a quick duct-tape fix while developing.~~
Removing that code seems to be the best option, as it also doesn’t affect other functionality.

This can also be observed in the systemd service log, since it only contains startup information and not the entire FTL.log.
In docker we are not capturing stdout, but tailing the logfile, but I intentionally wanted to keep the logfiles as-is, need to find a solution on how to implement this.

This currently changes webserver.log and the dnsmasq and webserver FIFOs to better align with FTL.log, but this can be reverted if necessary.

---

---


darkexplosiveqwx:
Both `--log-json` and `--log-journal` are functional to some degree and can be tested further.

I think that we can now start discussing what extra fields we might want/need. (Like the file paths shown in debug flags)

Regarding the blunt patch into dnsmasq to prevent it from messing with our file descriptors, I definitely need to guidance here on whether this is a good mean or if there is a better way.

yubiuser:
We probably don't want all `dnsmasq` log to be printed when `using --log-json`

(This was a container without any queries received...)

```
pihole  | {"timestamp":"2026-06-03T14:19:03.330Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:19:03.330Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:19:33.381Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:19:33.381Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:20:03.442Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:20:03.442Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:20:33.498Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:20:33.499Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:21:03.589Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:21:03.590Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:21:33.650Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:21:33.650Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:22:03.702Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:22:03.702Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:22:33.757Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:22:33.758Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:23:03.812Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:23:03.812Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:23:33.868Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:23:33.868Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:24:03.923Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:24:03.923Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:24:33.985Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:24:33.986Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:25:04.034Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:25:04.034Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:25:34.084Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:25:34.084Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:26:04.133Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:26:04.133Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:26:34.183Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:26:34.183Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:27:04.237Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:27:04.237Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"Pi-hole hostname pi.hole is 127.0.0.1"}
pihole  | {"timestamp":"2026-06-03T14:27:34.328Z","log_level":"INFO","service":"pihole-FTL","component":"dnsmasq","pid":"127M","message":"query[A] pi.hole from 127.0.0.1"}
```

rdwebdesign:
These are normal lines from `pihole.log`.

```
2026-06-03 16:21:16.922 query[A] pi.hole from 127.0.0.1
2026-06-03 16:21:16.923 Pi-hole hostname pi.hole is 192.168.0.201
```

If I'm not mistaken, they are generated by the container health check, every 30 seconds.

Do you think `pihole.log` should always be in plain text?

yubiuser:
> Do you think pihole.log should always be in plain text?

No. I think we should have an option to disable `dnsmasq` logs - imagine the example above with a few hundred queries from a real instance.

darkexplosiveqwx:
> Do you think pihole.log should always be in plain text?

I'm not sure if I understand the question, but what else should the logfile be besides plain text?

I do think that we should have all logs available with `--log-json`/`--log-journal`, this is why we can filter with `priority` and `component`.

Maybe we can add an option to exclude query-level information, but this information would also be added to pihole.log anyway.

Disk usage wise, I don't think it's so bad. (I have been using `--log-journal` for the last couple days with multiple active mobile devices.)

```
$ journalctl -b -u pihole-FTL.service | wc -l  # around 26h
449014

$ journalctl -u pihole-FTL.service | wc -l  # since May 24 11:57:49
2399128

$ journalctl --disk-usage
Archived and active journals take up 2.6G in the file system.
```

I think the ability to just filter for FTL logs again offsets the raw verbosity.

Edit: how it compares to other services:

```
$ journalctl -o json | jq -r '._SYSTEMD_UNIT // "unknown"'  | sort | uniq -c | sort -nr

2406426 pihole-FTL.service
2125350 unbound.service
 975965 wpa_supplicant@wlu1bros.service
 299943 tailscaled.service
 271666 user@1000.service
  99082 init.scope
  97450 unknown
  37606 hostapd@wlo0c.service
  31435 systemd-networkd.service
  21173 cron.service
  11202 iwd.service
   8259 avahi-daemon.service
   7550 caddy.service
   7129 update-dhcp6pd-set@eno0c.service
   4457 systemd-timesyncd.service
   3567 pihole-update-gravity.service
   3534 hostapd@wlu1ap.service
```

rdwebdesign:
@yubiuser

> imagine the example above with a few hundred queries from a real instance.

Currently (in `master`), these lines are already NOT printed to docker logs. There is no need to add them.

I think there are 2 distinct ideas being developed together:

1. The idea of adding an option to format Pi-hole logs as JSON. 
  Personally, I think if an user decides to use JSON format, all logs (`FTL.log`, `webserver.log` and `pihole.log`) should use this format.

2. The idea of generating docker logs using JSON format. 
  If I understood the idea correctly, this would send to docker log, the **same** lines we already send today, but formatted as JSON.
  This PR is changing what docker log "sees", also sending `pihole.log` lines to docker log.
  I don't think we need this.

darkexplosiveqwx:
> @yubiuser
> 
> 
> 
> > imagine the example above with a few hundred queries from a real instance.
> 
> 
> 
> Currently (in `master`), these lines are already NOT printed to docker logs. There is no need to add them.
> 
> 
> 
> I think there are 2 distinct ideas being developed together:
> 
> 
> 
> 1. The idea of adding an option to format Pi-hole logs as JSON. 
> 
>   Personally, I think if an user decides to use JSON format, all logs (`FTL.log`, `webserver.log` and `pihole.log`) should use this format.
> 
> 
> 
> 2. The idea of generating docker logs using JSON format. 
> 
>   If I understood the idea correctly, this would send to docker log, the **same** lines we already send today, but formatted as JSON.
> 
>   This PR is changing what docker log "sees", also sending `pihole.log` lines to docker log.
> 
>   I don't think we need this.


My idea was to not change the logfiles to JSON, but instead have a unified JSON stream in stdout where all three are included, which allows filtering with `component`.

By sending all logs to docker stdout, users (or rather machines) won't have to touch the files at all anymore and can just parse and filter the stdout.

DL6ER:
Thanks for pushing on this - structured logging is something we want, and the journald protocol work is genuinely solid. My feedback is mostly about scope, sequencing and architecture so this can land cleanly rather than growing into one big hard-to-review unit.

**1. Split JSON from journald.** Right now `src/journal/` is roughly 940 of the ~1240 added lines, i.e., a from-scratch reimplementation of the journal native protocol. That is the bulk of the review surface, and it is orthogonal to the JSON-to-stdout feature (which is under 200 lines). I would like to land JSON-to-stdout first as a small, self-contained PR and treat journald as a separate follow-up on top. It gets the common case in front of users quickly and keeps each diff reviewable.

**2. Prefer one FTL-owned logging pipeline over parallel duplicate streams.** As it stands the PR emits JSON/journald *in addition to* the existing file logging, so every line is produced twice and the formatting logic ends up split across dnsmasq and FTL. I would rather we converge on a single pipeline driven from FTL: a `log.format` enum in the config (`plain`/`json`, extensible) plus a destination (`-` for the stream, a path, or a unified file), with FTL owning the write. That avoids double-logging and gives us full control over format and destination in one place, and it folds `files.log.webserver = "-"` (see #2956) into the same mechanism instead of being a separate special case.

The dnsmasq side of this is smaller than it looks. `my_syslog()` already hands every formatted line to `FTL_dnsmasq_log()` (`src/dnsmasq/log.c`, the existing Pi-hole block around the `vsnprintf` into `buffer`). The clean mechanism is:

- Add a guarded early `return` *after* the Pi-hole diagnosis-warning block (i.e., after `dnsmasq_diagnosis_warning()`, not immediately after the handoff - otherwise we skip the diagnosis system for `LOG_WARNING` lines). That single return bypasses both dnsmasq's file-write path and the `log_fd == -1` syslog fallback in one shot.
- Stop emitting `log-facility=` from our generated config. On its own that would just make dnsmasq fall back to syslog (`log_fd == -1` -> `vsyslog`), which is why it only makes sense *combined* with the early return: together dnsmasq neither opens nor writes anything for the DNS log, with no dangling fd on an empty file and no syslog spam. FTL becomes the sole owner.

Two things to keep on the list for this:

- FTL needs its own `pihole.log` writer. Today `FTL_dnsmasq_log()` only pushes into the in-memory API ring buffer - it writes nothing to disk. Suppressing dnsmasq's writer does not hand us one; we have to add it, reproducing the async/non-blocking behaviour, the SIGHUP reopen for logrotate, and dnsmasq's exact timestamp/prefix format (the message handed to FTL is the bare line, no timestamp). Match that format and the file stays byte-for-byte identical.
- The early-startup safety net. We currently write `log-facility` even when logging is disabled specifically to persist early warnings/errors to disk before FTL is fully up. Both moves together remove that, so either we guarantee the FTL writer + SHM are up before dnsmasq can emit, or we make the early return conditional on an "FTL writer ready" flag and keep `log-facility` only for the bootstrap window.

**This need not be a breaking change.** Changing *who* writes `pihole.log` is invisible to users; only changing *what comes out* would break anything. If we keep the existing keys and defaults (plain format, current paths, preserved rotation) and have FTL reproduce dnsmasq's line format byte-for-byte, this can ship in a 6.x minor. A change to the default on-disk format, relocating/renaming the log files, or collapsing them into one unified file by default would be the breaking parts, and those are separable decisions we can defer. (Whether a major *feature* alone should bump the major version is a separate internal policy question and not a prerequisite here.)

**3. Do not delete dnsmasq's std-fd redirect - make it conditional, and take the option upstream.** The current change removes the `dup2(.../dev/null)` on all three std fds in `src/dnsmasq/dnsmasq.c` for everyone. That block's only job is detaching the daemon from the launching terminal - the fd-reuse safety people usually cite is a *separate* earlier block that we do not touch - so dropping it unconditionally regresses that detach for users who never asked for stream logging. Better: leave `stdin` redirected always, and free `stdout`/`stderr` only when the user opted into a destination that uses it. Longer term I would prefer a proper dnsmasq option for this (not redirecting specific std fds), floated on the dnsmasq mailing list. It is a general improvement that benefits every embedder and every container deployment, and it keeps our patch to a one-liner instead of carrying a local hack.

**Smaller notes on the current code:**

- `src/journal/*` carries `// SPDX-License-Identifier: BSD-3-Clause` rather than our usual EUPL header. If it is derived third-party code that is fine license-wise (BSD-3 is compatible), but please add a `README`/`LICENSE` in `src/journal/` recording the origin and the BSD copyright line, so the provenance is explicit.
- `log_web()` now takes `(priority, flag, ...)` and most call sites pass `DEBUG_NONE`. Since `log_web_debug()` is already a macro, a matching convenience macro for the plain info/error case would drop `DEBUG_NONE` from those call sites and read cleaner. It is only a handful of sites, so this is readability rather than diff size.
- `log_to_json()` builds a `cJSON` object and `cJSON_PrintUnformatted()`s it per line, i.e., several heap allocations per log line. Not urgent, but for query-level volume a static buffer via `cJSON_PrintPreallocated()` (or hand-emitting the JSON on the hot path) would avoid the churn.

Happy to discuss all of this further - none of it is set in stone. My current feeling is that a unified logging feature owned by FTL, where we have full control, is likely the better end state than coexisting with dnsmasq's own logging, even if it means duplicating some code from dnsmasq in the short term. What makes it feel cleaner to me is that we could package each output into its own function - `log_plain()` (or similar) alongside `log_json()` and later `log_journal()` - and select purely on the `log.format` enum value, so the destination and formatting live in one obvious place rather than being threaded through dnsmasq. Happy to help scope the JSON-first PR and sketch that pipeline if useful.

darkexplosiveqwx:
Thanks for the review.

This is exactly why it's still in draft: I'm still figuring some big parts out.
The current implementation works good, I tested it on my production-but-tinker machine. It is, however, by no means pretty nor fully thought out.

1.
   1. My initial implementation used libsystemd, which left the journal part fairly simple: just call sd_journal_send() as the sink. But after some testing, I noticed that dnsmasq and libsystemd had quite different assumptions. Instead of patching dnsmasq (I tried), which opened a whole can of worms, I went with changing the libsystemd side buy shipping an in-tree solution. This also simplified library linking and requires no dlopen.
   Having the in-tree solution of course inflates the LoC count.
   2. Because it is so large, I'd like to split it out. I have prepared https://github.com/darkexplosiveqwx/libjournal for that. I think that doing changes there would improve maintainability, reviewability, and I'd also like to reuse it for some of my project (or for anyone to use). It needs some cleanup, yes. At least: more clear README, descriptive file headers/comments and making it clear inside FTL that it is a seperate project (like by including README, LICENSE, etc.). There are also some critical bugfixes I need to do: I can pretty reliably get it to Segmentation Fault, which is obviously not ideal.

2.
   1. I intentionally chose a flag over a TOML option: a TOML option would be included in a teleporter backup, a flag not. What if a user imports a backup into docker, while originated on systemd-bare-metal? We'd need fallback, multiple options or a teleporter exclude. That is something I find too complex. The flag on the other hand can simply be added to the systemd unit or via a docker env. That looks like a better place for me, especially since it is so low-level. I am already skeptical all the file paths being included in the config/teleporter. IMO from a lifecycle perspective they fit better as "set once during startup", CLI flags fit the position, but they are too clunky for many options.
   2. Re-creting the dnsmasq log writer may look intimidating, but for long-term maintainability it would be clearer than having loads of patches. Additionally, it's not like we're starting from zero: We already have our logging pipline with some re-usable components and we can look at dnsmasq's source for the canonical format. If were already re-creating a cached FILE/fd (I didn't look at the source yet) and reloading with USR2, I think it makes sense to apply the same caching concept to FTL and webserver as well. `logrotate` is quickly updated with a `kill -USR2` in `postrotate` and `sharedscripts`. We need to add some guards to prevent interleaving then.
   3. One small thing I noticed: if no file logging is active, should we still have the FIFO? If yes, then how should it be titled, since there is no file path we can just copy?
   4. You mentioned:

      >The early-startup safety net. We currently write log-facility even when logging is disabled specifically to persist early warnings/errors to disk before FTL is fully up. Both moves together remove that, so either we guarantee the FTL writer + SHM are up before dnsmasq can emit, or we make the early return conditional on an "FTL writer ready" flag and keep log-facility only for the bootstrap window.

      This actually fits my model quite good: decide the logging path via somthing fixed on startup: flag, env and fail early; instead of utilizing the runtime-mutable TOML.

3. Commenting it out _works_, but it is bad nonetheless, yes. Changing the patch to be conditional, while always discarding the unneeded stdin is the better approach. Upstreaming a config option to dnsmasq would be ideal, but I am unsure how to go about submit something to dnsmasq-discuss and how to make it look worthwhile to Simon.

4. This PR already contains lots of miscellaneous fixes/small chanes for `log.c` such as: "Open logfiles in append instead of append+read mode", "change webserver.log format to match FTL.log", "enrich webserver and dnsmaq log with priority/severity", "Avoid NULL dereference when no web log file is configured". I found quite some more improvements, that I simply haven't commited (yet). What would you think about turning them into a separate PR?

5. 
   1.   
  
      > `log_web()` now takes `(priority, flag, ...)` and most call sites pass `DEBUG_NONE`. Since `log_web_debug()` is already a macro, a matching convenience macro for the plain info/error case would drop `DEBUG_NONE` from those call sites and read cleaner. It is only a handful of sites, so this is readability rather than diff size.
  
      no objections, will do.

   2.

      > `log_to_json()` builds a `cJSON` object and `cJSON_PrintUnformatted()`s it per line, i.e., several heap allocations per log line. Not urgent, but for query-level volume a static buffer via `cJSON_PrintPreallocated()` (or hand-emitting the JSON on the hot path) would avoid the churn.

      I chose the already embedded cJSON, because it does escaping automatically and I didn't want to reinvent it. I will take a look at more performent options inside cJSON. If that still has measurable performance implications, I will consider hand-rolling.

DL6ER:
**1.** I would actually prefer the external `libjournal` over the in-tree copy, provided you are willing to maintain it - it is the journal *protocol*, not `libsystemd`, so it has no glibc/systemd coupling and builds static for `musl` fine. We can make it available in the `ftl-build` image the same way we build the other static deps (see https://github.com/pi-hole/docker-base-images/blob/7e076b1989ea8cb2575b96a260df96dea50ffb1b/ftl-build/Dockerfile#L76-L81), and gate the journald driver behind a CMake library check so it compiles in when present and is simply absent otherwise. That keeps FTL's tree clean and lets you harden, test and reuse it independently. Just keep its `LICENSE`/`README` with it, and fix the reliable segfault before we depend on it. This also frees JSON-to-stdout to land first, since it needs none of it.

**2.1** I lean TOML over a flag. Any option is already settable via `FTLCONF_` env (so a container/unit sets it like a flag), env-forced options cannot be overwritten via API or teleporter import (`src/api/config.c`), so the docker-imports-a-bare-metal-backup case you worry about is already prevented, and `FLAG_RESTART_FTL` gives the "decide once at startup" property you want in 2.4. We also already keep the log *paths* in TOML, so the format is their natural sibling. Open to discuss, but that is my preference.

**2.2** Agreed, and I would unify the cached-fd + reopen-on-signal model across FTL, webserver and dnsmasq (dnsmasq already uses `SIGUSR2`, align the others on it). The interleaving guard is just a write lock around that shared writer.

**2.3** Keep the FIFO regardless - it backs the `/api/logs` endpoints and the live log view, independent of on-disk logging, and it is keyed by source, not a file path, so title it by component.

**2.4** Agreed; `FLAG_RESTART_FTL` gives "decide at startup, fail early" without leaving the config system.

**3.** Conditional patch plus always dropping stdin is the right interim. I can help with the dnsmasq side: what lands with Simon is a minimal, opt-in change with an unchanged default and a general motivation (embedded/containerised dnsmasq where the supervisor captures stdout/stderr). Best to float the shape on dnsmasq-discuss before writing much; happy to co-write it or send it and credit you.

**4.** Yes, please split those out - they are independent and mergeable now and should not wait on the design. The NULL-deref fix can even go first on its own. Just call out the `webserver.log` format change in the changelog since it is user-visible.

**5.1** Great. **5.2** `cJSON_PrintPreallocated()` keeps the escaping while dropping the per-line allocation, so try that before hand-rolling anything.

Nice work overall - misc fixes now, JSON next, `libjournal` as its own project, journald last gives us a path where each piece is reviewable on its own.

darkexplosiveqwx:
1. I chose the in-tree option because it is common practice in FTL and simplifies building. If you prefer to build a library, I will take a look at how to do it for ftl-build and how to make source builds not annoying. This will need some additional testing, so the plan to do it last fits.
2. To be honest, I previously did not take a look how `FLAG_RESTART_FTL` functions, but knowing that it does `execvp(argv[0], argv);` and that dnsmasq's file descriptor cleanup will run when it starts again, it seems to be a "clean enough" restart and a new exec() from the supervisor does not seem necessary. Using an enum-type TOMLconf option will also make possible values more explicit. We should clarify the recommended use via a environment variable in the description.
   One possible edge-case I can think of involving the conditional redirect of `stdout`: FTL starts with a non-JSON option set, dnsmasq redirects stdout; later the option is changed to JSON; after the restart, it will inherit the same redirected fd and no stdout is shown. Two options I could think of: never redirect stdout or make changing the log mode require a supervisor restart.
   Besides that, a TOMLconf enum seems as capable as a a CLI flag, while keeping it easier to understand.
3. The help is appreciated, thanks. I will take a look myself at dnsmasq-discuss over the coming days to look at their style and we can figure out how to proceed.
4. Will prepare the PR with the smaller fixes/improvements soon.

I imagine we will end up with 3 (or 4) PRs:
- Initial PR with fixes and webserver.log format change. Some parts of this will certainly be touched again later, but this will give a cleaner base.
- dnsmasq log refactoring, log mode config enum, JSON output; Depending, this may get split
- journald addition

I would keep this PR for the higher-level design discussion, even when it may end up unmerged when new PRs are easier to split up. This way the individual PRs can remain focussed.

DL6ER:
Bot-closed on merge conflicts, not on merit - but it should stay closed rather than be reopened. Work continues on #2960, then #2968 (JSON) and #2969 (journald).

PR #2901, opened 2026-05-28
Add answer CIDR cache exclusions

rz467fzs7d:
## Summary
- add `dns.cache.excludeAnswerCIDRs`, a restart-scoped JSON string array of CIDR ranges
- skip committing an upstream DNS response to FTL's cache when any A/AAAA answer matches one of the configured ranges
- add CIDR parsing/matching coverage for IPv4, IPv6, and invalid input

## Motivation
This helps deployments where Pi-hole forwards to a fake-IP resolver. The client still receives the upstream answer, but FTL avoids caching fake-IP answers that may outlive the upstream resolver's own mapping state.

The default is `198.18.0.0/15`, the RFC 2544 benchmarking range commonly used by fake-IP DNS implementations.

## Test
- `cc -Wall -Wextra -I src -o /tmp/cache_exclude_cidr_test test/cache_exclude_cidr_test.c src/dnsmasq/cache_exclude.c src/webserver/cJSON/cJSON.c && /tmp/cache_exclude_cidr_test`
- `git diff --check`

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -642,7 +642,22 @@ void initConfig(struct config *conf)
 	conf->dns.cache.rrtype.f = FLAG_RESTART_FTL;
 	conf->dns.cache.rrtype.d.s = (char*)"ANY";
 	conf->dns.cache.rrtype.c = validate_str_no_newline;
-	
+
+	conf->dns.cache.excludeAnswerCIDRs.k = "dns.cache.excludeAnswerCIDRs";
+	conf->dns.cache.excludeAnswerCIDRs.h = "Array of CIDR ranges. If any A or AAAA record in an upstream DNS answer matches one of these ranges, FTL will return the answer to the client but will not commit this answer to its DNS cache.\n\nThis is useful when Pi-hole is upstreamed to a fake-IP resolver where cached fake-IP answers may outlive the resolver's own mapping state.";
+	conf->dns.cache.excludeAnswerCIDRs.a = cJSON_CreateStringReference("Array of IPv4 or IPv6 CIDR ranges");
+	conf->dns.cache.excludeAnswerCIDRs.t = CONF_JSON_STRING_ARRAY;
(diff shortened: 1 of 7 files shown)


darkexplosiveqwx (changes requested):
Please use the PR template from https://github.com/pi-hole/.github/blob/master/.github/PULL_REQUEST_TEMPLATE.md.
The PR template also mentions the following issues: DCO and targeting `development` branch.

This PR mainly touches dnsmasq code. We want to make as litte changes a possible to dnsmasq to avoid future merge conflicts. If it is possible, it should be submitted upstream at [Dnsmasq-dicuss](https://lists.thekelleys.org.uk/mailman/listinfo/dnsmasq-discuss).

Some more issues are that `test/cache_exclude_cidr_test.c` is never used and that the config option is not added to `test/pihole.toml` and `src/api/docs/content/specs/config.yaml`

DL6ER:
I read your motivation but this is still somewhat obscure to me: we cache per domains per client per record so why is this bad? Note that this will introduce extra work as FTL will not be able to short-circuit queries for said combination (domain, client, type) but has to do all lengthy regex comparisons, etc. over and over again, always coming back to (very likely to guaranteed) the same outcome. Something is missing for me.

DL6ER:
Closing this one. It targets `master` rather than `development`, DCO fails, and nothing has moved since May.

Two things would block it as it stands. The shipped defaults are not the ones the description advertises: next to `198.18.0.0/15` the code adds 10/8, 172.16/12, 192.168/16, 169.254/16, `fc00::/7` and `fe80::/10`, which disables the cache for every private and link-local answer. And `no_cache_dnssec` gates the single `cache_end_insert()` at the end of `extract_addresses()`, so setting it suppresses caching for the entire reply rather than for the matching record.

For the fake-IP case, having the upstream resolver hand out TTL 0 solves this without a new option. Reopen if you want to take it further.

PR #2904, opened 2026-06-01
api: add adlist filter to top blocked domains endpoint

sushantbhadauria:
Add an optional `list` query parameter to GET /api/stats/top_domains that restricts results to domains belonging to a specific subscription list (adlist) identified by its integer ID.

- api/stats.c: parse `list` param in api_stats_top_domains(); extend get_top_domains() with adlist_id (-1 = all lists, unchanged behaviour); skip non-matching domains via gravityDB_domain_in_adlist()
- database/gravity-db.c/.h: add gravityDB_domain_in_adlist() which queries gravity WHERE domain = ?1 AND adlist_id = ?2
- api/padd.c, api/queries.c: update existing callers to pass -1
- api/docs/content/specs/stats.yaml: document the new list parameter

### DIFF
--- a/src/api/api.h
+++ b/src/api/api.h
@@ -35,7 +35,8 @@ int api_stats_top_domains(struct ftl_conn *api);
 int api_stats_top_clients(struct ftl_conn *api);
 int api_stats_recentblocked(struct ftl_conn *api);
 cJSON *get_top_domains(struct ftl_conn *api, const int count,
-                       const bool blocked, const bool domains_only);
+                       const bool blocked, const bool domains_only,
+                       const int adlist_id);
 cJSON *get_top_clients(struct ftl_conn *api, const int count,
                        const bool blocked, const bool clients_only,
                        const bool names_only, const bool ip_if_no_name);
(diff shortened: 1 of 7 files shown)


DL6ER:
What is the particular use case of this change?

Also please check out about the DCO complaint:
> There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.
>
> To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

https://github.com/pi-hole/FTL/pull/2904/checks?check_run_id=78914219332

DL6ER:
Closing this one. The branch is three months behind and `get_top_domains()` has been restructured since, so it no longer applies as a rebase, and the use case I asked about in June never arrived.

The approach would need rethinking regardless: it runs a full SQLite prepare/step/finalize per domain while holding the SHM lock, on the same connection the DNS threads use for their blocking lookups. Pulling the list's domains into a set once before taking the lock, or doing the whole computation as a database join, would be the way to go.

Happy to reopen if you want to pick it up with the use case spelled out.

PR #2907, opened 2026-06-09
macvendor: resolve sub-allocated blocks (MA-M/MA-S) via longest-prefix match

RamSet:

**AI ASSISTED**

Make the hardware-vendor lookup resolve devices that live in **sub-allocated** IEEE blocks, instead of stopping at the 24-bit OUI.

IEEE has sub-divided many OUIs into smaller **MA-M (28-bit)** and **MA-S / IAB (36-bit)** assignments. For a MAC in such a block, the 24-bit prefix is owned by *"IEEE Registration Authority"*, not the device maker. Today `getMACVendor()` truncates the address to `XX:YY:ZZ` and matches only that, so these devices show a **blank** (or useless) vendor in the Network table.
L
Example: `34:E1:D1:80:xx:xx` is a Hubitat device. `34:E1:D1` is just the IEEE registry placeholder; the real owner is the `34:E1:D1:8/28` assignment.

Wireshark's `manuf` (the source for `macvendor.db`) already carries the finer assignments, e.g.:

```
34:E1:D1:80/28    Hubitat    Hubitat Inc.
00:1B:C5:00:00/36 ...        Converging Systems Inc.
```

…but the generator flattened the mask away and the query never looked past 24 bits. Two coordinated changes:

- **`tools/macvendor.py`** — store each prefix as a bare uppercase hex string of its real nibble length (`/24`→6, `/28`→7, `/36`→9 hex), e.g. `34E1D18`, instead of the raw masked token `34:E1:D1:80/28`.
- **`getMACVendor()`** — normalize the queried MAC the same way (drop separators, upper-case) and pick the **longest matching prefix**:

  ```sql
  SELECT vendor FROM macvendor WHERE mac IN (
    substr(replace(upper(?1),':',''),1,9),
    substr(replace(upper(?1),':',''),1,7),
    substr(replace(upper(?1),':',''),1,6))
  ORDER BY length(mac) DESC LIMIT 1;
  ```

The most specific assignment wins; the common 24-bit case is unchanged. `updateMACVendorRecords()` reuses `getMACVendor()`, so both the insert path and the periodic refresh benefit.

## Testing

Built a `macvendor.db` with the updated generator against the live `manuf` (~57k rows) and ran the exact new query:

| MAC | Result | Block |
|---|---|---|
| `34:e1:d1:80:76:de` | **Hubitat Inc.** | MA-M /28 |
| `34:e1:d1:00:11:22` | **Tianjin Sublue …** | MA-M /28 (same parent, different owner) |
| `00:1b:c5:00:00:01` | **Converging Systems Inc.** | MA-S /36 |
| `98:48:27:c4:2f:f2` | **Tp-Link Technologies** | MA-L /24 (unchanged) |
| `de:ad:be:ef:00:01` | *no match* → NULL | — |

Row count is essentially unchanged (no bloat), and two devices sharing the `34:E1:D1` parent now resolve independently — impossible before.

## Notes

- This is a coordinated format change for the `macvendor` table key (colon → bare hex), so the generator and `getMACVendor()` must ship together. The hosted `macvendor.db` is rebuilt from this same `macvendor.py`, so a release stays internally consistent — flagging it so the `ftl.pi-hole.net` rebuild is confirmed to pick up the new format.
- Happy to add a `test/` case seeding a few MA-M/MA-S rows and asserting the longest-prefix result if that's preferred.

---

By submitting this pull request, I confirm the following:
- [x] I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/).
- [x] My commit is signed-off (DCO).
- [x] The change is made under the EUPL.

### DIFF
--- a/src/database/network-table.c
+++ b/src/database/network-table.c
@@ -1791,11 +1791,20 @@ static bool getMACVendor(const char *hwaddr, char vendor[MAXVENDORLEN])
 		return false;
 	}
 
-	// Only keep "XX:YY:ZZ" (8 characters)
-	char hwaddrshort[9];
-	strncpy(hwaddrshort, hwaddr, 8);
-	hwaddrshort[8] = '\0';
-	const char querystr[] = "SELECT vendor FROM macvendor WHERE mac LIKE ?;";
+	// Match the most specific IEEE assignment, not just the 24-bit OUI. Many
+	// OUIs are sub-divided into smaller MA-M (/28) and MA-S / IAB (/36) blocks,
+	// which the macvendor table stores in Wireshark manuf form (e.g.
+	// "34:E1:D1:80/28", "00:1B:C5:00:00/36"); the 24-bit prefix of such a block
+	// resolves only to "IEEE Registration Authority" (or to no row at all).
+	// hwaddr is a 17-char colon MAC (guaranteed by the length check above), so
+	// we reconstruct the candidate /24, /28 and /36 keys from it and let the
+	// longest match win - additive, requiring no change to the database itself.
+	const char querystr[] =
+		"SELECT vendor FROM macvendor WHERE mac IN ("
+		"substr(upper(?1),1,8),"                                   /* /24: 34:E1:D1        */
+		"substr(upper(?1),1,9)||substr(upper(?1),10,1)||'0/28',"   /* /28: 34:E1:D1:80/28  */
+		"substr(upper(?1),1,12)||substr(upper(?1),13,1)||'0/36'"   /* /36: 00:1B:C5:00:00/36 */
+		") ORDER BY length(mac) DESC LIMIT 1;";
 
 	sqlite3_stmt *stmt = NULL;
 	rc = sqlite3_prepare_v2(macvendor_db, querystr, -1, &stmt, NULL);
@@ -1805,11 +1814,11 @@ static bool getMACVendor(const char *hwaddr, char vendor[MAXVENDORLEN])
 		goto getMACVendor_end;
 	}
 
-	// Bind hwaddrshort to prepared statement
-	if((rc = sqlite3_bind_text(stmt, 1, hwaddrshort, -1, SQLITE_STATIC)) != SQLITE_OK)
+	// Bind the full MAC; the candidate prefixes are derived in SQL above
+	if((rc = sqlite3_bind_text(stmt, 1, hwaddr, -1, SQLITE_STATIC)) != SQLITE_OK)
 	{
-		log_err("getMACVendor(\"%s\" -> \"%s\"): Failed to bind hwaddrshort: %s",
-		        hwaddr, hwaddrshort, sqlite3_errstr(rc));
+		log_err("getMACVendor(\"%s\"): Failed to bind hwaddr: %s",
+		        hwaddr, sqlite3_errstr(rc));
 		goto getMACVendor_end;
 	}
 
--- a/test/test_suite.bats
+++ b/test/test_suite.bats
@@ -716,6 +716,50 @@ setup() {
   [[ ${lines[0]} == "/etc/pihole/pihole-FTL.db: SQLite 3.x database"* ]]
 }
 
+@test "MAC vendor lookup resolves MA-L, MA-M and MA-S blocks (longest prefix)" {
+  # The macvendor table stores Wireshark manuf keys verbatim: a plain "XX:XX:XX"
+  # for /24 (MA-L) and a masked form for the sub-divided blocks, e.g.
+  # "34:E1:D1:80/28" (MA-M) and "00:1B:C5:00:00/36" (MA-S). Seed a small db in
+  # that exact format.
+  DB=/tmp/macvendor_test.db
+  rm -f "${DB}"
+  ./pihole-FTL sqlite3 "${DB}" "CREATE TABLE macvendor (mac TEXT NOT NULL, vendor TEXT NOT NULL, PRIMARY KEY (mac));"
+  ./pihole-FTL sqlite3 "${DB}" "INSERT INTO macvendor (mac, vendor) VALUES ('98:48:27','Tp-Link'),('34:E1:D1:80/28','Hubitat'),('34:E1:D1:00/28','Tianjin Sublue'),('00:1B:C5:00:00/36','Converging Systems');"
+
+  # Mirrors the longest-prefix query used by getMACVendor(): reconstruct the
+  # candidate /24, /28 and /36 keys from the MAC and let the longest match win.
+  _macvendor_lookup() {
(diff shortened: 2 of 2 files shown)


RamSet:
‘ts been a while since I did one of these …

yubiuser (commented):
Nice change!

> Happy to add a test/ case seeding a few MA-M/MA-S rows and asserting the longest-prefix result if that's preferred.

This would be great. I think we should try to add as many tests for new/tweaked functions as we can.

RamSet:
Added tests for both tweaked functions in `8111a3b`:

- **`getMACVendor()` longest-prefix lookup** (BATS, `test_suite.bats`): builds a small `macvendor.db` in the new bare-hex key format and runs the exact lookup query, asserting MA-L (`/24`), MA-M (`/28`) and MA-S (`/36`) resolution, that two devices under the same 24-bit parent (`34:E1:D1`) resolve to *different* vendors, and that an unknown OUI returns no match.
- **`tools/macvendor.py` normalization** (BATS): self-test of `normalize_prefix()` across `/24`, `/28`, `/36`, IAB-style and separator/case variants.

To make the generator unit-testable I refactored it into pure functions (`normalize_prefix`/`parse_manuf`/`build_db`) and moved the network download under `__main__` with a lazy `requests` import, so the module imports with stdlib only. No change to the DB produced by live runs.

I couldn't find a deterministic way to drive `getMACVendor()` itself end-to-end through the network table (it fires on neighbour-cache discovery, which needs real ARP entries), so the BATS test exercises the identical query against a seeded DB. Happy to wire up a fuller integration test if you'd prefer a particular hook.

DL6ER:
Nice work - the longest-prefix idea is correct. No false positives are possible either: IEEE blocks nest on nibble boundaries, so if a MAC's first 7 hex match a stored `/28` key it's by definition inside that assignment. Refactoring `macvendor.py` to be importable for the unit test is a nice touch too.

I have one substantive concern, and I think it has a clean fix: **does the key format need to drop the colons?**

The new scheme rewrites the stored keys from colon form (`34:E1:D1`) to bare hex (`34E1D1`). That's what makes this ***a breaking format change*** rather than an additive one - the existing hosted `macvendor.db` already stores the /24 keys *with* colons. `getMACVendor()` and the DB ship through different channels (binary via the FTL release, DB via `ftl.pi-hole.net` into `/etc/pihole/macvendor.db`, not built locally for most users), so there's an unavoidable window where a new binary runs against an old DB.

With bare-hex what happens inside that breaking window is quite bad: *every* lookup misses (`34:E1:D1` len 8 vs queried `34E1D1` len 6), and because `updateMACVendorRecords()` runs `UPDATE network SET macVendor = ?1` **unconditionally** with an empty string when the lookup fails, it doesn't just fail to resolve new vendors - it **blanks the vendor column for every device** until the next 30-day refresh.

My point is: Your proposed logical change (longest-prefix match) works fine while *keeping* the colons - just different substring lengths to account for the separators:

``` sql
SELECT vendor FROM macvendor WHERE mac IN (
    substr(upper(?1),1,13),   -- /36  00:1B:C5:00:0
    substr(upper(?1),1,10),   -- /28  34:E1:D1:8
    substr(upper(?1),1,8))    -- /24  34:E1:D1
  ORDER BY length(mac) DESC LIMIT 1;
```

`hwaddr` is already a 17-char colon MAC (the `strlen != 17` guard), so no `replace()` needed. The generator emits `34:E1:D1:8` / `00:1B:C5:00:0` for the finer blocks and leaves /24 keys untouched, so there is no breaking change in sight here.

Minor while looking at this code: consider `INSERT OR IGNORE` in `build_db()` so an unexpected upstream prefix mask can't abort the whole generation.

RamSet:
Great catch on the breaking-window — you're right, and the `updateMACVendorRecords()` unconditional write makes it actively destructive, not just a silent miss. Pushed `address review` keeping the colon format:

- `getMACVendor()` now matches `substr(upper(?1),1,{13,10,8})` for MA-S/MA-M/MA-L. As you noted, `hwaddr` is already a 17-char colon MAC, so the `replace()` is gone.
- `macvendor.py` emits colon-grouped prefixes (`34:E1:D1:8`, `00:1B:C5:00:0`) and leaves /24 keys byte-identical — additive, not a re-key.
- `INSERT OR IGNORE` added in `build_db()`.
- Tests updated to the colon format.

Verified the backward-compat path explicitly: running the new query against an old /24-only DB still resolves /24 devices (no blanking), and a sub-allocated MAC simply falls back to the existing 24-bit row — same behaviour as today, never an empty string. So there's no bad window: an old DB keeps working unchanged, and finer resolution kicks in once the rebuilt DB lands.

DL6ER:
IMO there is no need to rewrite that database *at all*. Let me improve my late-night SQL suggestion from yesterday to make this clearer (I will suggest a direct code change rather than a "it might look like"):

```c
const char querystr[] =
    "SELECT vendor FROM macvendor WHERE mac IN ("
    "substr(upper(?1),1,8),"                                        /* /24: 34:E1:D1        */
    "substr(upper(?1),1,9)||substr(upper(?1),10,1)||'0/28',"        /* /28: 34:E1:D1:80/28  */
    "substr(upper(?1),1,12)||substr(upper(?1),13,1)||'0/36'"        /* /36: 00:1B:C5:00:00/36 */
    ") ORDER BY length(mac) DESC LIMIT 1;";
```

Example:
- `upper(?1)` on `34:e1:d1:80:76:de` = `34:E1:D1:80:76:DE`
- `substr(...,1,8)` = `34:E1:D1` matches existing `/24` key
- `substr(...,1,9)||substr(...,10,1)||'0/28'` = `34:E1:D1: + 8 + 0/28 = 34:E1:D1:80/28` matches existing /28 key

The final `ORDER BY length(mac) DESC `still picks the most specific: `/36` keys (len 18) > `/28` keys (len 14) > `/24` keys (len 8).

With this suggestion, the entire PR reduces to changing one string in `network-table.c`. No `macvendor.py` changes, no DB format change, no rebuild, no deployment coordination, no transition risk. I know that his may not be the most elegant implementation but, really, this is running not all that often and always outside any critical/hot DNS path so wasting (a few!) extra CPU cycles due to more string manipulations seems worthwhile given that much less battle-tested code needs to change, otherwise.

I have *nothing* against AI-assisted contributions, but I don't like needless bloating of the code just because it is not much work if an agent is doing it for you. Please check out my suggestion and tell the agent something like: "Avoid overengineering, YAGNI principle. Verify twice before changing things"

RamSet:
You're right, and point taken — the data was already in the shipped DB, so re-keying it was solving a problem I created. I verified your reconstruction query against a real `/etc/pihole/macvendor.db` (the manuf `/28` and `/36` rows are there verbatim), and it resolves Hubitat (`/28`), Converging Systems (`/36`) and Tp-Link (`/24`) with an unknown OUI returning nothing.

Force-pushed a single commit that reduces the change to exactly your suggestion:
- `network-table.c`: bind the full `hwaddr` and `IN (substr(...,1,8), …||'0/28', …||'0/36')` with `ORDER BY length(mac) DESC`. No `replace()`.
- `macvendor.py`: reverted — no generator/DB change at all.
- One BATS test seeding manuf-format MA-L/MA-M/MA-S rows: asserts each resolves, that two `/28`s under the same parent resolve independently, and that a `/24`+`/28` overlap picks the more specific one.

Net diff is the one query string plus the test. Appreciate the YAGNI nudge — fair call.

DL6ER (approved):

DL6ER:
Thank you!

PR #2929, opened 2026-06-29
fix(api-docs): correct three OpenAPI spec issues (#2867)

### LINKED ISSUE #2867: OpenAPI spec: three issues found while generating a Python client
Hi,

I've been generating a Python client from the FTL OpenAPI spec at `src/api/docs/content/specs/`. Three issues turned up that are worth fixing in the spec rather than working around them downstream. Happy to send PRs for any or all of these.

### 1. Tag capitalisation mismatch causes generators to emit duplicate classes

The root `tags:` block declares `Pi-hole configuration` (lowercase c). The `/config/*` operations carry `Pi-hole Configuration` (uppercase C). The `/teleporter` operations carry `Pi-hole configuration` (lowercase c).

OpenAPI tag values are case-sensitive. openapi-generator (and several others) treat the two strings as distinct tags and emit two API classes named identically: `PiHoleConfigurationApi`. The second silently shadows the first when imported, so the teleporter operations become unreachable through the standard import path.

Fix: normalise to a single capitalisation (lowercase to match the root declaration), or split teleporter into its own `Teleporter` tag if you'd prefer it as a separately-grouped workflow.

### 2. Several list-all endpoints are documented but not modelled

The endpoint descriptions for `GET /api/groups/{name}`, `GET /api/clients/{client}`, and `GET /api/lists/{list}` say the path identifier is optional and that omitting it returns the full collection. The runtime supports this. But OpenAPI 3.0 disallows optional path parameters, so the spec only models the single-item form, with the parameter marked `required: true`.

Consequence: every typed client refuses to call `/api/groups`, `/api/clients`, `/api/lists` — generators correctly emit a single method that requires the path parameter.

Fix: add separate `GET` operations under `/groups`, `/clients`, `/lists` with their own `operationId`s (e.g. `list_groups`, `list_clients`, `list_lists`) returning the same array schemas the existing single-item GETs already use.

### 3. `GET /config/{element}` declares no response schema

The `200` response for `get_config_elem` only has `description: OK` and no `content` block. Generators emit `-> None` and discard the JSON body, which is unfortunate because the body is a valid filtered config object.

Fix: attach a `content.application/json.schema` referencing the same `config` + `took` composition that `get_config` uses. The shape is documented as a subset of the full config object, so reusing the schema is correct.

### Reproducing

I hit all three while generating a Python client against FTL master (Core v6.4.1). Fixes verified by patching the bundled spec and regenerating; client now produces a clean `TeleporterApi`, working `list_*` methods, and a typed return for `get_config_elem`. Live-tested against a Pi-hole instance.

Happy to send a PR if any of this is welcome — just let me know which way you'd like the tag capitalisation to go.

DL6ER:
Addresses the three OpenAPI spec issues reported in #2867. All three are spec-only changes; no runtime behavior is altered.

### 1. Tag capitalization mismatch

The root `tags:` declaration in `main.yaml` and the teleporter operations use `Pi-hole configuration` (lowercase c), but `config.yaml` used `Pi-hole Configuration` (capital C). OpenAPI tags are case-sensitive, so generators treated these as two distinct tags and emitted two identically named `PiHoleConfigurationApi` classes, silently shadowing the teleporter operations.

Normalized `config.yaml` to `Pi-hole configuration`, matching the convention used by every other spec file (first letter of the phrase only). This is the direction suggested by @rdwebdesign in the issue thread.

### 2. List-all GET endpoints were not modelled

The runtime serves `GET /api/groups`, `GET /api/clients` and `GET /api/lists` (the collection form), because `api_list` reads all rows when no identifier is given (`src/api/list.c`). OpenAPI 3.0 forbids optional path parameters, so the collection form could not be expressed on the `/{name}` path and typed clients could only call the single-item form.

Added `list_groups`, `list_clients` and `list_lists` GET operations on the bare `/groups`, `/clients` and `/lists` paths, reusing the same array schemas and examples as the single-item GETs. `list_lists` also carries the optional `type` query parameter for parity with `get_lists`.

### 3. `get_config_elem` had no response schema

The `200` response for `GET /config/{element}` only had `description: OK` with no `content`, so generators emitted a `None`/`void` return type and discarded the JSON body. Attached the `config` + `took` schema composition, since the body is a documented filtered subset of `GET /config`.

### Out of scope

The separate `string | string[]` request-field discussion (raised later in the thread) is intentionally not included here; it is tracked separately as a design decision.

### Verification

- All four YAML files parse cleanly.
- Full OpenAPI resolution (openapi-spec-validator) passes with all cross-file `$ref`s resolved.


DL6ER:
The two `test_openapi.py` failures came from the new bidirectional cross-check, which compares the specs against FTL's `/api/endpoints` introspection output rather than against actual routing. At runtime `api_list` does serve `GET` on the bare `/api/clients`, `/api/groups` and `/api/lists` paths (routing matches by prefix; the `parameters` column is descriptive only), so the spec additions were correct. But the endpoint table only registered `HTTP_GET` on the `/{item}` rows, so introspection emitted `/api/groups/{name}` etc. and never the bare collection form -> the three documented collection GETs had no counterpart, and the count came out 89 vs 92.

rdwebdesign (approved):

PR #2932, opened 2026-07-02
Swap misaligned comments for domain-needed and expand-hosts.

0xpsyduck:
The inline configuration comments for `domain-needed` and `expand-hosts` were swapped, incorrectly matching the dnsmasq man page descriptions.

### DIFF
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -405,16 +405,16 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, boo
 
 	if(conf->dns.domainNeeded.v.b)
 	{
-		fputs("# Add the domain to simple names (without a period) in /etc/hosts in\n", pihole_conf);
-		fputs("# the same way as for DHCP-derived names\n", pihole_conf);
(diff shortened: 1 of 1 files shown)


yubiuser:
Good catch!

Please base on and target `development` branch

DL6ER:
The failing API tests (`test/api/test_api.py` counter mismatches -- `total_queries` 137 vs 135, `DNSKEY` 9 vs 7, `top_domain` "." vs "localhost") are not caused by this change.

They come from the old flaky upstream-DNSSEC detection in the test harness: the bats suite resolves `mask.icloud.com` upstream, and when the `icloud.com` -> `apple-dns.net` CNAME chain crosses DNSSEC-signed zones, dnsmasq fires two extra root DNSKEY validation queries, shifting several counters by +2. The `detect_upstream_dnssec` fixture tried to detect this at runtime, but the detection races the actual resolution and is unreliable.

This was made deterministic on `development` by commit 20164bff ("tests: resolve iCloud Private Relay zones from local pdns"), which resolves those zones from a local PowerDNS. This branch predates that commit. Please rebase onto the latest `development` and the tests should pass.

DL6ER (approved):

PR #2933, opened 2026-07-04
Fix prefix-match bug and improve fallthrough logging in redirect_root_handler

slmingol:
## Summary

`redirect_root_handler` (`src/webserver/webserver.c`) contains two issues:

### 1. Prefix-match false positive in host comparison

The existing check uses:
```c
strncmp(host, config.webserver.domain.v.s, host_len)
```
This compares only `host_len` bytes of the domain string. If the incoming `Host` header is shorter than `webserver.domain`, the comparison stops early and can return a false match — e.g. `Host: pi` matching domain `pi.hole` (first 2 bytes match).

**Fix:** add an exact-length equality check before `strncmp`:
```c
const size_t domain_len = strlen(config.webserver.domain.v.s);
if(host != NULL && host_len == domain_len &&
   strncmp(host, config.webserver.domain.v.s, host_len) == 0)
```

### 2. Silent fallthrough produces a confusing 403

When the host does not match `webserver.domain`, the handler returns `0`, falling through to CivetWeb's default handling. Because directory listing is disabled, CivetWeb generates a 403 and serves `error403.html` from the admin package — which contains the Pi-hole-branded "Access denied / Did you mean to go to your Pi-hole's dashboard?" message.

This makes the 403 look like an intentional access control decision when it is actually a side-effect of directory listing being disabled. Users deploying Pi-hole behind a reverse proxy where the proxy forwards a `Host` header that differs from `webserver.domain` hit this with no actionable error in the logs.

**Fix:** replace the bare `"Not redirecting %s"` log with one that names both the received host and the configured domain:
```
Not redirecting / (Host: "pi-hole5.bub.lan" != domain: "pi-hole5.svcs.lamolabs.com")
```

## What is not changed

The host-vs-domain check itself is preserved. The intentional behavior — not redirecting when the host is a blocked domain in IP blocking mode — is unchanged.

## Reproducing the confusing 403

```bash
# With webserver.domain = "pi.hole", this returns 308 (expected):
curl -si http://localhost:80/ -H "Host: pi.hole" | head -2

# This returns 403 with Pi-hole's branded error page (confusing):
curl -si http://localhost:80/ -H "Host: some-other-hostname" | head -2

# But /admin/ works fine regardless of Host:
curl -si http://localhost:80/admin/ -H "Host: some-other-hostname" | head -2
```

Discovered while deploying Pi-hole v6 in Docker behind Traefik (Pangolin) and pfSense/HAProxy, where the proxy's internal `Host` header differed from `webserver.domain`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/webserver/webserver.c
+++ b/src/webserver/webserver.c
@@ -167,7 +167,7 @@ static int redirect_root_handler(struct mg_connection *conn, void *input)
 	// (e.g. host "pi" incorrectly matching domain "pi.hole").
 	const size_t domain_len = strlen(config.webserver.domain.v.s);
 	if(host != NULL && host_len == domain_len &&
-	   strncmp(host, config.webserver.domain.v.s, host_len) == 0)
+	   strncasecmp(host, config.webserver.domain.v.s, host_len) == 0)
 	{
 		// 308 Permanent Redirect from http://pi.hole -> http://pi.hole/admin/
 		if(strcmp(uri, "/") == 0 || strcmp(uri, config.webserver.paths.prefix.v.s) == 0)
--- a/src/webserver/webserver.c
+++ b/src/webserver/webserver.c
@@ -160,10 +160,14 @@ static int redirect_root_handler(struct mg_connection *conn, void *input)
 		log_debug(DEBUG_API, "URI: %s", uri);
 	}
 
-	// Check if the requested host is the configured (defaulting to pi.hole)
-	// Do not redirect if the host is anything else, e.g. localhost or a
-	// blocked domain in IP blocking mode
-	if(host != NULL && strncmp(host, config.webserver.domain.v.s, host_len) == 0)
+	// Check if the requested host is the configured domain (defaulting to pi.hole).
+	// Do not redirect if the host is anything else, e.g. a blocked domain in
+	// IP blocking mode where the browser connects using the blocked hostname.
+	// Use an exact-length comparison to prevent a prefix-match false positive
+	// (e.g. host "pi" incorrectly matching domain "pi.hole").
+	const size_t domain_len = strlen(config.webserver.domain.v.s);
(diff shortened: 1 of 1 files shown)


DL6ER (changes requested):
Thanks for the fix -- the prefix-match bug is real and the exact-length guard (`host_len == domain_len`) is the right approach. I confirmed it also cleans up the empty-host / `host_len == 0` case, where the old `strncmp(..., 0)` matched unconditionally.

One request before merge: while you are touching this comparison, please make it case-insensitive. HTTP `Host` is case-insensitive, so `Host: PI.HOLE` should still match domain `pi.hole`, but `strncmp` rejects it. FTL already provides `strncasecmp` (wrapped in `FTL.h`), so:

```c
if(host != NULL && host_len == domain_len &&
   strncasecmp(host, config.webserver.domain.v.s, host_len) == 0)
```

Once that change lands we will approve and merge.

DL6ER:
One note on the red CI: the 7 failing API tests are unrelated to your change. They are a known-flaky upstream-DNSSEC test -- the bats suite resolves `mask.icloud.com` upstream, and when its CNAME chain crosses DNSSEC-signed zones dnsmasq fires two extra root DNSKEY validation queries, shifting several counters by +2 (`total_queries` 137 vs 135, `DNSKEY`, `top_domain`).

This was made deterministic on `development` by commit 20164bff ("tests: resolve iCloud Private Relay zones from local pdns"). This PR targets `master`, which has not received that fix yet, so the failure will clear once it propagates to `master` -- or you can rebase onto `development`. Note that FTL PRs are normally opened against `development` rather than `master`.

So actually two requests for changes here:
- the `strncmp` -> `strncasecmp` change from the review above
- rebasing your branch/fork on latest `development`

DL6ER (approved):

PR #2940, opened 2026-07-07
Add native DoT/DoH support for encrypted upstreams (part 1/2)

DL6ER:

Native **DoT** (DNS-over-TLS, [RFC 7858](https://www.rfc-editor.org/rfc/rfc7858.html)) and **DoH** (DNS-over-HTTPS, [RFC 8484](https://www.rfc-editor.org/rfc/rfc8484.html)) for FTL's **outbound** resolution - FTL talking to an encrypted upstream resolver - with **zero dnsmasq changes**.

**How it works.** dnsmasq cannot speak TLS to its upstreams, so FTL runs a small local forward-proxy. For every encrypted entry in `dns.upstreams`, FTL binds a loopback listener pair (UDP+TCP) in `127.47.11.0/24` and emits a `server=127.47.11.N#port` line into `dnsmasq.conf`. dnsmasq forwards the plaintext query to that listener as usual; a dedicated FTL worker thread then re-encrypts it, forwards it to the real resolver over DoT/DoH via mbedTLS, and hands the answer back. No query leaves the host in the clear.

**What is added** (new module `src/dotdoh/`, leaf units self-contained and unit-tested):
- `upstream_uri` - parser/classifier for `tls://`, `https://` and plaintext upstream entries.
- `framing` - DoT two-octet length prefix ([RFC 7858](https://www.rfc-editor.org/rfc/rfc7858.html)) and DoH HTTP/1.1 request/response ([RFC 8484](https://www.rfc-editor.org/rfc/rfc8484.html)). Bytes only; DNS message contents are never parsed here.
- `registry` - one loopback listener pair per upstream, bind-first (no `SO_REUSEADDR`, so nothing can co-bind).
- `tls_client` - fail-closed mbedTLS client: required chain + hostname verification, SNI, a pooled connection per upstream with a single reconnect. Never downgrades to plaintext.
- `edns_pad` - pads each outbound query to a 128-octet boundary with an EDNS(0) Padding option ([RFC 7830](https://www.rfc-editor.org/rfc/rfc7830.html) / [RFC 8467](https://www.rfc-editor.org/rfc/rfc8467.html)) so the ciphertext length no longer leaks the query size. Only the encrypted leg is padded. As a missing padding reduces obscuring the question length somewhat but does not weaken the integrity and/or encryption, this is implemented fail-open, so a query that cannot be padded is still sent rather than dropped.
- `proxy` - the forward-proxy worker thread, armed after dnsmasq startup so the listener fds are not reused.

**Config / API:**
- `dns.upstreams` now accepts `tls://` and `https://` URIs (validated by the parser).
- New `dns.upstreamCA` trust-anchor option (documented in the OpenAPI spec).
- The suggested-servers list gained v4/v6-**pinned** DoT/DoH URIs for the well-known resolvers (Google, Quad9, Cloudflare, OpenDNS), e.g. `tls://dns.google@8.8.8.8`. Pinning the address (`sni@ip`) avoids an untrusted bootstrap lookup of the resolver hostname before switching to it.

**Security posture:** encrypted upstreams are fail-closed - a failed certificate or hostname verification drops the answer rather than falling back to plaintext.

| Privacy feature | Shipped |
| --- | --- |
| CA trust-anchor chain verification - system trust store or a custom `dns.upstreamCA` bundle | :white_check_mark: |
| Server certificate hostname and SNI verification | :white_check_mark: |
| Strict privacy ([RFC 8310](https://www.rfc-editor.org/rfc/rfc8310.html)) - fail-closed, never downgrades to plaintext | :white_check_mark: |
| EDNS(0) query padding to a 128-octet boundary ([RFC 7830](https://www.rfc-editor.org/rfc/rfc7830.html) / [RFC 8467](https://www.rfc-editor.org/rfc/rfc8467.html)) | :white_check_mark: |

## How to test this locally

Set an encrypted upstream - via the web UI, a `PATCH /api/config`, or by editing `dns.upstreams` in `pihole.toml` - for example `tls://dns.google@8.8.8.8` (DoT) or `https://cloudflare-dns.com@1.1.1.1/dns-query` (DoH). FTL restarts, arms the proxy, and rewrites `dnsmasq.conf`. Then resolve normally; the query is carried to the upstream over TLS:

```
dig @127.0.0.1 example.com
```

FTL's log shows a `dotdoh: <DoT|DoH> upstream <name> armed` line for each encrypted upstream.

## How the CI tests this automatically

Two layers, both wired into the existing test job (`test/run.sh`):

1. **Unit / regression harness** - a standalone `dotdoh_regression` executable (same pattern as `tar_regression`/`gzip_regression`) exercises the URI parser and the DoT/DoH framing against a table of expected results. It is built in CI via `-DBUILD_DOTDOH_REGRESSION=ON` (`.github/Dockerfile` + `build.sh`) and run as a bats case in `test_suite.bats` ("DoT/DoH parser regression harness"). It can be built with ASan/UBSan locally via `./build.sh "-DDOTDOH_REGRESSION_SANITIZE=ON" test`.
2. **End-to-end** - `test/dotdoh.bats` configures a real encrypted upstream and resolves **real DoT and DoH** through a local TLS shim (`test/dotdoh_shim.py`, terminating TLS with the repo test certificate) into the recursor, asserting the decrypted answer. It runs **after** the pytest API suite on purpose: switching the upstream restarts FTL, which would otherwise perturb the query statistics those API tests assert on. `test_final.bats` still accounts for its config writes.

## Scope: part 1 of 2

This PR is **part 1 of 2** and deliberately covers **only the outbound direction** (FTL as a DoT/DoH *client* toward the upstream resolver).

**Part 2** - the **server** direction (FTL terminating inbound DoT/DoH from downstream clients and handing plaintext to dnsmasq) - will follow as a **separate** PR, submitted once this one is reviewed and merged, to keep each direction independently reviewable.

Much of the foundation laid here is deliberately reusable for part 2: the wire framing (`src/dotdoh/framing.c`), the URI/config plumbing, and the mbedTLS and test-shim scaffolding are shared between both directions - the module is named `dotdoh` (not `client`) for exactly this reason, so the server side can live beside it.

---

---

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template.
5. I have squashed any insignificant commits.

## Checklist:

- [x] I signed off all commits (DCO).
- [x] I signed all my commits.

### DIFF
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -25,7 +25,7 @@ RUN sed -i '/buffer_with_truncation /d' /bats/bats-core/libexec/bats-core/bats-f
 # Remove possible old build files
 RUN rm -rf cmake && \
 # Build and test FTL
-    bash build.sh "-DSTATIC=${STATIC} -DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON" ${BUILD_OPTS} && \
+    bash build.sh "-DSTATIC=${STATIC} -DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON" ${BUILD_OPTS} && \
 # Copy FTL binary to root directory
     cd / &&\
     cp /app/pihole-FTL . && \
--- a/build.sh
+++ b/build.sh
@@ -109,9 +109,9 @@ fi
 # They are gated behind CMake options so ordinary builds do not produce them.
 if [[ -n "${test}" ]]; then
     if [[ -n "${cmake_args}" ]]; then
-        cmake_args="${cmake_args} -DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON"
+        cmake_args="${cmake_args} -DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON"
     else
-        cmake_args="-DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON"
+        cmake_args="-DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON"
     fi
 fi
 
@@ -210,7 +210,7 @@ else
     cp pihole-FTL ../
     # Copy the regression test binaries alongside it so the bats tests can run
     # them from the repo root.
-    for regression_bin in tar_regression gzip_regression; do
+    for regression_bin in tar_regression gzip_regression dotdoh_regression; do
         if [[ -f "${regression_bin}" ]]; then
             cp "${regression_bin}" ../
         fi
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -303,6 +303,7 @@ add_executable(pihole-FTL
         $<TARGET_OBJECTS:config>
         $<TARGET_OBJECTS:tools>
         $<TARGET_OBJECTS:ntp>
+        $<TARGET_OBJECTS:dotdoh>
         )
 if(STATIC)
     set_target_properties(pihole-FTL PROPERTIES LINK_SEARCH_START_STATIC ON)
@@ -373,6 +374,7 @@ add_subdirectory(syscalls)
 add_subdirectory(config)
 add_subdirectory(tools)
 add_subdirectory(ntp)
+add_subdirectory(dotdoh)
 
 # Standalone regression harness for the in-memory tar parser (zip/tar.c).
 # It #includes tar.c directly to reach the file-internal helpers, stubs the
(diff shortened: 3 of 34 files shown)


DL6ER:
Review comments becoming less and less relevant the last two rounds, mostly edge case handling inside the tests, not the production code. I guess this means the PR is well converged by now. I did just fix the last ones.

rdwebdesign:
Does FTL always run the local forward-proxy or it is executed only when a DoT/DoH server is selected?

DL6ER:
> Does FTL always run the local forward-proxy or it is executed only when a DoT/DoH server is selected?

It starts one proxy per defined encrypted upstream. If there's are none, no proxies are started.

Erasure5959:
Maybe a silly question (I suspect this is more linked to libraries) - I assume DoQ/DoH3 are out of scope?

DL6ER:
You library-related-feeling is exactly right. This PR is implementing DoT/DoH in a very lightweight way without adding any additional dependencies.  Also the subsequently following server implementation (separate PR after this one) won't need extra libraries. Unfortunately, the currently used and battle-proven CivetWeb and MbedTLS libraries are not currently supporting what would be needed for QUIC/HTTP3.

Hence, a prerequisite is replacing mbedTLS by OpenSSL. This is already on the roadmap and actually already in the TOP3. This way, we gain QUIC (DoQ) rather easily, for DOH3, we'd still need another library (`nhttp3`) so I won't promise anything on this end right now.

Keep watching this space, the mbedTLS -> OpenSSL migration is likely one of the next things to come and DoQ can then follow afterwards.

darkexplosiveqwx:
Could it make sense for any of this to go into upstream dnsmasq?

DL6ER:
No, I don't think so. It is too large and Simon has declared multiple times that he thinks this is way out of scope and that this should be realized through specialized sidecars instead of dnsmasq-native features.

010011110:
this is a great feature i have been waiting for very long time. i have tested it and it works great. anything can be done to get this feature soon? but more important: will we also get the connection to clients support encryption, too?

DL6ER:
I have a series of changes I'd like to get reviewed/tested and merged in order:

1. DoH + DoT to upstream (FTL is client) - this PR
2. OpenSSL migration (#2941)
3. Opt-in experimental support for HTTP/2 (PR to be created, code already written)
4. DoH + DoT downstream to the clients (FTL is server). This is what you have asked for. The code is mostly written, PR to be opened.
5. DoQ client + server (code yet to be finished)

Once this all is tested and verified, we will probably aim for the next release. However, I ask for your patience as reviewing is a slow process so this won't land tonight or next week.

---

**edit** I started with the client feature as this is where encryption is probably more "needed". The upstream connection goes through the open Internet so encryption really brings confidentiality. Downstream to your clients happens in your local network where this isn't really needed, here the feature is more needed for modern apps and browsers which are expecting a DoT / DoT upstream. However, when your client is abroad, you should be using a VPN connection so even "plain" DNS is still encrypted in transit today.

DL6ER:
Failing tests are expected now until https://github.com/pi-hole/FTL/pull/2948 has been merged. The number of root DNSKEY queries in the test depends ICANN's published root key set, so the ongoing key-signing-key rollover silently shifted the counters (DNSKEY 9 -> 7, TOTAL 137 -> 135, TOP_DOMAIN . -> localhost). That is why the suite started failing on completely unrelated PRs. https://github.com/pi-hole/FTL/pull/2948 is fixing this by making the DNSSEC test hermetic, i.e., fully self-contained without depending on external (possibly changing) records.

PromoFaux (approved):
Been running this in prod now for nearly 2 weeks and have seen no issues. Shipit (to dev)

PR #2954, opened 2026-07-12
webserver: enable CivetWeb UNIX domain socket support (USE_X_DOM_SOCKET)

cyberb:
### What
Enable the `USE_X_DOM_SOCKET` compile-time definition for the bundled CivetWeb build and document the `x<path>` syntax in the `webserver.port` help text.

### Why
CivetWeb already implements listening on a UNIX domain socket via the `x<path>` syntax in `listening_ports`, and FTL passes `webserver.port` straight through to it. However the feature is gated behind `USE_X_DOM_SOCKET`, which FTL does not define, so `pihole-FTL` can only bind its web server to a TCP port.

Enabling it lets users run the web server (and REST API) entirely over a UNIX socket - e.g. behind a reverse proxy - without exposing any TCP port. For example:

```toml
[webserver]
port = "x/run/pihole/webserver.sock"
```

### Notes
- Default behaviour is unchanged (default `webserver.port` still uses TCP).
- One-line CMake change plus a documentation sentence; no C code is touched.


darkexplosiveqwx:
How do the DNS records `;; local.api.ftl.      		CH	TXT` and `;; domain.api.ftl.     		CH	TXT` behave with this change when a unix socket is configured?
I think it would be helpful for UNIX sockets to be supported when using [`pihole api`](https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/api.sh).

cyberb:
Thanks for taking a look, and good point.

You're right that the `*.api.ftl` CHAOS TXT records won't produce a usable value for a UNIX socket: `get_api_string()` synthesizes `http://<addr>:<port>/api/` from the listening port, and a socket has no `host:port` to format, so the discovery records (and therefore `pihole api` / `api.sh`, which rely on them) don't apply.

The intent of this PR is deliberately narrow: just enable the *transport* so CivetWeb can bind a socket. The main use case is running FTL's web/API behind a reverse proxy (nginx, Caddy, ...) with no TCP port exposed at all — in that setup clients reach the API through the proxy and never consult the `*.api.ftl` records, so the discovery gap doesn't bite. Anyone deliberately setting `webserver.port = "x/path"` is opting into exactly that kind of no-TCP deployment and can reasonably be expected to know the local auto-discovery / direct-CLI path doesn't apply.

So my suggestion would be to treat this as a documented limitation rather than block the transport support on it. I'm happy to:

- add a sentence to the `webserver.port` help text noting that `x<path>` sockets aren't advertised via the `*.api.ftl` records / usable by `pihole api`, and/or
- follow up separately to teach `get_api_string()` to emit a socket-aware form (which would also need a companion `curl --unix-socket` change in `pi-hole/pi-hole`'s `api.sh`) if you'd like first-class CLI support.

Would that scoping be OK with you?

darkexplosiveqwx:
Here is what I'm thinking:
- If a unix socket is configured in `webserver.port`, then `local.api.ftl. CH TXT` should be something like `unix:///run/pihole/webserver.sock` with `domain.api.ftl` being empty.
- `api.sh` (and ideally PADD too) should detect `local.api.ftl.` starting with `unix://...` and should use `curl --unix-socket ...`
- even though civetweb uses `x` for UNIX, it may be kind of a 'magic constant' to use it in the config. Maybe have the config be `unix:///...` and turn that into the `x` internally for a more clear API?
- I'm not sure if civetweb does any validation or if we should do some path validation (like enforcing absolute paths and checking if we can write the socket) in FTL first.
- When using a reverse proxy it may also be helpful to be able to override `domain.api.ftl` manually, since the client won't be using the port FTL is listening on. I tried to request this in https://discourse.pi-hole.net/t/allow-manual-configuration-of-domain-api-ftl/79096, but I didn't know how to explain it at that time so I gave up on it.
- (maybe make `domain.api.ftl` also empty if all TCP ports are bound to `::1` or any IP in the `127.0. 0.0/8` subnet, since thats explicitly loopback only. For reverse proxies, that should still be manually overridable)

This doesn't need to be all done here, but it is something to consider.

cyberb:
That all sounds right to me, and I like the direction — especially exposing the socket as `unix:///path` in the config and translating to CivetWeb's `x` internally rather than leaking the `x` magic constant into the public API. The `local.api.ftl` = `unix://...` / empty `domain.api.ftl` scheme (with `api.sh`/PADD using `curl --unix-socket`), the path validation, and the manual `domain.api.ftl` override for reverse proxies all make sense as the proper end-to-end design.

I'm happy to align with whatever shape the team prefers here. Since this touches the config surface (and `api.sh`/PADD in pi-hole/pi-hole), I'd rather not front-run the design — if you'd like to drive how it should look, I'm glad to help implement whichever pieces are useful, or to fold the `unix://` translation into this PR once there's agreement on the syntax. Just let me know how you'd like to proceed.

darkexplosiveqwx:
> I'm happy to align with whatever shape the team prefers here. Since this touches the config surface (and api.sh/PADD in pi-hole/pi-hole), I'd rather not front-run the design — if you'd like to drive how it should look, I'm glad to help implement whichever pieces are useful, or to fold the unix:// translation into this PR once there's agreement on the syntax. Just let me know how you'd like to proceed.

I am actually not part of the team - just an active community member with an interest in (system) design. Before you go ahead with my suggestions, I'd also much rather get feedback from the team.

DL6ER:
Thanks for the PR, and to @darkexplosiveqwx for the detailed back-and-forth - the shape you two converged on (a `unix:///path` syntax translated internally, socket-aware `*.api.ftl` records with a matching `curl --unix-socket` path in pi-hole, path validation, a manual `domain.api.ftl` override) is the right end-to-end design, so I won't repeat those points.

Stepping back, though, I am not convinced the use case justifies carrying this. A reverse proxy in front of FTL already works today by pointing it at a loopback TCP port (`127.0.0.1:<port>`), which is not exposed off-host. Switching that to a UNIX socket removes only the loopback port, which is not a meaningful reduction in attack surface on a single machine, and in exchange it breaks the on-box CLI/discovery (the `*.api.ftl` records go empty, so `pihole api`/`api.sh`/PADD stop working) and adds a config surface we would have to support indefinitely. So the practical benefit over "bind to loopback and proxy to it" looks thin, and I'd want to see actual, recurring demand for the socket specifically before committing to it.

On top of that there is a security angle nobody has raised yet, and it is what makes me most hesitant: we daemonize with `umask(0)`, and civetweb binds the UNIX socket with a plain `bind()` and no `chmod`, so it ends up mode 0777 - connectable by any local user or process. `webserver.acl` does not cover this either: the ACL matcher only understands IP subnets, so for a UNIX socket it either allows everything (the default) or denies the socket outright if an ACL is set. That leaves the session password as the only guard and exposes that surface to every local process by default, with no knob for owner/group/mode. Enabling the transport without a permission model is exactly the kind of open-ended surface I'd rather not add.

So I would prefer not to flip this flag on its own. And to be clear, the fact that it is cheap to implement - a one-line flag, or even the fuller design being quick to write now that tooling does most of the typing - is not itself an argument for it. Ease of writing the code doesn't change the long-term cost: every transport we expose is a surface we then have to document, support, and keep secure for years, regardless of how few lines it took. If real demand for a socket-only deployment does show up, we can revisit it as the complete design above - including proper socket ownership/permissions - but I don't want to enable it mainly because civetweb happens to support it and it is little work to turn on.

darkexplosiveqwx:
Sounds very reasonable.

We already have those user/group API permissions - kind of.
A unix socket would allow setting setting permissions on a filesystem level.
`pihole api` already works just by being user/group `pihole`.
We have: authenticated API on localhost and the password stored in the runtime file `/etc/pihole/cli_pw` with `-rw-r----- pihole:pihole` permissions.
This is one of the main arguments for a unix socket, but the current approach is arguably simpler: any API access needs a password, no matter local or external. Locally there is simply an easy way to obtain it.
UNIX sockets raise the question: should it be unauthenticated and ACL via filesystem permissions (for local tools), but if one adds a reverse proxy it would also be unauthenticated. We would need `cli_pw` again - complexity with little gain.

Regarding the other issues/potential improvements to `domain.api.ftl` I mentioned, maybe I care enough to submit something, but it has a very low priority.

cyberb:
Heads-up: I pushed a second commit to this PR that turned out to be **required** for socket-only operation.

While actually running FTL with `webserver.port = "x<path>"` (no TCP port at all), I found the **entire REST API returned 404** — every `/api/*` request fell through to static file serving, so the web UI rendered but couldn't read or write anything.

Root cause: `get_server_ports()` uses `mg_get_server_ports()`, which doesn't report UNIX domain sockets, so a socket-only config yields zero ports. The webserver-init guard

```c
if((ctx = mg_start2(...)) == NULL || !get_server_ports())
    return;
```

then returns **after** the server has started but **before** `mg_set_request_handler(ctx, "/api", api_handler, ...)`, leaving `/api` unregistered. The new commit only treats a NULL `mg_start2()` context as fatal and calls `get_server_ports()` for its side effect without gating handler registration on it.

So this PR now enables `USE_X_DOM_SOCKET` *and* makes it actually usable. Happy to split it into two PRs if you'd prefer to review them separately.

darkexplosiveqwx:
@cyberb The code is not the issue here.
We currently think that it is likely out of scope.
If you disagree please answer:
- Why does your setup require a UNIX socket (or anyone's setup)?
- Explain why this would not be possible (or severely restricted) using the loopback bind.
- Keep in mind that socket permissions are quite complicated to do (not necessarily code-wise, but design-wise), as explained by __DL6ER__ and me above.

cyberb:
Actually no problem, thanks!

DL6ER:
Seems we agreed to close this.

PR #2955, opened 2026-07-12
config: make the install/config directory a compile-time option (PIHOLE_INSTALL_DIR)

cyberb:
### What
Introduce a single compile-time base directory, `PIHOLE_INSTALL_DIR` (default `/etc/pihole`), and route the hardcoded configuration/state paths through it: `pihole.toml`, the legacy config, `dnsmasq.conf`, the hosts dir, `custom.list`, `dhcp.leases`, backups, `cli_pw`, `versions`, the inotify watch dir, and the embedded dnsmasq `CONFFILE`.

Packagers can now relocate the whole tree at configure time:

```sh
cmake -DPIHOLE_INSTALL_DIR=/opt/pihole/etc ...
```

### Why
Everything currently assumes a hardcoded `/etc/pihole`. Packages that install FTL under a different prefix - snaps, `/opt` installs, read-only-root or otherwise confined images - have to patch the sources. A single overridable base directory removes that need.

### Notes
- Default is unchanged, so existing builds and installs behave exactly as before.
- **Build-time knob only** - this deliberately does *not* add a runtime override. Runtime-settable paths (e.g. `files.database`, `files.log.*`) keep their existing config keys, and `files.pid` stays fixed per GHSA-6w8x-p785-6pm4.
- The default lives in a `#ifndef` guard in the new `src/install_paths.h`, and the top-level `CMakeLists.txt` exposes it as a cache variable + global compile definition so it also reaches the vendored dnsmasq `CONFFILE`.

Happy to adjust the naming (`PIHOLE_INSTALL_DIR` vs `PIHOLE_ETC_DIR`/`PIHOLE_CONFIG_DIR`) to your preference.


darkexplosiveqwx:
I think this should also override the `/etc/pihole/` of the default config options.
Like the default value of `files.gravity` being  `PIHOLE_INSTALL_DIR/gravity.db`

cyberb:
Good call — done. I pushed a change that also routes the `/etc/pihole` defaults of the runtime-settable options through `PIHOLE_INSTALL_DIR`: `files.database`, `files.tmp_db`, `files.gravity`, `files.macvendor` and `webserver.tls.cert`.

So now every `/etc/pihole` path in the tree derives from the base dir. I left the `/var/log/pihole` log defaults (`files.log.*`) alone since that's a different base directory — happy to give those the same treatment (e.g. a `PIHOLE_LOG_DIR`) if you'd like that in scope too.

DL6ER (changes requested):
The dual default works well: the CMake cache variable plus the `#ifndef` fallback in `install_paths.h` means every TU resolves the same value even when compiled without the `-D`, keeping `files.pid` fixed per GHSA-6w8x-p785-6pm4 is correct, and the default build stays byte-for-byte unchanged. One issue needs resolving before we merge, plus a few smaller things.

## Teleporter export and import become asymmetric for relocated installs

In `src/zip/teleporter.c` the archive stores each file under its absolute path minus the leading slash, but the import side matches hardcoded entry names:

```c
const char *extract_files[] = {
    "etc/pihole/pihole.toml",   // fixed
    "etc/pihole/dhcp.leases",   // fixed
    config.files.gravity.v.s... // derived, matches the export side
};
```

This PR makes the export names depend on `PIHOLE_INSTALL_DIR`: `dhcp.leases` via the changed `file_path = DHCPLEASESFILE`, and `pihole.toml` transitively through `GLOBALTOMLPATH`. So with `-DPIHOLE_INSTALL_DIR=/opt/pihole/etc` the export writes `opt/pihole/etc/pihole.toml` and `opt/pihole/etc/dhcp.leases`, while import still only accepts `etc/pihole/...`. The result is that config and DHCP-lease import silently no-op on a relocated build (they fall into the "Skipping file ..." branch), both on a same-host round-trip and when moving an archive between differently-configured installs.

We should settle the intended contract and make both sides match. If Teleporter archives are meant to stay portable across installs, i.e., the way backups get moved between machines, the canonical `etc/pihole/...` names should stay fixed on both sides regardless of `PIHOLE_INSTALL_DIR`, which argues for reverting the `dhcp.leases` change rather than extending it. If instead archives are treated as install-local, the import `extract_files[]` entries need to be derived from `PIHOLE_INSTALL_DIR` too. The fixed-name option is the more robust of the two, as it keeps existing archives importable on any build. Either way the current half-state is the one thing we cannot keep. The gravity and FTL-db entries already derive from the runtime config value symmetrically, so only `pihole.toml` and `dhcp.leases` are affected.

## Minor

- `src/install_paths.h`: the header comment says "Compile-time installation path *prototypes*", which is copy-paste from a function-prototype header, as this file declares no prototypes. "Compile-time installation paths" fits better.
- `src/config/dnsmasq_config.c`: line 248 now emits `GLOBALTOMLPATH`, but a few lines down the same generated header still hardcodes `"...in /etc/pihole/pihole.toml)"`. The partial change leaves the generated `dnsmasq.conf` header inconsistent on a relocated build, so both should use `GLOBALTOMLPATH`.
- `src/args.c`: the `--gen-x509` help examples still print `/etc/pihole/tls.pem`, which is now inconsistent with the relocated default we show for `webserver.tls.cert`. Low priority, but worth aligning here.
- The `/etc/pihole` mentions left in descriptive help and comment strings (the `cli_pw` help, the inotify debug description, the config-backup comment) are prose and fine to leave.

## Tests

Nothing in CI exercises a non-default `PIHOLE_INSTALL_DIR`, which is exactly the configuration where the teleporter gap shows up. A minimal relocated build plus a Teleporter export/import round-trip would catch it, worth adding once the contract above is settled.

## Naming

`PIHOLE_INSTALL_DIR` reads fine. Since the value is the config and state root rather than a general install prefix (the log dir lives elsewhere by design), `PIHOLE_CONFIG_DIR` would be marginally more precise, but we do not feel strongly either way.

jacklul:
Wouldn't also introducing an install prefix (+ support for it on the scripts side) be good here? This would open the possibilities for custom installs/packaging a bit more.

```
# defaults:
PIHOLE_INSTALL_PREFIX=
PIHOLE_INSTALL_DIR=$PIHOLE_INSTALL_PREFIX/etc/pihole
PIHOLE_SCRIPTS_DIR=$PIHOLE_INSTALL_PREFIX/opt/pihole
PIHOLE_BIN_DIR=$PIHOLE_INSTALL_PREFIX/usr/local/bin
PIHOLE_TMP_DIR=$PIHOLE_INSTALL_PREFIX/tmp
```

One could simple relocate it under new prefix:
```
PIHOLE_INSTALL_PREFIX=/opt
```

Or change locations of specifics only:
```
PIHOLE_INSTALL_DIR=/new/etc/pihole
```

I've been maintaining Entware package of Pi-hole for a while now and I simply [patch all the paths](https://github.com/jacklul/entware-pi-hole/blob/develop/patches/ftl-001-paths.patch) so this PR has caught my interest as it can make my work a bit easier.

Just dropping an idea here...

darkexplosiveqwx:
Compared to **jacklul**, I am doing some RPM packaging, but I am also patching many of the paths.
The main difference in RPM packaging is that it doesn't give Pi-hole its own prefix (like under /opt), but instead aims to follow the Filesystem Hierarchy Standard (FHS).
The paths I am patching:
`/usr/local/bin/pihole` -> `/usr/bin/pihole` for gravity
 `/var/www/html` -> `/usr/share/pihole/www` as the webroot default (on Fedora only Apache ("httpd") should own  `/var/www/html`)
`/etc/pihole/{pihole-FTL.db,pihole-tmp.db,gravity.db,dhcp.leases,config_backups/}` -> `/var/lib/pihole{pihole-FTL.db,pihole-tmp.db,gravity.db,dhcp.leases,config_backups/}`
`/etc/pihole/{macvendor.db,versions}` -> `/usr/share/pihole/{macvendor.db,versions}`

of course they are not that complicated to patch, but less downstream patches are always preffered.

For FHS the following variables may be useful:
| Variable | Upstream default | Intended FHS |
|---|---|---|
| `PIHOLE_BIN_DIR` | `/usr/local/bin` | `/usr/bin` |
| `PIHOLE_CONFIG_DIR` | `/etc/pihole` | `/etc/pihole` |
| `PIHOLE_DATA_DIR` | `/etc/pihole` | `/var/lib/pihole` |
| `PIHOLE_SHARE_DIR` | `/etc/pihole` | `/usr/share/pihole` |
| `PIHOLE_LOG_DIR` | `/var/log/pihole` | `/var/log/pihole` |

<details>
  <summary>How I imagine paths to be mapped using those variables</summary>

| Source file | Variable | Expanded path |
|---|---|---|
| `src/config/config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/pihole.toml` |
| `src/config/config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/pihole-FTL.conf` |
| `src/config/config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/setupVars.conf` |
| `src/config/config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/migration_backup_v6` |
| `src/config/config.c` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/tls.pem` |
| `src/config/dnsmasq_config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/dnsmasq.conf` |
| `src/config/dnsmasq_config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/dnsmasq.conf.temp` |
| `src/config/dnsmasq_config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/hosts` |
| `src/config/dnsmasq_config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/hosts/custom.list` |
| `src/config/dnsmasq_config.h` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/custom.list` |
| `src/config/dnsmasq_config.c` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/pihole.toml` |
| `src/config/password.c` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/cli_pw` |
| `src/config/inotify.c` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR` |
| `src/dnsmasq/CMakeLists.txt` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/dnsmasq.conf` |
| `src/config/config.c` | `PIHOLE_DATA_DIR` | `PIHOLE_DATA_DIR/pihole-FTL.db` |
| `src/config/config.c` | `PIHOLE_DATA_DIR` | `PIHOLE_DATA_DIR/pihole-tmp.db` |
| `src/config/config.c` | `PIHOLE_DATA_DIR` | `PIHOLE_DATA_DIR/gravity.db` |
| `src/config/dnsmasq_config.h` | `PIHOLE_DATA_DIR` | `PIHOLE_DATA_DIR/dhcp.leases` |
| `src/files.h` | `PIHOLE_DATA_DIR` | `PIHOLE_DATA_DIR/config_backups` |
| `src/api/info.c` | `PIHOLE_SHARE_DIR` | `PIHOLE_SHARE_DIR/versions` |
| `src/config/config.c` | `PIHOLE_SHARE_DIR` | `PIHOLE_SHARE_DIR/macvendor.db` |
| `src/config/config.c` | `PIHOLE_SHARE_DIR` | `PIHOLE_SHARE_DIR/www` |
| `src/config/config.c` | `PIHOLE_LOG_DIR` | `PIHOLE_LOG_DIR/FTL.log` |
| `src/config/config.c` | `PIHOLE_LOG_DIR` | `PIHOLE_LOG_DIR/pihole.log` |
| `src/config/config.c` | `PIHOLE_LOG_DIR` | `PIHOLE_LOG_DIR/webserver.log` |
| `src/config/legacy_reader.c` | `PIHOLE_LOG_DIR` | `PIHOLE_LOG_DIR/FTL.log` |
| `src/api/action.c` | `PIHOLE_BIN_DIR` | `PIHOLE_BIN_DIR/pihole` |
| `src/args.c` | `PIHOLE_CONFIG_DIR` | `PIHOLE_CONFIG_DIR/tls.pem` |
| `src/api/docs/content/specs/config.yaml` | same as `config.c` | same as `config.c` |

</details>

Compatibilty with **cyberb**'s and **jacklul**'s ideas:
If the FHS-compatible prefixes are not defined, the default value (`/etc/pihole/`, or var/log...) can be prefixed with a PIHOLE_INSTALL_PREFIX. A PIHOLE_INSTALL_PREFIX can also apply to all the tmp paths.

> (+ support for it on the scripts side)

This may be a bit more involved as the core scripts do not have a compilation/build step and are instead used directly.

**DL6ER** raised the concern of teleporter incompatibilities, it should always use the default `/etc/pihole` and then apply them to the configured path.

Again just some thoughts, open to some suggestions to fit you packaging needs.

DL6ER:
Two of the open points were ours to decide.

**Teleporter.** Archives should stay portable across installs, so the canonical `etc/pihole/...` entry names stay fixed on both sides regardless of `PIHOLE_INSTALL_DIR`. Decoupling the archive entry name from the on-disk source path is enough:

```c
mz_zip_writer_add_file(zip, "etc/pihole/pihole.toml", GLOBALTOMLPATH, ...);
mz_zip_writer_add_file(zip, "etc/pihole/dhcp.leases", DHCPLEASESFILE, ...);
```

Both names currently come from `file_path+1`, so a relocated build writes `opt/pihole/etc/...` while the import side still matches the fixed names in `extract_files[]`, and config and lease import silently no-op. This affects `pihole.toml` as well, transitively through `GLOBALTOMLPATH`.

**Scope.** We would like to take `PIHOLE_INSTALL_DIR` as it is. Where our files *should* live is a far broader question, touching the core scripts and the web tree rather than only FTL, and it deserves a proper discussion instead of being settled inside this PR. pi-hole/pi-hole#5844 is the open thread for it, with pi-hole/pi-hole#2976 as earlier background, and our Discourse is the better venue if it needs a wider audience. We are happy to take an FHS-style split as a follow-up once that is decided.

Remaining items, all small:

1. `src/api/docs/content/specs/config.yaml` still hardcodes the five defaults this PR makes configurable (`cert`, `database`, `tmp_db`, `gravity`, `macvendor`), plus two mentions in `lists.yaml`. We serve that spec ourselves, so a relocated install documents paths it does not use.
2. `src/config/dnsmasq_config.c:253` still hardcodes `"...in /etc/pihole/pihole.toml)"` just below the hunk that switched to `GLOBALTOMLPATH`.
3. The `--gen-x509` help in `src/args.c` still prints `/etc/pihole/tls.pem`.
4. `install_paths.h` says "Compile-time installation path *prototypes*", but declares none.
5. Worth rejecting a relative or trailing-slash value in CMake, as the teleporter strips the leading `/` by offset.

Please also rebase - the failing checks are `test/api/test_api.py` counter assertions on a July base and unrelated to your change, but that is not visible from the PR page.

cyberb:
Rebased onto current `development`.

**Teleporter:** took the fixed-name option — `pihole.toml` and `dhcp.leases` keep the canonical `etc/pihole/...` entry names regardless of `PIHOLE_INSTALL_DIR`, so archives stay portable and existing ones still import. Export and import now share the same two defines. Scope left as-is, no FHS split.

All five minor items done: specs (`config.yaml` + `lists.yaml`) are substituted at build time via `configure_file`, `dnsmasq_config.c` uses `GLOBALTOMLPATH`, `args.c` help uses `PIHOLE_INSTALL_DIR`, header comment fixed, and CMake now rejects a relative or trailing-slash value. Added a test asserting the exported archive uses the canonical names.

Kept the `PIHOLE_INSTALL_DIR` name — happy to rename if you prefer `PIHOLE_CONFIG_DIR`.

DL6ER:
Keep `PIHOLE_INSTALL_DIR`. The knob covers config, state and shared data alike - `pihole.toml`, `dnsmasq.conf`, the databases, `versions`, `config_backups` - while `PIHOLE_CONFIG_DIR` names only one of those roles, and it is the name an FHS split would want for the narrower directory. Renaming now would either force a second rename later or quietly change what the same variable means between releases.

The conversion itself is complete - every functional `/etc/pihole` under `src/` is gone and the teleporter entry names are right. Three things left, all small:

1. Two help strings still hardcode the path, and they are user-visible in `pihole.toml` and in our API docs, so a relocated build documents a directory it does not use. Same class as the specs last time, just missed: `webserver.api.cli_pw.h` ("stored in clear in /etc/pihole", `config.c:1261`) and `debug.inotify.h` ("monitoring of /etc/pihole filesystem events", `config.c:1646`).

2. `args.c:521-553` concatenates `PIHOLE_INSTALL_DIR` into the `printf` format string, so a path containing a `%` becomes a format specifier. `write_config_header()` does it the other way round and passes the path as an argument - please do the same here.

3. `test_teleporter_export_entry_names()` skips when `_teleporter_archive` is `None`, so it goes green without asserting anything the day the earlier test stops populating it. Better to fail there, or to export the archive within the test itself.

cyberb:
Pushed, rebased onto current `development`.

1. Both help strings follow `PIHOLE_INSTALL_DIR` now — `webserver.api.cli_pw.h` and `debug.inotify.h`. The only `/etc/pihole` left under `src/` are three code comments and the default in `install_paths.h`.

2. `--gen-x509` and `--read-x509` pass the directory as a `printf` argument, the way `write_config_header()` does.

   While looking for others of the same kind I found ~14 pre-existing sites that concatenate a now-relocatable path into a format string: `main.c:84` `log_info("Parsed config file "GLOBALTOMLPATH" successfully")`, and the same pattern with `DNSMASQ_TEMP_CONF`, `DNSMASQ_PH_CONFIG`, `DNSMASQ_HOSTSDIR`, `DNSMASQ_CUSTOM_LIST_LEGACY` and `SETUPVARS_CONF` in `dnsmasq_config.c`, `setupVars.c` and `api/teleporter.c`. They were harmless while the path was a fixed literal — this PR is what makes them reachable. Rather than rewrite all of them I made CMake reject a `%` in `PIHOLE_INSTALL_DIR`, alongside the existing absolute/trailing-slash checks: it closes the class at the source and keeps the diff to the paths conversion. Happy to convert the call sites instead if you would rather not constrain the value.

3. `test_teleporter_export_entry_names()` exports its own archive via `GET /api/teleporter` and asserts on that, so it can no longer go green without checking anything. It no longer reads `TestEndpointResponses._teleporter_archive` — `test_teleporter_import` still does, but that is pre-existing and untouched.

Kept `PIHOLE_INSTALL_DIR` as the name, per your reasoning.

DL6ER:
The relocation itself checks out - I built with `-DPIHOLE_INSTALL_DIR=/opt/pihole/etc` and the paths land everywhere they should, including the substituted API spec.

One gap: you pinned `ZIPNAME_TOML` and `ZIPNAME_DHCPLEASES`, and the new test says the names have to stay portable, but the gravity and FTL database entries still take theirs from the runtime path, on both export and import (`file_path = config.files.gravity.v.s` plus stripping the leading slash). Relocated, the archive then holds `opt/pihole/etc/gravity.db`, and importing that into a stock install skips it with a plain `log_info("Skipping file %s in Teleporter archive")` - no error, just a partial restore. The asymmetry predates your PR, but relocation is what makes it reachable. Either pin those two the same way or drop the portability wording from the test.

On the other sites that concatenate the path into a format string: leave them. The `%` check in `CMakeLists.txt` is the right mitigation and rewriting 14 call sites is churn. It does need a comment saying what it guards, though - as written nobody will guess, and someone will relax it later.

cyberb:
Both done, in two commits on top of the existing one so the delta is readable — happy to squash before merge.

**Database entry names** (`6d129bf`) — pinned rather than dropping the test wording. `ZIPNAME_GRAVITY` and `ZIPNAME_FTLDB` now hold `etc/pihole/gravity.db` and `etc/pihole/pihole-FTL.db` on export, and the import `extract_files[]` entry for gravity uses the same define instead of deriving from `config.files.gravity.v.s`. So all four entries are portable now, not two.

The import target is unaffected — gravity is still imported table-wise into whatever `files.gravity` points at, only the name inside the archive is fixed. That also lines the code up with things that already assumed the canonical name: the `"File etc/pihole/gravity.db in ZIP archive ..."` error strings in `test_and_import_database()`, and `TELEPORTER_FILES_EXPORT` / `TELEPORTER_FILES_IMPORT` in `test/api/libs/responseVerifyer.py`. Worth noting this closes the same hole for a *non*-relocated install that simply sets `files.gravity` elsewhere — it was reachable before this PR too, just not by default.

`test_teleporter_export_entry_names()` now asserts all four names.

**The `%` check** (`17c8cbf`) — comment added above it naming `main.c`'s `log_info("Parsed config file "GLOBALTOMLPATH" successfully")` as the reason and saying what has to change before it can be relaxed.

DL6ER (approved):
Looks good. I built this with `-DPIHOLE_INSTALL_DIR=/opt/pihole/etc` and the archive entry names stay `etc/pihole/...` while the on-disk paths move, which is exactly the property the test claims. Pinning the FTL database name as well was the right call.

One for later, not blocking: this is a silent break for anyone who moved `files.gravity`. Archives they exported before this carry the runtime-derived entry name, and the import now matches `ZIPNAME_GRAVITY` only, so their gravity database is skipped with nothing but the usual `Skipping file ...` line. Keeping the old name as an alias covers it:

```c
const char *extract_files[] = {
	ZIPNAME_TOML,
	ZIPNAME_DHCPLEASES,
	ZIPNAME_GRAVITY,
	config.files.gravity.v.s[0] == '/' ? config.files.gravity.v.s + 1 : config.files.gravity.v.s
};
```

The duplicate on a default install is harmless, the loop matches the first.

DL6ER:
The review side is done, this is approved. What holds the merge is that `development` requires verified signatures and all three commits here are unsigned.

The steps are the same ones I wrote out for another contributor recently: https://github.com/pi-hole/FTL/pull/2986#issuecomment-5302563987 - GPG or SSH, whichever you already have. Since there are three commits, re-sign them in one go:

```
git rebase --exec 'git commit --amend --no-edit -S' origin/development
git push --force-with-lease
```

Make sure the key is uploaded under Settings -> SSH and GPG keys as a *signing* key, otherwise GitHub still reports the commits as unverified. Let me know if you get stuck anywhere.

DL6ER:
For what it is worth, I could squash-merge this as it stands - GitHub signs that commit itself, so the requirement is satisfied that way. Setting signing up properly on your side is still the better route long-term, since it applies to everything you send us from here on.

DL6ER:
Heads-up: #3053 just landed and moved the embedded API documentation out of the source tree into the build directory. It touches the same `foreach` your `configure_file` block sits in.

I trial-merged your branch onto current `development` to check. It merges cleanly and the two compose the way they should: `OUTPUT_FILE` keeps the new `${CMAKE_CURRENT_BINARY_DIR}` and your substitution still applies on top. I built the result with `-DPIHOLE_INSTALL_DIR=/opt/pihole/etc` and the embedded spec carries `/opt/pihole/etc/tls.pem` while the source tree stays clean. So nothing is needed from you for correctness here.

Since you will be force-pushing for the signatures anyway, rebasing onto current `development` in the same go would be tidy.

cyberb:
Done

PR #2956, opened 2026-07-12
webserver: allow files.log.webserver = "-" to log to stderr

cyberb:
### What
Accept a single dash (`-`) for `files.log.webserver` to make the web server log to stderr instead of a file, mirroring the existing behaviour of `files.log.dnsmasq`.

### Why
`files.log.dnsmasq` already accepts `-` to print to stderr, but `files.log.webserver` only accepts a regular file path. That makes it impossible to run FTL without *any* on-disk log file - e.g. in a container or a confined package that ships all logs to the journal via stdout/stderr. With this change every FTL log destination has a uniform "no file" option.

### Changes
- `validate_webserver_logfile()` now accepts `-` (like `validate_filepath_dash`).
- `log_web()` writes to `stderr` when the path is `-` instead of opening a file.
- Help text / allowed-values updated.

### Notes
Default behaviour and any existing file path are unchanged.

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -1373,8 +1373,8 @@ void initConfig(struct config *conf)
 	conf->files.log.dnsmasq.c = validate_filepath_dash;
 
 	conf->files.log.webserver.k = "files.log.webserver";
-	conf->files.log.webserver.h = "The log file used by the webserver";
-	conf->files.log.webserver.a = cJSON_CreateStringReference("Any writable file");
+	conf->files.log.webserver.h = "The log file used by the webserver.\n\n A single dash (\"-\") makes the webserver log to stderr instead of a file.";
+	conf->files.log.webserver.a = cJSON_CreateStringReference("Any writable file, or \"-\" for stderr");
 	conf->files.log.webserver.t = CONF_STRING;
 	conf->files.log.webserver.f = FLAG_RESTART_FTL;
 	conf->files.log.webserver.d.s = (char*)"/var/log/pihole/webserver.log";
--- a/src/config/validator.c
+++ b/src/config/validator.c
@@ -356,6 +356,10 @@ bool validate_filepath_dash(union conf_value *val, const char *key, char err[VAL
 // execution.
 bool validate_webserver_logfile(union conf_value *val, const char *key, char err[VALIDATOR_ERRBUF_LEN])
 {
+	// Dash is allowed, this enables printing to stderr instead of a file
+	if(strlen(val->s) == 1 && val->s[0] == '-')
+		return true;
+
 	// Regular file-path validation first
 	if(!validate_filepath(val, key, err))
 		return false;
--- a/src/log.c
+++ b/src/log.c
@@ -359,8 +359,11 @@ void __attribute__ ((format (printf, 1, 2))) log_web(const char *format, ...)
 	// pihole-FTL instance is logging into the same file
 	const long pid = (long)getpid();
 
-	// Open web log file
-	FILE *weblog = fopen(config.files.log.webserver.v.s, "a+");
+	// Open web log file. A single dash ("-") means "write to stderr" instead
+	// of a file, mirroring files.log.dnsmasq.
+	const char *weblog_path = config.files.log.webserver.v.s;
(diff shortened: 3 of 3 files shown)


darkexplosiveqwx:
Hey,
while working on https://github.com/pi-hole/FTL/pull/2897 I found out that dnsmasq closes stdout and stderr on  during it's startup.
Does your patch work without something like https://github.com/pi-hole/FTL/pull/2897/changes/3c24e973a319b73d5f4c5ed22b0d37dd70a27eb7?
My PR would also allow logging without any files on disk, but utilizing a structured JSON stream for FTL, webserver and dnsmasq logs.

cyberb:
Thanks — and thanks for pointing at #2897, I hadn't seen it.

You're right that #2897 is the more complete answer here: reworking the logging destinations (JSON/journald) and removing the dnsmasq `dup2(.../dev/null)` on the std fds solves "no log files on disk" for FTL, the webserver and dnsmasq together, which this small change does not.

This PR is only meant as a minimal parity fix: `files.log.dnsmasq` already accepts `-` for stderr, and this gives `files.log.webserver` the same option, so a deployment can avoid the on-disk webserver log without any larger rework. If #2897 lands and reworks the log destinations anyway, this becomes redundant and I'm happy to close it.

So I'll defer to your preference: keep this as the minimal standalone change, or drop it in favour of #2897? Either is fine by me. I'll also double-check the stderr path against the dnsmasq std-fd closing you mentioned before pushing this further.

DL6ER:
Thanks for the PR, and thanks @darkexplosiveqwx for flagging the std-fd angle - it turns out to be the crux here.

I traced the stderr path and it does not survive our default start. The embedded dnsmasq redirects fd 2 to `/dev/null` during startup (`dup2(nullfd, STDERR_FILENO)` in `src/dnsmasq/dnsmasq.c`, unless `OPT_DEBUG`), and we lean on that on purpose, i.e., `go_daemon()` leaves the std streams to dnsmasq (see the note in `src/daemon.c`). We run dnsmasq with `-k`, not `-d`, so `OPT_DEBUG` is off in normal daemon mode *and* in plain `-f` foreground. The webserver thread is already up when that redirect happens, so `log_web()` writing to the process `stderr` at call time lands in `/dev/null`, silently.

That is also why this is not quite parity with `files.log.dnsmasq = "-"`, even though it looks like it: dnsmasq grabs `log_fd = dup(STDERR_FILENO)` in `log_start()` *before* the redirect, precisely so its own `-` keeps working. To match that we would have to capture the original fd early and write through the saved descriptor, plus serialize the per-line writes, since a shared unbuffered `stderr` across the civetweb worker threads would interleave lines. That is essentially re-deriving what #2897 already does centrally, only in a narrower spot.

So I agree with your own read: #2897 is the more complete answer, as it reworks the log destinations for FTL, the webserver and dnsmasq together rather than patching one stream. My current thinking is to converge on a single FTL-owned logging pipeline (format chosen by config, FTL owning the write) that gives a uniform "no file on disk" option across all logs, and to fold the goal of this PR into that effort. I have left the detailed reasoning over in #2897.

I will close this one to keep things focused - not because the idea is wrong, but because it is better solved centrally there. Much appreciated either way, and thanks for getting the ball rolling on it.

cyberb:
Wrapping this up: agreed this is best solved centrally in #2897.

One concrete data point from testing a socket-only, no-file-on-disk setup that's relevant to that effort: **dnsmasq closes stdin/stdout/stderr during startup** (the `dup2(.../dev/null)` you mentioned), so *both* `files.log.webserver = "-"` (stderr) and `files.log.ftl = /dev/stdout` silently drop everything logged after boot — only startup lines survive. So `-`/stderr as proposed here isn't actually a reliable sink on its own; it needs the fd-close removal from #2897 to work.

What did work for us without touching dnsmasq: an empty `files.log.ftl`, which trips FTL's existing `vsyslog()` fallback (`Opening of FTL's log file failed, using syslog instead!`) so the log goes to `/dev/log` → journald and survives the fd close. Sharing in case a syslog destination is useful alongside the JSON/journald work in #2897. Thanks again.

PR #2957, opened 2026-07-13
Fix EDNS MAC persistence in network table

arinl:

Fix persistence of dnsmasq's forwarded EDNS MAC when ECS supplies the logical client address.

FTL already uses the forwarded MAC in memory for client and group identity. During a network-table update, however, a MAC that is not yet in the database falls through to the recent-IP, DHCP-lease, and mock-address paths. The mock path replaces the formatted MAC with `ip-<address>`, so IPv4 and IPv6 ECS addresses from one device are persisted as separate synthetic devices. An existing mock mapping also remains associated with the address instead of being replaced by the forwarded MAC.

- Use recent-IP, DHCP-lease, and mock-address fallbacks only when no six-byte MAC is available.
- Leave the existing fallback behavior unchanged for clients without a MAC.
- Add a regression test that begins with an existing mock IPv4 mapping, sends IPv4 and IPv6 ECS queries carrying the same forwarded MAC, and verifies that both addresses converge on one MAC row with no matching mock rows left behind.

### Scope and trust boundary

This changes network-table identity persistence only. DNS responses and live client-group selection are unchanged.

Forwarded EDNS MAC data is unauthenticated and is already used for live client-group resolution. This change persists that same identity, so an incorrect or spoofed forwarded MAC can also persist in the network table. Deployments using this metadata must trust the forwarding DNS server.

### Evidence and validation

The focused regression fails against unmodified `development`:

```text
ip-192.168.47.97|192.168.47.97
ip-fe80::b167:af1e:968b:dead|fe80::b167:af1e:968b:dead
mac_rows=0
mock_rows=2
```

With this change:

```text
02:00:00:00:00:01|192.168.47.97
02:00:00:00:00:01|fe80::b167:af1e:968b:dead
mac_rows=1
mock_rows=0
```

Additional validation:

- ARM64 architecture, static executable, library, binary-classification, backtrace, and crash-handler checks passed.
- 190 BATS tests passed.
- 149 API tests passed.
- 9 final log-validation tests passed.
- An isolated x86_64 Pi-hole 2026.07.2 container using FTL v6.7 reproduced the control result and persisted both addresses under the forwarded MAC after applying the patch.
- The patched FTL binary integrity check passed.

None. This restores the behavior described by the existing fallback comments and does not add or change a user-facing option.

Development and test preparation were assisted by OpenAI Codex. I reviewed the final diff and validated it with the upstream test suite and an isolated Pi-hole deployment.

---

---

### DIFF
--- a/src/database/network-table.c
+++ b/src/database/network-table.c
@@ -875,7 +875,7 @@ static bool add_FTL_clients_to_network_table(sqlite3 *db, const enum arp_status 
 			if(dbID >= 0)
 				log_debug(DEBUG_ARP, "Network table: Client with MAC %s is network ID %i", hwaddr, dbID);
 		}
-		if (dbID == DB_NODATA)
+		else
 		{
 			//
 			// Variant 2: Try to find a device using the same IP address within the last 24 hours
@@ -897,7 +897,7 @@ static bool add_FTL_clients_to_network_table(sqlite3 *db, const enum arp_status 
 		// if DHCP server is enabled
 		//
 		bool dhcp_lease = false;
-		if (dbID == DB_NODATA && config.dhcp.active.v.b)
+		if (snap_hwlen != 6 && dbID == DB_NODATA && config.dhcp.active.v.b)
 		{
 			log_debug(DEBUG_ARP, "Network table: DHCP server enabled, checking leases for IP %s", ipaddr);
 			FILE *fp = fopen(DHCPLEASESFILE, "r");
@@ -963,7 +963,7 @@ static bool add_FTL_clients_to_network_table(sqlite3 *db, const enum arp_status 
 		// Only try this when there is no EDNS(0) MAC address available
 		// nor a corresponding DHCP lease
 		//
-		if (dbID < 0 && dhcp_lease == false)
+		if (snap_hwlen != 6 && dbID < 0 && dhcp_lease == false)
 		{
 			unlock_shm();
 			dbID = find_device_by_mock_hwaddr(db, ipaddr);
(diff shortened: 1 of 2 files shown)


DL6ER:
Thanks for this, and in particular for being upfront about the OpenAI Codex assistance - the transparency is appreciated and does not count against the PR in the slightest. What matters to us is that a human understood and validated the change, and the evidence here (control vs. patched runs, the isolated container repro, and the full test suite) shows exactly that.

The fix is correct. The three fallbacks - recent-IP, DHCP-lease, and mock-address - were always meant to be "no MAC of last resort" paths; the comments on variants 2 and 4 even say *"Only try this when there is no EDNS(0) MAC address available"*. The guards were keyed on `dbID` rather than on whether we actually hold a six-byte MAC, so a forwarded MAC that was not yet in the database would fall through to the mock path and be rewritten as `ip-<address>`, splitting one device across its IPv4 and IPv6 ECS addresses. Gating all three on `snap_hwlen != 6` restores the documented intent, and the subsequent `INSERT OR REPLACE` on `network_addresses` together with the mock-device cleanup then collapse the stale `ip-` rows, i.e., precisely what your regression test asserts.

The trust-boundary note is the right framing, too: forwarded EDNS MAC data is unauthenticated and we already trust it for live client-group resolution, so persisting the same identity does not widen the trust boundary - it only makes the stored view consistent with the in-memory one.

I'll run it through our CI before merging, but this looks good to me.

arinl:
Thanks for the detailed review and confirmation. I’m glad the fix matches the intended fallback behavior. 😊

DL6ER (approved):

PR #2958, opened 2026-07-14
Change webserver.log format to match FTL.log & Move multiple API-related logsites from FTL.log to webserver.log

darkexplosiveqwx:

**Please make sure you**

---

Discussed and split out from https://github.com/pi-hole/FTL/pull/2897.

The FIFO priority adding allows us to to some colorization in the web interface.

This makes webserver.log overall a more useful log (was quite empty before).
I saw that c358cac3e6c55ce7e41dc00252c318e5a6eca0ef changed a single a single call from `log_web()` (webserver.log) to `log_debug()` (FTL.log) (because it caused a conflict on rebase), now that I added a conditional debug macro we can properly move most API logging to webserver.log.

---

---

### DIFF
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -1790,7 +1790,7 @@ static bool special_domain(const queriesData *query, const char *domain)
 	// _dns.resolver.arpa.     86400   IN      SVCB    2 dns.google. alpn="h2,h3" key7="/dns-query{?dns}"
 	//
 	// RFC 9462, Section 4 says:
-	// 
+	//
 	// If the recursive resolver that receives this query has no Designated
 	// Resolvers, it SHOULD return NODATA for queries to the "resolver.arpa"
 	// zone, to provide a consistent and accurate signal to clients that it
@@ -2040,7 +2040,7 @@ static bool FTL_check_blocking(const char *domainstr, queriesData *query, client
 	}
 
 	// when we reach this point: the query is not in FTL's cache (for this client)
-	
+
 	// Check exact whitelist for match
 	const char *blockedDomain = domainstr;
 	PERF_START(_pcb_allow);
@@ -4261,13 +4261,15 @@ static void _query_set_dnssec(queriesData *query, const enum dnssec_status dnsse
 }
 
 // Add dnsmasq log line to internal FIFO buffer (can be queried via the API)
-void FTL_dnsmasq_log(const char *payload, const int length)
+void FTL_dnsmasq_log(const char *payload, const int priority, const int length)
 {
 	// Lock SHM
 	lock_shm();
 
-	// Add to FIFO buffer
-	add_to_fifo_buffer(FIFO_DNSMASQ, payload, NULL, length);
+	// Add to FIFO buffer. dnsmasq has no FTL debug flags, so its LOG_DEBUG is
+	// a plain "DEBUG" rather than the DEBUG_ANY catch-all priostr() maps to
+	const char *prio = priority == LOG_DEBUG ? "DEBUG" : priostr(priority, DEBUG_NONE);
+	add_to_fifo_buffer(FIFO_DNSMASQ, payload, prio, length);
 
 	// Unlock SHM
 	unlock_shm();
--- a/src/log.c
+++ b/src/log.c
@@ -131,7 +131,7 @@ unsigned int get_year(const time_t timein)
 	return tm.tm_year + 1900;
 }
 
-static const char * __attribute__((const)) priostr(const int priority, const enum debug_flag flag)
+const char * __attribute__((const)) priostr(const int priority, const enum debug_flag flag)
 {
 	switch (priority)
 	{
--- a/src/log.c
+++ b/src/log.c
@@ -131,6 +131,29 @@ unsigned int get_year(const time_t timein)
 	return tm.tm_year + 1900;
 }
 
+static void get_idstr(char *idstr, size_t size)
+{
+	const int pid = getpid(); // Get the process ID of the calling process
+	const int mpid = main_pid(); // Get the process ID of the main FTL process
+	const int tid = gettid(); // Get the thread ID of the calling process
+
+	// There are four cases we have to differentiate here:
+	if(pid == tid)
+		if(is_fork(mpid, pid))
+			// Fork of the main process
+			snprintf(idstr, size, "%i/F%i", pid, mpid);
+		else
+			// Main process
+			snprintf(idstr, size, "%iM", pid);
+	else
+		if(is_fork(mpid, pid))
+			// Thread of a fork of the main process
+			snprintf(idstr, size, "%i/F%i/T%i", pid, mpid, tid);
+		else
+			// Thread of the main process
+			snprintf(idstr, size, "%i/T%i", pid, tid);
+}
+
 const char * __attribute__((const)) priostr(const int priority, const enum debug_flag flag)
 {
 	switch (priority)
@@ -256,27 +279,8 @@ void __attribute__ ((format (printf, 3, 4))) _FTL_log(const int priority, const 
 	// Get and log PID of current process to avoid ambiguities when more than one
 	// pihole-FTL instance is logging into the same file
 	char idstr[42];
-	const int pid = getpid(); // Get the process ID of the calling process
-	const int mpid = main_pid(); // Get the process ID of the main FTL process
-	const int tid = gettid(); // Get the thread ID of the calling process
-
+	get_idstr(idstr, sizeof(idstr));
 	const char *prio = priostr(priority, flag);
-
-	// There are four cases we have to differentiate here:
-	if(pid == tid)
-		if(is_fork(mpid, pid))
-			// Fork of the main process
-			snprintf(idstr, sizeof(idstr)-1, "%i/F%i", pid, mpid);
-		else
-			// Main process
-			snprintf(idstr, sizeof(idstr)-1, "%iM", pid);
-	else
-		if(is_fork(mpid, pid))
-			// Thread of a fork of the main process
-			snprintf(idstr, sizeof(idstr)-1, "%i/F%i/T%i", pid, mpid, tid);
-		else
-			// Thread of the main process
-			snprintf(idstr, sizeof(idstr)-1, "%i/T%i", pid, tid);
 
 	// Print to stdout before writing to file
 	if((!daemonmode || cli_mode) && print_stdout)
@@ -341,43 +345,71 @@ void __attribute__ ((format (printf, 3, 4))) _FTL_log(const int priority, const 
 	}
 }
 
-void __attribute__ ((format (printf, 1, 2))) log_web(const char *format, ...)
+void __attribute__ ((format (printf, 3, 4))) _log_web(const int priority, const enum debug_flag flag, const char *format, ...)
 {
+	// We have been explicitly asked to not print anything to the log
+	if(!print_log && !print_stdout)
+		return;
+
 	char timestring[TIMESTR_SIZE];
 	const time_t now = time(NULL);
 	va_list args;
-
-	// Add line to FIFO buffer
-	char buffer[MAX_MSG_FIFO + 1u];
-	va_start(args, format);
-	const size_t len = vsnprintf(buffer, MAX_MSG_FIFO, format, args) + 1u; /* include zero-terminator */
-	va_end(args);
-	add_to_fifo_buffer(FIFO_WEBSERVER, buffer, NULL, len > MAX_MSG_FIFO ? MAX_MSG_FIFO : len);
 
 	// Get human-readable time
 	get_timestr(timestring, now, true, false);
 
 	// Get and log PID of current process to avoid ambiguities when more than one
 	// pihole-FTL instance is logging into the same file
-	const long pid = (long)getpid();
+	char idstr[42];
+	get_idstr(idstr, sizeof(idstr));
+	const char *prio = priostr(priority, flag);
 
-	// Open web log file
-	FILE *weblog = fopen(config.files.log.webserver.v.s, "a+");
+	// Open web log file (if a path is configured)
+	FILE *weblog = NULL;
+	if(print_log && config.files.log.webserver.v.s != NULL)
+		weblog = fopen(config.files.log.webserver.v.s, "a+");
 
-	// Write to web log file
-	if(weblog != NULL)
+	// _FTL_log() prints these itself, so do not print them twice
+	const bool relay = print_log && weblog == NULL && priority <= LOG_WARNING;
+
+	// Print to stdout before writing to file
+	if(!relay && (!daemonmode || cli_mode) && print_stdout)
 	{
-		fprintf(weblog, "[%s %ld] ", timestring, pid);
+		// Only print time/ID string when not in direct user interaction (CLI mode)
+		if(!cli_mode)
+			printf("%s [%s] %s: ", timestring, idstr, prio);
 		va_start(args, format);
-		vfprintf(weblog, format, args);
+		vprintf(format, args);
 		va_end(args);
-		fputc('\n',weblog);
-		fclose(weblog);
+		printf("\n");
 	}
-	else if(!daemonmode)
+
+	// Print to log file or FIFO
+	if(print_log)
 	{
-		printf("!!! WARNING: Writing to web log file failed!\n");
-		syslog(LOG_ERR, "Writing to web log file failed!");
+		// Add line to FIFO buffer. It lives in shared memory and is
+		// independent of the log file, so it is filled even when we relay
+		char buffer[MAX_MSG_FIFO + 1u];
+		va_start(args, format);
+		const size_t len = vsnprintf(buffer, MAX_MSG_FIFO, format, args) + 1u; /* include zero-terminator */
+		va_end(args);
+		add_to_fifo_buffer(FIFO_WEBSERVER, buffer, prio, len > MAX_MSG_FIFO ? MAX_MSG_FIFO : len);
+
+		if(relay)
+		{
+			// No web log available - keep severe messages durable in FTL.log/syslog
+			_FTL_log(priority, flag, "%s", buffer);
+		}
+		else if(weblog != NULL)
+		{
+			// Write to web log file
+			fprintf(weblog, "%s [%s] %s: ", timestring, idstr, prio);
+			va_start(args, format);
+			vfprintf(weblog, format, args);
+			va_end(args);
+			fputc('\n',weblog);
+			fclose(weblog);
+		}
 	}
 }
 
--- a/src/log.c
+++ b/src/log.c
@@ -31,7 +31,6 @@
 
 static bool print_log = true, print_stdout = true;
 static bool ftl_log_available = true;
-static const char *process = "";
 bool debug_flags[DEBUG_MAX] = { false };
 
 void clear_debug_flags(void)
@@ -46,7 +45,7 @@ void log_ctrl(bool plog, bool pstdout)
 	print_stdout = pstdout;
 }
 
-void init_FTL_log(const char *name)
+void init_FTL_log(void)
 {
 	// Open the log file in append/create mode
 	if(config.files.log.ftl.v.s != NULL)
@@ -65,15 +64,6 @@ void init_FTL_log(const char *name)
 		// Close log file
 		if(logfile != NULL)
 			fclose(logfile);
-	}
-
-	// Store process name (if available), strip path if found
-	if(name != NULL)
-	{
-		if(strrchr(name, '/') != NULL)
-			process = strrchr(name, '/')+1;
-		else
-			process = name;
 	}
 }
 
--- a/patch/civetweb/0001-Increase-niceness-of-all-civetweb-threads-as-DNS-ope.patch
+++ b/patch/civetweb/0001-Increase-niceness-of-all-civetweb-threads-as-DNS-ope.patch
@@ -2,14 +2,13 @@ diff --git a/src/webserver/civetweb/civetweb.c b/src/webserver/civetweb/civetweb
 index da254bb4..b1e5b9a6 100644
 --- a/src/webserver/civetweb/civetweb.c
 +++ b/src/webserver/civetweb/civetweb.c
-@@ -244,6 +244,7 @@ static void DEBUG_TRACE_FUNC(const char *func,
+@@ -244,5 +244,6 @@ static void DEBUG_TRACE_FUNC(const char *func,
  
  #else
  #include "log.h"
 +#include <sys/resource.h>
  #define DEBUG_TRACE(fmt, ...)                                                  \
- 	if(debug_flags[DEBUG_WEBSERVER]) {\
- 		log_web("DEBUG: " fmt " (%s:%d)", ##__VA_ARGS__, short_path(__FILE__), __LINE__); }
+ 		log_web_debug(DEBUG_WEBSERVER, fmt " (%s:%d)", ##__VA_ARGS__, short_path(__FILE__), __LINE__);
 @@ -2890,6 +2891,9 @@ mg_set_thread_name(const char *name)
  	 */
  	(void)prctl(PR_SET_NAME, threadName, 0, 0, 0);
--- a/patch/civetweb/0001-Log-debug-messages-to-webserver.log-when-debug.webse.patch
+++ b/patch/civetweb/0001-Log-debug-messages-to-webserver.log-when-debug.webse.patch
@@ -2,7 +2,7 @@ diff --git a/src/webserver/civetweb/civetweb.c b/src/webserver/civetweb/civetweb
 index 3df8eab9..9b0c6308 100644
 --- a/src/webserver/civetweb/civetweb.c
 +++ b/src/webserver/civetweb/civetweb.c
-@@ -239,9 +239,10 @@ static void DEBUG_TRACE_FUNC(const char *func,
+@@ -239,9 +239,9 @@ static void DEBUG_TRACE_FUNC(const char *func,
  #endif
  
  #else
@@ -10,8 +10,7 @@ index 3df8eab9..9b0c6308 100644
  #define DEBUG_TRACE(fmt, ...)                                                  \
 -	do {                                                                       \
 -	} while (0)
-+	if(debug_flags[DEBUG_WEBSERVER]) {\
-+		log_web("DEBUG: " fmt " (%s:%d)", ##__VA_ARGS__, short_path(__FILE__), __LINE__); }
++		log_web_debug(DEBUG_WEBSERVER, fmt " (%s:%d)", ##__VA_ARGS__, short_path(__FILE__), __LINE__);
  #endif /* DEBUG */
  #endif /* DEBUG_TRACE */
  
--- a/src/api/2fa.c
+++ b/src/api/2fa.c
@@ -308,7 +308,7 @@ enum totp_status verifyTOTP(const uint32_t incode)
 				continue;
 			}
 			const char *which = i == -1 ? "previous" : i == 0 ? "current" : "next";
-			log_debug(DEBUG_API, "2FA code from %s time step is valid", which);
+			log_web_debug(DEBUG_API, "2FA code from %s time step is valid", which);
 			last_counter = counter;
 			pthread_mutex_unlock(&totp_lock);
 			return TOTP_CORRECT;
@@ -317,7 +317,7 @@ enum totp_status verifyTOTP(const uint32_t incode)
 
 	if(reused)
 	{
-		log_warn("2FA code has already been used, please wait %lu seconds",
+		log_web(LOG_WARNING, "2FA code has already been used, please wait %lu seconds",
 		         (unsigned long)(RFC6238_X - (now % RFC6238_X)));
 		pthread_mutex_unlock(&totp_lock);
 		return TOTP_REUSED;
--- a/src/api/action.c
+++ b/src/api/action.c
@@ -120,7 +120,7 @@ static int run_and_stream_command(struct ftl_conn *api, const char *path, const 
 			        WCOREDUMP(status) ? "(core dumped)" : "");
 		}
 
-		log_debug(DEBUG_API, "Gravity return code: %d", code);
+		log_web_debug(DEBUG_API, "Gravity return code: %d", code);
 
 		// Close the reading end of the pipe
 		close(pipefd[0]);
@@ -181,7 +181,7 @@ int api_action_flush_logs(struct ftl_conn *api)
 		                       "Flushing the logs is not allowed",
 		                       "Check setting webserver.api.allow_destructive");
 
-	log_info("Received API request to flush the logs");
+	log_web(LOG_INFO, "Received API request to flush the logs");
 
 	// Flush the logs
 	if(flush_dnsmasq_log())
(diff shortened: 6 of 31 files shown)


yubiuser:
I like the idea in general but think this is a breaking change, because users might monitor `FTL.log` but a lot has been moved to the webserver log.
We should also think about how to handle this on docker, because we only stream `FTL.log` and important errors are now moved to the webserver log.

darkexplosiveqwx:
> I like the idea in general but think this is a breaking change, because users might monitor `FTL.log` but a lot has been moved to the webserver log. We should also think about how to handle this on docker, because we only stream `FTL.log` and important errors are now moved to the webserver log.

This is a breaking change, yes. Both for `webserver.log`'s format and the move of some calls. It has been intentionally been split out as to not drown in a giant logging PR.

I moved quite a lot to `webserver.log` and some sites, especially more severe ones that may affect FTL in its entirety, might need to be moved back to `FTL.log`. I am still thinking about some cases.
For docker it may be possible to tail both `FTL.log` and `webserver.log`. They now share the same format. When no debug flags are set `webserver.log` is pretty empty anyway. (On my machine it mostly only contains "Initializing HTTP server on ports ..." and "Authentication required, redirecting to /admin/login"; the latter is got debug-gated after c358cac3e6c55ce7e41dc00252c318e5a6eca0ef)

DL6ER:
CI builds are failing. I'm - in principle - fine with a breaking change (a.k.a. FTL v7.0) if we have real good reasons for it. We should be very clear and sure about the format we change into because we obviously don't want to change it again (= another breaking change) in the nearer future. I will try to do a comprehensive review sometime over the next days.

darkexplosiveqwx:
> CI builds are failing.

One line the tests are looking for moved from FTL to webserver.log. Since I haven't decided yet on whether to move all the current calls, I didn't adjust the tests yet.

> I'm - in principle - fine with a breaking change (a.k.a. FTL v7.0) if we have real good reasons for it. We should be very clear and sure about the format we change into because we obviously don't want to change it again (= another breaking change) in the nearer future. I will try to do a comprehensive review sometime over the next days.

On its own I don't think this deserves a v7.0. Maybe in tandem with some other larger change. Arguably the move to OpenSSL https://github.com/pi-hole/FTL/pull/2941 would be also a breaking change, since it changes `--read-x509`'s output and has noticeable implications to building from source.
The move of logsites is still up for debate, but not doing the file format would be very inconvenient for structured logging. Before, only some calls included a manual priority in the format string and now all calls include their severity. I simply matched `FTL.log`'s format, because this allows one to reuse their existing parser instead of having another different format.

You decide whether this deserves a major v7 bump, or if a v6.x is fine, since we do include less severe breaking changes in minor (not patch) releases.

The later discussed change of `FTL.log` and `webserver.log` using a cached fd is also technically breaking, since it will require `SIGUSR2` for logrotation. Owning `dnsmasq`'s log will also not be without any breaking change, as it would break anyone using a custom `log-facility` dnsmasq config.

Just saying that retaining 100% backwards compatibility will be difficult and that they are not very severe changes (but breaking nonetheless). Not like v5 -> v6.

Looking at https://github.com/pi-hole/FTL/pulls?q=is%3Apr+label%3A%22Breaking+change%22 , there have been three breaking changes released into a FTL minor.

rdwebdesign:
@yubiuser 

> We should also think about how to handle this on docker, because we only stream `FTL.log` and important errors are now moved to the webserver log.

We probably can stream the 2 logs, using something like this:

```bash
FTLstartFrom=$(stat -c%s "${FTLlogFile}")
WEBstartFrom=$(stat -c%s "${WEBlogFile}")

if [ "${TAIL_FTL_LOG:-1}" -eq 1 ]; then
    # Start tailing the FTL log file from the EOF position we recorded on container start
    tail -F -c +$((FTLstartFrom + 1)) -- "${FTLlogFile}" &
    tail -F -c +$((WEBstartFrom + 1)) -- "${WEBlogFile}" &
else
    ...
```

DL6ER:
Oh yes, I definitely see a combination satisfying a breaking version jump. There is not only the upgrade/migration to openSSL but also on its way (80% done): full up and down DoH, DoT, and DoQ support. DoQ heavily depends on OpenSSL as mbedTLS cannot do QUIC, and I wouldn't want to implement this when we can get it "for free" using OpenSSL, which is also fixing some quirks we have on the HTTP server with mbedTLS (like 100% CPU spinning threads).

The breaking changes you have found either had a migration logic (so no users actually saw a difference, just working with the API or directly in the config file) changed. However, the change of the log file cannot really have some migration code, so the format definitely be different.

darkexplosiveqwx:
> Oh yes, I definitely see a combination satisfying a breaking version jump. There is not only the upgrade/migration to openSSL but also on its way (80% done): full up and down DoH, DoT, and DoQ support. DoQ heavily depends on OpenSSL as mbedTLS cannot do QUIC, and I wouldn't want to implement this when we can get it "for free" using OpenSSL, which is also fixing some quirks we have on the HTTP server with mbedTLS (like 100% CPU spinning threads).

Thinking some more about it, we already have some headline features that could satisfy a major bump: built-in encrypted DNS, openSSL, prometheus metrics and also the major adminLTE 4 migration, and then structured journald/JSON logging. My point was about a 7.0 just having a logfile change would be underwhelming. With those features combined it does make sense.

Will there be a `development-v7` branch or is v7 planned with a more regular cycle (unlike the longer v6 cycle)?

DL6ER:
I seriously hope there will be no long release cycle this time. But I guess OpenSSL and DoH/DoT/DoQ will require longer testing periods each by the core team before they will approve. There is nothing wrong about this, but I wouldn't expect being able to cut the v7.0 release before ... September earliest (but this is just completely my feeling, not much more than a best guess after looking into a crystal ball oracle).

darkexplosiveqwx:
> There is nothing wrong about this, but I wouldn't expect being able to cut the v7.0 release before ... September earliest (but this is just completely my feeling, not much more than a best guess after looking into a crystal ball oracle).

September sounds like an optimistic target, the last v6.y cycles were: 94, 44, 83 and 31 days respectively, going from v6.6 -> v6.7 through v6.3 -> v6.4. With v6.7 having been released at the beginning of July, sometime in September would be 2-3 Months after. Nothing here screams like needing a two year beta test like v6 (yet), but taking one, two months more wouldn't still be unreasonable for testing. 

Seeing DoH/DoT merged to `development` makes me assume that no `development-v7` will exist.

DL6ER:
The mechanical bulk reads well.

Two things to settle first. Both are the concrete follow-up to what @yubiuser raised above about important errors leaving `FTL.log`, and they are about *error durability* rather than the format break itself, which we have accepted for the v7.0 bump:

1. Errors can silently disappear when `files.log.webserver` is empty.
    `_log_web()` writes to the in-memory FIFO plus the web log file, nothing else. When the web log path is unset - a state we support, i.e., the reason for the NULL guard - a relocated `log_web(LOG_ERR, ...)` ends up only in the volatile ~512-line FIFO ring: not persisted, and in daemon mode not on stdout or syslog either. Those lines used to be `log_err()` into `FTL.log` and always durable. Note that tailing both logs, as suggested above, does not cover this case, as there is no web log to tail. I would like `_log_web()` to fall back to `FTL.log` (or stdout) when no web log is configured, so severe messages never vanish. We can discuss this further - it is just something that came to my mind during review. I can also see a point for leaving the web log undefined on purpose, e.g., for privacy or simply not needing it, so I have no hard opinion yet.

3. Route the critical sites by severity, not by source file.
    Several moved sites are process-health signals, not web events, e.g., OOM (`Memory allocation failed in ...` in `src/api/stats.c`, the `Failed to allocate memory ...` sites in `src/webserver/webserver.c`), database-open failures, and the teleporter import errors. A teleporter failure is most likely a config- or data-related problem that merely happens to be triggered through the web API, not a webserver problem, and someone debugging a failed import will look in `FTL.log`, not `webserver.log`. So `LOG_ERR`/`LOG_CRIT` for these - OOM, DB failures, teleporter - should stay in `FTL.log`; only the INFO/DEBUG and genuinely web-scoped warnings should move to `webserver.log`.

DL6ER:
I missed the CodeQL status when I commented above.

image

CodeQL is unhappy about two `cpp/path-injection` alerts from the new env override:

- `src/log.c:54` (`init_FTL_log`) and `src/log.c:304` (`_FTL_log`): `config.files.log.ftl.v.s` reaches `fopen()` straight from `FTLCONF_files_log_ftl`.

These are new with this PR - `development` has no `cpp/path-injection` alerts; the flow appears only because `getLogFilePathENV()` feeds the env var into `fopen()` without the `validate_filepath` the normal load runs. That is also why the required CodeQL check fails while `Analyze (c-cpp)` passes.

The exposure is operator-level, i.e., whoever sets that env already controls the FTL process, so it is not remotely reachable, and I do not consider it a real vulnerability. I would still rather fix it than suppress it, because the env path is currently the one config source that skips validation: `config.files.log.ftl.c` is already `validate_filepath`, but `getLogFilePathENV()` `strdup()`s the value in without ever running it.

Please run the env value through `validate_filepath()` in `getLogFilePathENV()` before accepting it, and on failure return `false` so we fall through to TOML/legacy, with a `log_warn` naming the rejected path. Concretely, wrap the value in a temporary `union conf_value` and pass it with a key and a `char err[VALIDATOR_ERRBUF_LEN]` buffer, the same way the normal validated load calls it. That keeps the early ENV path consistent with the regular config load and doubles as the sanitizer, so both alerts clear without a dismissal.

darkexplosiveqwx:
Invalid environment variables are now rejected (it intentionally runs without write access so we see what file it tried to write)
```sh
$ FTLCONF_files_log_ftl='@@@' ./pihole-FTL -f
ERROR: Opening of FTL log (/var/log/pihole/FTL.log) failed: Permission denied
Using syslog instead!
2026-07-26 14:08:50.348 CEST [126365M] INFO: ########## FTL started on frodo-blue! ##########
2026-07-26 14:08:50.348 CEST [126365M] INFO: FTL branch: log-misc
2026-07-26 14:08:50.348 CEST [126365M] INFO: FTL version: v6.7-57-g2f695e90-dirty
2026-07-26 14:08:50.348 CEST [126365M] INFO: FTL commit: 2f695e90-dirty
2026-07-26 14:08:50.348 CEST [126365M] INFO: FTL date: 2026-07-26 13:30:12 +0200
2026-07-26 14:08:50.348 CEST [126365M] INFO: FTL user: john-blue
2026-07-26 14:08:50.348 CEST [126365M] INFO: Compiled for x86_64 (compiled locally) using cc (GCC) 16.1.1 20260515 (Red Hat 16.1.1-2)
2026-07-26 14:08:50.349 CEST [126365M] INFO: 1 FTLCONF environment variable found (0 used, 1 invalid, 0 ignored)
2026-07-26 14:08:50.349 CEST [126365M] ERROR:   [✗] FTLCONF_files_log_ftl files.log.ftl: not a valid file path ("@@@"), using default instead
```

Regarding missing errors/warnings, do you imaginine something like this:
```c
	// Write to web log file
	if(weblog != NULL)
	{
		fprintf(weblog, "%s [%s] %s: ", timestring, idstr, prio);
		va_start(args, format);
		vfprintf(weblog, format, args);
		va_end(args);
		fputc('\n',weblog);
		fclose(weblog);
	}
	else if(config.files.log.webserver.v.s == NULL && priority >= LOG_WARNING) {
		_FTL_log(priority, flag, "%s", buffer);
	}
``` 

I could also see the point for intentionally disabling the log - but I would image one being aware of the consequences of disabling it.

DL6ER:
`2f695e907` is the split I had in mind - malformed uploads stay web-scoped, resource/DB/filesystem failures go back to `FTL.log`.

On CodeQL: my suggestion does not work, the two alerts are unchanged at `src/log.c:54` and `src/log.c:302`. A project-local boolean predicate is not a dataflow barrier for `cpp/path-injection`, so the flow `getenv` -> `strdup` -> `config.files.log.ftl.v.s` -> `fopen` is still reported no matter where we put the call. `validate_filepath()` is also just a character whitelist that permits `.` and `/`, so it does not actually stop traversal and would not be a correct barrier even if CodeQL modelled it. Please keep it anyway - it is the right thing for consistency with the regular load, and it rejects the `@@@` case you showed - but I will dismiss both alerts instead of chasing them. `FTLCONF_files_log_ftl` already reaches `fopen()` on `development` through `readEnvValue()`; your commit only makes the flow local enough for CodeQL to follow it, so there is no new exposure.

On the fallback, the direction is right but the snippet has two problems. First, the comparison is inverted: syslog priorities ascend as severity drops (`LOG_CRIT` 2 < `LOG_ERR` 3 < `LOG_WARNING` 4 < `LOG_INFO` 6), so `priority >= LOG_WARNING` catches WARNING/NOTICE/INFO/DEBUG and skips exactly the ERROR and CRIT lines we are trying to save. It needs to be `priority <= LOG_WARNING`. Second, `config.files.log.webserver.v.s == NULL` will practically never be true: the default is `/var/log/pihole/webserver.log`, `initConfig()` always seeds `v` from `d`, and we have no empty-means-disabled handling for that key (unlike `files.pcap`, which uses `validate_filepath_empty`). Setting `files.log.webserver = ""` passes `validate_webserver_logfile()` and then simply fails in `fopen()`. So the case that actually loses messages is the *open* failing - empty path, unwritable directory, read-only filesystem - and there we currently drop the line in daemon mode without even the syslog fallback that `_FTL_log()` has. Please gate on `weblog == NULL` instead, and rather than duplicating the formatting, just hand the already-formatted `buffer` over:

```c
	else if(priority <= LOG_WARNING)
	{
		// No web log available - keep severe messages durable
		_FTL_log(priority, flag, "%s", buffer);
	}
```

That reuses the whole stdout/`FTL.log`/syslog chain and lets the ad-hoc `"Writing to web log file failed!"` printf go away.

Two things are still open from my second point, all in `src/webserver/webserver.c`: the OOM sites (lines 64, 73, 82, 91, 100, 360, 506, 665, 715, 814, 836, 948) and the startup failures - 619 `Not starting web server as webserver.port is empty`, 652 `Initializing HTTP library failed!`, 658 `Failed to build web paths ...` and the whole `mg_start2()` failure block at 871-874. Those are process-health signals: "web interface will not be available" is what someone greps `FTL.log` for. Line 874 also became circular - we now write "Hint: Check the webserver log at ..." *into* the webserver log. Worth noting that a bad `files.log.webserver` is itself a plausible reason for that block to run, and then all four lines disappear in daemon mode, i.e., no web interface and nothing in the logs.

`src/api/stats_database.c` keeps its 21 SQL prepare/bind errors in `webserver.log` while teleporter's went back. I can follow that - those are read-only per-request queries and the database open itself still errors into `FTL.log` - but let us write the rule down somewhere so it does not read as arbitrary later.

Two smaller things: `_log_web()` does not honour `print_log`/`print_stdout` outside CLI mode, so `log_ctrl(false, false)` (e.g., `src/api/action.c:49`) no longer silences everything, and it never prints to stdout in non-daemon mode - so `pihole-FTL -f` no longer shows the relocated errors on the console the way `log_err()` did. Routing through `_FTL_log()` as above fixes the second half for free. And `FTLCONF_files_log_ftl=""` is skipped by `getLogFilePathENV()` but accepted by the regular load, which is a small inconsistency.

darkexplosiveqwx:
The syslog priorites trip me up every time, the above was simply a sketch so I didn't verify anything. 
https://github.com/pi-hole/FTL/pull/2958/changes/86044964056f40b4657ca91c4c107b0cf5c831b3 now relays weblog messages if writing failed and `priority <= LOG_WARNING`. #2960's https://github.com/pi-hole/FTL/commit/1b37397b5cb575b83e811d5810fe290f12ebb440 logs an additional line ifthe initial open failed. This is not included here because the current `_log_web()` is mostly stateless and the other PR is meant for the initial open + cached FD.

DL6ER:
`2f695e907` and `014c223f7` cover the routing split, and the fallback in `860449640` is what I had in mind. Two leftovers from the last conflict resolution, then what is still open.

**1. `get_server_ports()` picked up duplicated lines.** The `https_port` assignment appears twice, and the `Web server ports:` header is now printed both before the loop, as on `development`, and again inside it for `i == 0`. The reformatting of the following `log_info()` argument list is unrelated churn as well.

**2. Two `test_suite.bats` assertions were missed.** `2/12` (`listening_ports`) and `11/12` (`proxy_protocol_secret`) still grep `FTL.log` while the other ten moved to `webserver.log`. All twelve come from the same `print_webserver_opts(true, ...)` loop and `test/pihole.toml` sets `debug.webserver = true`, so they land in the web log like the rest. This is the only thing CI fails on (`test_suite.bats:1847`, 191 tests / 1 failure on every architecture); `11/12` is next in line.

Deferring the open-failure notice to #2960 makes sense - that one needs the cached descriptor. The other two points are independent of it, and `_log_web()` on the #2960 branch does not address them either.

**3. `print_log` is ignored.** `_log_web()` returns early only when *both* flags are false, then writes FIFO and file unconditionally, while `_FTL_log()` gates both on `print_log`. No call site trips over it today, so this is consistency rather than a bug - fine by me if it lands in #2960 together with the rest of the write path, I would just like it not to fall between the two PRs.

**4. No stdout outside CLI mode.** This one belongs here, as it only exists because the `log_err()` sites moved. `_FTL_log()` prints when `(!daemonmode || cli_mode) && print_stdout`, so the relocated errors used to show on the console of a `pihole-FTL -f` run and now only do when the fallback fires. Mirroring that condition restores it.

**5. The rule is still only in the commit messages.** A comment above the `log_web`/`log_web_debug` macros would do: web log for anything scoped to a request or to the web server's own operation, `FTL.log` for process health.

**6. Four sites still read the other way to me:** the pipe-creation failure and `gravity failed with signal %d` in `src/api/action.c`; `Failed to open database in networkTable_readDevices()` and the gravity attach/detach errors in `src/api/network.c`, while the other database-open failures went back; the three-way TLS split, where generation failure goes to `FTL.log` but certificate-not-readable and the renewal failure in `webserver_thread()` go to `webserver.log`, although all three mean HTTPS is broken; and the base32 errors in `src/api/2fa.c`, which operate on the *configured* secret rather than on request data.

**7. Minor:** `getLogFilePathENV()` does not check `strdup()`, so on failure we return `true` with `v.s == NULL` and never look at the TOML file. And everything from `log_http_message()` now enters as `LOG_INFO` although that is CivetWeb's error channel - severity is exactly what the web interface is meant to color by.

darkexplosiveqwx:
The review comments should be addressed now.
I made `_log_web()` behave more similarly to `_FTL_log` for print_log, stdout & friends behavior.

The base branch change (which got changed back) was me trying to use `gh stack`, but stacked PRs are not possible across forks (yet).

darkexplosiveqwx:
Rebased on latest `development` and squashed some iterations down.
How I squashed:

`d1155271 Open logfiles in append instead of append+read mode` : dropped, made redundant with #2960
`5eda3141 change webserver.log format to match FTL.log` : new commit: `breaking: make webserver.log more useful & align with FTL.log`
`72e96fe0 enrich webserver and dnsmaq log with priority/severity` : split: squashed web part into `5eda3141` and keep dnsmasq part as own commit
`a4a0700b Avoid NULL dereference when no web log file is configured` : squashed into `5eda3141`
`a382df0b make non-debug log_web() more convenient` : squashed into `5eda3141`
`1af64080 Remove unused process variable and name parameter from init_FTL_log()` : kept as-is
`4b063387 fix: make FTLCONF_files_log_ftl work early` : kept
`093a9427 Reorder some logsites from FTL to webserver log` : squashed into `5eda3141`
`3f955e02 Don't move to webserver.log` : squashed into `5eda3141`
`9bde50ef adjust tests for the move FTL -> webserver.log` : squashed into `5eda3141`
`7e579cb8 Keep process-health error sites in FTL.log` : squashed into `5eda3141`
`c1169158 validate the the filepath in getLogFilePathENV()` : squashed into `4b063387`
`400e11d9 accept empty files_log_ftl in ENV` : squashed into `4b063387`
`86044964 weblog: route errors to FTL.log if webserver.log is unavailible` : squashed into `4b063387`
`014c223f move even more severe messages back to FTL.log` : squashed into `5eda3141`
`1b8fef3c Remove duplicate https_port assignment and duplicated port header in get_server_ports()` : folded in, was branch-internal rebase/merge cleanup 
`11c25459 Fix two missed webserver option log assertions` : squashed into `5eda3141`
`8902a7eb honour print_log in _log_web()` : squashed into `5eda3141`
`3c00390a print to stdout outside CLI mode in _log_web()` : squashed into `5eda3141`
`f7b53686 document the log_web() vs FTL.log routing rule` : squashed into `5eda3141`
`b7dd66b9 Keep process-health log sites in FTL.log` : squashed into `5eda3141`
`45891833 check strdup() in getLogFilePathENV()` : squashed into `4b063387`
`3c2abaa2 log HTTP error channel messages with LOG_ERR` : squashed into `5eda3141`
`75862942 assert webserver option 12/12 sentinel in webserver.log` : squashed into `5eda3141`
`f9014a65 make the final diff more consistant` : folded in, was branch-internal rebase/merge cleanup

New commits:
`aa3e227a enrich dnsmasq log with priority/severity`
`a4042e72 breaking: make webserver.log more useful & align with FTL.log`
`8cfc9d48 Remove unused process variable and name parameter from init_FTL_log()`
`e7684e9d fix: make FTLCONF_files_log_ftl work early`

DL6ER:
The rebase came out clean and the tree builds warning-free here. Six of my seven points are done: the `get_server_ports()` duplication is gone, all twelve option assertions plus the sentinel moved to `webserver.log`, `print_log` is honored, stdout works again, the routing rule is written down above the macros, `strdup()` is checked, and `log_http_message()` enters as `LOG_ERR`. Thanks - that reads well.

Four things left.

**1. The fallback prints to stdout twice.**

`_log_web()` prints to stdout, and then, when the web log cannot be opened, hands the message to `_FTL_log()`, which prints to stdout under the very same condition. On a `pihole-FTL -f` run with an unwritable `files.log.webserver` every relayed error shows up twice on the console.

Deciding whether we relay *before* printing avoids it:

```c
	// Open web log file (if a path is configured)
	FILE *weblog = NULL;
	if(print_log && config.files.log.webserver.v.s != NULL)
		weblog = fopen(config.files.log.webserver.v.s, "a+");

	// _FTL_log() prints these itself, so do not print them twice
	const bool relay = print_log && weblog == NULL && priority <= LOG_WARNING;

	if(!relay && (!daemonmode || cli_mode) && print_stdout)
```

**2. `src/webserver/x509.c` still reads against the rule.**

The new comment says TLS problems belong in `FTL.log`, but this file moved them the other way, and it is inconsistent within one function: `set_name()` and the `fchmod()` warning in `write_to_file()` are reached only from `generate_certificate()`, whose six other failure sites all stayed `log_err()`.

`read_certificate()` and friends are the bigger point. "Could not read certificate file" and "Cannot parse certificate" mean HTTPS is broken, which is the process-health case. They are also what `pihole-FTL --read-x509` prints, and that only stays clean because `args.c` disables file logging right before the call.

I would revert x509.c to `log_err()`/`log_warn()`/`log_info()` throughout.

**3. `DEBUG_GENERIC` inside `enum debug_flag`.**

Putting it after `DEBUG_MAX` means `DEBUG_MAX` is no longer the maximum, while still being the bound `debug_flags[]` is sized with. Nothing indexes the array with it today, so this is not a bug, but it is a trap for whoever adds the next flag.

The only consumer is `FTL_dnsmasq_log()` wanting a plain "DEBUG" out of `priostr()`. Handling `LOG_DEBUG` with `flag == DEBUG_NONE` inside `priostr()` would do the same without touching the enum.

**4. Two nits in `getLogFilePathENV()`.**

`err[]` is filled and then dropped. We cannot log there yet, so that is fine, but a short comment saying so would help - the regular env parse reports the same rejection later anyway.

The `CONF_STRING_ALLOCATED` free and the restore-to-default on the `strdup()` failure path are both unreachable, since `getLogFilePath()` memsets `config` and sets `t = CONF_STRING` immediately before the call.

---

Two more things, neither about the code:

CI has not run on `e52a24c8` yet, only DCO and the conflict check. The bats changes look right to me and `test/pihole.toml` does set `debug.webserver = true`, but nothing has verified them since the rebase.

`prio` is now populated for the dnsmasq and webserver logs where it was null before. `logs.yaml` describes the field as "(if available)", so the spec still holds, but colorization is the whole motivation here - is there a web interface PR to link? The documentation PR field is also still empty, and the `webserver.log` format change is something users will look up.

darkexplosiveqwx:
Thanks for the review. All four code points are addressed, each as a separate commit on top of `e52a24c8` for now, but I will squash them into the appropriate commit from the existing 4 if they look good to you:

**1. The fallback prints to stdout twice.**

Fixed in `c5cbe160`. `_log_web()` now decides whether to relay *before* printing anything: the web log file is opened up front, `relay` is computed as `print_log && weblog == NULL && priority <= LOG_WARNING`, and the stdout print is gated on `!relay`. In the relay case `_FTL_log()` owns the whole message (stdout, FIFO, FTL.log/syslog), so nothing is printed twice. 
I also skip the `FIFO_WEBSERVER` add when relaying, so the line does not show up in both `/api/logs/ftl` and `/api/logs/webserver`. This can be changed if desired.

**2. `src/webserver/x509.c` still reads against the rule.**

Reverted in `d6a90514` to `log_err()`/`log_warn()`/`log_info()` throughout. The diff against `development` for this file is now empty.

**3. `DEBUG_GENERIC` inside `enum debug_flag`.**

Fixed in `90d24c61`. `DEBUG_GENERIC` is gone; `priostr()` maps the generic syslog level to a plain `"DEBUG"` for `flag == DEBUG_NONE`, and `FTL_dnsmasq_log()` passes `DEBUG_NONE`. `DEBUG_MAX` is the maximum again.

**4. Two nits in `getLogFilePathENV()`.**

Fixed in `2bbda9e8`. Added a comment that `err[]` is filled but dropped because logging is not available at that stage and the regular env parse reports the same rejection later. The `CONF_STRING_ALLOCATED` free and the restore-to-default on the `strdup()` failure path are removed - they were unreachable, as you noted.


---

On the non-code points:

- Build & Test CI does not run on my commits automatically, since they are from a fork branch. Full runs need to be triggered by members manually.

- I have an idea for the web PR, but haven't started anything particular yet. One thing that needs consideration is the highlighting in dnsmasq's log: Blocked queries are already colored red and doing the same for errors may create confusion.

- Regarding the docs PR: There was no previous section describing logging, so nothing that needs immediate updating. My plan was to write a new docs page covering all my logging PRs: on-disk format, JSON format, available journald fields.

src/log.c
```diff
@@ -158,6 +171,10 @@ static const char * __attribute__((const)) priostr(const int priority, const enu
 			return "INFO";
 		// debug-level messages
 		case LOG_DEBUG:
+			// The generic syslog level gets a plain "DEBUG", FTL debug flags
+			// get the specific flag name
+			if(flag == DEBUG_NONE)
+				return "DEBUG";
 			return debugstr(flag);
```
DL6ER, line 174: `DEBUG_ANY` is `#define DEBUG_ANY 0` in `log.h` and `DEBUG_NONE` is `0`, so this also catches every `log_debug(DEBUG_ANY, ...)` and turns its `DEBUG_ANY:` prefix into `DEBUG:`. That is roughly 30 call sites across 11 files, `daemon.c` alone has 11. `config.c:64` sets `debug_flags[DEBUG_ANY]` as soon as any debug flag is on, so they do show up in practice, and nothing in the test suite greps for it.

```suggestion
		case LOG_DEBUG:
			return debugstr(flag);
```

src/dnsmasq_interface.c
```diff
@@ -4261,13 +4261,13 @@ static void _query_set_dnssec(queriesData *query, const enum dnssec_status dnsse
...
 // Add dnsmasq log line to internal FIFO buffer (can be queried via the API)
-void FTL_dnsmasq_log(const char *payload, const int length)
+void FTL_dnsmasq_log(const char *payload, const int priority, const int length)
 {
 	// Lock SHM
 	lock_shm();
 
 	// Add to FIFO buffer
-	add_to_fifo_buffer(FIFO_DNSMASQ, payload, NULL, length);
+	add_to_fifo_buffer(FIFO_DNSMASQ, payload, priostr(priority, DEBUG_NONE), length);
```
DL6ER: dnsmasq does need the plain string - it calls `my_syslog(LOG_DEBUG, ...)` in six places - so the special case is better placed here, where it cannot collide with `DEBUG_ANY`.

```suggestion
	// Add to FIFO buffer. dnsmasq has no FTL debug flags, so its LOG_DEBUG is
	// a plain "DEBUG" rather than the DEBUG_ANY catch-all priostr() maps to
	const char *prio = priority == LOG_DEBUG ? "DEBUG" : priostr(priority, DEBUG_NONE);
	add_to_fifo_buffer(FIFO_DNSMASQ, payload, prio, length);
```

src/log.c
```diff
@@ -341,43 +339,77 @@ void __attribute__ ((format (printf, 3, 4))) _FTL_log(const int priority, const
...
+			if(weblog != NULL)
+			{
+				fprintf(weblog, "%s [%s] %s: ", timestring, idstr, prio);
+				va_start(args, format);
+				vfprintf(weblog, format, args);
+				va_end(args);
+				fputc('\n',weblog);
+				fclose(weblog);
+			}
+		}
```
DL6ER, line 402: On your FIFO question: I would keep the add on both paths. The ring lives in shared memory and works even when the file does not, so skipping it empties `/api/logs/webserver` exactly when the web log is broken - which is where someone would go looking. Showing the line in both views is the smaller cost.

Hoisting it also drops the duplicated `buffer`. `relay` implies `weblog == NULL`, so there is still exactly one `fclose()`.

```suggestion
		// Add line to FIFO buffer. It lives in shared memory and is
		// independent of the log file, so it is filled even when we relay
		char buffer[MAX_MSG_FIFO + 1u];
		va_start(args, format);
		const size_t len = vsnprintf(buffer, MAX_MSG_FIFO, format, args) + 1u; /* include zero-terminator */
		va_end(args);
		add_to_fifo_buffer(FIFO_WEBSERVER, buffer, prio, len > MAX_MSG_FIFO ? MAX_MSG_FIFO : len);

		if(relay)
		{
			// No web log available - keep severe messages durable in FTL.log/syslog
			_FTL_log(priority, flag, "%s", buffer);
		}
		else if(weblog != NULL)
		{
			// Write to web log file
			fprintf(weblog, "%s [%s] %s: ", timestring, idstr, prio);
			va_start(args, format);
			vfprintf(weblog, format, args);
			va_end(args);
			fputc('\n',weblog);
			fclose(weblog);
		}
```

DL6ER (commented):
`90d24c61e` needs one more pass - see the two suggestions below. Dropping the enum member is right, the replacement just happens to land on `DEBUG_ANY` too.

Yes to squashing all four into the existing commits once that is settled. Please do not bother preserving authorship on the suggestions, just absorb them.

Build & Test is approved from my side, but the runs simply do not start - most likely the Actions trouble GitHub has been having all week. Your next push should pick it up.

DL6ER (approved):
All four points are addressed and the suggestions landed cleanly. Build \& Test finally ran this time, green across the whole matrix, so the moved `test_suite.bats` assertions are actually verified rather than just plausible.

The two CodeQL path-injection alerts are dismissed as false positives: `FTLCONF_files_log_ftl` is operator input, and the same flow already exists on `development`.

PR #2960, opened 2026-07-16
breaking: utilize a cached file descriptor for logging & own dnsmasq.log

darkexplosiveqwx:
## Breaking changes

**`files.log.dnsmasq = "-"` is no longer supported.** The `"-"` value used to select stderr logging via dnsmasq's `log-facility`. Since FTL writes `pihole.log` itself, this value is retired: when it is encountered, FTL logs a hint and migrates existing `-` values to the default path (`/var/log/pihole/pihole.log`) instead of silently creating a file named `-`.

**Custom `log-facility` settings in dnsmasq configs no longer take effect.** Since FTL owns `pihole.log` and dnsmasq's own file-write path is bypassed, neither `log-facility=<path>` nor `log-facility=-` in a custom dnsmasq config (e.g. `dnsmasq.d`) is honored anymore - every dnsmasq line goes to `files.log.dnsmasq` regardless of what the option points at, including stderr. This is also why `files.log.dnsmasq = "-"` (which generated `log-facility=-`) is retired: the stderr channel it selected no longer exists in the daemon path.

**dnsmasq's `log-async` option is dropped from the generated config.** The option drove dnsmasq's asynchronous writer queue, which this PR bypasses entirely, so it no longer had any effect. The trade-off: every query-log line is now a blocking `write()` on the DNS thread. A dedicated asynchronous writer for FTL is planned in the future.

**FTL owns `pihole.log` and keeps descriptors open, so rotation now needs `SIGUSR2`.** This also applies to `FTL.log` and `webserver.log`, which previously re-opened per line and needed no postrotate. The companion [pi-hole/pi-hole#6672](https://github.com/pi-hole/pi-hole/pull/6672) adds `kill -USR2` to the `FTL.log` and `webserver.log` logrotate stanzas.

FTL writes `FTL.log` and `webserver.log` today by opening, writing and closing the file for every line. This PR replaces that with a cached file descriptor per log file and extends the same mechanism to `pihole.log`, which from now on is written by FTL itself instead of by dnsmasq.

dnsmasq's own stderr/file/syslog write paths are bypassed: every dnsmasq line is routed through FTL and appended to `pihole.log` via the cached descriptor. This gives all three logs a consistent on-disk format and makes FTL the single owner of all on-disk logging.

The mechanism is the design discussed in https://github.com/pi-hole/FTL/pull/2897, and this PR builds on top of https://github.com/pi-hole/FTL/pull/2958.

- **Cached descriptors:** a `struct log_fd` per log file caches the path, descriptor, reopen flag and a per-file mutex. Writes are plain `O_APPEND` `write()`s (atomic per line); the mutex only guards the `SIGUSR2` reopen.
- **Fork safety:** `pthread_atfork()` handlers take and release all three per-file mutexes around `fork()`, registered before any threads or dnsmasq forks exist. A TCP-query fork can never inherit a locked log mutex and hang the first log write there.
- **`SIGUSR2` handling:** FTL registers its own `SIGUSR2` handler (replacing dnsmasq's) that marks all three logs for reopen. `write_log_line()` tests the reopen *before* the `fd == -1` fast path, so a failed initial open (missing directory, transient `EACCES`, rotation timing) is recovered instead of disabling that log for the life of the process. `SIGUSR2` in the TCP-query forks is ignored.
- **dnsmasq routing:** `my_syslog()` logs through `FTL_dnsmasq_log()`/`FTL_write_dnsmasq_log()`; dnsmasq's own file and syslog fallback paths are bypassed. `echo_stderr` is kept so `dnsmasq --test` errors still reach stderr (used by config testing).
- **Fallbacks:** a failed `pihole.log` open is reported at startup (the message adapts to `misc.hide_dnsmasq_warn`); while `pihole.log` is unavailable, dnsmasq warnings and errors fall back to syslog so they are not silently dropped for the lifetime of the process. `webserver.log` messages keep their existing relay to `_FTL_log()` for `<= LOG_WARNING` when no descriptor is available.
- **Privilege handling:** `FTL_fork_and_bind_sockets()` now chowns `webserver.log` and `pihole.log` alongside `FTL.log` when dropping from root, so freshly created files are `pihole`-owned.
- **Flush:** `pihole.log` is truncated through the cached descriptor on the flush timer instead of the `fopen("w")` + close + reopen dance.
- **Timestamps:** the dnsmasq timestamp buffer uses `ctime_r()` (locale-independent, thread-safe) so `pihole.log` lines stay byte-identical to what we wrote before, in any locale.

- Start/restart FTL, enable query logging, and confirm replies land in `pihole.log` with the same format as before.
- `logrotate --force` the Pi-hole log files (or send `kill -USR2` manually) and confirm FTL reopens the fresh files instead of writing into the rotated ones.
- Set `files.log.dnsmasq` to `"-"` via `pihole-FTL --config files.log.dnsmasq "-"` and verify that it is rejected.
- Run `pihole-FTL --config` / `--teleporter` etc. as a non-root user and confirm no log files are created.

## Link documentation PRs if any are needed to support this PR:

A docs PR covering the many logging changes is planned but not yet opened.

---

---

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -28,10 +28,87 @@
 #include "database/query-table.h"
 // runGC()
 #include "gc.h"
+// open(), O_WRONLY, O_CREAT, O_APPEND, O_CLOEXEC
+#include <fcntl.h>
 
 static bool print_log = true, print_stdout = true;
-static bool ftl_log_available = true;
 bool debug_flags[DEBUG_MAX] = { false };
+
+// Per-file log state: fd, path, writer-preferenced lock, reopen flag
+struct log_fd {
+	int fd;
+	const char *path;
+	pthread_mutex_t lock;
+	volatile sig_atomic_t reopen_needed;
+};
+
+static struct log_fd ftl_log = { .fd = -1, .lock = PTHREAD_MUTEX_INITIALIZER };
+static struct log_fd webserver_log = { .fd = -1, .lock = PTHREAD_MUTEX_INITIALIZER };
+static struct log_fd dnsmasq_log = { .fd = -1, .lock = PTHREAD_MUTEX_INITIALIZER };
+
+// dnsmasq forks per TCP query while FTL threads may be mid-write; without
+// atfork handling the child would inherit a locked log mutex and hang on the
+// first my_syslog() there.  Lock/unlock all log mutexes around fork().
+static void log_atfork_prepare(void)
+{
+	pthread_mutex_lock(&ftl_log.lock);
+	pthread_mutex_lock(&webserver_log.lock);
+	pthread_mutex_lock(&dnsmasq_log.lock);
+}
+static void log_atfork_parent(void)
+{
+	pthread_mutex_unlock(&ftl_log.lock);
+	pthread_mutex_unlock(&webserver_log.lock);
+	pthread_mutex_unlock(&dnsmasq_log.lock);
+}
+static void log_atfork_child(void)
+{
+	pthread_mutex_unlock(&ftl_log.lock);
+	pthread_mutex_unlock(&webserver_log.lock);
+	pthread_mutex_unlock(&dnsmasq_log.lock);
+}
+
+// Return 1 if this fd is associated with any logfile to avoid
+// dnsmasq closing it during initialization
+int __attribute__((pure)) is_log_fd(const int fd)
+{
+	return fd == ftl_log.fd || fd == webserver_log.fd || fd == dnsmasq_log.fd;
+}
+
+// Writer-preferenced per-file lock: only the fd for this specific log is
+// held, so writes to different files never contend.  The reopen flag is
+// per-file so SIGUSR2 only touches the fd that actually needs it.
+static bool write_log_line(struct log_fd *log, const char *line, size_t len)
+{
+	if(log->fd == -1)
+		return false;
+
+	pthread_mutex_lock(&log->lock);
+
+	if(log->reopen_needed)
+	{
+		log->reopen_needed = 0;
+		close(log->fd);
+		log->fd = open(log->path, O_WRONLY|O_CREAT|O_APPEND|O_CLOEXEC, S_IRUSR|S_IWUSR|S_IRGRP);
+	}
+
+	ssize_t written = 0;
+	while(written < (ssize_t)len)
+	{
+		ssize_t rc = write(log->fd, line + written, len - written);
+		if(rc == -1)
+		{
+			if(errno == EINTR)
+				continue;
+			pthread_mutex_unlock(&log->lock);
+			return false;
+		}
+		written += rc;
+	}
+
+	pthread_mutex_unlock(&log->lock);
+	return true;
+}
 
 void clear_debug_flags(void)
 {
@@ -45,26 +122,61 @@ void log_ctrl(bool plog, bool pstdout)
 	print_stdout = pstdout;
 }
 
-void init_FTL_log(void)
+// Open cached log fds from config paths.
+// open_log_fds(true):  open FTL.log only (called early, before full config)
+// open_log_fds(false): open webserver.log + pihole.log (called after config)
+void open_log_fds(bool ftl)
 {
-	// Open the log file in append/create mode
-	if(config.files.log.ftl.v.s != NULL)
+	if(ftl)
 	{
-		FILE *logfile = NULL;
-		if((logfile = fopen(config.files.log.ftl.v.s, "a+")) == NULL)
+		// FTL.log - path is known from getLogFilePath()
+		if(config.files.log.ftl.v.s != NULL)
 		{
-			printf("ERROR: Opening of FTL log (%s) failed: %s\nUsing syslog instead!\n",
-			       config.files.log.ftl.v.s, strerror(errno));
-			syslog(LOG_ERR, "Opening of FTL\'s log file failed, using syslog instead!");
-			ftl_log_available = false;
+			ftl_log.path = config.files.log.ftl.v.s;
+			ftl_log.fd = open(ftl_log.path, O_WRONLY|O_CREAT|O_APPEND|O_CLOEXEC, S_IRUSR|S_IWUSR|S_IRGRP);
+			if(ftl_log.fd == -1)
+			{
+				printf("ERROR: Opening of FTL log (%s) failed: %s\nUsing syslog instead!\n",
+				       ftl_log.path, strerror(errno));
+				syslog(LOG_ERR, "Opening of FTL\'s log file failed, using syslog instead!");
+			}
 		}
-		else
-			ftl_log_available = true;
-
-		// Close log file
-		if(logfile != NULL)
-			fclose(logfile);
+		return;
 	}
+
+	// webserver.log + pihole.log - paths are known after readFTLconf()
+	if(config.files.log.webserver.v.s != NULL)
+	{
+		webserver_log.path = config.files.log.webserver.v.s;
+		webserver_log.fd = open(webserver_log.path, O_WRONLY|O_CREAT|O_APPEND|O_CLOEXEC, S_IRUSR|S_IWUSR|S_IRGRP);
+	}
+
+	// pihole.log (dnsmasq) - FTL owns this file from now on
+	if(config.files.log.dnsmasq.v.s != NULL)
+	{
+		dnsmasq_log.path = config.files.log.dnsmasq.v.s;
+		dnsmasq_log.fd = open(dnsmasq_log.path, O_WRONLY|O_CREAT|O_APPEND|O_CLOEXEC, S_IRUSR|S_IWUSR|S_IRGRP);
+	}
+
+	// Register atfork handlers once, before any threads or dnsmasq forks
+	// exist, so a TCP-query fork can never inherit a locked log mutex.
+	// Invariant: fork() is never called from inside a log write, so these
+	// handlers only need to cover the case where a thread holds a log mutex
+	// at the moment of the fork.
+	static bool atfork_registered = false;
+	if(!atfork_registered)
+	{
+		atfork_registered = true;
+		pthread_atfork(log_atfork_prepare, log_atfork_parent, log_atfork_child);
+	}
+}
+
+// Signal that log fds need to be reopened (called from SIGUSR2 handler path)
+void mark_log_reopen(void)
+{
+	ftl_log.reopen_needed = 1;
+	webserver_log.reopen_needed = 1;
+	dnsmasq_log.reopen_needed = 1;
 }
 
 // Return time(NULL) but with (up to) nanosecond accuracy
@@ -254,6 +366,36 @@ const char *debugstr(const enum debug_flag flag)
 	}
 }
 
+// Write a dnsmasq log line to pihole.log in dnsmasq's exact on-disk format.
+// The message is the bare body (no timestamp, no prefix) as handed to
+// FTL_dnsmasq_log() from my_syslog().  We reproduce dnsmasq's format:
+//   "Mon Jan  1 12:00:00 2024 dnsmasq-dhcp[12345]: <message>\n"
+// where the func suffix (e.g. "-dhcp", "-tftp") comes from the priority
+// bits extracted in my_syslog().
+void FTL_write_dnsmasq_log(const char *message, const char *func)
+{
+	if(dnsmasq_log.fd == -1)
+		return;
+
+	time_t now = time(NULL);
+	char *ts = ctime(&now);
+
+	char line[2048];
+	int off = snprintf(line, sizeof(line), "%.20s dnsmasq%s[%d]: ", ts + 4, func ? func : "", getpid());
+
+	const char *msg = message ? message : "";
+	off += snprintf(line + off, sizeof(line) - off, "%s", msg);
+
+	// Clamp to buffer end - snprintf returns would-be length on truncation
+	if(off >= (int)sizeof(line))
+		off = sizeof(line) - 1;
+
+	if(off > 0 && line[off - 1] != '\n')
+		line[off++] = '\n';
+
+	write_log_line(&dnsmasq_log, line, off);
+}
+
 void __attribute__ ((format (printf, 3, 4))) _FTL_log(const int priority, const enum debug_flag flag, const char *format, ...)
 {
 	char timestring[TIMESTR_SIZE];
@@ -294,40 +436,25 @@ void __attribute__ ((format (printf, 3, 4))) _FTL_log(const int priority, const 
 		va_end(args);
 		add_to_fifo_buffer(FIFO_FTL, buffer, prio, len > MAX_MSG_FIFO ? MAX_MSG_FIFO : len);
 
-		bool logged = false;
-		if(ftl_log_available && config.files.log.ftl.v.s != NULL)
+		// Format full line and write to cached fd.  Bounded to one 2048-byte
+		// write per line so O_APPEND guarantees a line never interleaves with
+		// another thread's - lines past the cap are truncated (a deliberate
+		// trade-off of the cached-fd design vs. the old unbounded vfprintf()).
+		char line[2048];
+		int off = snprintf(line, sizeof(line), "%s [%s] %s: ", timestring, idstr, prio);
+		va_start(args, format);
+		off += vsnprintf(line + off, sizeof(line) - off, format, args);
+		va_end(args);
+
+		// Clamp to buffer end - snprintf returns would-be length on truncation
+		if(off >= (int)sizeof(line))
+			off = sizeof(line) - 1;
+
+		line[off++] = '\n';
+
+		if(!write_log_line(&ftl_log, line, off))
 		{
-			// Open log file
-			FILE *logfile = fopen(config.files.log.ftl.v.s, "a+");
-
-			// Write to log file
-			if(logfile != NULL)
-			{
-				// Prepend message with identification string and priority
-				fprintf(logfile, "%s [%s] %s: ", timestring, idstr, prio);
-
-				// Log message
-				va_start(args, format);
-				vfprintf(logfile, format, args);
-				va_end(args);
-
-				// Append newline character to the end of the file
-				fputc('\n', logfile);
-
-				// Close file after writing
-				fclose(logfile);
-
-				logged = true;
-			}
-			else if(!daemonmode)
-			{
-				printf("!!! WARNING: Writing to FTL\'s log file failed!\n");
-				syslog(LOG_ERR, "Writing to FTL\'s log file failed!");
-			}
-		}
-		if(!logged)
-		{
-			// Syslog logging
+			// Fallback: syslog if fd is unavailable or write failed
 			va_start(args, format);
 			vsyslog(priority, format, args);
 			va_end(args);
@@ -354,16 +481,8 @@ void __attribute__ ((format (printf, 3, 4))) _log_web(const int priority, const 
 	get_idstr(idstr, sizeof(idstr));
 	const char *prio = priostr(priority, flag);
 
-	// Open web log file (if a path is configured)
-	FILE *weblog = NULL;
-	if(print_log && config.files.log.webserver.v.s != NULL)
-		weblog = fopen(config.files.log.webserver.v.s, "a+");
-
-	// _FTL_log() prints these itself, so do not print them twice
-	const bool relay = print_log && weblog == NULL && priority <= LOG_WARNING;
-
 	// Print to stdout before writing to file
-	if(!relay && (!daemonmode || cli_mode) && print_stdout)
+	if((!daemonmode || cli_mode) && print_stdout)
 	{
 		// Only print time/ID string when not in direct user interaction (CLI mode)
 		if(!cli_mode)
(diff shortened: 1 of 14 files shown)


DL6ER:
Had a look at this one while reviewing #2958, even though it is still a draft. The shape of the cached-fd rework is fine, but a few things would bite in production.

**1. `FTL.log` and `webserver.log` stop being written after the first rotation.** Our shipped `advanced/Templates/logrotate` rotates both with `create 640 pihole pihole` and no postrotate, with the comment "FTL.log and webserver.log are opened and closed for each line, therefore no postrotate is needed". With a cached descriptor we keep writing into the renamed `FTL.log.1` - which logrotate then compresses while we still hold it open - and the freshly created file stays empty until FTL restarts. This needs a companion pi-hole PR adding `kill -USR2` to both stanzas, and the two have to land together. `pihole.log` already has that postrotate, and `create 640` matches the new `S_IRUSR|S_IWUSR|S_IRGRP` mode, so that part lines up.

**2. `write_log_line()` can never recover a log whose initial open failed.** The `if(log->fd == -1) return false;` fast path sits before the `reopen_needed` check, so `SIGUSR2` cannot revive it. A missing `/var/log/pihole` at first start, or a transient `EACCES`, disables that log for the lifetime of the process. Testing the reopen flag first fixes it.

**3. The dnsmasq timestamp buffer is too small for non-English locales.** `char ts_buf[16]` fits `"Jan  1 12:00:00"` exactly in the C locale, but `init_locale()` calls `setlocale(LC_ALL, "")`, so `%b` is the localized abbreviation - six bytes for `ru_RU`, more elsewhere. When it does not fit, `strftime()` returns 0 and leaves the buffer contents *unspecified*, so `pihole.log` gets garbage. dnsmasq itself sidesteps this with `ctime(&time_now) + 4` and `%.15s`, which is locale-independent and keeps the on-disk format byte-identical to what we write today.

**4. dnsmasq messages have no fallback left.** `my_syslog()` returns before dnsmasq's own syslog path and `FTL_write_dnsmasq_log()` returns silently when the descriptor is -1, so if `pihole.log` cannot be opened every dnsmasq message is lost. `_FTL_log()` and `_log_web()` both fall back, and `open_log_fds()` warns for `webserver.log` but says nothing for `pihole.log`.

**5. The per-file mutex is not fork-safe.** dnsmasq forks per TCP query while our threads are running. If a fork happens while another thread holds `dnsmasq_log.lock`, the child inherits it locked and the first `my_syslog()` there blocks forever, hanging that query. Our SHM lock is process-shared and robust for exactly this reason. Since every write is `O_APPEND`, and a single `write()` to a regular file is atomic, the lock only guards the reopen - `pthread_atfork()` handlers or an atomic descriptor swap would be enough.

**6. New root-owned log files.** `open_log_fds(false)` runs before we drop privileges, so a fresh install now gets root-owned `webserver.log` and `pihole.log`; dnsmasq used to create `pihole.log` after dropping to `pihole`, and `webserver.log` was created on the first request. `FTL_fork_and_bind_sockets()` chowns only `files.log.ftl` and should cover the other two.

**7. `log-async` is now dead code in the generated config.** With `my_syslog()` bypassed, dnsmasq's async queue is never used, so query logging became a blocking `write()` in the DNS thread - which is the stall that queue exists to prevent. Either drop the `log-async` lines or say why the synchronous write is acceptable for us.

**Smaller:** removing `validate_filepath_dash` silently retires `files.log.dnsmasq = "-"`, which is a user-visible value and belongs on the breaking-change list; five of the new comments contain em dashes while the rest of the tree is ASCII; `log->fd` is read outside the lock as a fast path while `flush_dnsmasq_log()` reassigns it under the lock; and `off` is used as an offset into `line` before it is clamped.

darkexplosiveqwx:
> 1. FTL.log and webserver.log stop being written after the first rotation. Our shipped advanced/Templates/logrotate rotates both with create 640 pihole pihole and no postrotate, with the comment "FTL.log and webserver.log are opened and closed for each line, therefore no postrotate is needed". With a cached descriptor we keep writing into the renamed FTL.log.1 - which logrotate then compresses while we still hold it open - and the freshly created file stays empty until FTL restarts. This needs a companion pi-hole PR adding kill -USR2 to both stanzas, and the two have to land together. pihole.log already has that postrotate, and create 640 matches the new S_IRUSR|S_IWUSR|S_IRGRP mode, so that part lines up.


Will take a look regarding the code later, but a Core PR already exists. It did get linked, but after that gotten mostly buried.
Here is an explicit mention again: https://github.com/pi-hole/pi-hole/pull/6672

DL6ER:
**1. `struct log_fd.path` outlives the string it points at.** The path is cached once in `open_log_fds()` as a pointer straight into `config.files.log.*.v.s` and dereferenced much later, in the reopen inside `write_log_line()` and in `flush_dnsmasq_log()`. `files.log.ftl` becomes `CONF_STRING_ALLOCATED` whenever it comes from ENV, TOML or the legacy reader, and it carries `FLAG_FTL_LOG`, not `FLAG_RESTART_FTL`. `free_config()` only spares it while terminating:

```c
case CONF_STRING_ALLOCATED:
    if(terminating && conf_item->f & FLAG_FTL_LOG)
        continue;
    free(conf_item->v.s);
```

and the config replacement path calls it with `terminating == false`. So changing `files.log.ftl` at runtime frees the string, and the next `SIGUSR2` reopens a freed pointer. Even setting that aside, we would keep writing to the old file, whereas today `_FTL_log()` reopens the config value per line and picks the change up at once. Re-reading the config value at reopen time, or `strdup()`ing into `struct log_fd` and swapping on change, both solve it.

**2. `ctime()` is not thread-safe.** It returns a pointer to a static buffer shared with `localtime()` and `asctime()`. dnsmasq can use that idiom because it is single-threaded; `FTL_write_dnsmasq_log()` runs on the DNS thread while the webserver, database and NTP threads format their own timestamps, so concurrent callers tear the string. `ctime()` can also return NULL, and `+ 4` on that is undefined. `ctime_r()` with a 26-byte buffer gives the same output and the same locale independence.

**3. dnsmasq messages still have no fallback.** The new open-time warning is good, but it is a one-time notice: `FTL_write_dnsmasq_log()` discards `write_log_line()`'s return value, so with no descriptor every dnsmasq line is dropped for the life of the process. `_FTL_log()` falls back to `vsyslog()` and `_log_web()` to `_FTL_log()` for `priority <= LOG_WARNING`; dnsmasq lines deserve at least the same for warnings and errors.

**4. The early-startup window is still uncovered.** Moving `open_log_fds(false)` to just after `readFTLconf()` covers dnsmasq's runtime, but `write_dnsmasq_config()`, and with it `test_dnsmasq_config()`, is called from *inside* `readFTLconf()` (`config.c:1913`, `1943`, `1976`). Anything emitted there hits a descriptor that is still -1 and, per point 3, disappears. That is exactly what `log-facility` used to catch. Either open `pihole.log` before the first `write_dnsmasq_config()`, or give the writer a fallback.

**5. `flush_dnsmasq_log()` does not check its reopen.** If that `open()` fails, `pihole.log` is dead until the next `SIGUSR2` with nothing in any log explaining why. An `ftruncate()` under the lock would also be simpler than the `fopen("w")` plus close plus reopen dance.

Smaller: `validate_filepath_dash` is still removed outright, which retires `files.log.dnsmasq = "-"` and belongs on the breaking-change list; the `off` clamps in all three writers only guard the upper bound, while `snprintf()` returns a negative value on an encoding error; `is_log_fd()` keeps `__attribute__((pure))` although the descriptors it reads are reassigned on reopen from another thread; and the atfork comment should state the invariant it relies on, namely that we never fork from inside a log write. Leaving dnsmasq's now-dead write path in place after the early `return` is the right call for future merges.

Please rebase onto `development`. #2958 was squash-merged, so this branch still carries its original commits and that is what conflicts; a tree diff also shows it reverting newer work such as the `FTL_parse_pseudoheaders()` signature. pi-hole/pi-hole#6672 needs to land in the same release as this.

DL6ER:
Careful work, and the `pthread_atfork` handling in particular is right - a TCP fork inheriting a locked log mutex is a real hazard and you covered it. A few things before this can go in.

Why is `log-async` dropped? Every dnsmasq log line now takes a blocking `write()` on the DNS thread. dnsmasq has the async queue precisely so a slow or full disk cannot stall the resolver, which on an SD-card install with query logging is the case I worry about. If it is genuinely no longer needed, say why in the description - I would rather not lose it as a side effect.

Removing `validate_filepath_dash` means `files.log.dnsmasq = "-"` stops validating, so anyone using it has their value rejected on upgrade with nothing explaining why. There should at least be a hint in the log when the old value is encountered. The description is still the unfilled template, so this is not called out as breaking anywhere but the commit subject - please write it up there.

`_log_web()` lost the `relay` guard, so when `webserver.log` is unavailable a `<= LOG_WARNING` message prints to stdout once from `_log_web()` and again from the `_FTL_log()` relay. I read that as an unintended side effect rather than intent.

`readFTLconf()` calling `open_log_fds(false)` needs addressing too: `readFTLconf` runs on every CLI path, so `pihole-FTL --config ...` now opens both log files with `O_CREAT`. As a normal user that warns on every invocation, as root it can leave the files root-owned.

One question: the per-file mutex only serializes threads within one process, but dnsmasq forks per TCP query and the children inherit the descriptors. Do you see interleaving in practice, or is `O_APPEND` enough here?

darkexplosiveqwx:
> Removing `validate_filepath_dash` means `files.log.dnsmasq = "-"` stops validating, so anyone using it has their value rejected on upgrade with nothing explaining why. There should at least be a hint in the log when the old value is encountered. The description is still the unfilled template, so this is not called out as breaking anywhere but the commit subject - please write it up there.

> `_log_web()` lost the `relay` guard, so when `webserver.log` is unavailable a `<= LOG_WARNING` message prints to stdout once from `_log_web()` and again from the `_FTL_log()` relay. I read that as an unintended side effect rather than intent.

> `readFTLconf()` calling `open_log_fds(false)` needs addressing too: `readFTLconf` runs on every CLI path, so `pihole-FTL --config ...` now opens both log files with `O_CREAT`. As a normal user that warns on every invocation, as root it can leave the files root-owned.

Adressed in three new commits. If it looks good, I will trim some comments, squash down and rebase on development for good measure.


> One question: the per-file mutex only serializes threads within one process, but dnsmasq forks per TCP query and the children inherit the descriptors. Do you see interleaving in practice, or is `O_APPEND` enough here?

The per-file mutex is only needed to avoid a race conditioning when re-opening the file after `USR2`. individual `write()` calls with `O_APPEND` do not interleave, with the exception being certain network file systems.

To quote [`man write()`](https://man7.org/linux/man-pages/man3/write.3p.html) and [`man open()`](https://man7.org/linux/man-pages/man2/open.2.html)
>       If the O_APPEND flag of the file status flags is set, the file
>       offset shall be set to the end of the file prior to each write and
>       no intervening file modification operation shall occur between
>       changing the file offset and the write operation.


>  O_APPEND
> 
>       The file is opened in append mode.  Before each write(2),
>       the file offset is positioned at the end of the file, as if
>       with lseek(2).  The modification of the file offset and the
>       write operation are performed as a single atomic step.
>
>       O_APPEND may lead to corrupted files on NFS filesystems if
>       more than one process appends data to a file at once.  This
>       is because NFS does not support appending to a file, so the
>       client kernel has to simulate it, which can't be done
>       without a race condition.



> Why is `log-async` dropped? Every dnsmasq log line now takes a blocking `write()` on the DNS thread. dnsmasq has the async queue precisely so a slow or full disk cannot stall the resolver, which on an SD-card install with query logging is the case I worry about. If it is genuinely no longer needed, say why in the description - I would rather not lose it as a side effect.

Since we bypass dnsmasq's writer, the `log-async` option no longer has any effect. You yourself asked to remove it in https://github.com/pi-hole/FTL/pull/2960#issuecomment-5162197148. I agree that is it bad to do potentially blocking operations on the DNS thread. I am currently experimenting with an asynchronous writer for FTL using a Vyukov bounded MPMC, but it will definitely be out of scope and not ready for this PR. This also avoids all the locking gymnastics for the file re-open, since a singular thread will be soley responsible for writing the file.


I additionally updated the PR description (finnally!) to inform users eventually stumbling onto this breaking change what is going on.

DL6ER:
Nothing blocking from me on the code. Three things from your side:

1. Squash and rebase as you offered on 1 September - then I approve and merge this.
2. The installer migration in https://github.com/pi-hole/pi-hole/pull/6672. Detail is in my comment there; without it that PR only reaches fresh installs, and everyone who updated to v6.4.3 keeps a logrotate file with no `postrotate` for `FTL.log` and `webserver.log`.
3. A docs PR. `docs/ftldns/signals.md` still says `SIGUSR2` reopens the file from `files.log.dnsmasq` - it is ours now and reopens all three. The three breaking changes want writing up there too.

Only 1 gates the merge here. 2 and 3 are for v7.0, which is a way off yet, so no rush on either.

I will assign myself so `stale_pr.yml` cannot close this the way it closed #2897.

darkexplosiveqwx:
Squashed down and rebased on `development`.

DL6ER (commented):
**1. `files.log.ftl` set at runtime never reaches the writer.** `log->path` is only assigned in `open_log_fds()`, which runs at startup, so the reload logs the change while FTL keeps appending to the old file and `SIGUSR2` reopens the cached path. The other two already carry `FLAG_RESTART_FTL`; in `getLogFilePath()`:

```c
config.files.log.ftl.f = FLAG_FTL_LOG | FLAG_RESTART_FTL;
```

I built that and it restarts FTL as they do.

**2. The crash handler can now block, so the atfork invariant does not hold.** `write_log_line()` takes a process-wide mutex (`log.c:89`), and the crash handler emits dozens of `log_info()` calls before forking for addr2line (`signals.c:651`) - that fork runs `log_atfork_prepare()` and takes all three log mutexes. A thread parked in `write()` on a stalled disk hangs us before the backtrace, and a SIGSEGV taken inside `write_log_line()` deadlocks on the handler's first `log_info()`. `signals.c:1251` keeps the SIGTERM handler free of `log_info()` for this reason; #3029 is the same class.

**3. The early `open_log_fds(false)` uses the default paths.** `config.c:1926` runs before `readFTLtoml()`, so with `files.log.dnsmasq` pointing elsewhere we `O_CREAT` an empty `/var/log/pihole/pihole.log` on every start - never chowned, since `FTL_fork_and_bind_sockets()` uses the configured path - and the dnsmasq `--test` output lands there too. Calling it after the TOML read but before `write_dnsmasq_config()` covers the same window with the right paths.

**4. The `"-"` fallback only redirects the writes.** `config.files.log.dnsmasq.v.s` keeps `"-"`, so `writeFTLtoml()` writes it back and the value never heals, `api/logs.c:86` reports `"file": "-"`, and `chown_pihole("-")` (`dnsmasq_interface.c:3752`) fails. Rewriting the value to the default fixes all three, and it is what actually retires `"-"` - which #3075 needs.

**5. `flush_dnsmasq_log()` ignores `reopen_needed`.** The `ftruncate()` at `log.c:1021` hits whatever the cached fd points at, so a flush between logrotate's `SIGUSR2` and the next write empties the rotated `pihole.log.1` and reports success.

**6. Two warnings point at the wrong place.** `log.c:212` promises dnsmasq warnings are "relayed to the FTL log", but the fallback (`dnsmasq_interface.c:4267`) is `syslog()`. The `hide_dnsmasq_warn` branch is wrong too - it only selects the query in `count_messages()` (`message-table.c:1042`). The `webserver.log` message is accurate.

**Smaller:** `_FTL_log()`/`_log_web()` truncate at 2048 bytes where `vfprintf()` was unbounded; a failed `write()` to `webserver.log` on a valid fd is dropped, while `_FTL_log()` falls back to `vsyslog()`. And `log: make new comments more concise` reworks comments from `c186a794`, `4b54100a` and `9f34029b`, so it splits rather than folds into the first - it also drops the invariant from point 2 and leaves three blank lines at `log.c:421`.

pi-hole/pi-hole#6672 is merged, so only the docs PR is left from my 6 September list. #3079 test-merges onto this branch cleanly, so the order between them is free.

darkexplosiveqwx:
I amended your suggestions directly into the existing commits.

Regarding 2.: I couldn't find a solution that’s reasonable here. `_log_FTL()` was already very much async-signal-unsafe before, and the pthread mutex just makes it more visible. Creating a new logging API that just skips the mutex will change the race of a deadlock to a race of writing a wrong fd or cause EBADF. I already dislike the mutex + atfork machinery and luckily can be avoided with my asynchronous logger. I am intentionally deferring this, as it would otherwise really clutter this PR again. I hope that I am not over-selling my design here by ignoring flaws, but it does seem to solve most of our problems. The producer will only collect data (PID, TID, timestamp), format the message (via *printf, but for async-signal-safety we can use fixed strings or hand-roll a minimal formatter) and then send off the message into the Vyukov ring. The singular logger thread then takes care of writing to the sinks and handling the reopen. No need for any locks or atfork handling.

If you can think of a solution that fits well into the scope of this PR then I am welcome to accept it.

I am still thinking of whether to do the asynchronous logger directly after this PR or whether to put it at the end of my existing logging stack (JSON + journal).

DL6ER (commented):
On point no. 2: The `malloc` arena lock predates your PR, `fopen()`/`vfprintf()` take it, and `write_log_line()` is closer to async-signal-safe than what the code was before. However, I think `pthread_atfork()` adds a new issue that doesn't exist in `development` right now: The crash handler logs its way through the backtrace and then forks for calling `addr2line`. The new `log_atfork_prepare()` now takes all three mutexes; a thread parked in `write()` on `webserver.log` or `pihole.log` hangs us before the trace. In non-debug mode, the latter is still a rather high-volume output, so the likeliness to hit here is not vanishing IMO.
The forked tracing child only does `dup2()`, `open()`, `close()` and `execvp()`, so it never needs the handlers. `posix_spawn()` with `posix_spawn_file_actions_adddup2()`/`addopen()` expresses the same thing and skips `atfork` on both `glibc` and `musl` (see below). I put my suggested diff inline.

Where the async logger sits in your stack is your call. We are kind of overdue to merge this one here IMO, so I am fine with doing it as follow-up.

My PoC: `./poc` hangs, `SPAWN=1 ./poc` prints `OKAY`, on `glibc` and `musl` alike:

```c
#define _GNU_SOURCE
#include <pthread.h>
#include <spawn.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/wait.h>

static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
static void prepare(void) { pthread_mutex_lock(&lock); }
static void release(void) { pthread_mutex_unlock(&lock); }
static void *hold(void *a) { (void)a; pthread_mutex_lock(&lock); pause(); return NULL; }

int main(void)
{
	pthread_t th;
	pid_t pid;
	char *av[] = { "true", NULL };

	pthread_atfork(prepare, release, release);
	pthread_create(&th, NULL, hold, NULL);
	sleep(1);

	if(getenv("SPAWN"))
		posix_spawnp(&pid, "true", NULL, NULL, av, environ);
	else if((pid = fork()) == 0)
		_exit(0);

	waitpid(pid, NULL, 0);
	puts("OKAY");
	return 0;
}
```

src/log.c
```diff
@@ -28,11 +28,114 @@
...
+	pthread_mutex_unlock(&ftl_log.lock);
+	pthread_mutex_unlock(&webserver_log.lock);
+	pthread_mutex_unlock(&dnsmasq_log.lock);
+}
+static void log_atfork_child(void)
+{
+	pthread_mutex_unlock(&ftl_log.lock);
+	pthread_mutex_unlock(&webserver_log.lock);
+	pthread_mutex_unlock(&dnsmasq_log.lock);
+}
```
DL6ER, line 69: These also fire on the crash path, and it hangs there. `handle_SIGSEGV()` logs its way through the backtrace and then `run_addr2line_object()` calls `fork()` (`signals.c:651`), so `log_atfork_prepare()` takes all three mutexes at that point. A thread parked in `write()` on `webserver_log` or `dnsmasq_log` is enough - FTL.log's own lock is not involved, so the handler logs happily right up to the fork and then blocks forever.

With one thread holding `dnsmasq_log.lock`, a `SIGSEGV` gets as far as

> Backtrace (X frames from signal context):

and stops there: no frames, no teardown, process still alive with the main thread in `futex_wait_queue`.

That child only does `dup2()`, `open()`, `close()` and `execvp()`, so it never touches a log mutex. `posix_spawnp()` does the same thing without running the handlers - glibc and musl both implement it with `clone(CLONE_VM|CLONE_VFORK)`. Replacing `signals.c:651-674` with:

```c
	// posix_spawn() rather than fork(): glibc and musl implement it with
	// clone(CLONE_VM|CLONE_VFORK) and do not run the pthread_atfork()
	// handlers, so a thread holding a log mutex cannot stall us here
	posix_spawn_file_actions_t fa;
	if(posix_spawn_file_actions_init(&fa) != 0)
	{
		close(pipefd[0]);
		close(pipefd[1]);
		return A2L_RUN_SPAWN_FAIL;
	}

	// adddup2() clears O_CLOEXEC on the duplicate so fds 1/2 survive the exec
	posix_spawn_file_actions_adddup2(&fa, pipefd[1], STDOUT_FILENO);
	posix_spawn_file_actions_addopen(&fa, STDERR_FILENO, "/dev/null", O_WRONLY, 0);
	posix_spawn_file_actions_addclose(&fa, pipefd[0]);
	posix_spawn_file_actions_addclose(&fa, pipefd[1]);

	pid_t pid = 0;
	const int spawn_err = posix_spawnp(&pid, "addr2line", &fa, NULL, argv, environ);
	posix_spawn_file_actions_destroy(&fa);
	if(spawn_err != 0)
	{
		close(pipefd[0]);
		close(pipefd[1]);
		return spawn_err == ENOENT ? A2L_RUN_MISSING : A2L_RUN_SPAWN_FAIL;
	}
```

plus `#include <spawn.h>`. `environ` is already visible through `<unistd.h>` since `FTL.h` defines `_GNU_SOURCE`. Everything below stays as it is: both implementations report a failed exec back through `posix_spawnp()`, so the `ENOENT` branch catches a missing `addr2line` and the `WEXITSTATUS(status) == 127` check remains as the fallback. The `A2L_RUN_SPAWN_FAIL` comment wants `posix_spawnp()` rather than `fork()`.
`posix_spawn()` is seemingly not on the async-signal-safe list either, so this might not be the general fix - but in my experiments it keeps the log locks out of the one fork the crash handler makes. It is the best I am able to come up with right now.

darkexplosiveqwx:
If I understand correctly that `pthread_atfork()` is the immediate issue, then what about using `_Fork()`? It got officially standardized in The Open Group Specification Issue 8 (POSIX.1-2024) and is specifically designed for async-signal-safety by being `fork()` without calling `pthread_atfork()` handlers (which may or may not do unsafe stuff). 

To quote [The Open Group Base Specification Issue 8](https://pubs.opengroup.org/onlinepubs/9799919799/id=62):

"The _Fork() function shall be equivalent to fork(), except that fork handlers established by means of the pthread_atfork() function shall not be called and _Fork() shall be async-signal-safe."

It would be a one-line change and not use the more complex `posix_spawnp()`.

Despite being standardized to POSIX in 2024, it has been available in Linux Standard C libraries for a bit longer. It is available since [Glibc 2.34, released August 1, 2021](https://lists.gnu.org/archive/html/info-gnu/2021-08/msg00001.html) and [musl 1.2.2, released January 15, 2021](https://musl.libc.org/releases.html).

DL6ER:
`_Fork()` is the better answer and I was simply not aware of this method's existence. Please take that instead of my suggestion and thank you for suggesting it.

On glibc: 2.34 is from August 2021, so the only distributions that would miss it are Debian 11 and Ubuntu 20.04 at 2.31 - and both are end of life by now, bullseye only recently (since 2026-08-31) and focal since May 2025. Debian 12 ships 2.36 and Ubuntu 22.04 ships 2.35, so every target we still support clears the bar. As this will be part of a breaking change, that's fine with me. And users can still use FTL on their devices (using the pre-compiled binaries), they will merely lose the ability to compile from source on EOL distros. An acceptable compromise, I'd say.

darkexplosiveqwx:
> `_Fork()` is the better answer and I was simply not aware of this method's existence. Please take that instead of my suggestion and thank you for suggesting it.

Already committed. The multiple force-pushes are simply from me messing a comment up multiple times.

DL6ER (approved):
Thanks for the long haul on this one.

PR #2963, opened 2026-07-21
ci: build riscv64 natively on the RISE project runner

PromoFaux:

riscv64 is the only architecture we still cross-build under QEMU, and it's comfortably the slowest leg of the build. The [RISE project](https://riscv-runners.riseproject.dev/docs/getting-started/labels) now offers native RISC-V GitHub Actions runners, so this points the riscv64 matrix entry at their early-access `ubuntu-26.04-riscv` label (Spacemit K3 hardware) and skips QEMU setup and the `--platform` flags on that one native leg via a `native` matrix flag.

Very much an experiment: I want to see whether the job actually gets picked up by their runners and, if so, how the native build times stack up against emulation. Happy to fall back to the GA `ubuntu-24.04-riscv` label, or revert entirely, depending on how it behaves.

---

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,8 +99,13 @@ jobs:
             build_opts: ""
           - platform: linux/riscv64
             bin_name: pihole-FTL-riscv64
-            runner: ubuntu-24.04-arm
+            # Native RISC-V runner from the RISE project (Scaleway EM-RV1) so we
+            # build natively instead of under QEMU emulation. The early-access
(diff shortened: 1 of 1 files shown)


yubiuser:
It seems, it was not picked up...

DL6ER:
Yeah, I think we are at a dead end here. Closing this one. We can later revisit once outside of early-non-working-preview state. But I somehow have the feeling that they will want to see some money at some point. If it is not a fixed amount (but usability-driven) and cheap enough, it may be worth deciding to invest some money here to reduce the pain slow RISCV64 is causing all the time...

PR #2976, opened 2026-07-26
Serve HTTP/2 and HTTP/3 on the webserver, add DoH2/DoH3 support

DL6ER:

An in-process TLS terminator fronting CivetWeb: it terminates TLS (OpenSSL) and serves HTTP/1.0, 1.1, 2 (`nghttp2`) and 3 (OpenSSL QUIC + `nghttp3`), forwarding plain HTTP/1.1 to the backend. It also adds DoH-over-HTTP/2 and DoH-over-HTTP/3 upstream clients to `dotdoh`.

The large `src/webserver/civetweb/` diff is *not* new code we we will have to carry/review: it is generic `PROXY-v2` support the backend needs so the terminator can hand the real client address (ultimately needed for client-aware blocking). That feature is proposed upstream as https://github.com/civetweb/civetweb/pull/1413; here we vendor it as `proxy_v2.inl` plus a ~9-line hook in `civetweb.c`, carried as a `patch/civetweb/` patch until upstream accepts it. civetweb/civetweb#1413 sits on civetweb/civetweb#1412 (a build and macOS-CI fix); the macOS-CI part there may end up folded into @yubiuser's build PR https://github.com/civetweb/civetweb/pull/1392.

1. Build FTL and start it with a TLS port, e.g. `webserver.port = "80,443s"`. (Already included in the default set of parameters, so there is no change here visible to users and neither a need to change any configuration already present to get this running)
2. `curl -k --http1.1 https://<pi>/admin/`, then `--http2` and `--http3` - each returns the web UI; the plain-HTTP port is unaffected.
3. For encrypted upstreams, arm `tls://`, `https://` and `h3://` upstreams and confirm they resolve. The `dotdoh` and HTTP/2 paths are also extensively covered by new tests in `test/dotdoh.bats` and `test/http2_test.sh` in CI.

This PR also builds FTL with `-ffunction-sections -fdata-sections -Wl,--gc-sections`, letting the linker drop unreferenced code. On its own that only strips FTL's own dead code; the full effect - notably trimming the large unused parts of OpenSSL, about 815 KB off the stripped static binary - needs the bundled static libraries compiled with matching section flags, which is https://github.com/pi-hole/docker-base-images/pull/177. The two are independent and non-breaking: either can be merged on its own and everything keeps building unchanged; only together do they realise the size reduction.

Note that there is no explicit need to configure HTTP 1.1/2 for normal `https://` requests because negotiation via ALPN will automatically find out what the best protocol is supported by this server.

Follow-up pull requests will be
- the still-missing downstream (from my local clients to the pi-hole) direction for DoT, DoH-over-HTTP1/2/3
- extension to up- and downstream DoQ support, including DoQ-over-HTTP3. Pi-hole then really covers all standardized alternative DNS protocols for the foreseeable future

---

---

## Checklist:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -572,18 +572,28 @@ if(LIBSSL AND LIBCRYPTO AND USE_TLS)
         message(STATUS "Building FTL with HTTP/2 support: NO (nghttp2 not found)")
     endif()
 
-    # HTTP/3 for the TLS terminator (OpenSSL QUIC + nghttp3). Optional and
-    # additionally requires QUIC support in the linked OpenSSL; without it the
-    # terminator serves HTTP/1.1 and (if built) HTTP/2 only. HTTP/3 also needs
-    # OpenSSL >= 4.0 for SSL_get_peer_addr() to forward the real client address;
-    # terminator.c raises a #error against an older OpenSSL.
+    # HTTP/3 for the TLS terminator (OpenSSL QUIC + nghttp3). Requires nghttp3
+    # AND OpenSSL >= 4.0, which provides SSL_get_peer_addr() (the per-connection
+    # QUIC peer getter used to forward the real client IP). Against an older
+    # OpenSSL such as 3.5.x, HTTP/3 and QUIC are disabled at build time and the
+    # terminator serves HTTP/1.1 and (if built) HTTP/2 only.
+    include(CheckCSourceCompiles)
+    check_c_source_compiles("
+#include <openssl/opensslv.h>
+#if OPENSSL_VERSION_NUMBER < 0x40000000L
+#error OpenSSL < 4.0 has no QUIC peer-address API
+#endif
+int main(void) { return 0; }
+" OPENSSL_HAS_QUIC_PEER_ADDR)
     find_library(LIBNGHTTP3 NAMES libnghttp3${LIBRARY_SUFFIX} nghttp3)
-    if(LIBNGHTTP3)
+    if(LIBNGHTTP3 AND OPENSSL_HAS_QUIC_PEER_ADDR)
         message(STATUS "Building FTL with HTTP/3 support: YES")
         target_compile_definitions(webserver PRIVATE HAVE_HTTP3)
         # The dotdoh client speaks DoH3 (HTTP/3 over QUIC) on h3:// upstreams.
         target_compile_definitions(dotdoh PRIVATE HAVE_HTTP3)
         target_link_libraries(pihole-FTL ${LIBNGHTTP3})
+    elseif(LIBNGHTTP3)
+        message(STATUS "Building FTL with HTTP/3 support: NO (OpenSSL < 4.0 lacks the QUIC peer-address API)")
     else()
         message(STATUS "Building FTL with HTTP/3 support: NO (nghttp3 not found)")
     endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -220,12 +220,14 @@ if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
     endif()
 endif()
 
-# Emit each function/variable in its own section and let the linker drop the
-# unreferenced ones (-Wl,--gc-sections). Combined with the static libraries
-# built with the same flags (see the ftl-build image), this strips dead code -
-# including large unused parts of OpenSSL - from the shipped binary.
+# Emit each function/variable in its own section so the linker can drop the
+# unreferenced ones at link time (-Wl,--gc-sections, added as a link option
+# below). Combined with the static libraries built with the same flags (see the
+# ftl-build image), this strips dead code - including large unused parts of
+# OpenSSL - from the shipped binary.
 if (CMAKE_C_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
-    set(HARDENING_FLAGS "${HARDENING_FLAGS} -ffunction-sections -fdata-sections -Wl,--gc-sections")
+    set(HARDENING_FLAGS "${HARDENING_FLAGS} -ffunction-sections -fdata-sections")
+    add_link_options(-Wl,--gc-sections)
 endif()
 
 # -FILE_OFFSET_BITS=64: used by stat(). Avoids problems with files > 2 GB on 32bit machines
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -221,11 +221,14 @@ if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
 endif()
 
 # Emit each function/variable in its own section so the linker can drop the
-# unreferenced ones at link time (-Wl,--gc-sections, added as a link option
-# below). Combined with the static libraries built with the same flags (see the
-# ftl-build image), this strips dead code - including large unused parts of
-# OpenSSL - from the shipped binary.
-if (CMAKE_C_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
+# unreferenced ones at link time (-Wl,--gc-sections). Combined with the static
+# libraries built with the same flags (see the ftl-build image), this strips dead
+# code - including large unused parts of OpenSSL - from the shipped binary. Probe
+# the linker for the flag rather than assuming it from the compiler: it is a GNU
+# ld/gold/lld feature, so e.g. Apple's ld64 (which spells it -dead_strip) skips it.
+include(CheckLinkerFlag)
+check_linker_flag(C "-Wl,--gc-sections" HAVE_LD_GC_SECTIONS)
+if (HAVE_LD_GC_SECTIONS)
     set(HARDENING_FLAGS "${HARDENING_FLAGS} -ffunction-sections -fdata-sections")
     add_link_options(-Wl,--gc-sections)
 endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -220,6 +220,14 @@ if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
     endif()
 endif()
 
+# Emit each function/variable in its own section and let the linker drop the
+# unreferenced ones (-Wl,--gc-sections). Combined with the static libraries
+# built with the same flags (see the ftl-build image), this strips dead code -
+# including large unused parts of OpenSSL - from the shipped binary.
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
+    set(HARDENING_FLAGS "${HARDENING_FLAGS} -ffunction-sections -fdata-sections -Wl,--gc-sections")
+endif()
+
 # -FILE_OFFSET_BITS=64: used by stat(). Avoids problems with files > 2 GB on 32bit machines
 # We define HAVE_POLL_H as this is needed for the musl builds to succeed
 set(CMAKE_C_FLAGS "-std=c17 -pipe ${WARN_FLAGS} -D_FILE_OFFSET_BITS=64 ${HARDENING_FLAGS} ${DEBUG_FLAGS} ${CMAKE_C_FLAGS} -DHAVE_POLL_H ${SQLITE_DEFINES}")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -208,10 +208,16 @@ if(STATIC)
 else()
     message(STATUS "Compiling dynamically linked executable")
 endif()
-# -pie -fPIE: (Dynamic) position independent executable
-
+# Position-independent executable (image ASLR). -fPIE compiles the objects
+# position-independent; the executable is then linked PIE via -pie (dynamic
+# build) or -static-pie (static build, set on the link line below). A plain
+# -static silently overrides -pie and yields a non-PIE binary with a fixed load
+# address, so the static build - the one we actually ship - must use -static-pie.
 if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
-    set(HARDENING_FLAGS "${HARDENING_FLAGS} -pie -fPIE")
+    set(HARDENING_FLAGS "${HARDENING_FLAGS} -fPIE")
+    if(NOT STATIC)
+        set(HARDENING_FLAGS "${HARDENING_FLAGS} -pie")
+    endif()
 endif()
 
 # -FILE_OFFSET_BITS=64: used by stat(). Avoids problems with files > 2 GB on 32bit machines
@@ -308,7 +314,7 @@ add_executable(pihole-FTL
 if(STATIC)
     set_target_properties(pihole-FTL PROPERTIES LINK_SEARCH_START_STATIC ON)
     set_target_properties(pihole-FTL PROPERTIES LINK_SEARCH_END_STATIC ON)
-    target_link_libraries(pihole-FTL -static-libgcc -static)
+    target_link_libraries(pihole-FTL -static-libgcc -static-pie)
     set(LIBRARY_SUFFIX "${CMAKE_STATIC_LIBRARY_SUFFIX}")
 else()
     find_library(LIBMATH m)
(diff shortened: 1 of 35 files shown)


darkexplosiveqwx:
This is looking quite nice and it no longer needs the legacy engine header. Compiles successfully on Fedora.

One thing about OpenSSL 3.5 compatibility:
You `#error` at https://github.com/pi-hole/FTL/blob/34cacc2127786e846fbfaae264ff636978802472/src/webserver/terminator.c#L43-L48 and https://github.com/pi-hole/FTL/blob/34cacc2127786e846fbfaae264ff636978802472/src/dotdoh/quic_client.c#L35-L39


So the only part hard-depending on OpenSSL 3 is HTTP/3 as said in:
https://github.com/pi-hole/FTL/blob/34cacc2127786e846fbfaae264ff636978802472/src/CMakeLists.txt#L575-L580

I propose a CMake option to turn it off to allow building with older openssl versions without patching the code:
```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 31a2d0b3..96fbc890 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -545,6 +545,7 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 endif()
 
 option(USE_TLS "Build FTL with TLS support, if available" ON)
+option(USE_HTTP3 "Build FTL with HTTP/3 support, if available" ON)
 
 find_library(LIBSSL NAMES libssl${LIBRARY_SUFFIX} ssl)
 find_library(LIBCRYPTO NAMES libcrypto${LIBRARY_SUFFIX} crypto)
@@ -578,7 +579,7 @@ if(LIBSSL AND LIBCRYPTO AND USE_TLS)
     # OpenSSL >= 4.0 for SSL_get_peer_addr() to forward the real client address;
     # terminator.c raises a #error against an older OpenSSL.
     find_library(LIBNGHTTP3 NAMES libnghttp3${LIBRARY_SUFFIX} nghttp3)
-    if(LIBNGHTTP3)
+    if(LIBNGHTTP3 AND USE_HTTP3)
         message(STATUS "Building FTL with HTTP/3 support: YES")
         target_compile_definitions(webserver PRIVATE HAVE_HTTP3)
         # The dotdoh client speaks DoH3 (HTTP/3 over QUIC) on h3:// upstreams.
```

DL6ER:
Yes, I started with OpenSSL 3.5.7 as QUIC was announced to be available here, but then (during implementing exactly this PR here -> HTTP/3) found that it is actually not complete. Only OpenSSL 4.0.0+ added a very necessary dependency (`SSL_get_peer_addr()`) that is being able to get peer's addresses in QUIC mode (also relevant for HTTP/3). Instead of trying to patch the crypto library myself to retrofit the function, I decided switching to 4.0 right from the beginning would be the better strategy.

The `#ifdef`s you found have actually designed for minimum 3.5.7 requirement before but I decided to simply increment them now, too, to match my "start from 4.0.0 idea. But I rather built an automatic-detector for the installed OpenSSL version to make it easier to compile FTL from sources on such systems.

We could but I actually don't want to go down a halfway-functional path and explain users that HTTP/3 and DoQ is *in principle* available when comoiled with OpenSSL 3.5.* but often used features like client-aware blocking would simply magically not work with OpenSSL 3.5.7. That feels wrong and will likely cause a notable support burden on us.

yubiuser:
> On its own this drops FTL's own dead code

Does it also say which part of FTlLs code is dead? Could it be removed right from the source code as well?

DL6ER:
> Does it also say which part of FTLs code is dead?

No, unfortunately not. The compiler will complain if we have functions defined that are really not used anywhere. However, if users decide to build FTL with more or less functions, it can happen that some functions are unused. This will have more effect on nettle and OpenSSL where now all the many algorithms, which are not used by FTL/TLS, will be stripped from the final binary.

> Could it be removed right from the source code as well?

yubiuser:
Really nice

```
 Connected to nanopi.lan (10.0.1.24) port 443

* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://nanopi.lan/
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: nanopi.lan]
* [HTTP/2] [1] [:path: /]
* [HTTP/2] [1] [user-agent: curl/8.11.0-DEV]
* [HTTP/2] [1] [accept: */*]
> GET / HTTP/2
> Host: nanopi.lan

```

```
* Connected to nanopi.lan (10.0.1.24) port 443
* using HTTP/3
* [HTTP/3] [0] OPENED stream for https://nanopi.lan/
* [HTTP/3] [0] [:method: GET]
* [HTTP/3] [0] [:scheme: https]
* [HTTP/3] [0] [:authority: nanopi.lan]
* [HTTP/3] [0] [:path: /]
* [HTTP/3] [0] [user-agent: curl/8.11.0-DEV]
* [HTTP/3] [0] [accept: */*]
> GET / HTTP/3
> Host: nanopi.lan
```

yubiuser:
I've selected IPv4 `https://dns.quad9.net@9.9.9.9/dns-query` as upstream and FTL logged

`WARNING: Invalid IPv6 address when trying to resolve hostname: https://dns.quad9.net@9.9.9.9/dns-query`

Why is it trying to resolve IPv6 when I only want to use the IPv4 address?

____

And it's logged again despite I switched the upstream to `h3://cloudflare-dns.com/dns-query` and FTL was restarted

```
2026-07-28 14:01:41.568 UTC [522517/T533038] INFO:  -> Known forward destinations: 4
2026-07-28 14:01:41.582 UTC [522517/T533038] INFO: Compiled 0 allow and 0 deny regex for 5 clients in 0.1 msec
2026-07-28 14:01:44.125 UTC [522517/T533040] WARNING: Invalid IPv6 address when trying to resolve hostname: https://dns.quad9.net@9.9.9.9/dns-query
2026-07-28 14:01:44.126 UTC [522517/T533040] WARNING: Invalid IPv6 address when trying to resolve hostname: h3://cloudflare-dns.com/dns-query

```

src/args.c
```diff
@@ -1122,8 +1128,9 @@ void parse_args(int argc, char *argv[])
 				printf("Compression: %sYes%s\n", green, normal);
 			else
 				printf("Compression: %sNo%s\n", red, normal);
-			if(mg_check_feature(MG_FEATURES_HTTP2))
-				printf("                 HTTP2: %sYes%s, ", green, normal);
+			// Likewise CivetWeb's native HTTP/2 stays off; the backend serves it.
+			if(webserver_have_http2())
+				printf("                 HTTP2: %sYes%s %s(via TLS backend)%s, ", green, normal, blue, normal);
```
yubiuser, line 1133: Do you want to add HTTP3 via TLS backend info here?

darkexplosiveqwx: I think not.
HTTP/2 and TLS are features civetweb could provide itself, "(via TLS backend)" is added to acknowledge that the feature is available, but not via civetweb's own implementation. Civetweb has no HTTP/3 support at all, so there is no clarification needed.

DL6ER: The `(via TLS backend)` note exists to disambiguate a feature CivetWeb could self-provide: HTTP/2 (even if the "experimental" implementation is very incomplete). CivetWeb has no HTTP/3 at all, so nothing to disambiguate on this end. The HTTP/2, HTTP/3 truth is slightly lower in the full version output.

src/webserver/webserver.c
```diff
@@ -611,6 +699,87 @@ static void print_webserver_opts(const bool debug, const size_t idx, const char
...
+			strncat(backend, ",", backend_len - strlen(backend) - 1);
+		strncat(backend, ent, backend_len - strlen(backend) - 1);
+	}
+	free(copy);
+
+	// Append the loopback plaintext backend CivetWeb serves the terminator on.
+	// Port 0 lets the kernel pick a free port; it is read back after mg_start2().
+	if(backend[0] != '\0')
+		strncat(backend, ",", backend_len - strlen(backend) - 1);
+	strncat(backend, "127.0.0.1:0", backend_len - strlen(backend) - 1);
```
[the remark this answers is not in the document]
DL6ER: The length handed to `strncat()` cannot actually underflow here, as `strncat()` keeps `strlen(backend) <= backend_len - 1`, so `backend_len - strlen(backend) - 1` bottoms out at *zero* rather than wrapping. We would still rather not depend on that invariant holding for every future caller, so the four appends now go through a small `str_append()` helper that becomes a *no-op* once the buffer is full. That is more robust than the inline arithmetic, as `strncat()` never sees a computed length again, and it also keeps the pattern from tripping the scanner. Done in 030fca9bd.

src/CMakeLists.txt
```diff
@@ -220,6 +220,16 @@ if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
...
 
+# Emit each function/variable in its own section so the linker can drop the
+# unreferenced ones at link time (-Wl,--gc-sections, added as a link option
+# below). Combined with the static libraries built with the same flags (see the
+# ftl-build image), this strips dead code - including large unused parts of
+# OpenSSL - from the shipped binary.
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
+    set(HARDENING_FLAGS "${HARDENING_FLAGS} -ffunction-sections -fdata-sections")
+    add_link_options(-Wl,--gc-sections)
+endif()
```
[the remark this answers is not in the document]
DL6ER: We only target Linux here, i.e., GNU ld/gold/lld, where `--gc-sections` is always available - Mach-O/ld64 (which spells it `-dead_strip`) is not a platform we build for. We would rather keep the plain GNU/Clang compiler-ID guard than add a linker-capability probe for a case that does not arise, as the probe is extra machinery buying us nothing on the toolchains we actually use. If a non-GNU-ld toolchain ever needs it, adding the probe at that point is cheap.

DL6ER:
This is a *cosmetic* bug, but a real one, and it is fixed now. We are not actually trying to use IPv6: encrypted upstreams are recorded in the upstream table by their configured URL, e.g., `https://dns.quad9.net@9.9.9.9/dns-query`, rather than the internal loopback tuple, so the API and PADD show the real upstream instead of a `127.x` address. The periodic hostname resolver then passed that whole URL to its PTR lookup, which decides IPv4 vs IPv6 by whether the string contains a `:`, and the `://` sent it down the IPv6 path - hence the warning. The `https://...@9.9.9.9...` and the `h3://cloudflare-dns.com...` lines are the same case. These URLs have no PTR record to resolve, so we now skip them in the resolver. No query was ever affected by this, only the log line.

yubiuser:
> This is a cosmetic bug, but a real one, and it is fixed now. 

Did you forget to push the fix?

```
2026-07-28 20:03:05.982 UTC [648948M] INFO: FTL commit: 030fca9b
....
2026-07-28 20:03:07.198 UTC [648948/T649009] INFO: Compiled 0 allow and 0 deny regex for 5 clients in 0.1 msec
2026-07-28 20:03:09.070 UTC [648948/T649011] WARNING: Invalid IPv6 address when trying to resolve hostname: h3://cloudflare-dns.com/dns-query
2026-07-28 20:03:20.843 UTC [648948/T649008] INFO: Received 8/8 valid NTP replies from de.pool.ntp.org
```

P.S. Merge conflicts.

yubiuser:
I noticed that my browser does not use h3 to connect to the dashboard, it only uses h2 or h1.1. Do we need to add the `Alt-Svc: h3=":443"` header?

DL6ER:
> Did you forget to push the fix?

I ... am ... well ... yes ...

yubiuser:
Not sure if this is something that changed with this branch, but I get a strange login behavior. I was logged in via `https` but closed the tab without logging out. Opened a new tab and navigated to the non-http address which showed the login page. Instead of logging in I followed the link to the HTTPS page and was automatically logged in. Is this expected?

Peek 2026-07-29 12-18

DL6ER:
The `Alt-Svc` change is pushed and there is a new test alongside it.

Re login behavior:

What you describe is expected and there since ancient PHP times. The reason is that the session cookie is marked `Secure` so your browser is actually instructed to **never** submit this cookie in plain. Hence, FTL just sees "there is no session cookie" and sends you to the login page. When you switch back to HTTPS, the browser is sending the already existing and still valid (not yet expired) Session cookie and you are recognized again without having to re-login. This is by construction and cannot (and IMO shouldn't) be changed.

The new last commit fixed a malfunctioning `./build.sh debug` thing I have found while working on the new feature and wanting to have the debugger attached.

darkexplosiveqwx:
`./build.sh -DUSE_TLS=OFF` fails on Fedora 44, GCC `16.1.1`:

```
/home/shared/Projects/fleet/pi-hole-FTL/src/webserver/terminator.c: In function ‘terminator_proxy_token_hex’:
/home/shared/Projects/fleet/pi-hole-FTL/src/webserver/terminator.c:3341:6: error: function might be candidate for attribute ‘const’ [-Werror=suggest-attribute=const]
 3341 | bool terminator_proxy_token_hex(char *out, size_t outsz)
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

at the end of `src/webserver/terminator.c` for the `terminator_proxy_token_hex()` stub we can either add:
- ```C
  __attribute__((const))
  ```
- ```C
  #pragma GCC diagnostic push
  #pragma GCC diagnostic ignored "-Wsuggest-attribute=const"
  ```

rdwebdesign:
> Re login behavior:
> 
> What you describe is expected and there since ancient PHP times. The reason is that the session cookie is marked `Secure` so your browser is actually instructed to **never** submit this cookie in plain. Hence, FTL just sees "there is no session cookie" and sends you to the login page. When you switch back to HTTPS, the browser is sending the already existing and still valid (not yet expired) Session cookie and you are recognized again without having to re-login. This is by construction and cannot (and IMO shouldn't) be changed.

I agree. 
This is the correct behavior.

The browser considers `http://domain` as a different website than `https://domain`. They are independent and they can't share a session cookie.

DL6ER:
Fixed in `d03cb341e`. GCC 16's analysis is stricter than the compilers we build against here and (correctly) sees that the no-TLS stub of `terminator_proxy_token_hex()` has no observable effect, so it suggests `__attribute__((const))`, which `-Werror` then turns into a hard error. We cannot mark the shared declaration `const`, since the real TLS build genuinely has side effects, so the stub now clears the caller's output buffer instead - a real side effect, and the correct empty-token result when there is no terminator.

yubiuser:
> The Alt-Svc change is pushed and there is a new test alongside it.

For anyone coming here and testing h3 in real browsers. Usually browsers reject QUIC with self-signed CAs (see [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1985341)). In FF, one can enable `network.http.http3.disable_when_third_party_roots_found` which allows http3 connections, ~~but somehow the dashboard fails to load completely. Maybe still a bug at our side.~~ (Bug fixed with the two following commits)

I was unable to get http3 working on Chrome-based browsers.

yubiuser (approved):
Tested all new function as much as I could. But manually reviewing > 6000 lines of code is impossible. I'm approving. When @darkexplosiveqwx does not find more compiler-related issues this could go to `development` I would say.

DL6ER:
Than you. These latest PRs actually lay the basis for the following pieces:
- DNS DoT/DoH server/downstream part (clients in your local network -> your Pi-hole)
- DoQ implementation for server and client

Optional follow-ups are
- Remove the Lua engine from the webserver -> all-static web interface. Preliminary tests suggest decrease of I/O activity by ~75%, so potentially big improvements for slowish SD cards and friends.
   Even when we plan to keep `lua` as FTL subcommand (there is ongoing work in core to port some scripts over to Lua), removing the CGI-feature will both make external contributions easier (not that many people are familiar with Lua in Kepler style) as well as easier to do future migrations. This deserves to be included in a breaking/major release nonetheless.
- Switch to https://github.com/DL6ER/civetweb/tree/pi-hole which is our own "maintained" fork of CivetWeb. Maintained in this case means that we grabbed latest `master` and integrated all fixes from open pull requests after having reviewed them for correctness ourselves. CI on this branch is green and future security updates will be shipped.

yubiuser:
> Remove the Lua engine from the webserver -> all-static web interface. Preliminary tests suggest decrease of I/O activity by ~75%, so potentially big improvements for slowish SD cards and friends.

I would keep it as subcommand for now, but try to switch to all static.

DL6ER:
@darkexplosiveqwx did you see the question from yubiuser? 

> When @darkexplosiveqwx does not find more compiler-related issues this could go to development I would say.

https://github.com/pi-hole/FTL/pull/2976#pullrequestreview-4816022709

We can also merge and resolve possible additional issues separately later if you are too busy right now.

darkexplosiveqwx:
> @darkexplosiveqwx did you see the question from yubiuser? 
> 
> 
> 
> > When @darkexplosiveqwx does not find more compiler-related issues this could go to development I would say.
> 
> 
> 
> https://github.com/pi-hole/FTL/pull/2976#pullrequestreview-4816022709
> 
> 
> 
> We can also merge and resolve possible additional issues separately later if you are too busy right now. 


Must've missed it. The GitHub mobile app does some aggressive collapsing, even for unread comments - it also hid the Copilot session.

I haven't had any other issues while compiling on Fedora 44. No blocker from my side.

Issue #2978, opened 2026-07-26
Internal PTR resolver can assign a delayed UDP response to the next client


brealorg:
### Versions

- Pi-hole Core: v6.4.3
- Web interface: v6.6
- FTL: v6.7
- Exact FTL source commit used for reproduction: `fa65a88f8cdef1013594d4de14108077954faea4`
- `src/resolve.c` blob at that commit: `079d5451e93035876a280104488c17461ad3969e`

### Platform

Original observation:

- OS: Debian 12
- Platform: native Raspberry Pi / arm64
- Internal resolver topology: FTL queries Pi-hole on `127.0.0.1`, with local reverse forwarding and a local Unbound resolver

Deterministic reproduction:

- Full FTL daemon with embedded dnsmasq
- Pi-hole's official build/runtime image: `ghcr.io/pi-hole/ftl-build:v2.19`
- Isolated Docker network with two client addresses and a controlled UDP DNS server
- Independent `AF_PACKET` capture used to validate the packet sequence

### Expected behavior

A UDP DNS response should only be accepted for the request it matches. At minimum, the internal resolver should correlate the response using the DNS transaction ID and the echoed question before parsing or storing its answer.

A delayed response to request A must not be accepted while the resolver is waiting for request B.

### Actual behavior / bug

The internal PTR resolver reuses one UDP socket across sequential client lookups. Its UDP path sends a query and consumes a single datagram with `recvfrom()`, then parses that datagram as the active response without visibly checking the received response against the active request.

In the v6.7 source used here, the UDP receive path does not validate:

- source address and port against the intended endpoint;
- `QR` and opcode;
- DNS transaction ID;
- `QDCOUNT`;
- echoed QNAME;
- QTYPE;
- QCLASS.

The `recvfrom()` byte count is also discarded. Answer parsing starts at an offset based on the outbound query length, and parser bounds use the full receive buffer rather than the actual datagram length.

This allows the following sequence:

```text
send PTR request A
A exceeds the resolver's receive timeout
send PTR request B on the same UDP socket
delayed response A arrives
response A is consumed and parsed as the response to B
A's PTR hostname is stored for client B
correct response B arrives later, after B's lookup has already completed
```

### Deterministic full-daemon reproduction

The controlled server handled two reverse names:

```text
A = 2.251.29.172.in-addr.arpa
B = 3.251.29.172.in-addr.arpa
```

[...]

PR #2979, opened 2026-07-26
Validate internal PTR DNS responses

brealorg:

The internal PTR resolver reuses one UDP socket across sequential lookups. If a lookup times out, a delayed response from that lookup can remain queued and be consumed while the resolver is waiting for the next lookup.

This change correlates received UDP DNS datagrams with the active PTR request before parsing them. It:

- validates the response source address and port;
- validates `QR`, opcode, transaction ID and `QDCOUNT`;
- decodes and compares the echoed QNAME, QTYPE and QCLASS;
- discards unrelated datagrams while keeping one absolute `CLOCK_MONOTONIC` deadline;
- uses the actual `recvfrom()` byte count as the UDP parser boundary;
- validates into a local DNS header and only copies it out after the response is accepted;
- preserves the secure query ID introduced by #3030 and derives `request_id` from the final `dns.id`;
- reads both the TCP length prefix and announced DNS message body completely under one absolute deadline, rejects replies shorter than a DNS header, and bounds parsing by the received length;
- adds a deterministic regression harness under `test/` for the stale-response sequence.

The independent `return NULL;` → `return false;` cleanup identified during review was split into #3027 and has already landed separately. The three remaining cases in this `bool` resolver function were corrected here during the rebase as requested.

The regression harness lets lookup A time out, then releases A's delayed response while lookup B starts on the same UDP socket. The server sends the stale A response before the correct B response.

On the pre-rebase baseline used to demonstrate the regression (`a17efdd3307a233172d2f3b0d1e2ead5f9c85302`):

```text
FIRST_LOOKUP_TIMEOUT=PASS
SECOND_LOOKUP_RESULT=late-a.example
SERVER_SEQUENCE=PASS
PTR_RESPONSE_REGRESSION=FAIL
```

With this change applied:

```text
FIRST_LOOKUP_TIMEOUT=PASS
SECOND_LOOKUP_RESULT=answer-b.example
SERVER_SEQUENCE=PASS
PTR_RESPONSE_REGRESSION=PASS
```

The harness is integrated into the normal test suite as `PTR stale-response regression harness`.

The current PR head (`3126faff259335facba39794013b76da180d9b9b`) passed the full GitHub CI matrix. An isolated local merge of this head into current `development` at `c4144a786bf1ceea0cae6d8aec8763b9f24a4927` was also built and tested in the current `ghcr.io/pi-hole/ftl-build:v2.25` environment:

```text
193 BATS tests, 0 failures
151 pytest API tests passed
12 dotdoh client tests, 0 failures
25 dotdoh server tests, 0 failures
9 final log-validation tests, 0 failures
TLS terminator test: PASS
```

The original UDP regression harness and a temporary TCP fault-injection harness also passed under ASan/UBSan. The TCP audit covered fragmented length-prefix and body reads, EOF after a partial prefix, replies shorter than a DNS header, oversized announced lengths, premature body EOF, and byte trickling constrained by the single 2000 ms deadline.

The earlier full-daemon reproduction documented on #2978 remains independent corroboration of the same stale-response mechanism.

## Additional information

**Related issue or feature (if applicable):** Refs #2978

---

5. I have squashed any insignificant commits.
6. My change does not modify `src/dnsmasq/`.

## Checklist:

- [x] I signed off all commits (DCO).
- [x] I cryptographically signed all commits.

### DIFF
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -35,7 +35,7 @@ RUN rm -rf cmake && \
 # Start from clean counters so the statistics below describe this build alone
     ccache --zero-stats && \
 # Build and test FTL
-    bash build.sh "-DSTATIC=${STATIC} -DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON -DCMAKE_C_COMPILER_LAUNCHER=ccache" ${BUILD_OPTS} && \
+    bash build.sh "-DSTATIC=${STATIC} -DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON -DBUILD_PTR_RESPONSE_REGRESSION=ON -DCMAKE_C_COMPILER_LAUNCHER=ccache" ${BUILD_OPTS} && \
 # Report what the cache saved, so the CI log shows whether it is worth keeping
     ccache --show-stats && \
 # Copy FTL binary to root directory
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ pihole-FTL*
 # Generated standalone regression test binaries (built via ./build.sh test)
 /tar_regression
 /gzip_regression
+/dotdoh_regression
+/ptr_response_regression
 
 # Generated pihole.toml file
 pihole.toml
--- a/build.sh
+++ b/build.sh
@@ -113,9 +113,9 @@ fi
 # They are gated behind CMake options so ordinary builds do not produce them.
 if [[ -n "${test}" ]]; then
     if [[ -n "${cmake_args}" ]]; then
-        cmake_args="${cmake_args} -DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON"
+        cmake_args="${cmake_args} -DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON -DBUILD_PTR_RESPONSE_REGRESSION=ON"
     else
-        cmake_args="-DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON"
+        cmake_args="-DBUILD_TAR_REGRESSION=ON -DBUILD_GZIP_REGRESSION=ON -DBUILD_DOTDOH_REGRESSION=ON -DBUILD_PTR_RESPONSE_REGRESSION=ON"
     fi
 fi
 
@@ -212,7 +212,7 @@ echo "Copying compiled pihole-FTL binary to repository root"
 cp pihole-FTL ../
 # Copy the regression test binaries alongside it so the bats tests can run them
 # from the repo root.
-for regression_bin in tar_regression gzip_regression dotdoh_regression; do
+for regression_bin in tar_regression gzip_regression dotdoh_regression ptr_response_regression; do
     if [[ -f "${regression_bin}" ]]; then
         cp "${regression_bin}" ../
     fi
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -482,6 +482,14 @@ if(BUILD_PTR_RESPONSE_REGRESSION)
     target_include_directories(ptr_response_regression PRIVATE ${PROJECT_SOURCE_DIR}/src)
     target_compile_options(ptr_response_regression PRIVATE -fvisibility=hidden)
     target_link_libraries(ptr_response_regression Threads::Threads)
+
+    # Optional ASan/UBSan instrumentation for local runs:
+    #   ./build.sh "-DPTR_RESPONSE_REGRESSION_SANITIZE=ON" test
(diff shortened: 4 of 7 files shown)


brealorg:
A new production recurrence has been captured on the unpatched FTL v6.7 installation.

The full timestamped evidence and interpretation are recorded on issue #2978:

https://github.com/pi-hole/FTL/issues/2978#issuecomment-5145402831

Key observation: two IPv6 PTR timeouts were immediately followed by the IPv4 PTR lookup for `192.168.1.78` being reported with residual `ip6.arpa` question/answer data, while embedded dnsmasq simultaneously returned the correct `Stormtab.localdomain` PTR.

There is no packet capture from that exact production event, so it is documented as strong independent corroboration rather than standalone wire-level proof. No change to the draft patch is requested solely from this recurrence.

DL6ER:
Confirmed the mechanism on #2978, so this is the right layer to fix it.

Using `recvfrom()`'s length as the parser boundary is a second fix in its own right: the existing `bufend = buf + sizeof(buf)` let the answer parser walk past the datagram into stale buffer content. The TCP path carries the same over-read and `prefix` is known there, so please use `buf + prefix` instead of leaving that path out of scope.

Smaller things:

- Timeouts log through `strsockerr(EAGAIN)`, an errno string for an errno nothing returned. "Timed out after 2000 ms" is clearer.
- `validate_udp_ptr_response()` fills the caller's `dns` before any check passes, so a rejected packet clobbers the header copy. Harmless while the loop continues, but validating into a local and copying out on success is cleaner.
- A peer sending unrelated datagrams keeps the loop spinning until the deadline. Fine on loopback, worth a comment saying so.
- `source_len < sizeof(source)` compares `socklen_t` against `size_t`.
- The four `return NULL` to `return false` corrections are independent of the feature. Split them into their own PR and I will merge that one right away.

Drop the PoC wording from the comments before this goes ready - the bug story belongs in the PR and the commit message, not in the code.

You wrote on #2978 that the deterministic harness can be supplied separately. I would rather have it in `test/` as a regression test than kept aside, if you are willing to adapt it.

Worth deciding separately: `connect()`ing the socket would let the kernel drop foreign sources for free, but not a stale reply from the right source, so the validation is needed either way.

brealorg:
Split the four independent bool-return cleanups into #3027 as requested.

DL6ER:
The conflict in `src/resolve.c` is my doing - #3030 landed this morning and changed the query ID in the same function. When you rebase, `dns.id` is no longer `htons(random())`, so `request_id` should pick up whatever the ID ends up being rather than reintroducing the old call. The green checks here ran against the base from before that merge, so they do not say anything about the rebased state yet.

One thing left in the TCP path. `prefix` is the length the peer announces, but the `recv()` below reads into `sizeof(buf)` and its return value is discarded, so nothing checks that `prefix` bytes actually arrived. On a stream socket a short read is normal, and `bufend = buf + prefix` then covers bytes that were never received - the same over-read you fixed for UDP by using the real `recvfrom()` length. Reading exactly `prefix` bytes in a loop and bounding by what came back would close it.

Three `return NULL` are also left in this `bool` function: the `hname` allocation, the UDP send failure and the truncated-response path. I said four last time, which was the count in the TCP block you moved to #3027 - these are the rest.

brealorg:
Rebased onto current `development` and addressed the remaining review points:

- preserved the secure query ID path from #3030 and derive `request_id` from the final `dns.id`;
- read the TCP DNS body to the announced length before parsing;
- corrected the three remaining `return NULL;` cases in the `bool` resolver function;
- adapted the deterministic PTR regression harness to the secure query-ID path.

The exact rebased tree passed locally in the current FTL build image: 193 BATS tests, 151 pytest API tests, 12 DoT/DoH client tests, 25 DoT/DoH server tests, 9 final validation tests, and the TLS terminator test.

Fresh GitHub workflows are now awaiting maintainer approval.

DL6ER:
I have pushed the small things from my review straight onto the branch - it was quicker than writing them up, and easier to read as a diff. Please look the commit over before we merge, and say so if you disagree with any of it.

The one with substance: the TCP length prefix was still read with a single unchecked `recv()`, and the body loop got a fresh `SO_RCVTIMEO` per call. Both now go through `recv_all()` under one deadline, the same budget the UDP path uses.

The rest is a shared `RESOLVER_TIMEOUT_MS`, two unreachable `EINTR` branches (`FTLrecvfrom()` and `FTLrecv()` already retry), a dead initializer, and moving the CMake block below the dotdoh one so that comment stops describing the wrong option. I also tried to drop `-fvisibility=hidden` as unexplained and found out why it is there: without it `--gc-sections` keeps `resolveClients()` and the harness no longer links. It stays, now with a comment saying so.

brealorg:
I reviewed the current follow-up commits through `3126faff` and have no objections.

I also tested an isolated merge of this head into current `development` at `c4144a78`. It merged cleanly and passed the full amd64 CI-equivalent suite. The original UDP regression harness and a temporary TCP fault-injection harness both passed under ASan/UBSan; the TCP audit covered fragmented prefix/body reads, EOF after a partial prefix, short and oversized announced lengths, premature body EOF, and a byte-trickling peer constrained to the single 2000 ms deadline.

The `recv_all()` single-deadline change, short-header guard, and `recv_nowarn()` logging cleanup all look good to me. Thanks for tightening this up — good to merge from my side.

DL6ER (approved):
Mechanism confirmed on #2978 - the shared UDP socket is what needed the correlation.

I pushed the rest of the review onto the branch: the TCP read under one deadline, a length lower bound, `.gitignore` entries, an ASan option, a duplicate log line. Shout if you disagree.

Thanks for the harness.

PR #2988, opened 2026-07-31
synthesize a HTTPS record for pi.hole

darkexplosiveqwx:

**Please make sure you**

---

FTL now includes native HTTP/2 and HTTP/3 support, it makes sense to advertise those capabilities via DNS, more specifically the HTTPS RR. Modern browsers are already querying for it in addition to A/AAAA. This allows them to connect directly using QUIC instead of upgrading later. 

This PR adds an automatically synthesized RFC 9460 HTTPS RR for `pi.hole.` if any native TLS listener is active. 
example:
```
1 . alpn="h2,h3" port=8086 ipv4hint=127.0.0.1 ipv6hint=::1
```

`alpn` is decided based on compile-time support.
The IP hints are localized and match the A/AAAA data.

Limitation I am working on:

This currently doesn't work with CNAMEs targeting `pi.hole`. It is not as simple as https://github.com/pi-hole/FTL/pull/2585, since dnsmasq's cache differs for host (A/AAAA) and generic Resource Records. 

---

---

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -576,8 +576,6 @@ if(LIBSSL AND LIBCRYPTO AND USE_TLS)
         target_compile_definitions(webserver PRIVATE HAVE_HTTP2)
         # The dotdoh client auto-negotiates HTTP/2 (DoH2) on https:// upstreams.
         target_compile_definitions(dotdoh PRIVATE HAVE_HTTP2)
-        # h2 is dynamically enabled in the HTTPS RR
-        target_compile_definitions(core PRIVATE HAVE_HTTP2)
         target_link_libraries(pihole-FTL ${LIBNGHTTP2})
     else()
         message(STATUS "Building FTL with HTTP/2 support: NO (nghttp2 not found)")
@@ -602,8 +600,6 @@ int main(void) { return 0; }
         target_compile_definitions(webserver PRIVATE HAVE_HTTP3)
         # The dotdoh client speaks DoH3 (HTTP/3 over QUIC) on h3:// upstreams.
         target_compile_definitions(dotdoh PRIVATE HAVE_HTTP3)
-        # h3 is dynamically enabled in the HTTPS RR
-        target_compile_definitions(core PRIVATE HAVE_HTTP3)
         target_link_libraries(pihole-FTL ${LIBNGHTTP3})
     elseif(LIBNGHTTP3)
         message(STATUS "Building FTL with HTTP/3 support: NO (OpenSSL < 4.0 lacks the QUIC peer-address API)")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -576,6 +576,8 @@ if(LIBSSL AND LIBCRYPTO AND USE_TLS)
         target_compile_definitions(webserver PRIVATE HAVE_HTTP2)
         # The dotdoh client auto-negotiates HTTP/2 (DoH2) on https:// upstreams.
         target_compile_definitions(dotdoh PRIVATE HAVE_HTTP2)
+        # h2 is dynamically enabled in the HTTPS RR
+        target_compile_definitions(core PRIVATE HAVE_HTTP2)
         target_link_libraries(pihole-FTL ${LIBNGHTTP2})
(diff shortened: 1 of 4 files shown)


rdwebdesign:
Can you please add some information to the PR template explaining the reasons for this change and adding a few details about the code?

DL6ER:
Two things need to change before this can go ready, both about where the advertised values come from.

`webserver_have_http2()` and `webserver_have_http3()` are pure `#ifdef` wrappers, so the record advertises `h3` whenever FTL was *built* with HTTP/3 support, whether or not a QUIC listener is actually running. Clients then attempt QUIC and fall back. This has to key off the runtime listener state instead.

`get_https_port()` returns a single global port, while #3017 serves every configured TLS port - the record would then advertise one arbitrary port out of several.

On top of that, the config keys around this are not settled yet. #3012 is still a draft for exactly that reason: the current shape, `dns.doh` as a boolean (false: no DoH, true: served on all webserver ports) next to `dns.dot` and `dns.doq` as uint16 ports, is not what the team wants. Whatever comes out of that discussion decides what this record should advertise, so I would rather have this one wait for it than rework it twice.

The encoding itself checks out: SvcParamKeys in ascending order as RFC 9460 requires, the port in network byte order, and `"sbt"` giving SvcPriority 1 with a single zero byte for TargetName `.`. Worst case is 44 of the 64 bytes, and `alpn[6]` fits `h2` plus `h3` exactly.

Smaller things: the BATS test pins `port=443` and both ALPNs, so it encodes the CI build configuration and will need updating along with the points above. The description says CNAMEs targeting `pi.hole` do not work, but the test asserts an HTTPS RR through `pihole.mydomain.net` - which of the two is it? And please drop the whitespace-only hunks in `dnsmasq_interface.c` and `test_suite.bats`, and fill in the description rdwebdesign asked for.

DL6ER:
```
   ✗ Pi-hole uses interface-dependent replies inside CNAME chains
     (from function `__assert_line' in file /bats//bats-assert/src/assert_line.bash, line 252,
      from function `assert_line' in file /bats//bats-assert/src/assert_line.bash, line 131,
      in test file test/test_suite.bats, line 1357)
       `assert_line --index 1 "1 . alpn=\"h2,h3\" port=443 ipv4hint=127.0.0.1 ipv6hint=::1"' failed
     
     -- line differs --
     index    : 1
     expected : 1 . alpn="h2,h3" port=443 ipv4hint=127.0.0.1 ipv6hint=::1
     actual   :
     --
     
     
     ═══════════════════════════════════════════════════════════════════════════════
                                   BATS TEST FAILURE DEBUG
     ═══════════════════════════════════════════════════════════════════════════════
     
        TEST DESCRIPTION:
        Pi-hole uses interface-dependent replies inside CNAME chains
     
        COMMAND EXECUTED:
        bash -c dig HTTPS pihole.mydomain.net @127.0.0.1 +short
     
        OUTPUT CAPTURED:
        pi.hole.
     
     ═══════════════════════════════════════════════════════════════════════════════
```

darkexplosiveqwx:
I added the BATS test for CNAME domains because that's functionality I'd like to have, however after a bit of trying, getting the right approach is quite involved because we can't do cache injections like for host-type RRTYPEs.

I'd like to hold this until after https://github.com/pi-hole/FTL/issues/2996 is addressed to make it more uniform and reduce the dnsmasq hooks necessary, especially for CNAME handling.

Waiting until after https://github.com/pi-hole/FTL/pull/3017 would probably be also a good idea at that point. One potential issue I might see is that TLS is only served on all TCP and not all UDP/QUIC ports. This record is specifically meant to improve HTTP/3 discovery, but also adding the remaining HTTP/2 ports shouldn't hurt.

https://github.com/pi-hole/FTL/pull/3012 got closed, so I'll assume to work with the current non-configurable DoT/DoQ ports.

DL6ER's suggestion to use runtime state is also more preferable and should be implemented.


Edit: https://github.com/pi-hole/FTL/pull/3012 is irrelevant here, as this is for regular HTTPS, not encrypted DNS. It is only relevant for the _dns.resolver.arpa SVCB record, which is not added here.

PR #2989, opened 2026-07-31
Add inbound DoT/DoH server for clients -> Pi-hole

DL6ER:

This adds a native *inbound* encrypted-DNS server, so Pi-hole terminates DoT and DoH from downstream clients itself instead of needing a separate proxy in front of it. It is the server-side counterpart to the outbound DoT/DoH client (encrypted upstreams) already on this line of work **and makes the DoH/DoT work complete in all directions.**

The implementation is fully cross-compatible. You may chose to
- continue as before (only use plaintext queries) even with the features enabled,
- switch the upstream connection to DoH/DoT and leave your clients on plaintext,
- use DoT/DoH from your clients connecting to your Pi-hole and keep an unencrypted upstream connection (maybe a local `unbound`), or
- or use encryption in and out everywhere.

Everything will work without the user having to tweak any configuration (well, you need to specify encrypted upstreams obviously if you want to go down this path).

The next step after this PR is adding full DoQ-support but this will happen in a dedicated PR and will be much smaller.

### DoH

DoH is served *natively* in the TLS frontend. I initially reused CivetWeb's request handling on `/dns-query`, but it turned out to be really slow, adding tens of milliseconds per query even on a fast machine. So the frontend now parses `/dns-query` itself with a minimal, purpose-built HTTP implementation and resolves through the same path as DoT. It accepts both standardized methods,
- a `POST` of `application/dns-message` and
- a base64url `GET ?dns=`,

over HTTP/1.1, HTTP/2 and HTTP/3. DoH now resolves in ~2 ms/query. Over HTTP/3 the resolve runs on a small worker pool off the QUIC event loop, so one slow resolve cannot stall the other connections sharing that loop.

The HTTP surface is deliberately small and bounded: requests are read non-blocking against an absolute deadline (so a slow-drip client cannot pin a handler thread), `Transfer-Encoding` is refused, bodies are size-capped, connections are limited per source, a `405` carries `Allow`, and `Expect: 100-continue` is honored. The base client PROXY v2 header CivetWeb needs for the UI and API stays; only the DoH destination-address extension it used to need is gone, as the terminator now learns the connected address directly and injects it as the private EDNS option, so `pi.hole` localization over DoH is unchanged.

### DoT

DoT is written as a single-threaded, event-driven `poll()` reactor: each connection is a non-blocking state machine (TLS handshake -> read query -> resolve over a non-blocking loopback socket -> write answer -> keep-alive) and one thread multiplexes them all. This is more robust than a multi-threaded thread-per-connection design, as a flood costs a small state record rather than a thread stack, and a slow or stalled resolve cannot block other connections by design.

It is also the shape DoQ (DNS-over-QUIC) will need, so the TLS I/O sits behind a few small helpers a QUIC transport can later reuse. Each connection carries an absolute per-query deadline plus an absolute lifetime cap (both swept on every wake-up) so slow-loris cannot pin a slot.

### Security aspects

Both paths hand the decrypted query to dnsmasq over a loopback TCP connection and carry the real downstream client in a Pi-hole-private EDNS option. The option is signed with a cryptographic per-query HMAC-SHA256 signature. A `root` (or other system-level) process able to sniff the loopback hop cannot learn the key and can neither forge nor usefully replay a client attribution; it is trusted only when the query source is loopback, kept out of the ECS field, and neutralized in place before it can be forwarded upstream. This preserves per-client behavior (groups, rate limits, logging) with *no dnsmasq patch at all*.

Answers are padded to a 468-octet boundary ([RFC 8467](https://www.rfc-editor.org/rfc/rfc8467.html)) when the client's query asked for padding, section 1 says:

>   Padding DNS messages is useful only when transport is encrypted using
>   protocols such as DNS over Transport Layer Security [[RFC7858](https://www.rfc-editor.org/rfc/rfc7858)], DNS
>   over Datagram Transport Layer Security [[RFC8094](https://www.rfc-editor.org/rfc/rfc8094)], or other encrypted
>   DNS transports specified in the future.

Block padding closes a specific crypto-weakness where the query may be guessed from the exact length of the datagramms. It is used in all major encryption methods. TLS uses AES which does NOT do it by default as transort encryption tries to stay as small as possible, applications have to add this themselves. 

### New config options
`dns.doh` and `dns.dot` (both default on). Can be switched off if not needed, saving around 40 KB of RAM memory and leaving the relevant ports untouched.

Make sure a TLS `webserver.port` is configured so a certificate exists. If you are using a self-signed certificate, it may be necessary to add the certificate to your device's certificate storage for it to accept a connection to your Pi-hole.

DoH over HTTP/1.1, HTTP/2 and HTTP/3 tests:

``` bash
# Build the raw wire-format DNS query the POST body needs. `query.bin` is not a
# repo file; this helper (shipped in the PR, used by the bats tests) emits one:
python3 test/dotdoh_query.py emit example.com query.bin

for v in --http1.1 --http2 --http3-only; do
  curl $v -sS --cacert /etc/pihole/tls.pem --resolve pi.hole:443:127.0.0.1 \
       -H 'Content-Type: application/dns-message' \
       --data-binary @query.bin https://pi.hole/dns-query -o answer.bin \
       -w "negotiated HTTP %{http_version}\n"
done

# Check an answer (optional): prints the resolved IP or fails on a bad reply.
python3 test/dotdoh_query.py check answer.bin <expected-ip>
```

DoT (`kdig` from `knot-dnsutils`):

``` bash
kdig -d @<pi-hole-ip> +tls-ca=/etc/pihole/tls.pem +tls-host=pi.hole example.com
```

(you may need to download the CA to your test machine if it not the same as your Pi-hole)

Then confirm the query is attributed to the *real* client, not to loopback: The query log should show the real address. A non-encrypted `GET http://<ip>/dns-query` must return `426`, a wrong method `405`, a bad media type `415`.

New tests on the CI:
- `test/dotdoh_server.bats` covers DoH `POST` and `GET`, DoH over HTTP/2 and HTTP/3 (version-asserted, with a real-client attribution check), DoT, IPv6 DoT, keep-alive, survival of a malformed frame, the `426`/`405`/`415`/`400` error paths, `Cache-Control`, and real-client attribution.
- `test/dotdoh_regression.c` (built with `-DBUILD_DOTDOH_REGRESSION=ON`) unit-tests the framing, the source-address filter (serves loopback/local, refuses public in every non-`LISTEN_ALL` mode, and restricts to the configured interface under `LISTEN_SINGLE`), and the EDNS padding (query and the 468-octet response case).

---

---

## Follow-ups

- [x] #2996 - make the `pi.hole` address selection for multi-homed hosts consistent and subnet-aware across plain and encrypted DNS (deferred here; the encrypted-path loopback -> NODATA stopgap is included here)

## Checklist:


darkexplosiveqwx:
The `pi.hole` special DNS record replies with the loopback IPs when queries via  encrypted DNS, even from non-loopback origins.
```lisp
$ kdig @192.168.178.138 -p 53335 pi.hole
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 62165
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; EDE: 29 (Synthesized): 'synthesized'

;; QUESTION SECTION:
;; pi.hole.            		IN	A

;; ANSWER SECTION:
pi.hole.            	0	IN	A	192.168.178.138

;; Received 69 B
;; Time 2026-08-01 09:49:03 CEST
;; From 192.168.178.138@53335(UDP) in 0.4 ms

$ kdig @192.168.178.138 +tls pi.hole
;; TLS session (TLS1.3)-(ECDHE-X25519)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 53662
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; EDE: 29 (Synthesized): 'synthesized'
;; PADDING: 395 B

;; QUESTION SECTION:
;; pi.hole.            		IN	A

;; ANSWER SECTION:
pi.hole.            	0	IN	A	127.0.0.1

;; Received 468 B
;; Time 2026-08-01 09:49:08 CEST
;; From 192.168.178.138@853(TLS) in 48.3 ms

$ kdig @192.168.178.138 +https -p 8086 pi.hole
;; TLS session (TLS1.3)-(HYBRID-X25519-MLKEM768)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; HTTP session (HTTP/2-POST)-(192.168.178.138/dns-query)-(status: 200)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 0
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; EDE: 29 (Synthesized): 'synthesized'
;; PADDING: 395 B

;; QUESTION SECTION:
;; pi.hole.            		IN	A

;; ANSWER SECTION:
pi.hole.            	0	IN	A	127.0.0.1

;; Received 468 B
;; Time 2026-08-01 09:49:22 CEST
;; From 192.168.178.138@8086(HTTPS) in 46.6 ms

$ kdig @192.168.178.138 +https-get -p 8086 pi.hole
;; TLS session (TLS1.3)-(HYBRID-X25519-MLKEM768)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; HTTP session (HTTP/2-GET)-(192.168.178.138/dns-query)-(status: 200)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 0
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; EDE: 29 (Synthesized): 'synthesized'
;; PADDING: 395 B

;; QUESTION SECTION:
;; pi.hole.            		IN	A

;; ANSWER SECTION:
pi.hole.            	0	IN	A	127.0.0.1

;; Received 468 B
;; Time 2026-08-01 09:55:48 CEST
;; From 192.168.178.138@8086(HTTPS) in 49.9 ms

$ sudo ss -tulnp | rg FTL
udp   UNCONN 0      0                                0.0.0.0:53335      0.0.0.0:*    users:(("pihole-FTL",pid=16747,fd=22))    
udp   UNCONN 0      0                                   [::]:53335         [::]:*    users:(("pihole-FTL",pid=16747,fd=24))    
udp   UNCONN 0      0                                      *:8086             *:*    users:(("pihole-FTL",pid=16747,fd=44))    
tcp   LISTEN 0      128                              0.0.0.0:853        0.0.0.0:*    users:(("pihole-FTL",pid=16747,fd=37))    
tcp   LISTEN 0      32                               0.0.0.0:53335      0.0.0.0:*    users:(("pihole-FTL",pid=16747,fd=23))    
tcp   LISTEN 0      200                            127.0.0.1:41469      0.0.0.0:*    users:(("pihole-FTL",pid=16747,fd=40))    
tcp   LISTEN 0      128                                 [::]:853           [::]:*    users:(("pihole-FTL",pid=16747,fd=42))    
tcp   LISTEN 0      32                                  [::]:53335         [::]:*    users:(("pihole-FTL",pid=16747,fd=25))    
tcp   LISTEN 0      200                                    *:8085             *:*    users:(("pihole-FTL",pid=16747,fd=39))    
tcp   LISTEN 0      128                                    *:8086             *:*    users:(("pihole-FTL",pid=16747,fd=41))    

```

darkexplosiveqwx:
After your force-push, `pi.hole` gets correctly localized:
```lisp
$ kdig @192.168.178.138 -p 53335 pi.hole
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 4979
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; EDE: 29 (Synthesized): 'synthesized'

;; QUESTION SECTION:
;; pi.hole.            		IN	A

;; ANSWER SECTION:
pi.hole.            	0	IN	A	192.168.178.138

;; Received 69 B
;; Time 2026-08-01 14:52:34 CEST
;; From 192.168.178.138@53335(UDP) in 0.8 ms
$ kdig @192.168.178.138 +tls pi.hole
;; TLS session (TLS1.3)-(ECDHE-X25519)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 46697
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; EDE: 29 (Synthesized): 'synthesized'
;; PADDING: 395 B

;; QUESTION SECTION:
;; pi.hole.            		IN	A

;; ANSWER SECTION:
pi.hole.            	0	IN	A	192.168.178.138

;; Received 468 B
;; Time 2026-08-01 14:52:38 CEST
;; From 192.168.178.138@853(TLS) in 48.4 ms
$ kdig @192.168.178.138 +https -p 8086 pi.hole
;; TLS session (TLS1.3)-(HYBRID-X25519-MLKEM768)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; HTTP session (HTTP/2-POST)-(192.168.178.138/dns-query)-(status: 200)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 0
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; EDE: 29 (Synthesized): 'synthesized'
;; PADDING: 395 B

;; QUESTION SECTION:
;; pi.hole.            		IN	A

;; ANSWER SECTION:
pi.hole.            	0	IN	A	192.168.178.138

;; Received 468 B
;; Time 2026-08-01 14:52:44 CEST
;; From 192.168.178.138@8086(HTTPS) in 47.8 ms
```

There is however still an issue with CNAMEs targeting `pi.hole`:
```lisp
$ kdig @192.168.178.138 -p 53335 pihole.mydomain.net
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 17126
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 2; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR

;; QUESTION SECTION:
;; pihole.mydomain.net.		IN	A

;; ANSWER SECTION:
pihole.mydomain.net.	0	IN	CNAME	pi.hole.
pi.hole.            	0	IN	A	192.168.178.138

;; Received 85 B
;; Time 2026-08-01 14:50:51 CEST
;; From 192.168.178.138@53335(UDP) in 2.6 ms
$ kdig @192.168.178.138 +tls pihole.mydomain.net
;; TLS session (TLS1.3)-(ECDHE-X25519)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 8470
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 2; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; PADDING: 379 B

;; QUESTION SECTION:
;; pihole.mydomain.net.		IN	A

;; ANSWER SECTION:
pihole.mydomain.net.	0	IN	CNAME	pi.hole.
pi.hole.            	0	IN	A	127.0.0.1

;; Received 468 B
;; Time 2026-08-01 14:50:58 CEST
;; From 192.168.178.138@853(TLS) in 48.6 ms
$ kdig @192.168.178.138 +https -p 8086 pihole.mydomain.net
;; TLS session (TLS1.3)-(HYBRID-X25519-MLKEM768)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; HTTP session (HTTP/2-POST)-(192.168.178.138/dns-query)-(status: 200)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 0
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 2; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; PADDING: 379 B

;; QUESTION SECTION:
;; pihole.mydomain.net.		IN	A

;; ANSWER SECTION:
pihole.mydomain.net.	0	IN	CNAME	pi.hole.
pi.hole.            	0	IN	A	127.0.0.1

;; Received 468 B
;; Time 2026-08-01 14:51:01 CEST
;; From 192.168.178.138@8086(HTTPS) in 50.4 ms
```

DL6ER:
Confirmed and fixed in `570321e63`. A direct `pi.hole` query was already localised to the address the client connected to, but a CNAME whose target is `pi.hole` was not: the CNAME target is answered from the shared `pi.hole` cache record, a different code path than the direct reply, which never saw the connected-address hint.

We now carry that hint through the cache path as well, so `pihole.mydomain.net` -> `pi.hole` resolves to the address the client reached over DoT/DoH, matching both the direct query and plain DNS. Two related things tightened along the way: an explicit `dns.reply.host.force4`/`force6` still takes precedence, i.e., a forced host reply is *not* overridden for encrypted clients (the connected address only replaces the interface-derived default), and the hint is now tied to the query it belongs to so it cannot leak across queries via the upstream-reply CNAME path.

New end-to-end tests cover both DoT and DoH. Another spin on your setup would be welcome.

darkexplosiveqwx:
<details><summary>My originally reported issue is fixed</summary>

```lisp
$ kdig @192.168.100.11 -p 53335 pihole.mydomain.net AAAA
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 46636
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 2; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR

;; QUESTION SECTION:
;; pihole.mydomain.net.		IN	AAAA

;; ANSWER SECTION:
pihole.mydomain.net.	0	IN	CNAME	pi.hole.
pi.hole.            	0	IN	AAAA	2001:db8:100::11

;; Received 97 B
;; Time 2026-08-02 09:44:44 CEST
;; From 192.168.100.11@53335(UDP) in 2.5 ms
$ kdig @192.168.178.138 -p 53335 pihole.mydomain.net
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 56627
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 2; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR

;; QUESTION SECTION:
;; pihole.mydomain.net.		IN	A

;; ANSWER SECTION:
pihole.mydomain.net.	0	IN	CNAME	pi.hole.
pi.hole.            	0	IN	A	192.168.178.138

;; Received 85 B
;; Time 2026-08-02 09:48:27 CEST
;; From 192.168.178.138@53335(UDP) in 1.5 ms
$ kdig @192.168.178.138 +tls pihole.mydomain.net
;; TLS session (TLS1.3)-(ECDHE-X25519)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 61008
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 2; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; PADDING: 379 B

;; QUESTION SECTION:
;; pihole.mydomain.net.		IN	A

;; ANSWER SECTION:
pihole.mydomain.net.	0	IN	CNAME	pi.hole.
pi.hole.            	0	IN	A	192.168.178.138

;; Received 468 B
;; Time 2026-08-02 09:48:32 CEST
;; From 192.168.178.138@853(TLS) in 47.8 ms
$ kdig @192.168.178.138 +https -p 8086 pihole.mydomain.net
;; TLS session (TLS1.3)-(HYBRID-X25519-MLKEM768)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; HTTP session (HTTP/2-POST)-(192.168.178.138/dns-query)-(status: 200)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 0
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 2; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; PADDING: 379 B

;; QUESTION SECTION:
;; pihole.mydomain.net.		IN	A

;; ANSWER SECTION:
pihole.mydomain.net.	0	IN	CNAME	pi.hole.
pi.hole.            	0	IN	A	192.168.178.138

;; Received 468 B
;; Time 2026-08-02 09:48:41 CEST
;; From 192.168.178.138@8086(HTTPS) in 49.0 ms
$ kdig @192.168.178.138 +https-get -p 8086 pihole.mydomain.net
;; TLS session (TLS1.3)-(HYBRID-X25519-MLKEM768)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; HTTP session (HTTP/2-GET)-(192.168.178.138/dns-query)-(status: 200)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 0
;; Flags: qr aa rd ra; QUERY: 1; ANSWER: 2; AUTHORITY: 0; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR
;; PADDING: 379 B

;; QUESTION SECTION:
;; pihole.mydomain.net.		IN	A

;; ANSWER SECTION:
pihole.mydomain.net.	0	IN	CNAME	pi.hole.
pi.hole.            	0	IN	A	192.168.178.138

;; Received 468 B
;; Time 2026-08-02 09:48:46 CEST
;; From 192.168.178.138@8086(HTTPS) in 50.5 ms
$ pihole-FTL -vv | head -n7
****************************** FTL **********************************
Version:         v6.7-83-g570321e6
Branch:          new/dotdoh-server
Commit:          570321e6 (2026-08-01 22:18:37 +0200)
Architecture:    x86_64 (compiled locally)
Compiler:        clang version 22.1.8 (Fedora 22.1.8-4.fc44)
GLIBC version:   2.43
```

</details>


There is some more unexpected behaviour of the `pi.hole` record.
Some are already unexpected over plain DNS (even without this PR), one behaviour when encrypted is argubly preferable and one is undoubtably unwantedv behaviour when encrypted.
Most importantly, the record return different things when queried over plain DNS vs. encrypted DNS.


Let's create a dummy interface with multiple IP addresses to better showcase it:

```sh
sudo modprobe dummy
sudo ip link add dummy0 type dummy
sudo ip link set dummy0 up
# Host-only IPv4 addresses (/32)
sudo ip addr add 192.168.100.10/32 dev dummy0
sudo ip addr add 192.168.100.11/32 dev dummy0
# Shared IPv4 subnet (/30) - different network
sudo ip addr add 192.168.101.1/30 dev dummy0
sudo ip addr add 192.168.101.2/30 dev dummy0
# Host-only IPv6 addresses (/128)
sudo ip -6 addr add 2001:db8:100::10/128 dev dummy0
sudo ip -6 addr add 2001:db8:100::11/128 dev dummy0
# Shared IPv6 subnet (/100) - different prefix
sudo ip -6 addr add 2001:db8:101::1/100 dev dummy0
sudo ip -6 addr add 2001:db8:101::2/100 dev dummy0
```

```sh
$ ip a show dummy0
6: dummy0: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ether 32:67:dc:30:c6:b4 brd ff:ff:ff:ff:ff:ff
    inet 192.168.100.10/32 scope global dummy0
       valid_lft forever preferred_lft forever
    inet 192.168.100.11/32 scope global dummy0
       valid_lft forever preferred_lft forever
    inet 192.168.101.1/30 scope global dummy0
       valid_lft forever preferred_lft forever
    inet 192.168.101.2/30 scope global secondary dummy0
       valid_lft forever preferred_lft forever
    inet6 2001:db8:101::2/100 scope global 
       valid_lft forever preferred_lft forever
    inet6 2001:db8:101::1/100 scope global 
       valid_lft forever preferred_lft forever
    inet6 2001:db8:100::11/128 scope global 
       valid_lft forever preferred_lft forever
    inet6 2001:db8:100::10/128 scope global 
       valid_lft forever preferred_lft forever
    inet6 fe80::3067:dcff:fe30:c6b4/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever
```

```sh
IPS=(
  192.168.100.10
  192.168.100.11
  192.168.101.1
  192.168.101.2
  2001:db8:101::2
  2001:db8:101::1
  2001:db8:100::11
  2001:db8:100::10
)
```

When queried via regular DNS:
```bash
for ip in "${IPS[@]}"; do
    echo "=== Querying $ip ==="
    kdig @"$ip" -p 53335 pi.hole A +short
    echo
done

for ip in "${IPS[@]}"; do
    echo "=== Querying $ip ==="
    kdig @"$ip" -p 53335 pi.hole AAAA +short
    echo
done
```

They all get answered with the firstmost IPv4/IPv6 address in enumeration, no matter if the source is from IPv4 or IPv6:
```sh
=== Querying 192.168.100.10 ===
192.168.100.10

=== Querying 192.168.100.11 ===
192.168.100.10

=== Querying 192.168.101.1 ===
192.168.100.10

=== Querying 192.168.101.2 ===
192.168.100.10

=== Querying 2001:db8:101::2 ===
192.168.100.10

=== Querying 2001:db8:101::1 ===
192.168.100.10

=== Querying 2001:db8:100::11 ===
192.168.100.10

=== Querying 2001:db8:100::10 ===
192.168.100.10

=== Querying 192.168.100.10 ===
2001:db8:101::2

=== Querying 192.168.100.11 ===
2001:db8:101::2

=== Querying 192.168.101.1 ===
2001:db8:101::2

=== Querying 192.168.101.2 ===
2001:db8:101::2

=== Querying 2001:db8:101::2 ===
2001:db8:101::2

=== Querying 2001:db8:101::1 ===
2001:db8:101::2

=== Querying 2001:db8:100::11 ===
2001:db8:101::2

=== Querying 2001:db8:100::10 ===
2001:db8:101::2
```

When queried via DNS over TLS:
```bash
for ip in "${IPS[@]}"; do
    echo "=== Querying $ip ==="
    kdig @"$ip" +tls pi.hole A +short
    echo
done

for ip in "${IPS[@]}"; do
    echo "=== Querying $ip ==="
    kdig @"$ip" +tls pi.hole AAAA +short
    echo
done
```


If the requested address family matches the one queried over, they get answered with the exact IP address the query arrived from. If the address family mismatches, then it replies with the loopback address:
```sh
=== Querying 192.168.100.10 ===
192.168.100.10

=== Querying 192.168.100.11 ===
192.168.100.11

=== Querying 192.168.101.1 ===
192.168.101.1

=== Querying 192.168.101.2 ===
192.168.101.2

=== Querying 2001:db8:101::2 ===
127.0.0.1

=== Querying 2001:db8:101::1 ===
127.0.0.1

=== Querying 2001:db8:100::11 ===
127.0.0.1

=== Querying 2001:db8:100::10 ===
127.0.0.1

=== Querying 192.168.100.10 ===
::1

=== Querying 192.168.100.11 ===
::1

=== Querying 192.168.101.1 ===
::1

=== Querying 192.168.101.2 ===
::1

=== Querying 2001:db8:101::2 ===
2001:db8:101::2

=== Querying 2001:db8:101::1 ===
2001:db8:101::1

=== Querying 2001:db8:100::11 ===
2001:db8:100::11

=== Querying 2001:db8:100::10 ===
2001:db8:100::10
```


There are a few options we can go about here, applying both to plain DNS and encrypted DNS: 

- return only the firstmost address in enumeration (current plain behavior; simple, fast, predictable-ish (magic decision), may return an address unusable for the client)
- return always all addresses of that interface (simple, may return an address unusable for the client, in addition to usable)
- return only the exact address the client connected to (only possible if family matches, address is known to be usable by the client)
- return all addresses matching the client's subnet (only possible if family matches, most likely to be usable by the client)
- return all addresses of the same scope (IPv6-only; meaning all ULA, LL, GUA based on the address connected to; argubly just a broader subnet match)

Limitations:

- cross-family queries will always be a problem because we have fundamentally too little information about the client (we may do a database lookup, but I dont think it's worth it because it is limited to same-segment if we want cross-family information and not every client exists with that information)
- returning the loopback address is plain wrong
- returning no cross-family IP is also a bad option because it limits discoverability and no client expects to need to query DNS over both address families.

My evaluation:

I think matching the client's subnet would be the best option, because it returns the addresses most likely to be usable. This however doesn't solve the issue for cross-family queries and would create an asymmetry.
For cross-family queries I actually prefer returning all addresses is better than returing a magic one. It is less arbitrary and client have multiple IPs to fallback to if one is not reachable. There is still the option of leaking addresses, but this is also the case for the magic-address one if the "wrong one" happens to be chosen. Leakage is also less an issue as one might think, as it only applies to the singular interface the client's query arrived from and not to any other interfaces. If we focus on consitancy for same-family and cross-family queries I would also preferr all addresses over magic address.
The advanced static and non-localized `dns.reply.host.*` options will continue working as before. My usecase of `pi.hole` however has always been for multiple interfaces with dynamic adresses each.

Note that since all those suggestions also affect the `pi.hole` record via non-encrypted transport they are better resolved in a seperate PR, but they are still relevant to understand my findings and that's why they are included here.

DL6ER:
You are right that returning the loopback in the cross-family case is unwanted. I have pushed a small mitigation (2c370df): a `pi.hole`/`<hostname>` query for a family we did not receive a connected address for (e.g., AAAA arriving over an IPv4 transport) now answers NODATA instead of `127.0.0.1`/`::1`. A forced fixed reply (`dns.reply.host.force4`/`force6`) still wins and stays positive.

The broader question of which address(es) `pi.hole` should return for a multi-homed host affects plain DNS as well, so I have split it out into #2996 to resolve separately, as you suggested.

DL6ER:
Follow-up: a few performance fixes to the inbound DoT/DoH server, measured in the `ftl-build` container (sequential queries, local answer - relative numbers only).

- **No more `fork()` per query.** We keep the loopback connection to our own DNS listener open and reuse it instead of opening one per query (dnsmasq forks per TCP connection). DoT went from ~675 ms/q and collapsing under load to *2.97 ms/q* sustained; plain UDP unchanged (~1.9 ms/q).
- **Nagle off on the webserver (`tcp_nodelay`).** CivetWeb left it off, stalling small TLS responses ~40 ms on the client's delayed ACK. A well-behaved DoH client dropped from *45.5 -> 14.0 ms/q*; the rest is inherent HTTPS cost (DoT is ~3 ms). Helps the UI/API too.
- Plus: one shared helper for the loopback handoff (also saves ~192 KiB/DoT-conn and the per-worker DoH buffers), pooled DoT buffers, and hardened EDNS attribution.

Commits: `5738ae21d`, `570da4c2c`, `5fb1dbc1a`, `ae7a00d3e`.

DL6ER:
More follow-up on the inbound DoT/DoH server - latency, and getting DoH off CivetWeb.

We stop forking a dnsmasq TCP child per query (keep the loopback DNS connection open and reuse it) and disable Nagle on the terminator's client sockets - a small TLS response otherwise stalls ~40 ms on the client's delayed ACK.

I initially planned to leverage CivetWeb's HTTP parsing for DoH, but it turned out to be really slow, adding sometimes 40 ms and more even on a fast machine. So we replaced it with a minimal HTTP implementation in the front terminator that serves `/dns-query` natively over HTTP/1.1, HTTP/2 and HTTP/3, both GET and POST (RFC 8484) - low attack surface, with countermeasures (per-request deadlines, `Transfer-Encoding` refused, bounded bodies, source-filtered). HTTP/3 resolves off the QUIC event loop on a small worker pool, so one slow resolve cannot stall other connections. DoH now resolves in ~2.0 ms/query.

With DoH no longer on CivetWeb, the PROXY v2 destination-address extension it needed is gone - the terminator learns the connected address directly and injects it as the private EDNS option, so `pi.hole` localisation over DoH is unchanged; the base client PROXY v2 stays for the UI and API.

DL6ER:
Squashed this down from 28 commits to 4 logical ones (server engine, dnsmasq attribution/localisation, CivetWeb PROXY v2, native DoH terminator) - IT IS 1:1 the same code, just easier to review. That also folds in automatic code review comments e.g. from CodeQL

yubiuser:
Tried DoH with

`curl -k --resolve nanopi.lan:433:127.0.0.1 -v --cacert ./tls_nanopi.pem https://nanopi.lan/dns-query?dns=AAABAAABAAAAAAAAA3d3dwtoaXNvbHV0aW9ucwNjb20AAAEAAQ`

and it showed the real client



Maybe we should amend the web interace to indicate DoH/DoT?

DL6ER:
> Maybe we should amend the web interace to indicate DoH/DoT?

Possible and I thought about it before but it comes at a price: As the Query log now fully uses SQLite3 in an in-memory table, we need to change the database `query_storage` table format appending another column. Is this something we want to do? It won't be a breaking change as we only amend, but it will be ~ 2 additional bytes per row.

Sketching this, I'd say it'd add about ~ 300 lines on top of this PR as we need to touch database migration code, API changes (printing the new column, make filtering by the new column possible), new code that actually sets this new property and saves it in the database + test changes and quite a few more private EDNS extensions as the embedded `dnsmasq` doesn't actually know which way a query used, so we'd have to invent another new EDNS option to store this fact, string helpers and a new enum.

yubiuser:
We can wait until a user requests it 😉

.devcontainer/devcontainer.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "FTL x86_64 Build Env",
-  "image": "ghcr.io/pi-hole/ftl-build:nightly",
```
yubiuser, line 3: Why did you change this? That we, we need to remember to update the devcontainer for each bump of the build-container.

src/config/config.c
```diff
@@ -611,6 +611,20 @@ void initConfig(struct config *conf)
...
 
+	conf->dns.doh.k = "dns.doh";
+	conf->dns.doh.h = "Enable the inbound DNS-over-HTTPS (DoH) server. When enabled, FTL answers DoH queries (RFC 8484: an HTTP POST with an application/dns-message body, or a GET with a base64url-encoded ?dns= parameter, to /dns-query) on the Pi-hole HTTPS webserver, letting downstream clients use this Pi-hole directly as their encrypted resolver. This requires a TLS-enabled webserver.port (e.g. \"443s\") with a valid certificate, see config option webserver.port.";
+	conf->dns.doh.t = CONF_BOOL;
+	conf->dns.doh.d.b = true;
+	conf->dns.doh.f = FLAG_RESTART_FTL;
+	conf->dns.doh.c = validate_stub;
+
+	conf->dns.dot.k = "dns.dot";
+	conf->dns.dot.h = "Enable the inbound DNS-over-TLS (DoT) server on port 853. When enabled, FTL terminates DoT connections directly (RFC 7858) so downstream clients can use this Pi-hole as their encrypted resolver. Requires a valid TLS certificate (the same one configured for the webserver).";
```
yubiuser, line 622: This needs to be documented. And a change to the docker repo: documentation +  example `compose.yaml`

Should this port be user changeable?

src/dotdoh/server.h
```diff
@@ -0,0 +1,77 @@
...
+*  (c) 2026 Pi-hole, LLC (https://pi-hole.net)
+*  Network-wide ad blocking via your own hardware.
+*
+*  FTL Engine
+*  Inbound DoT/DoH server
+*
+*  Public interface of the inbound encrypted-DNS server (serving downstream
+*  clients). Throughout this module "client" always means the downstream device
+*  that queries us - as in Pi-hole's client tables (clientsData/getClient) - never
+*  our own TLS-client role (that is the outbound "upstream" side). DoH reuses the
```
yubiuser: Is this still a valid comment after

> So the frontend now parses /dns-query itself with a minimal, purpose-built HTTP implementation and resolves through the same path as DoT.

yubiuser (commented):
Test DoT (with `kdig`) and DoH with `POST` method - both worked.

____

Do we expect any interference when a query already carries EDNS information (like ECS)?

.devcontainer/devcontainer.json
DL6ER: I think I needed a fresh image because of the latest additions and nightly is only built once per day. Seems I forgot reverting it which I will do.

DL6ER:
> Do we expect any interference when a query already carries EDNS information (like ECS)?

No harmful interference, by design. Injects preserve the client's OPTs. So everything like ECS, cookie, etc. is left intact. We just add. Parsing keeps them in separate fields, I think I noted this down in the source code *somewhere* but no blaming you didn't see it (and maybe I forgot). The consumption prioritizes out own fields, so `private_client` wins over ECS for the query's client identity. However, even with DoT/DoH, a client that also seds ECS is still attributed to the real device, not the subnet if `dns.EDNS0ECS` is `true. There is a dedicated neutralization step so our options's payload is fully zeroed out before `dnsmasq` forwards anything upstream. Real client IPs cannot leak this way. ECS-to-upsteam behavior of `dnsmasq` is not changed.

The implementations are deliberately orthogonal.

src/dotdoh/server.h
DL6ER: Yes? I think so?

src/config/config.c
DL6ER: You mean the `docs` repo, right? Just asking because there is so much going on here right now, I might have misunderstood the comment.

---

re port quote from RFC 7858:

> A DoT server **MUST** listen on TCP port 853 by default, *unless* it has a mutual agreement with its clients to use another port; changing it requries a configuration option in both client and server.

For DoQ (#3000 :partying_face:), RFC 9250 says exactly the same as port 853 is shared between DoT (TCP) and DoQ (UDP).

So ... do we want to add the port now? This will make the PR larger.
My preference is keeping this a follow-up (next to the DoH/DoT display on the Query Log) PR and do it consistently for DoT and (then) DoQ at the same time.

---

Additional information for me to pick up later now that I re-checked the RFCs:


For DoT, it boils down to both need to know the custom port.

For DoQ, RFC 9462 specifies `DDR/SVCB` with: `_dns.resolver.arpa` with `alpn=doq` and a `port=<whatever>` key. Or, again, manual configuration on both sides.

DL6ER:
Why this?...

> Base automatically changed from development to special/CI_development [6 minutes ago](https://github.com/pi-hole/FTL/pull/2989#event-29003367163)

yubiuser:
> Why this?...
> 
> > Base automatically changed from development to special/CI_development [6 minutes ago](https://github.com/pi-hole/FTL/pull/2989#event-29003367163)

I guess it's the new "stacked" feature.

DL6ER:
I cannot change it back... but you were right. Once I "unstacked", I was able to manually set it back.

src/config/config.c
yubiuser: > You mean the docs repo, right? Just asking because there is so much going on here right now, I might have misunderstood the comment.

Yes, the `doc` repo. I think the lastest changes deserve a whole section quite high, maybe as menu in https://docs.pi-hole.net/ftldns/. And likely here: https://docs.pi-hole.net/main/prerequisites/#ports

Plus https://github.com/pi-hole/docker-pi-hole/blob/dd91b4847d97f0aac68bfefd1c108ed0627e6c68/README.md?plain=1#L46

____

> My preference is keeping this a follow-up (next to the DoH/DoT display on the Query Log) PR and do it consistently for DoT and (then) DoQ at the same time.

Agreed.

src/dotdoh/server.h
yubiuser: I meant

>  DoH reuses the pi.hole HTTPS server (civetweb) for TLS termination and is served as a request handler; DoT has its own raw-TLS listener as it is not HTTP.

I thought we don't use `civetweb` for DoH anymore as it was slow and you added the 
>  minimal, purpose-built HTTP implementation 

This sounded like `civetweb` is not involved anymore in DoH handling

.devcontainer/devcontainer.json
DL6ER: Undone

yubiuser (dismissed):
I trust @DL6ER to open a `https://github.com/pi-hole/docs` PR soon and finish it before v7 is released.

src/dotdoh/server.h
DL6ER: @yubiuser - all comments should be fixed now. CivetWeb is not involved at all here any longer

yubiuser:
Can you please squash down again?

DL6ER:
done

yubiuser:
What is this?

DL6ER:
Good question, locally, it showed 4 commits. Maybe my local `development` branch is off, I will check

DL6ER:
done²

yubiuser (approved):

PR #2992, opened 2026-08-01
fix(dnsmasq): answer NODATA instead of forwarding for locally-known names

### LINKED ISSUE #2841: Local host overrides for public domain intermittently ignored, clients receive upstream NXDOMAIN instead of local answer
Edit: Removed DNSSEC ref from the title as was a red herring.

### Versions

- Pi-hole:  v6.4.1
- Web:  v6.5
- FTL:  v6.6

### Platform

- OS and version:  Debian-based Linux on multiple hosts

- Platform:  Native installs on multiple LAN hosts / VMs

### Expected behavior

A hostname configured as a local Pi-hole override should be answered locally every time, regardless of which LAN client asks the question.

If a local host record exists for a name under a public domain, Pi-hole should return the configured local A record instead of falling through to upstream/public recursion. In short, the override should actually override.

Related issues:
- [DNS returns NXDOMAIN on entries from dnsmasq](https://github.com/pi-hole/FTL/issues/2560)
- [Local addresses defined in dnsmasq config return NXDOMAIN](https://github.com/pi-hole/FTL/issues/2497)
- [Pi-Hole v6 not resolving names for LAN devices after upgrade](https://github.com/pi-hole/pi-hole/issues/6153)

### Actual behavior / bug

The same Pi-hole server returns different answers for the same locally configured hostname depending on which client asks.

For the same query name and same Pi-hole IP:
- querying from the Pi-hole host itself returns `NOERROR` with the configured local A record
- querying from another Linux LAN client returns `NXDOMAIN` with the public authoritative SOA for the parent domain

This should not happen for a local host override.

Important context: this setup worked until a few weeks ago, and the breakage appears to line up with recent Pi-hole v6-era behavior changes around local DNS / local-domain handling. Similar reports describe local DNS records or split-DNS forwarding working before upgrade and failing afterward:
- [Local DNS Records not working anymore after updating to v6](https://www.reddit.com/r/pihole/comments/1iu6iil/local_dns_records_not_working_anymore_after/)
- [Pihole not forwarding local domain after update to V6](https://discourse.pi-hole.net/t/pihole-not-forwarding-local-domain-after-update-to-v6/76113)
- [Local DNS records not resolving - Beta 6.0](https://discourse.pi-hole.net/t/local-dns-records-not-resolving/66521)

I initially suspected DNSSEC because disabling DNSSEC made some cases start working again, but the client-dependent behavior still reproduces even after:
- setting `dns.domain.name = ""`
- setting `dns.dnssec = false`
- setting `dns.localise = false`

### Steps to reproduce

1. Configure local host overrides in Pi-hole for names under a public domain, for example:
   - `host-a.example.test -> 10.0.10.10`
   - `node-1.lab.example.test -> 10.0.20.20`
2. Confirm the records exist in FTL config under `dns.hosts`.
3. From the Pi-hole host itself, query the Pi-hole LAN IP:
   - `dig @10.0.20.53 node-1.lab.example.test`
4. Observe `NOERROR` with the configured local A record.
5. From another Linux client on the LAN, query the same Pi-hole IP for the same name:
   - `dig @10.0.20.53 node-1.lab.example.test`
6. Observe `NXDOMAIN` with the public SOA for `example.test` instead of the local A record.

Optional additional attempts that did **not** resolve the issue:
1. Disable DNSSEC
2. Set `dns.domain.name` to an empty string
3. Set `dns.localise` to `false`
4. Restart `pihole-FTL`
5. Repeat the same tests
6. Observe that the same client-dependent mismatch can still occur

References:
- [Pi-hole FTL configuration](https://docs.pi-hole.net/ftldns/configfile/)
- [DNSSEC validation and local domains](https://discourse.pi-hole.net/t/dnssec-validation-and-local-domains/51186)
- [Understanding DNSSEC validation using Pi-hole's Query Log](https://pi-hole.net/blog/2021/12/12/understanding-dnssec-validation-using-pi-holes-query-log/)

## Debug Token

- URL:  _Please replace with your `pihole -d` upload token_

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Additional context

This was reproduced across multiple Pi-hole instances configured as LAN DNS servers.

The local records were present in `pihole-FTL --config` under `dns.hosts`, so this does not appear to be a missing-record problem. Pi-hole’s FTL config docs describe these as local host mappings managed by FTL:
- [Pi-hole FTL configuration](https://docs.pi-hole.net/ftldns/configfile/)

Behavior differed between clients:
- macOS with Firefox appeared to continue working
- Linux clients, including Home Assistant OS and another Linux machine, hit the failing path

The strongest evidence is that the same Pi-hole server returned different results for the same query at nearly the same time:

From the Pi-hole host:
```text
dig @10.0.20.53 node-1.lab.example.test
-> NOERROR
-> ANSWER SECTION: node-1.lab.example.test. 0 IN A 10.0.20.20
```

From another Linux client:
```text
dig @10.0.20.53 node-1.lab.example.test
-> NXDOMAIN
-> AUTHORITY SECTION: example.test. IN SOA <public authoritative SOA>
```

This makes it look like Pi-hole/FTL is varying behavior by client/query path, or incorrectly bypassing local host overrides for some clients. Recent v6 reports about local entries returning `NXDOMAIN` unexpectedly seem related.

vrkansagara:
## Summary

Fixes #2841 — local host overrides (`/etc/hosts`, DHCP leases, Pi-hole's `dns.hosts` / `custom.list`) placed under a publicly-signed parent domain were intermittently shadowed by an upstream `NXDOMAIN`, so the same query returned a different answer depending on which client asked. Root cause was diagnosed by @DL6ER in the issue thread as two related gaps in dnsmasq's local-name handling; this PR implements both fixes described there.

### 1. `answer_request()` in `src/dnsmasq/rfc1035.c`

Previously, if a client queried a locally-configured name for a record type with no local answer (e.g. `AAAA`/`HTTPS`/`SVCB` when only an `A` record is configured), the function gave up (`return 0`) and let the query fall through to upstream forwarding. Since the name only exists locally, upstream then correctly-but-misleadingly returns `NXDOMAIN` for the *type*, which callers/resolvers interpret as "domain doesn't exist" rather than "domain exists, wrong type". This explains the client-dependent behavior: `systemd-resolved` and other modern Linux resolvers routinely send `AAAA`/`HTTPS` alongside `A`, so they hit this path far more than clients that don't.

Fix: before giving up, check whether the queried name has *any* local record (`F_HOSTS`/`F_DHCP`/`F_CONFIG`). If so, answer `NODATA` (`NOERROR`, no records) ourselves instead of forwarding — this correctly tells the resolver "this name exists, just not for this type."

### 2. `process_reply()` in `src/dnsmasq/forward.c`

dnsmasq already had a safety net for this: if an upstream reply for a locally-known name comes back `NXDOMAIN`, it gets converted to `NODATA`. But that conversion was gated behind `!bogusanswer`, i.e. it was skipped whenever DNSSEC validation flagged the reply as bogus. For a local override under a publicly-signed domain, the upstream's signed `NXDOMAIN` proof will often fail validation *precisely because* it contradicts the locally-known answer — so the one case this safety net exists for is the one case it was disabled for.

Fix: moved the `NXDOMAIN`→`NODATA` conversion so it runs unconditionally, before the `bogusanswer` gate, and clears `bogusanswer` once it applies (so the later DNSSEC bogus-check doesn't turn the corrected `NODATA` back into `SERVFAIL`/`NXDOMAIN`). Local host records are authoritative for a domain's *existence* regardless of upstream DNSSEC status.

## Why this fixes the reported behavior

Both gaps independently caused the same symptom (client-dependent `NXDOMAIN` for a locally-overridden name):
- Gap 1 fires per-query-type, explaining why it depended on which record types a given client's stack queries for.
- Gap 2 fires once DNSSEC validation is involved, explaining why disabling DNSSEC "fixed" some but not all cases in the original report.

With both fixed, a local host override is always authoritative for its name's existence: unsupported record types get `NODATA` directly from FTL, and any upstream `NXDOMAIN` that does get forwarded (e.g. via the type-gap before this fix, in transitional states, or via other code paths) is corrected to `NODATA` regardless of DNSSEC bogus status.

## Test plan

- [x] Clean CMake build (`-DCMAKE_BUILD_TYPE=Debug`) compiles `src/dnsmasq/forward.c` and `src/dnsmasq/rfc1035.c` with no new warnings, and links `pihole-FTL` successfully.
- [ ] Manual reproduction per the issue's steps (local `A`-only override under a public domain, query for `AAAA`/`HTTPS` from a Linux client, confirm `NODATA` instead of `NXDOMAIN`).
- [ ] Manual reproduction with DNSSEC enabled, confirm `NODATA` survives instead of being overwritten by the bogus-answer check.

Fixes #2841.

### DIFF
--- a/src/dnsmasq/forward.c
+++ b/src/dnsmasq/forward.c
@@ -817,21 +817,31 @@ static size_t process_reply(struct dns_header *header, time_t now, struct server
       return resize_packet(header, n, pheader, plen);
     }
   
+  /* If we forwarded a query for a locally known name (because it was for
+     an unknown type) and the answer is NXDOMAIN, convert that to NODATA,
+     since we know that the domain exists, even if upstream doesn't. Local
+     host records (/etc/hosts, DHCP leases, Pi-hole's dns.hosts/custom.list)
+     are authoritative for the *existence* of the domain regardless of what
+     upstream DNSSEC validation says about it, so this must happen before
+     the bogusanswer gate below: otherwise a signed upstream NXDOMAIN proof
+     that fails validation (because it contradicts our local knowledge)
+     bypasses this safety net entirely and the client is handed the bogus
+     NXDOMAIN instead of the correct NODATA. */
+  if (!(header->hb3 & HB3_TC) &&
+      rcode == NXDOMAIN && extract_name(header, n, NULL, daemon->namebuff, EXTR_NAME_EXTRACT, 0) &&
+      (check_for_local_domain(daemon->namebuff, now) || lookup_domain(daemon->namebuff, F_CONFIG, NULL, NULL)))
+    {
+      header->hb3 |= HB3_AA;
+      SET_RCODE(header, NOERROR);
+      cache_secure = 0;
+      bogusanswer = 0;
+      rcode = NOERROR;
+    }
+
   if (header->hb3 & HB3_TC)
     log_query(F_UPSTREAM, NULL, NULL, "truncated", 0);
   else if (!bogusanswer || (header->hb4 & HB4_CD))
     {
-      if (rcode == NXDOMAIN && extract_name(header, n, NULL, daemon->namebuff, EXTR_NAME_EXTRACT, 0) &&
-	  (check_for_local_domain(daemon->namebuff, now) || lookup_domain(daemon->namebuff, F_CONFIG, NULL, NULL)))
-	{
-	  /* if we forwarded a query for a locally known name (because it was for 
-	     an unknown type) and the answer is NXDOMAIN, convert that to NODATA,
-	     since we know that the domain exists, even if upstream doesn't */
-	  header->hb3 |= HB3_AA;
-	  SET_RCODE(header, NOERROR);
-	  cache_secure = 0;
-	}
-      
       if (daemon->doctors && do_doctor(header, n, daemon->namebuff))
 	cache_secure = 0;
       
--- a/src/dnsmasq/rfc1035.c
+++ b/src/dnsmasq/rfc1035.c
@@ -2357,11 +2357,26 @@ size_t answer_request(struct dns_header *header, char *limit, size_t qlen,
 	    }
 	}
     }
-  
-  if (!ans)
-    return 0; /* failed to answer a question */
 
-  /* We found a negative record. See if we have an SOA record to 
+  if (!ans)
+    {
+      /* We have no answer for this specific query type, but the name is
+	 locally configured (F_HOSTS/F_DHCP/F_CONFIG), e.g. only an A record
+	 exists and the client asked for AAAA/HTTPS/SVCB. The domain does
+	 exist, just not for this type, so answer NODATA ourselves rather
+	 than forwarding: forwarding would let an upstream authoritative
+	 server return NXDOMAIN for a name that's actually a local override,
+	 which incorrectly makes the whole domain look non-existent. */
+      if (qclass == C_IN && cache_find_by_name(NULL, name, now, F_HOSTS | F_DHCP | F_CONFIG))
+	{
+	  ans = 1;
+	  sec_data = 0;
+	}
+      else
+	return 0; /* failed to answer a question */
+    }
+
+  /* We found a negative record. See if we have an SOA record to
      return in the AUTH section. 
      
      For FORWARD NEG records, the addr.rrdata.datalen field of the othewise


DL6ER:
Thanks for the effort, but we cannot take this.

We have a strict policy to not carry any change in `src/dnsmasq/` that deviates from upstream dnsmasq. That directory is a vendored copy of Simon Kelley's tree and we merge his commits into it as they are, so a fix living only in our copy is either lost on the next sync or turns every future merge into a conflict we have to resolve by hand. The more important reason is fragmentation: our users would run a resolver behaving differently from every other dnsmasq out there, which makes bug reports on either side hard to reason about. All fixes to this code have to go through the `dnsmasq-discuss` mailing list. Once they appear in dnsmasq master, we usually merge them rather quickly.

For this particular bug, that has already happened - I sent the patch to Simon and the list on 2026-05-16: https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2026q2/018476.html

I compared it against yours, and the two are close, which is not surprising as both implement what I described in #2841:

- `forward.c` is effectively identical. Both hoist the NXDOMAIN-to-NODATA conversion above the `bogusanswer` gate, add the `HB3_TC` guard and set `rcode = NOERROR`. Yours additionally clears `bogusanswer`, and that is a real difference: without clearing it, the DNSSEC block further down turns the corrected NOERROR into SERVFAIL, so the client gets SERVFAIL instead of NODATA. On the other hand, clearing it also drags the reply through `do_doctor()`/`extract_addresses()`, i.e., we would extract and cache the content of a reply we know did not validate. That trade-off is worth discussing, but on the list, not here.

- `rfc1035.c` implements the same idea differently. Mine iterates the cached records for the name and requires the matching entry to actually carry `F_HOSTS`/`F_DHCP`/`F_CONFIG`, clears `auth`, and emits `log_query(F_NEG | F_CONFIG | ...)`. Yours passes the source flags straight into `cache_find_by_name()` as the type mask and does not log. The latter matters for us in particular: `FTL_hook()` sits inside `log_query()`, so without that call the synthesized NODATA never reaches Pi-hole's query log and the query is shown without a reply.

So nothing here is wrong as such, it is simply not the route this code takes. If you want the `bogusanswer` part considered, please reply to my patch on `dnsmasq-discuss` - Simon has the final word on it. I am closing this PR.

Unrelated to the above, for future contributions: we branch off and target `development`, `master` only ever receives release merges.

PR #2993, opened 2026-08-01
fix(dnsmasq): restore INSECURE result for empty DS replies without NSEC records

### LINKED ISSUE #2942: FTL v6.7 DNSSEC PTR Regression
### Versions
- Core version is v6.4.3 (Latest: v6.4.3)
- Web version is v6.6 (Latest: v6.6)
- FTL version is v6.7 (Latest: v6.7)

### Platform
- OS and version:  Ubuntu 26.04 LTS (GNU/Linux 7.0.0-27-generic x86_64)
- Platform:  VM

### Expected behavior

PTR (reverse DNS) lookups for RFC1918 private address space (e.g. 192.168.0.0/24), forwarded via a dns.revServers conditional-forward rule to a local authoritative resolver, should resolve successfully with DNSSEC enabled — as they did on FTL v6.6.2 and earlier.

Since private-use reverse zones (168.192.in-addr.arpa, 10.in-addr.arpa, etc.) are never delegated/signed in the public DNSSEC root, the resolver should recognise this and treat the zone as insecure rather than attempting full chain validation, per RFC 8501-style handling. This is the behavior observed on older dnsmasq/FTL builds, evidenced by the log line:

```
Insecure reply received for DS 168.192.in-addr.arpa, assuming that's OK for a RFC-1918 address.
reply 168.192.in-addr.arpa is no DS
validation result is INSECURE
reply 192.168.0.22 is <<resolved hostname>>.lan
```

### Actual behavior / bug

On FTL v6.7 (embedded dnsmasq updated to v2.93 in [#2890](https://github.com/pi-hole/FTL/pull/2890)), the same PTR query against the same upstream/config returns SERVFAIL to the client. The query log shows the DNSSEC validator attempting to walk the DS chain up to 168.192.in-addr.arpa and then giving up outright, rather than recognizing the RFC1918 special case:

```
dnssec-query[DS] 168.192.in-addr.arpa to 1.1.1.1
validation 22.0.168.192.in-addr.arpa is ABANDONED
```

No "Insecure reply received for DS ... RFC-1918 address" message is ever logged post-upgrade — that code path/message appears to be missing or broken in dnsmasq v2.93 as embedded in FTL v6.7. This affects every RFC1918 reverse-zone PTR query when DNSSEC is enabled and a rev-server/dns.revServers conditional forward is configured — confirmed across 192.168.0.x addresses in general, not just one host.

Config (dns.revServers, dns.dnssec, dns.upstreams, dns.bogusPriv) was verified byte-identical between a working (FTL 6.6.2) and broken (FTL 6.7) instance via diff of pihole.toml. System clock/NTP sync and DNSSEC root trust anchor file were also verified identical and correct on both. The only variable that changed was the FTL/embedded-dnsmasq version.

Note: querying the upstream authoritative resolver (OPNsense/dnsmasq at 192.168.0.1) directly, bypassing Pi-hole entirely, returns the correct PTR answer instantly and correctly on both instances — confirming the upstream/network path is fine and the regression is isolated to FTL's own DNSSEC validation logic.

### Steps to reproduce

1. Set up a Pi-hole instance with dns.dnssec = true and a conditional forward (dns.revServers) pointing to a local authoritative resolver covering an RFC1918 subnet, e.g.:

`   dns.revServers = ["true,192.168.0.0/24,192.168.0.1,lan"]
`

2. Confirm the local authoritative resolver (e.g. OPNsense/dnsmasq/router) answers PTR queries for that range correctly and directly (e.g. dig `@192.168.0.1 22.0.168.192.in-addr.arpa PTR` — should return a valid answer).
3. On a Pi-hole running FTL v6.6.2 or earlier, run: `dig @<pihole-ip> 22.0.168.192.in-addr.arpa PTR` — observe a correct NOERROR response with the expected PTR record, and validation result is INSECURE in the query log.
4. Upgrade the same instance to FTL v6.7 (no other config changes).
5. Re-run: `dig @<pihole-ip> 22.0.168.192.in-addr.arpa PTR` — observe SERVFAIL, with validation ... is ABANDONED in the query log instead of the previous Insecure reply received for DS ... RFC-1918 address / validation result is INSECURE sequence.

## Debug Token

- URL:  https://tricorder.pi-hole.net/03xHA0Np/

## Screenshots
Demonstration of difference in response between PiHole versions:
Core version is v6.4.2 vs v6.4.3
Web version is v6.5.1 vs v6.6
FTL version is v6.6.2 vs v6.7

Image

Image

vrkansagara:
## Summary

Fixes #2942 — a regression in FTL v6.7's embedded dnsmasq (v2.93) that broke DNSSEC-validated PTR lookups for RFC-1918 private address ranges forwarded via `rev-server`/`dns.revServers` to a local authoritative resolver.

This PR is a straight cherry-pick of @DL6ER's own fix commit (`4c6704ee`), already pushed to the `fix/dnssec-empty-ds-rfc1918-2942` branch in this repo and confirmed working by the reporter in the issue thread. It is opened here as an upstreamable PR since that branch had no PR associated with it yet — original authorship (and `Signed-off-by`) is preserved via the cherry-pick.

### Why it's broken

Upstream dnsmasq commit `707bc2d25` ("Fix DNSSEC fail with CNAME replies to DS queries") removed this shortcut from `dnssec_validate_reply()`:

```c
/* Empty DS without NSECS */
if (qtype == T_DS)
  return STAT_INSECURE;
```

Many public resolvers serve the reverse zones of private address space (e.g. `168.192.in-addr.arpa`) as empty zones per RFC 6303, answering a `DS` query with a bare `NXDOMAIN`/`NODATA` that carries no `SOA`, `NSEC`, or `RRSIG`. Without the shortcut, `prove_non_existence()` returns `DNSSEC_FAIL_NONSEC`, and `zone_status()` is then asked about the very name whose `DS` is currently being resolved — returning `STAT_NEED_DS` for that same name. `dnssec_validate_ds()` propagates `STAT_NEED_DS`, the identical `DS` query is re-issued, the self-dependency is detected as a loop, and validation is abandoned (`STAT_ABANDONED` -> `SERVFAIL`). The `bogus-priv` RFC-1918 escape hatch in `dnssec_validate_ds()` never gets a chance to fire because it's gated on `rc == STAT_INSECURE`.

### The fix

Restores the shortcut in `src/dnsmasq/dnssec.c`, scoped to the primary query (`j == 0`) so it doesn't regress the CNAME-to-DS handling that `707bc2d25` was itself fixing (which operates on CNAME targets, `j > 0`, via `prim_ok`):

```c
if (qtype == T_DS && j == 0)
  return STAT_INSECURE;
```

An empty `DS` reply with no non-existence proof for the queried name itself is again reported as `STAT_INSECURE`, letting the `bogus-priv`/domain-specific-server heuristics decide instead of looping into `STAT_ABANDONED`.

## Test plan

- [x] Cherry-picked cleanly onto current `master`, no conflicts.
- [x] `src/dnsmasq/dnssec.c` and the `dnsmasq` build target compile with zero new warnings.
- [x] Reporter (`jelliuk`) already confirmed on the issue thread that checking out the original `fix/dnssec-empty-ds-rfc1918-2942` branch (same commit) resolves the SERVFAIL regression in their environment.

Fixes #2942.

### DIFF
--- a/src/dnsmasq/dnssec.c
+++ b/src/dnsmasq/dnssec.c
@@ -2304,7 +2304,20 @@ int dnssec_validate_reply(time_t now, struct dns_header *header, size_t plen, ch
 	  {
 	    if (rc_nsec & DNSSEC_FAIL_WORK)
 	      return STAT_ABANDONED;
-	    
+
+	    /* Empty DS reply with no NSEC/NSEC3 proof. This is what many
+	       public resolvers return for the reverse zones of RFC-1918
+	       private address space, which they serve as empty zones (RFC 6303),
+	       ie a bare NXDOMAIN/NODATA with no records to prove non-existence.
+	       Treat the answer to the DS query itself (j == 0) as insecure so
+	       the caller's bogus-priv/domain-specific-server heuristics can act
+	       on it. Without this, zone_status() below returns STAT_NEED_DS for
+	       the very name we're validating the DS for, and the resulting
+	       self-dependent DS query is detected as a loop and abandoned.
+	       CNAME targets (j > 0) are handled via prim_ok above. */
+	    if (qtype == T_DS && j == 0)
+	      return STAT_INSECURE;
+
 	    if ((rc_nsec & (DNSSEC_FAIL_NONSEC | DNSSEC_FAIL_NSEC3_ITERS)) &&
 		!STAT_ISEQUAL((rc = zone_status(name, qclass, keyname, now)), STAT_SECURE))
 	      {


DL6ER:
Same answer as on #2992: we do not carry any change in `src/dnsmasq/` that deviates from upstream dnsmasq. All fixes to that code go through the `dnsmasq-discuss` mailing list first, and we merge them once they show up in dnsmasq master.

There is also nothing to compare here - this is my own commit `4c6704ee` verbatim. The branch it sits on exists for one purpose, namely letting the reporter confirm the diagnosis, which @jelliuk did. It is not a staging area, and the missing PR is not an oversight: the patch has been with Simon and on the list since 2026-07-12, see https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2026q3/018512.html

I asked for review there precisely because it touches the DS/CNAME logic that `707bc2d25` reworked, and I am not going to push a DNSSEC validation change out to all our users ahead of that review. Once it lands upstream, it comes to us rather quickly.

Closing this one as well. A general request: please do not open further PRs patching `src/dnsmasq/` - the answer will be the same regardless of the fix.

PR #2997, opened 2026-08-02
Introduce agents.md

PromoFaux:

Adds an `AGENTS.md` file following the [agents.md](https://agents.md) standard: a predictable place for AI coding agents to find the repository layout, build/test commands and contribution requirements (DCO sign-off, branch targeting, code style) without them having to rediscover it every session. It is just standard markdown, and no tooling or workflow changes come with it.

---

---


AGENTS.md
```diff
@@ -0,0 +1,57 @@
+# AGENTS.md
+
+## Project overview
+
+FTL (`pihole-FTL`) is Pi-hole's DNS/DHCP engine. It embeds a modified `dnsmasq`, adds a query database, statistics, and a REST API, and serves the web interface via an embedded web server. It is written in C, built with CMake, and targets Linux only.
+
+## Repository layout
+
+- `src/` - FTL source code (C)
+- `src/dnsmasq/` - embedded dnsmasq. This is vendored upstream code that is regularly synced with dnsmasq upstream. Do not make changes here unless strictly necessary; Pi-hole-specific hooks live in `src/dnsmasq_interface.c`. Any unavoidable modification to dnsmasq source must be minimal and clearly marked so it survives upstream merges.
```
yubiuser: A patch must be written if any of the vendor files is modified. Patches live in `/patch`

AGENTS.md
```diff
@@ -0,0 +1,57 @@
...
+- Tests exercise a real FTL instance (DNS resolution, API, configuration), so they must run on Linux with the built binary present.
+- Add or extend tests for any behavioural change.
+- Fix any compiler warnings your change introduces; CI treats them as errors.
+
+## PR instructions
+
+- Base all work on the `development` branch; pull requests target `development`.
+- Read the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/)
+- Every commit must be signed off (DCO): use `git commit -s`.
+- Run the test suite before committing.
```
yubiuser, line 47: Any commit must be cryptographically signed. 
https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

yubiuser (dismissed):

DL6ER (commented):
Four things should be fixed before this lands, since the whole point of the file is that agents will act on it without verifying.

**1. `./build.sh test` is destructive on a normal host.** We should make the container the requirement for the suite.

**2. The `src/dnsmasq/` section contradicts our PR template.** Item 6 of the template says changes must not touch `src/dnsmasq/` at all - fixes go through `dnsmasq-discuss` first and come back to us via a merge once they are in dnsmasq master. AGENTS.md softens that to "unless strictly necessary" and then routes such a change into `patch/<library>/`, which does not exist for dnsmasq: `patch/` only holds `civetweb`, `lua` and `sqlite3`.

**3. The patch rule only applies to three trees.** `civetweb`, `lua` and `sqlite3` have a `patch/<lib>/` directory plus a matching `patch/<lib>.sh`. The other vendored code (`src/tre-regex`, `src/webserver/cJSON`, `src/zip/miniz`, `src/config/tomlc17`, `src/database/sqlite3_rsync.c`) has no patch mechanism at all, so naming the three explicitly is more robust than the current phrasing.

As written, AGENTS.md presents `patch/` as the normal route for touching vendored code. It is not. The default for *every* external component we vendor is to send the change to the mother project, not to carry a patch here. `dnsmasq` (item 2 above) is only the strictest instance of that rule, not a special case: the same expectation applies to CivetWeb, Lua and SQLite. We have upstream-first fixes pending in all of them, and that is deliberately the path of least resistance.

A patch in `patch/<lib>/` is justified only for changes that are genuinely Pi-hole specific and therefore have little to no chance of being accepted upstream - our URI rewriting, the CSRF token in `request_info`, the bundled Lua script loading, and so on. Anything that is a plain bug fix or a general improvement belongs upstream, where it benefits everyone and where we stop owning it.

The reason is maintainability: every patch we carry is something that has to be re-applied and re-verified on the next drop, and a growing stack makes importing a new upstream release progressively more painful and more likely to break things quietly. Keeping the vendored trees close to their origin is what makes a bump a merge instead of an excavation.

**4. `cmake/` is the build directory, not CMake modules.** It is gitignored (`/cmake/`), created by `build.sh`, and removed again by `build.sh clean` - it is not part of a checkout. The CMake logic lives in the per-directory `CMakeLists.txt` files plus `src/FindDBus.cmake` and `src/gen_version.cmake`. `tools/` could be added to the layout while we are in there.

It would also help to link `.github/PULL_REQUEST_TEMPLATE.md`

Nobody should be invited "fixing" warnings in `src/dnsmasq/`.

yubiuser (dismissed):

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+- `src/webserver/` - embedded web server (CivetWeb) and Lua handling
+- `test/` - test suite (BATS-based, plus API and regression tests)
+- `tools/` - standalone helper scripts (TLS cert generation, MAC vendor lookup, TOML-to-Markdown docs)
+- `cmake/` - build output directory created by `build.sh`; gitignored, not part of a checkout
+
+## Vendored code policy
+
+FTL vendors several third-party libraries. For all of them, the default is to send a fix upstream, not to carry it as a local patch; a patch is only justified for something genuinely Pi-hole specific with little or no chance of upstream acceptance.
+
+- `src/dnsmasq/` is a verbatim copy of upstream dnsmasq and must not be modified at all, no exceptions. Send fixes to the [`dnsmasq-discuss` mailing list](https://lists.thekelleys.org.uk/cgi-bin/mailman/listinfo/dnsmasq-discuss); they get merged in once they land in dnsmasq master. Pi-hole-specific behavior belongs in `src/dnsmasq_interface.c`, never in the vendored tree.
```
DL6ER:
```suggestion
- `src/dnsmasq/` tracks upstream dnsmasq and must not be modified, no exceptions. The `FTL_*()` calls already present in it are the complete set of Pi-hole additions and are not a precedent for more: every one of them is a call into `src/dnsmasq_interface.c`, which is where Pi-hole-specific behavior belongs.  Send `dnsmasq` related fixes to the [`dnsmasq-discuss` mailing list](https://lists.thekelleys.org.uk/cgi-bin/mailman/listinfo/dnsmasq-discuss). They reach us through a merge once they land in dnsmasq master.
```

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+- CivetWeb (`src/webserver/`), the bundled Lua, and SQLite3 are the only vendored components with a local patch mechanism: genuinely Pi-hole-specific changes (URI rewriting, the CSRF token, bundled Lua script loading, etc.) live as patch files under `patch/<library>/`, applied via the matching `patch/<library>.sh` script during the build.
```
DL6ER:
```suggestion
- CivetWeb (`src/webserver/`), the bundled Lua, and SQLite3 are the only vendored components with a local patch mechanism: genuinely Pi-hole-specific changes (URI rewriting, the CSRF token, bundled Lua script loading, etc.) live as patch files under `patch/<library>/`, applied via the matching `patch/<library>.sh` script manually after each drop-in copy from upstream (only done by the FTL maintainers team). What is committed in the tree is the already-patched source.
```

DL6ER (approved):

DL6ER:
Good to be merged from my side but maybe someone else should press the merge button because I changed something herein now, too

PR #3000, opened 2026-08-05
Add DNS-over-QUIC (RFC 9250) up- and downstream

DL6ER:

This adds **DNS-over-QUIC (RFC 9250)** in both directions, on top of the inbound DoT/DoH server in #2989 (this PR is stacked on `new/dotdoh-server` and should be reviewed after it).

DoQ is the last IETF standards-track encrypted DNS transport we did not speak, and it is by far the cheapest one left to add: the DNS message rides a QUIC stream with the same 2-byte length prefix DoT already uses, so on top of the OpenSSL QUIC stack the DoH3 work brought in there is no HTTP layer at all. Up- and downstream land together because they share that stack.

**Outbound.** `doq://host` becomes a `dns.upstreams` scheme, defaulting to UDP port 853; the `quic://` spelling AdGuard and dnsproxy configs use is accepted as an alias so a copy-pasted resolver address just works. The exchange is fail-closed exactly like DoT/DoH/DoH3 - a bad chain, a reset stream or a timeout drops the query so dnsmasq fails over to the next server rather than downgrading to plaintext. Per RFC 9250 Sec. 4.2.1 the Message ID goes out as 0 and dnsmasq's own ID is restored on the answer, which is what several public DoQ resolvers enforce.

**Inbound.** `dns.doq` (default on, like `dns.dot`) brings up a listener on UDP/853 - the same port number DoT uses on TCP, which does not collide because the transports differ. It is a single-threaded event loop in the shape of the DoT listener: OpenSSL owns the QUIC transport (handshake, loss recovery, flow control, address validation) while every in-flight query is a small non-blocking state machine, so a slow resolve never stalls another connection and a flood costs a bounded state record rather than a thread stack. Queries are attributed to the real downstream client through the same private-EDNS handoff DoT and DoH use, and answers are padded per RFC 8467 when the client asked for it.

**Shared plumbing.** The socket, handshake and timer code the DoH3 client already carried moves into `quic_common.c`, so both QUIC clients share one context, one trust store and one fail-closed verify instead of two copies. DoQ is gated on a new `HAVE_QUIC` (OpenSSL >= 4.0) rather than `HAVE_HTTP3`, so a build without nghttp3 still speaks DoQ.

On hardening: QUIC address validation (Retry) stays on so we cannot be used to amplify towards a forged source, 0-RTT is explicitly disabled because a replayed early-data query would be answered and logged twice, only the `doq` ALPN is accepted, and connections, per-source connections, concurrent streams and per-connection queries are each capped. RFC 9250 Sec. 5.5.2 makes `edns-tcp-keepalive` a protocol error on DoQ, so a query carrying it closes the connection.

One wrinkle needed solving. OpenSSL's QUIC API exposes the peer address but no per-connection *local* address, and the listener is wildcard-bound, so - unlike DoT and DoH - we cannot simply read off which of our addresses a client reached. Conveying nothing is not an option: the answer is then built from the interface of our own loopback handoff, so `pi.hole` resolves to 127.0.0.1 for every DoQ client, and a CNAME chain reaching it writes that into the *shared* cache record other clients are served from. So we ask the kernel which source address it would use to reach that peer - the address it would itself put on a reply datagram, and therefore the one a plain-DNS answer is built from - and convey that as the private EDNS destination option. On a normal LAN this is exactly the address the client used. Should OpenSSL ever expose the real per-connection local address, this becomes a one-line change.

Automated coverage (all of it runs in CI):

- `test/dotdoh_regression.c` - `doq://` and `quic://` URI parsing: defaults, `sni@ip` pinning, bracketed IPv6, and the rejections (empty host, port 0/out of range, a path, `doqx://` not prefix-matching `doq`).
- `test/dotdoh.bats` - outbound: a query resolves over the DoQ proxy path against the aioquic DoQ server in `test/dotdoh_shim.py` (port 8854), the forwarded query is padded (RFC 8467), and every forwarded query carries Message ID 0 (RFC 9250). That the answer is still accepted proves we map dnsmasq's ID back.
- `test/dotdoh_server.bats` - inbound: the listener comes up on UDP/853, a query resolves, it is attributed to the real downstream client (sourced from `127.0.0.4`, which no other test uses, so the API evidence can only come from this query), several queries multiplex over one connection, a client not offering the `doq` ALPN is refused, a malformed query does not wedge the reactor, an IPv6 query resolves, and - the one that pins the destination hint down - `pi.hole/AAAA` over a v4 DoQ transport returns NODATA rather than leaking `::1`. I verified that last one discriminates by suppressing the hint and re-running: it fails with `expected NODATA but got 1 answer record(s)` while its DoT twin still passes.

Manual, against a running Pi-hole:

1. Outbound - point Pi-hole at a public DoQ resolver and confirm queries flow and are attributed to the real upstream, not the internal loopback tuple:
   ```
   pihole-FTL --config dns.upstreams '["doq://dns.adguard.com"]'
   dig @127.0.0.1 example.com
   ```
   The query log / `/api/queries` must show the `doq://` upstream, not `127.47.11.N`. Cross-check the resolver itself with `kdig +quic @94.140.14.14 A example.com`.
2. Inbound - with `dns.doq = true` and a valid `webserver.tls.cert`, query the Pi-hole from another machine:
   ```
   kdig +quic @<pi-hole-ip> -p 853 A example.com
   ```
   The answer must be correct and the query must appear in the dashboard under the *querying device's* IP, not under the Pi-hole itself. `q -t A example.com quic://<pi-hole-ip>` works as a second client.
3. Fail-closed check - point a `doq://` upstream at a host presenting the wrong certificate and confirm the query is dropped (dnsmasq fails over) rather than answered.
4. `pi.hole` over DoQ - from another machine, `kdig +quic @<pi-hole-ip> -p 853 A pi.hole` must return the Pi-hole address that machine reaches, not 127.0.0.1.
5. Build check - configure without `nghttp3`. The build must still report `DoQ (DNS-over-QUIC) support: YES` and `HTTP/3 support: NO`, and DoQ must work. Against an OpenSSL older than 4.0 the build must succeed with `DoQ (DNS-over-QUIC) support: NO`; `-Werror` makes that configuration easy to break, so it is worth checking.

---

---

## Checklist:

### DIFF
--- a/src/dotdoh/doq_client.c
+++ b/src/dotdoh/doq_client.c
@@ -1,0 +1,345 @@
+/* Pi-hole: A black hole for Internet advertisements
+*  (c) 2026 Pi-hole, LLC (https://pi-hole.net)
+*  Network-wide ad blocking via your own hardware.
+*
+*  FTL Engine
+*  Outbound DoQ client (DNS-over-QUIC, RFC 9250)
+*
+*  DoQ is the cheapest of the encrypted transports to speak: the DNS message goes
+*  on a QUIC stream with the same 2-byte length prefix DoT uses, so there is no
+*  HTTP layer at all. Like the other clients it is strict and fail-closed - a bad
+*  chain, a stream reset or a timeout returns -1 and the query is dropped, so FTL
+*  fails over to the next server instead of downgrading to plaintext.
+*
+*  Each exchange runs on its own short-lived QUIC connection, mirroring the DoH3
+*  client: many workers then run concurrently without sharing a connection
+*  object, at the cost of one 1-RTT handshake per query. Reusing connections
+*  across queries would need a shared, locked connection object; that is a
+*  follow-up, not a correctness matter.
+*
+*  This file is copyright under the latest version of the EUPL.
+*  Please see LICENSE file for your rights under this license. */
+
+#include "FTL.h"
+#include "log.h"
+#include "doq_client.h"
+
+// DoQ needs OpenSSL QUIC, but - unlike DoH3 - no HTTP/3 library.
+#if defined(HAVE_TLS) && defined(HAVE_QUIC)
+
+// dot_frame/dot_deframe (RFC 9250 reuses the DoT framing), DNS_MSG_MAX
+#include "framing.h"
+// edns_remove_option()
+#include "edns_pad.h"
+// shared QUIC socket/handshake/timer plumbing
+#include "quic_common.h"
+
+#include <openssl/quic.h>
+#include <openssl/bio.h>
+#include <openssl/ssl.h>
+
+#include <stdlib.h>
+#include <string.h>
+#include <pthread.h>
+#include <unistd.h>
+
+// Overall budget (ms) for one DoQ exchange (connect + handshake + query +
+// answer), so a bad upstream cannot pin a worker; on expiry, dnsmasq fails over.
+#define DOQ_EXCHANGE_TIMEOUT_MS 10000
+
+// ALPN token for DNS-over-QUIC (RFC 9250 Sec. 4.1.2). QUIC mandates ALPN, so an
+// upstream that does not speak DoQ fails the handshake instead of half-working.
+static const unsigned char DOQ_ALPN[] = { 3, 'd', 'o', 'q' };
+
+// EDNS(0) TCP Keepalive (RFC 7828). RFC 9250 Sec. 5.5.2 forbids it on a DoQ
+// connection in either direction, and a conformant server aborts the connection
+// with DOQ_PROTOCOL_ERROR when it sees one - as our own inbound listener does.
+// It is a hop-by-hop option, so stripping it here is also what RFC 7828 wants of
+// a forwarder.
+#define EDNS_OPT_TCP_KEEPALIVE 11
+
+// Per-upstream DoQ pool. In the connection-per-exchange model it holds only the
+// immutable upstream descriptor and the stats (lock-guarded); the QUIC
+// connection is short-lived on the worker stack for one exchange.
+struct doq_pool {
+	struct upstream_uri u;      // upstream descriptor (immutable)
+	int max;                    // connection cap (immutable; informational here)
+	pthread_mutex_t lock;       // guards stats
+	struct dotdoh_stats stats;
+};
+
+struct doq_pool *doq_pool_new(const struct upstream_uri *u, int max_conns)
+{
+	if(quic_client_ctx() == NULL || u == NULL || max_conns < 1)
+		return NULL;
+	struct doq_pool *p = calloc(1, sizeof(*p));
+	if(p == NULL)
+		return NULL;
+	p->u = *u;
+	p->max = max_conns;
+	pthread_mutex_init(&p->lock, NULL);
+	return p;
+}
+
+void doq_pool_free(struct doq_pool *p)
+{
+	if(p == NULL)
+		return;
+	pthread_mutex_destroy(&p->lock);
+	free(p);
+}
+
+// Write the whole framed query to the request stream and close our sending side
+// with a FIN, which is how DoQ delimits the query (RFC 9250 Sec. 4.2). Returns
+// false on a fatal stream/connection error or on running out of time.
+static bool doq_send_query(SSL *conn, SSL *stream, int fd,
+                           const uint8_t *buf, size_t len, uint64_t deadline)
+{
+	size_t off = 0;
+	while(off < len)
+	{
+		size_t written = 0;
+		const int r = SSL_write_ex(stream, buf + off, len - off, &written);
+		if(r == 1 && written > 0)
+		{
+			off += written;
+			continue;
+		}
+		if(r != 1)
+		{
+			const int err = SSL_get_error(stream, r);
+			if(err != SSL_ERROR_WANT_READ && err != SSL_ERROR_WANT_WRITE)
+				return false;
+		}
+		// Blocked (or a zero-byte accept): wait for progress rather than spin.
+		if(quic_now_ms() >= deadline)
+			return false;
+		quic_wait(conn, fd, deadline);
+		SSL_handle_events(conn);
+	}
+	// FIN: no more data on this stream. The peer needs it to start resolving.
+	return SSL_stream_conclude(stream, 0) == 1;
+}
+
+// Read the length-prefixed answer from the request stream into buf. Returns the
+// DNS message length and sets *msg_off, or -1 on a protocol error, a reset, a
+// premature FIN or the deadline expiring.
+static ssize_t doq_read_answer(SSL *conn, SSL *stream, int fd, uint8_t *buf,
+                               size_t bufsz, size_t *msg_off, uint64_t deadline)
+{
+	size_t have = 0;
+	for(;;)
+	{
+		const ssize_t alen = dot_deframe(buf, have, msg_off);
+		if(alen < 0)
+			return -1;  // zero-length frame: protocol error
+		if(alen > 0)
+			return alen; // complete answer buffered
+
+		if(have >= bufsz)
+			return -1; // no room left and still incomplete
+
+		size_t nread = 0;
+		const int r = SSL_read_ex(stream, buf + have, bufsz - have, &nread);
+		if(r == 1 && nread > 0)
+		{
+			have += nread;
+			continue;
+		}
+		if(r != 1)
+		{
+			const int err = SSL_get_error(stream, r);
+			if(err != SSL_ERROR_WANT_READ && err != SSL_ERROR_WANT_WRITE)
+				// ZERO_RETURN (FIN before a full answer), a reset or a
+				// connection error - none of them can ever complete.
+				return -1;
+		}
+		if(quic_now_ms() >= deadline)
+			return -1;
+		quic_wait(conn, fd, deadline);
+		SSL_handle_events(conn);
+	}
+}
+
+// One exchange over a freshly handshaked connection. Returns the answer length
+// written into answer[], or -1.
+static ssize_t doq_exchange_on_conn(SSL *conn, int fd, const uint8_t *query, size_t qlen,
+                                    uint8_t *answer, size_t answer_sz, uint64_t deadline)
+{
+	// Framing and receive buffers: 64 KiB each is too much for a worker's thread
+	// stack, so keep one set per worker thread (as the proxy does).
+	static _Thread_local uint8_t sendbuf[2 + DNS_MSG_MAX];
+	static _Thread_local uint8_t recvbuf[2 + DNS_MSG_MAX];
+
+	ssize_t flen = dot_frame(query, qlen, sendbuf, sizeof(sendbuf));
+	if(flen < 0)
+		return -1;
+
+	// A downstream client's edns-tcp-keepalive rides through dnsmasq untouched; it
+	// must not reach a DoQ upstream. Strip it from our framed copy (the helper
+	// keeps the padded length intact by absorbing the freed bytes into the
+	// padding, so this is usually length-neutral) and correct the prefix if not.
+	const size_t stripped = edns_remove_option(sendbuf + 2, qlen, EDNS_OPT_TCP_KEEPALIVE);
+	if(stripped != qlen)
+	{
+		sendbuf[0] = (uint8_t)((stripped >> 8) & 0xff);
+		sendbuf[1] = (uint8_t)(stripped & 0xff);
+		flen = (ssize_t)(2 + stripped);
+	}
+
+	// RFC 9250 Sec. 4.2.1: the DNS Message ID MUST be 0 on a QUIC stream (the
+	// stream itself correlates query and answer). Remember the ID dnsmasq chose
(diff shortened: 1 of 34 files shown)


test/dotdoh_query.py
```diff
@@ -303,9 +303,139 @@ async def run():
...
+    cfg.verify_mode = ssl.CERT_NONE
+    cfg.server_name = server_name
+
+    family = socket.AF_INET6 if ":" in host else socket.AF_INET
+    sock = socket.socket(family, socket.SOCK_DGRAM)
+    sock.setblocking(False)
+    if source:
+        sock.bind((source, 0))
+    addr = (host, port)
+
```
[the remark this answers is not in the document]
DL6ER: Python accepts a 2-tuple for `AF_INET6` (`bind(('::1', 0))` yields `('::1', 47700, 0, 0)`), and `test/dotdoh_server.bats` already exercises this exact path with source `::1` as part of the passing suite, so the IPv6 DoQ test is not broken.

src/dotdoh/doq_server.c
```diff
@@ -0,0 +1,1241 @@
...
+	                                          s->conn->dest[0] != '\0' ? s->conn->dest : NULL,
+	                                          s->wbuf, WBUF_SZ);
+	if(flen < 0)
+		return -1;
+	s->wlen = (size_t)flen;
+	s->woff = 0;
+
+	// One query per stream (RFC 9250 Sec. 4.2): anything after it is ignored.
+	s->have = 0;
+
```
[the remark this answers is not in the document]
DL6ER: Trailing bytes are now a stream error rather than silently dropped; not requiring the client's FIN before answering stays deliberate, as the client is obliged to send it but nothing is gained by withholding an answer we can already produce.

darkexplosiveqwx:
I tried building this branch inside a Fedora 45 container for testing, because 44 is on openSSL 3.5, but 45 also introduces nettle 4.0, which introduces breaking changes.

### 1. `src/edns0.c` — `hmac_sha256_digest()` (build failure)

From nettle's NEWS:

> The `_digest` functions for hash algorithms, MACs and AEADs no longer take the desired digest size as argument, instead, they always produce the full-size digest. The typedef `nettle_hash_digest_func` has also been changed accordingly.

The build stops at `src/edns0.c:136`:

```
cc1: error: passing argument 2 of 'nettle_hmac_sha256_digest' makes pointer from integer without a cast [-Wint-conversion]
       136 |         hmac_sha256_digest(&ctx, DOTDOH_MAC_LEN, out);
```

This call (from the inbound DoT/DoH work) uses the three-argument nettle API. Nettle 4.0 removed the digest-length argument, leaving only the two-argument form:

```c
void hmac_sha256_digest(struct hmac_sha256_ctx *ctx, uint8_t *digest);
```

Two facts matter for the fix. The second-argument length is gone, *and* the function now always writes the full 32-byte digest — but `out` here is declared `unsigned char out[DOTDOH_MAC_LEN]` (16 bytes). So dropping the length argument is not enough; nettle 4 would write past the buffer. The fix must go through a full-size scratch buffer and truncate:

```c
#if NETTLE_VERSION_MAJOR >= 4
	unsigned char full[32];
	hmac_sha256_digest(&ctx, full);          // full 32 bytes
	memcpy(out, full, DOTDOH_MAC_LEN);       // truncate to 16
#else
	hmac_sha256_digest(&ctx, DOTDOH_MAC_LEN, out);
#endif
```

The earlier `f70ff5a6` "Make FTL nettle v4.0 compatible" commit covered `2fa.c`, `files.c`, and the digest calls in `password.c`, but the `edns0.c` call was added later in the stack and slipped through. Also note `src/files.c:784` already carries this same `#if NETTLE_VERSION_MAJOR >= 4` pattern; this call should follow it.

#### I also let an agent scan for more nettle 4.0 changes based on NEWS and it found:

### 2. `src/config/password.c` — `base64_decode_update()` input length (runtime)

Nettle 4.0 also changed the length semantics of `base64_decode_update` — same signature, different contract:

> The `dst_length` argument to `base16_decode_update` and `base64_decode_update` is now both an input and output argument. On input it must now hold the size of the destination buffer, and decoding fails if that is not sufficient. Previously, `dst_length` was an output only, and it was required that the destination buffer was large enough for any input of the given `src_length`.

The helper at `src/config/password.c:176` initializes the length to `0`:

```c
size_t salt_len = 0;   // likewise hash_len = 0
*salt = base64_decode(salt_base64, &salt_len);
```

On nettle 4.0 those `0` values are now treated as an input buffer size, so the decode fails (returns 0, `*dst_length` stays 0). I verified this end-to-end rather than in isolation: replicating FTL's exact `base64_encode` / `base64_decode` / `parse_PHC_string` flow against `nettle-devel-4.0` on Fedora 45, a PHC string decodes to `salt_len=0`/`hash_len=0` (expected 16/32). The `salt_len != SALT_LEN` guard in `parse_PHC_string()` then returns false, and `verify_password()` reports `PASSWORD_INCORRECT` (`src/config/password.c:470`) for a password hash that FTL itself generated. The same PHC string parses fine once the buffer capacity is passed in (`salt_len=16`/`hash_len=32`). On nettle 3.x the original code returns the decoded length correctly, so this only reproduces on distros pairing OpenSSL 4.0 with nettle 4.0 — exactly the F45 target this PR needs for DoQ.

The fix is version-agnostic because the signature is unchanged; pass the buffer capacity as input and read the decoded length back out:

```c
const size_t decodelen = BASE64_DECODE_LENGTH(strlen(data));
uint8_t *decoded = calloc(decodelen, sizeof(uint8_t));

size_t dest_len = decodelen;
struct base64_decode_ctx ctx;
base64_decode_init(&ctx);
base64_decode_update(&ctx, &dest_len, decoded, strlen(data), data);
base64_decode_final(&ctx);

// The caller expects the number of decoded bytes, not the buffer size
*length = dest_len;
```

010011110:
issue 2 predates this pr and is unrelated to doq, i am taking care of it in a separate pr

darkexplosiveqwx:
> issue 2 predates this pr and is unrelated to doq, i am taking care of it in a separate pr
> 

I just checked this PR's diff again and Issue 1 was also not introduced in this PR (this PR doesn't touch edns0.c), but it is at least related to this PR's dotdoh theme. May be included in this PR or a new one.

I wrote the second one simply also down while noticing nettle API incompatibilities. Thanks for taking care of it early before it gets forgotten.

DL6ER:
You are right, and both are fixed - in `development` rather than here, since both came in before this PR.

Issue 1 is `25c947a80`, issue 2 `base64_decode()` in `password.c`. This branch was still rebased on a commit that predates them, so a build here would have kept failing at `edns0.c:136`; it is rebased on current `development` now and picks both up.

The MAC fix went a different way than a scratch buffer: the output parameter is widened to `SHA256_DIGEST_SIZE` so nettle 4.0 writes the full digest in place, with a `static_assert` pinning that the MAC is a truncated prefix of it. `<nettle/version.h>` is included there too - without it `NETTLE_VERSION_MAJOR` is undefined and the guard quietly keeps the old branch.

Thanks for building on 45. Neither would have surfaced otherwise.

src/dotdoh/server.c
```diff
@@ -212,19 +376,6 @@ static int loopback_connect(void)
 // the terminator's per-connection detached handler threads (and from restartable
 // h3 workers), neither of which lives for the whole life of the process, so
 // without this each such thread would leak its loopback fd.
```
darkexplosiveqwx: The _Thread_local file descriptor below got removed, so this comment has become stale.

darkexplosiveqwx (changes requested):

DL6ER:
Good catch, thank you. That paragraph described the thread-local fd and its thread-exit destructor, both of which went when DoH moved onto the shared pool, so it was describing the opposite of what the code does now.

Looking for others of the same kind turned up two more. `drive_read()`'s keep-alive reuse branch is unreachable, because the socket goes back to the pool as soon as the answer is written - sockets still come back warm, from the pool now rather than from the connection. And `dotdoh_loopback_take()`'s contract in the header mentioned neither -2 nor that -1 reserves the slot anyway, which is the one that bites: a caller that opens its own socket still has to release it.

All three in 1a88fd350.

src/dotdoh/doq_client.c
```diff
@@ -0,0 +1,345 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <pthread.h>
+#include <unistd.h>
+
+// Overall budget (ms) for one DoQ exchange (connect + handshake + query +
+// answer), so a bad upstream cannot pin a worker; on expiry, dnsmasq fails over.
+#define DOQ_EXCHANGE_TIMEOUT_MS 10000
+
+// ALPN token for DNS-over-QUIC (RFC 9250 Sec. 4.1.2). QUIC mandates ALPN, so an
```
darkexplosiveqwx, line 50: There is no Section 4.1.2 in the RFC.
It's just 4.1:

> **4.1. Connection Establishment**
DoQ connections are established as described in the QUIC transport specification [RFC9000]. During connection establishment, DoQ support is indicated by selecting the Application-Layer Protocol Negotiation (ALPN) token "doq" in the crypto handshake.

Ref: https://www.rfc-editor.org/info/rfc9250/#section-4.1

src/dotdoh/doq_server.c
```diff
@@ -0,0 +1,1272 @@
...
+#include <poll.h>
+#include <pthread.h>
+#include <signal.h>
+#include <sys/prctl.h>
+#include <time.h>
+
+// RFC 9250 Sec. 4.1.1: DoQ uses UDP port 853 by default - the same number DoT
+// uses on TCP, which does not collide.
+#define DOQ_PORT 853
+// ALPN token for DNS-over-QUIC (RFC 9250 Sec. 4.1.2). QUIC mandates ALPN, so a
```
darkexplosiveqwx, line 65: See https://github.com/pi-hole/FTL/pull/3000#discussion_r4076311169

src/dotdoh/doq_server.c
```diff
@@ -0,0 +1,1272 @@
...
+	}
+
+	// Answer complete: hand the socket back so the next query reuses it rather
+	// than forking another dnsmasq child.
+	dotdoh_loopback_give(s->upfd);
+	s->upfd = -1;
+	s->up_pooled = false;
+	s->up_ev = 0;
+
+	// RFC 8467 Sec. 4: pad the answer only if the query asked for it.
```
darkexplosiveqwx, line 890: RFC 8467 Sec. 4 is only "4.  Padding Strategies".
It itself references RFC 7830 Sec. 4, which is directly relevant:
> **4.  Usage Considerations**
>
> Responders MUST pad DNS responses when the respective DNS query included the "Padding" option, unless doing so would violate the maximum UDP payload size.
>
> Responders MAY pad DNS responses when the respective DNS query indicated EDNS(0) support of the Requestor and the "Padding" option was not included.

Ref: https://www.rfc-editor.org/info/rfc7830/#section-4 & https://www.rfc-editor.org/info/rfc8467/#section-4

src/dotdoh/dot_server.c
```diff
@@ -317,13 +331,111 @@ static struct dot_conn *conn_new(int cfd, const char *client, const char *dest)
...
+	// Sample the query's EDNS state before dot_servfail() overwrites the buffer
+	// it sits in.
+	bool query_do = false;
+	const bool query_edns = edns_query_opt(c->abuf, c->qsave, &query_do);
+
+	ssize_t slen = dot_servfail(c->abuf, c->qsave, c->abuf, ABUF_SZ);
+	if(slen < 0)
+		return -1;
+	// RFC 6891 Sec. 6.1.1: answer an EDNS query with an OPT, whether or not it
+	// padded - replying without one marks us EDNS-lame. RFC 8467 Sec. 4: only
```
darkexplosiveqwx, line 391: See https://github.com/pi-hole/FTL/pull/3000#discussion_r4076435141

test/dotdoh_query.py
```diff
@@ -355,6 +485,54 @@ def main():
...
+                         timeout=8.0)
+        except SystemExit as exc:
+            msg = str(exc.code) if exc.code is not None else ""
+            if msg.startswith("SKIP:"):
+                raise
+            # No answer (or a reset stream) is a perfectly good outcome here; the
+            # follow-up query in the bats test proves the listener still serves.
+        print("OK")
+    elif cmd == "doqalpn":
+        # QUIC mandates ALPN and RFC 9250 Sec. 4.1.2 defines exactly one token for
```
darkexplosiveqwx, line 531: See https://github.com/pi-hole/FTL/pull/3000#discussion_r4076311169

darkexplosiveqwx (changes requested):

src/dotdoh/doq_client.c
darkexplosiveqwx: Dan has informed me that Section 4.1.2 was present in an early draft of the RFC.

But then the reference would've been "draft-ietf-dprive-dnsoquic-00" and not RFC9250.

Especially since the RFC is out of draft since May 2022, this should be corrected.

https://author-tools.ietf.org/iddiff?url1=rfc9250&url2=draft-ietf-dprive-dnsoquic-00&difftype=--html

PR #3005, opened 2026-08-06
Optional microsecond timestamp in the log (`debug.extra`)

DL6ER:

Our log timestamps carry millisecond resolution. That is fine for everyday use, but too coarse whenever one is actually chasing a timing problem - a burst of related log lines regularly ends up sharing the very same millisecond, so the log no longer tells us in which order things happened.

Rather than adding yet another config option for this, we reuse `debug.extra`, the flag that already exists for exactly this kind of temporary investigation ("*Temporary flag that may print additional information [...] meant to be used whenever needed for temporary investigations*"). When it is enabled, `get_timestr()` prints six instead of three fractional digits:

```
2026-08-06 08:01:40.086 CEST [12345M] INFO: ...     (default)
2026-08-06 08:01:40.086300 CEST [12345M] INFO: ...  (debug.extra = true)
```

As every timestamped log line goes through `get_timestr()`, this covers `FTL.log`, `webserver.log` and the foreground/stdout output alike. The default output is bit-for-bit unchanged, so nothing that reads our logs today is affected unless the flag is deliberately turned on.

The `debug.extra` help text mentions the added resolution so the behavior is discoverable from `pihole-FTL --config debug.extra` instead of only from the source.

Note that the handful of lines emitted *before* the config has been read still use millisecond resolution - the flag simply is not known yet at that point.

1. Build FTL and start it as usual. The timestamps in `/var/log/pihole/FTL.log` have three fractional digits, as before:
   ```
   2026-08-06 08:01:28.828 CEST [12345M] INFO: ########## FTL started on host! ##########
   ```
2. Enable the flag and restart:
   ```
   sudo pihole-FTL --config debug.extra true
   sudo service pihole-FTL restart
   ```
   From the point where the config has been read, all timestamps carry six fractional digits:
   ```
   2026-08-06 08:01:40.086300 CEST [12345M] DEBUG_ANY: * EXTRA:          YES  *
   ```
   The same can be done without touching the config file by starting FTL with `FTLCONF_debug_extra=true`.
3. Check `/var/log/pihole/webserver.log` while the flag is set - it uses the same time string and switches along.
4. Turn the flag off again and confirm the log is back to milliseconds.

Automated coverage: the existing `test/` suite runs with `debug.extra` unset, so the log format it sees is unchanged; `test_final.bats` continues to scan `FTL.log` for unexpected `WARNING`/`ERROR`/`CRIT` lines as before.

---

---

## Checklist:

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -1661,7 +1661,7 @@ void initConfig(struct config *conf)
 	conf->debug.webserver.c = validate_stub; // Only type-based checking
 
 	conf->debug.extra.k = "debug.extra";
-	conf->debug.extra.h = "Temporary flag that may print additional information. This debug flag is meant to be used whenever needed for temporary investigations. The logged content may change without further notice at any time.";
+	conf->debug.extra.h = "Temporary flag that may print additional information. This debug flag is meant to be used whenever needed for temporary investigations. The logged content may change without further notice at any time. Enabling this flag also raises the resolution of all log timestamps from milliseconds to microseconds.";
 	conf->debug.extra.t = CONF_BOOL;
 	conf->debug.extra.d.b = false;
 	conf->debug.extra.c = validate_stub; // Only type-based checking
--- a/src/log.c
+++ b/src/log.c
@@ -106,11 +106,16 @@ void get_timestr(char timestring[TIMESTR_SIZE], const time_t timein, const bool 
 	{
 		struct timeval tv;
 		gettimeofday(&tv, NULL);
-		const int millisec = tv.tv_usec/1000;
 
-		snprintf(timestring, TIMESTR_SIZE, "%d-%02d-%02d%c%02d%c%02d%c%02d.%03i%c%s",
+		// debug.extra bumps the resolution to microseconds so events
+		// happening within the same millisecond can still be ordered
+		const bool micros = config.debug.extra.v.b;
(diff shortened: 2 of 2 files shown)


yubiuser (commented):
`test/pihole.toml` needs adjustment

DL6ER:
Parked as a draft for now: the open chain of logging PRs (#2958, #2960, #2968, #2969) introduces `get_timestr_iso8601()` for the JSON/journald output, which still formats milliseconds. Once that series has landed, this should be revisited so both timestamp paths gain the higher resolution together.

DL6ER:
Rebuilt on top of `development` as a single commit. `test/pihole.toml` now carries the extended `debug.extra` help text, and the comment in `src/ntp/client.c` that still claimed `get_timestr()` only prints milliseconds is updated as well.

The structured logging PRs (#2968, #2969) leave the body of `get_timestr()` alone and format their own ISO 8601 timestamp, so there is no reason to keep this parked behind them. The higher resolution for that path can come with their rebase.

darkexplosiveqwx:
My current plan for logging is to do the asynchronous logger first (I created `log-async-v2`), because it touches all the same stuff as structured logging and I believe it would otherwise more annoying to review. Structured logging itself becomes much simpler with a centralized writer system.

My current approach stores the timestamp as a `double` and  doesn't touch (but uses) `get_timestr()`, they are compatible with each other.

This one may go first.

darkexplosiveqwx (approved):

PR #3012, opened 2026-08-08
dotdoh: make the inbound DoT and DoQ listener ports configurable

DL6ER:

Follow-up to #3000. `dns.dot` and `dns.doq` were booleans with the port hardcoded to 853, so the listeners could only be turned on or off. Both become `CONF_UINT16` defaulting to `853`, with `0` disabling the listener - the same shape `dns.port` already has for plain DNS, so whether a listener runs and which port it uses stay a single setting rather than two.

The defaults are the assignments from RFC 7858, Sec. 3.1 (TCP/853 for DoT) and RFC 9250, Sec. 4.1.1 (UDP/853 for DoQ). They still do not collide, as one is TCP and the other UDP.

Each listener reads its port once when its thread starts instead of dereferencing `config` at every use, so the bind and the log line cannot disagree should the configuration be replaced underneath.

The DoQ socket drops `SO_REUSEADDR`. It bought nothing on UDP, which has no `TIME_WAIT` to work around, but Linux does let two UDP sockets share an address once both set it - so `dns.doq = 53` would have bound alongside dnsmasq and quietly taken a share of its datagrams. Without the option that bind fails with `EADDRINUSE` and says so in the log, which is what the DoT listener has always done on TCP. That is also why the keys need no cross-key port validation: a collision is now a visible failure on both transports.

There is deliberately no boolean-compatibility shim. Neither key has shipped in a release, and the TOML reader treats a type mismatch as an absent key. Anyone who set `dns.dot = false` while testing #3000 therefore gets the listener back on 853, with only a `DEBUG_CONFIG` line to say so, and `pihole.toml` is rewritten with the number. `dns.doh` stays a boolean, as it has no port of its own and rides `webserver.port`.

Set a non-default port and restart (both keys are `FLAG_RESTART_FTL`):

```
pihole-FTL --config dns.dot 8853
pihole-FTL --config dns.doq 8853
```

`FTL.log` should then show `dotdoh: DoT server listening on port 8853` and `dotdoh: DoQ server listening on UDP port 8853`, and a query confirms it end to end:

```
python3 test/dotdoh_query.py dot 127.0.0.1 8853 pi.hole 127.0.0.2 test/test.crt 127.0.0.1
```

Setting either key to `0` should leave the corresponding listener absent from the log entirely, and pointing `dns.doq` at `53` should now fail the bind with a logged error instead of shadowing dnsmasq.

`test_suite.bats` gains CLI assertions that both keys read back as `853` and that a boolean is rejected as a 16-bit unsigned integer; neither reads nor rejected writes touch `pihole.toml`, so the write counters in `test_final.bats` are unaffected. Beyond that, `dotdoh_server.bats` (35 tests) exercises both listeners on the default port and `test_suite.bats` diffs `test/pihole.toml` against the config FTL regenerates, which pins the new types and help text. Note that no automated test binds a *non-default* port - doing so needs an FTL restart mid-suite, which would perturb shared state and those write counters, so it is worth a reviewer checking the manual steps above.

## Additional information

**Related issue or feature (if applicable):** follow-up to #3000

---

## Checklist:

- [ ] I based my code and PRs against the repository&#39;s `development` branch.
- [ ] I have read the above and my PR is ready for review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ pihole-FTL*
 # Generated standalone regression test binaries (built via ./build.sh test)
 /tar_regression
 /gzip_regression
+/dotdoh_regression
 
 # Generated pihole.toml file
 pihole.toml
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -611,6 +611,17 @@ if(LIBSSL AND LIBCRYPTO AND USE_TLS)
 #endif
 int main(void) { return 0; }
 " OPENSSL_HAS_QUIC_PEER_ADDR)
+
+    # DoQ (RFC 9250) rides on OpenSSL's QUIC alone - it carries DNS directly on
+    # QUIC streams, so unlike DoH3 it needs no HTTP/3 library. Gate it on the
+    # OpenSSL version only, so a build without nghttp3 still speaks DoQ.
+    if(OPENSSL_HAS_QUIC_PEER_ADDR)
+        message(STATUS "Building FTL with DoQ (DNS-over-QUIC) support: YES")
+        target_compile_definitions(dotdoh PRIVATE HAVE_QUIC)
+    else()
+        message(STATUS "Building FTL with DoQ (DNS-over-QUIC) support: NO (OpenSSL < 4.0)")
+    endif()
+
     find_library(LIBNGHTTP3 NAMES libnghttp3${LIBRARY_SUFFIX} nghttp3)
     if(LIBNGHTTP3 AND OPENSSL_HAS_QUIC_PEER_ADDR)
         message(STATUS "Building FTL with HTTP/3 support: YES")
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -298,9 +298,9 @@ components:
                 doh:
                   type: boolean
                 dot:
-                  type: boolean
+                  type: integer
                 doq:
-                  type: boolean
+                  type: integer
                 blocking:
                   type: object
                   properties:
@@ -774,8 +774,8 @@ components:
             upstreams: [ "127.0.0.1#5353", "8.8.8.8" ]
             upstreamCA: ""
             doh: true
-            dot: true
-            doq: true
+            dot: 853
+            doq: 853
             CNAMEdeepInspect: true
             EDNS0ECS: true
             ignoreLocalhost: false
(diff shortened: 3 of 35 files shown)


rdwebdesign:
Suggestion:

We could use `dns.dot_port` and `dns.doq_port`. The current names seems like the options are boolean, like `dns.doh`.

DL6ER:
You are right that `dns.dot = 853` next to a boolean `dns.doh` reads oddly.

`dns.doh` stays a boolean because DoH has no port of its own - it rides `webserver.port`. DoT and DoQ do have one, which is why they became ports, following `dns.port` for plain DNS: the key *is* the port, `0` disables it. If we rename, it would be `dns.dotPort`/`dns.doqPort` rather than snake_case, as no key under `dns.` currently carries an underscore.

That said, we are reconsidering whether custom ports are worth having at all - see below.

DL6ER:
Stepping back from the naming: it is not clear this change earns its keep.

A custom port is only useful if clients can be told about it. Android's Private DNS takes a hostname and nothing else, so it always uses 853 - and that is the largest group of DoT clients pointed at a Pi-hole by some distance. Firefox, stubby and systemd-resolved accept a port, but they are a thin slice next to it.

The other route is advertising it, via an RFC 9462 SVCB record under `_dns.resolver.arpa`. That is cheaper than it sounds: `dns-rr=_dns.resolver.arpa,64,<hex>` in the generated `dnsmasq.conf` would do it, with no SVCB-serving code in FTL. It raises its own questions, though:

- We currently answer NODATA for the whole `resolver.arpa` zone on purpose (`dns.specialDomains.designatedResolver`), which is the correct "no Designated Resolver" signal per RFC 9462, Sec. 4. Advertising means turning that into a three-way choice.
- `port=` is a generic SvcParam, so this is not a DoQ question - DoT needs the same treatment, and DoH would too if it ever moved off `webserver.port`.
- A client that discovers us this way still has to trust our certificate. With a self-signed or private-CA certificate the upgrade fails, so discovery alone does not make the port reachable.

So the choice is roughly: keep ports and accept they are undiscoverable for most clients, add DDR advertisement and take on the questions above, or drop the change and leave DoT and DoQ on their assigned 853.

Our inclination is to drop it and keep 853, and revisit if someone actually needs a different port.

DL6ER:
This will be kept as draft as it is not decided what to do with it. IMO this is feature creep to some extend without a clear application scenario.

rdwebdesign:
> IMO this is feature creep to some extend without a clear application scenario.

I agree. We probably don't need this.

DL6ER:
Agreed, feature creep without a clear scenario. Closing - easy to revive if one turns up.

PR #3017, opened 2026-08-09
webserver: serve every configured TLS port, and DoH behind a reverse proxy

DL6ER:

Three gaps in how the front TLS terminator exposes ports, the first raised in review of the terminator work.

**1. Only the first TLS port was served.** `split_terminator_ports()` captured the port under `if(tls_port == 0)`, so every later `...s` entry in `webserver.port` was dropped from the list handed to CivetWeb while nothing else picked it up. The port simply vanished, with no diagnostic. `"80or,443os,8080,4443s"` - the example in `webserver.port`'s own help text - silently lost 4443.

The terminator now takes an array of listeners and its accept thread polls all of them, so every configured TLS port is served and reported in the API port list. Overlapping entries are collapsed first, which matters more than it sounds: the shipped default `"80o,443os,[::]:80o,[::]:443os"` names one dual-stack socket and then its IPv6 half, so binding both naively yields `EADDRINUSE` on every start. A bare port therefore supersedes any address-scoped entry for that port.

HTTP/3 stays on the first TLS port. The QUIC listener is a single socket wired to one `quic_fd`, and giving it N listeners is a much larger change; Alt-Svc names the HTTP/3 port explicitly (`h3=":443"`), so advertising it from the other TLS ports stays correct rather than misleading.

**2. IPv4 and IPv6 halves of a port could not coexist.** `fill_bind_addr()` maps `0.0.0.0` to the v4-mapped `::ffff:0.0.0.0`, which accepts IPv4 only, and every socket was bound dual-stack - so `"0.0.0.0:443s,[::]:443s"` served IPv4 only and there was no way to express the split. `IPV6_V6ONLY` is now set for an explicit IPv6 literal and cleared otherwise, which lets the two halves coexist as separate sockets.

A side effect is that a configuration naming *only* a bracketed TLS entry, e.g. `"[::]:443s"`, no longer answers over IPv4 through v4-mapped addresses. That is what `webserver.port` documents (*"[::]:80 will bind to port 80 IPv6 only"*) and what CivetWeb already does for plaintext ports, so this aligns the terminator with both. Nothing released changes: the terminator only exists in `development`, so the dual-stack-always behavior was never in a release. The default port list is unaffected either way, as it names both a bare and a bracketed entry.

**3. DoH could not sit behind a reverse proxy.** Plaintext `/dns-query` was always refused with 426, so a Pi-hole behind an external TLS terminator (nginx, Traefik, Caddy, HAProxy) could not expose DoH at all. Two settings now cover that deployment:

- **`webserver.proxySecret`** - a shared secret authenticating the proxy. It is handed to CivetWeb as `proxy_protocol_secret`, so a PROXY protocol v2 header carrying it is believed, and the client address and TLS status it announces replace the transport peer. FTL's own terminator then uses the same secret instead of its per-boot token. Empty by default, in which case the token stays per-boot and internal, exactly as today. Write-only, so it is masked in API output.
- **`dns.dohReverseProxy`** - off by default; serves DoH on such an authenticated connection.

Trust rests on the proxy **proving who it is**, not on the hop being assumed safe. An unauthenticated plaintext `/dns-query` is still refused with 426, so enabling this cannot expose cleartext DNS on a port something else can reach - which matters, because the shipped default `webserver.port` has `80o` on all interfaces.

An unauthenticated `X-Forwarded-For` is deliberately never consulted. It is trivially spoofable, and client identity is load-bearing in FTL: per-client rate limiting (`client->rate_limit`) and per-client group policy both key on it, so a spoofed header would let any host on that port exhaust a victim's rate-limit budget or select their policy. Because the PROXY v2 address is authenticated instead, queries are attributed to the **real client** rather than to the proxy - so there is no attribution or rate-limit-sharing caveat.

Set a port list, restart, and check both the listener lines in `FTL.log` and what actually answers:

```
pihole-FTL --config webserver.port '80o,443os,4443s'
curl -sk -o /dev/null -w '%{http_code} %{http_version}\n' https://127.0.0.1:443/api/info/ftl
curl -sk -o /dev/null -w '%{http_code} %{http_version}\n' https://127.0.0.1:4443/api/info/ftl
```

Measured here, one FTL start per row:

| `webserver.port` | listeners | IPv4 | IPv6 |
| --- | --- | --- | --- |
| `80o,443os,[::]:80o,[::]:443os` (default) | `*#443` | 200 | 200 |
| `80o,0.0.0.0:443s,[::]:443s` | `0.0.0.0#443` + `::#443` | 200 | 200 |
| `80o,[::]:443s` | `::#443` | - | 200 |
| `80o,0.0.0.0:443s` | `0.0.0.0#443` | 200 | - |
| `80o,443os,4443s` | `*#443` + `*#4443` | 200 | 200 |

The default row is the important regression check: one dual-stack listener and no `ERROR: Terminator: bind()` in the log.

DoH behind a proxy needs a real PROXY v2 header, so a plain `curl` cannot exercise it - point an actual proxy at Pi-hole, or use a small script that prepends the header. With `webserver.proxySecret` set and `dns.dohReverseProxy true`, measured here:

| request to `http://127.0.0.1/dns-query` | result |
| --- | --- |
| PROXY v2 with the correct secret + `PP2_TYPE_SSL` | `200 application/dns-message` |
| no PROXY header (plain cleartext) | `426` |
| PROXY v2 with a **wrong** secret | `426` |
| PROXY v2, correct secret, no TLS marker | `426` |

With an off-subnet announced client the request is refused with `403` by the usual `dns.listeningMode` policy, the same rule a plain DNS client faces. Attribution was confirmed by the query being recorded against the announced client address rather than the proxy.

Automated coverage: the full suite passes (`test_suite.bats` 191, `dotdoh.bats` 12, `dotdoh_server.bats` 25, `test_final.bats` 9, pytest 149) with only the three known version-string failures that also fail on a clean `development` build in the same container. `test_suite.bats` diffs `test/pihole.toml` against the regenerated config, which pins the new key, and `test_openapi.py` type-checks it against the updated spec. The build was also checked in the no-nghttp2, no-nghttp3, no-QUIC and TLS-only configurations, since none of those are covered by CI. No automated test binds a second TLS port or exercises plaintext DoH, so the manual steps above are worth running.

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -297,6 +297,8 @@ components:
                   type: string
                 doh:
                   type: boolean
+                dohReverseProxy:
+                  type: boolean
                 dot:
                   type: boolean
                 blocking:
@@ -462,6 +464,8 @@ components:
               type: object
               properties:
                 domain:
+                  type: string
+                proxySecret:
                   type: string
                 acl:
                   type: string
@@ -772,6 +776,7 @@ components:
             upstreams: [ "127.0.0.1#5353", "8.8.8.8" ]
             upstreamCA: ""
             doh: true
+            dohReverseProxy: false
             dot: true
             CNAMEdeepInspect: true
             EDNS0ECS: true
@@ -876,6 +881,7 @@ components:
               expire: 365
           webserver:
             domain: pi.hole
+            proxySecret: ''
             acl: "+0.0.0.0/0,::/0"
             port: 80,[::]:80
             threads: 0
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -612,6 +612,13 @@ void initConfig(struct config *conf)
 	conf->dns.doh.f = FLAG_RESTART_FTL;
 	conf->dns.doh.c = validate_stub;
 
+	conf->dns.dohReverseProxy.k = "dns.dohReverseProxy";
+	conf->dns.dohReverseProxy.h = "Serve DoH to a trusted reverse proxy that terminates TLS in front of Pi-hole (nginx, Traefik, Caddy, HAProxy, ...). The client-facing connection stays HTTPS; the proxy handles the encryption and forwards plain HTTP to Pi-hole. This requires webserver.proxySecret to be set and the proxy to send a PROXY protocol v2 header carrying that secret: only an authenticated proxy is believed, and the client address it announces is used to attribute the query. A plain HTTP request that is not authenticated this way is still refused, so enabling this alone never exposes DNS queries in cleartext.";
+	conf->dns.dohReverseProxy.t = CONF_BOOL;
+	conf->dns.dohReverseProxy.d.b = false;
+	conf->dns.dohReverseProxy.f = FLAG_RESTART_FTL;
+	conf->dns.dohReverseProxy.c = validate_stub;
+
 	conf->dns.dot.k = "dns.dot";
 	conf->dns.dot.h = "Enable the inbound DNS-over-TLS (DoT) server on port 853. When enabled, FTL terminates DoT connections directly (RFC 7858) so downstream clients can use this Pi-hole as their encrypted resolver. Requires a valid TLS certificate (the same one configured for the webserver).";
 	conf->dns.dot.t = CONF_BOOL;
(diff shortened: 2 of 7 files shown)


darkexplosiveqwx (commented):
I think this PR would benefit quite a bit from being split into multiple commits.

Your description already describes three distinct problems (and solutions):
1. Only the first TLS port was served.
2. IPv4 and IPv6 halves of a port could not coexist.
3. DoH could not sit behind a reverse proxy.

If 1. and 2. are too inter-mangled to split reasonably, I'd still say that the third point deserves its own commit at least.

DL6ER:
Split into three commits as you suggested: binding an explicit IPv6 entry IPv6-only comes first, as serving several ports relies on it, then the multi-port listener, then DoH behind a reverse proxy. The end result is unchanged apart from two comments, and each commit builds on its own.

darkexplosiveqwx:
Thanks for split into multiple commits, that made a review more manageable.

#### **1.**  I think multiple QUIC listeners are desired, but not necessarily in this PR. 

This PRs goal is to address the inconsistency of HTTP capabilities between the first and subsequent ports. HTTP/3 is still a gap here.
The `Alt-Svc` header explicitly supports multiple values:
>  The Alt-Svc field value can have multiple values:
>
> Alt-Svc: h2="alt.example.com:8000", h2=":443"
>
> When multiple values are present, the order of the values reflects the server's preference (with the first value being the most preferred alternative).

[RFC7838, Section 3](https://datatracker.ietf.org/doc/html/rfc7838#section-3)

My idea is to advertise every port where the QUIC listener was successfully initialized, while remaining the user's `webserver.ports` order.

#### **2.** Reject unsupported DoH methods when using a reverse proxy

Every recognized non-POST method with a query string is processed as GET. Consequently, authenticated PUT, DELETE, HEAD, and ordinary OPTIONS requests can resolve DNS and return 200; HEAD may also return a response body.
Using a reverse proxy should be subject to the same validation as queries using FTL's terminator by explicitly rejecting unsupported methods.

#### **3.** Support DoH via a reverse proxy even without `HAVE_TLS`

Currently the `webserver.proxySecret` setup is guarded with `HAVE_TLS`. I think it makes sense for DoH via reverse proxy to be available even when FTL was not build with OpenSSL, because FTL does no TLS handling itself in that case.

#### **4.** Prevent cross-client caching of reverse-proxy DoH responses

The reverse-proxy DoH handler emits `Cache-Control: max-age=...`, allowing a shared cache to reuse GET responses across clients. DoH answers can depend on the authenticated client address, including client groups and blocking policy. Native DoH correctly emits `private, max-age=....`
Use the same private caching policy here and add a reverse-proxy GET test that verifies the response headers.

#### **5.** Fully validate secure-port entries

For example, 44s3 becomes TLS port 44 and 443xs becomes TLS port 443, potentially opening a wildcard listener from a typo.
The terminator should match civetweb's validation of port entries. This is not a new issue here, but more prevalent now that the terminator doesn't only serve the first HTTPS port.

Issue #3015, opened 2026-08-09
Config API: rapid dns.hosts mutations can silently lose committed deletes (dnsmasq-test waitpid race); v6.7 can also spin on ECHILD


poindexter12:
**Versions:** Reproduced on `pihole/pihole:2025.03.0` and `pihole/pihole:2026.07.2` (FTL v6.7), Docker on macOS and ubuntu-latest CI runners.

## Symptom 1: committed deletes silently revert ("resurrection")

Under bursts of strictly **serialized** REST mutations to `/api/config/dns/hosts/...` (PUT to create, DELETE to remove — never concurrent, each DELETE verified by polling the list endpoint until the entry is gone), a deleted entry can later re-materialize. Concrete evidence from our reproduction: after a test run that created and deleted 30 `stress-dns-N.local` entries with per-delete verification, one entry (`10.0.0.3 stress-dns-3.local`) was present again with no further writes happening — the DELETE had returned 204 **and** the list endpoint had confirmed removal before the entry returned.

Downstream this surfaces as `bad_request / Item already present (Uniqueness of items is enforced)` on later PUTs of the same entry. Other API clients appear to hit the same class, e.g. kubernetes-sigs/external-dns#5251 (AAAA records that "did not get deleted" + `Item already present` crash loops).

Reproduction rate for us: ~90% per run of a Terraform-provider stress suite that does bulk create → verified bulk delete → immediate re-create cycles (30 entries, one HTTP request at a time). A plain `curl` loop with ~10–20 ms between operations does **not** reproduce it — the window seems to need back-to-back requests over a keep-alive connection (Go HTTP client).

## Symptom 2: v6.7 can spin forever in the config-test wait loop

With `debug.config = true`, during such bursts FTL logged an unbounded stream of:

```
DEBUG_CONFIG: Waiting for dnsmasq test returned: No child process
```

(thousands of identical lines with the same timestamp), after which the API stopped responding until restart.

Looking at `src/config/dnsmasq_config.c` in the **v6.7 tag**, the loop saves `const int err = errno;` but the ECHILD escape checks the **global** `errno` again:

```c
const int err = errno;
log_debug(DEBUG_CONFIG, "Waiting for dnsmasq test returned: %s", strerror(err));
...
if(errno == ECHILD)   // <- errno may have been clobbered by log_debug/strerror
```

[...]

PR #3011, opened 2026-08-08
Validate internationalized domain names with libidn2

DL6ER:
# What does this implement/fix?

Since v6.7.1, `valid_domain()` accepts every well-formed UTF-8 sequence where the name is handed to dnsmasq, e.g., in `dns.hosts`. That is more than dnsmasq loads: a label whose punycode form is longer than 63 characters, mixed LTR/RTL scripts or an emoji pass and end as `bad name` in the log.

A name with non-ASCII bytes is now converted with `idn2_to_ascii_8z()` and the same `IDN2_NONTRANSITIONAL` flag dnsmasq's `canonicalise()` uses, and the punycode result has to pass the existing ASCII checks. FTL therefore accepts exactly what dnsmasq will load. The `allow_utf8` parameter is gone again.

The second commit lets gravity take such names as well and stores them in punycode, which is what a query carries. An internationalized line in an adlist used to be counted as invalid, now it blocks.

Pure ASCII names never reach libidn2: a list of 5 million domains parses as fast as before (1.6 s), with 1% internationalized names it takes 0.4 s longer.

## How to test the change during review

```
pihole-FTL --config dns.hosts '[ "2.2.2.2 äste.com steä.com" ]'
```

is accepted, the same goes for `PATCH /api/config`. `ex!ample.com`, malformed UTF-8 and names IDNA2008 rejects are refused. The BATS tests *Internationalized domain names are accepted, invalid ones are not* and *Gravity stores internationalized domains as punycode* cover all of these.

## Additional information

**Related issue or feature (if applicable):** N/A

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** N/A

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.2)
5. I have squashed any insignificant commits. ([`git rebase`](https://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
[...]

PR #3020, opened 2026-08-09
Serialize config changes so concurrent API writes cannot drop each other

DL6ER:

Changing the config is a read-copy-modify-install sequence: `duplicate_config()` takes a private copy, the copy is changed, `replace_config()` installs it. Both helpers lock the SHM for themselves, but nothing held a lock across the whole sequence. Two writers can therefore each copy the config, change their own copy and install it one after the other - the one installing last silently drops the other's change.

That is reachable from ordinary API use: we run 50 webserver worker threads, and `reread_config()` (housekeeping thread, on a `pihole.toml` change) as well as a Teleporter import are writers, too. When the dropped change is a `DELETE`, the record is back after we answered 204, and the next `PUT` of the same record fails with `Item already present` - the report in #3015. The same overlap also let several threads rotate `pihole.toml`, write the single `pihole.toml.tmp` and run `dnsmasq --test` on the shared temporary file at the same time.

Two commits:

1. Serialize the whole mutate-and-write-back sequence on a new config lock, taken by every in-process writer. It is recursive because a config change can nest: setting `webserver.api.password` runs the legacy password upgrade, which writes the config out on its own. Initialized at runtime, as musl has no `PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP`.
2. `getpwnam()`, `getpwuid()` and `getgrgid()` return pointers into a buffer libc shares between all threads. `chown_pihole()` and `/api/info/database` run on worker threads, so this was an actual heap corruption, not a theoretical one - `gdb` caught two workers aborting inside `realloc()` from `getgrgid()` while rotating `pihole.toml`. Those call sites now use the `_r` variants; the startup, CLI and crash-handler users are single-threaded and stay as they are.

The `waitpid()`/`ECHILD` half of #3015 is already fixed on `development` (we check the saved `err`, not the global `errno`), it is only missing from the v6.7 tag. The endless `Waiting for dnsmasq test returned: No child process` the reporter saw is a consequence of several threads testing the same dnsmasq config file at once, so it should go away with this as well.

New pytest case `test/api/test_m_mutations.py::TestConcurrentConfigMutations` - four connections doing strictly serialized `PUT`/`DELETE` pairs on a config array in parallel, asserting nothing is left behind. On `development` this fails within a few hundred requests, usually by taking FTL down; with this branch it passes.

```
pytest test/api/test_m_mutations.py::TestConcurrentConfigMutations -v
```

Manually, against the endpoint from the issue:

```
for i in $(seq 1 40); do
  curl -sX PUT    "http://pi.hole/api/config/dns/hosts/10.0.0.$i%20stress-$i.local"
  curl -sX DELETE "http://pi.hole/api/config/dns/hosts/10.0.0.$i%20stress-$i.local"
done &   # start four of these in parallel
curl -s "http://pi.hole/api/config/dns/hosts"   # must be empty again
```

I ran this at 4, 8 and 12 concurrent writers with 50 rounds each: no lost deletes, no resurrected records, no crash. `test_m_mutations.py` and `test_z_auth*.py` are otherwise unchanged (the one remaining failure in the former is the pre-existing IDN validation issue with the umlaut record seeded in `test/pihole.toml`).

## Additional information

**Related issue or feature (if applicable):** #3015

---

## Checklist:

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -271,9 +271,11 @@ int api_handler(struct mg_connection *conn, void *ignored)
 		                      api.request->local_uri_raw);
 	}
 
-	// Restart FTL if requested
+	// Restart FTL if requested. Config changes go through the delay so that a
+	// burst of them results in a single restart, /api/action/restartdns is an
+	// explicit request and restarts right away.
 	if(api.ftl.restart)
-		restart_ftl(api.ftl.restart_reason);
+		request_restart(api.ftl.restart_reason);
 
 	return ret;
 }
--- a/src/api/config.c
+++ b/src/api/config.c
@@ -741,7 +741,7 @@ static int api_config_get(struct ftl_conn *api)
 	JSON_SEND_OBJECT(json);
 }
 
-static int api_config_patch(struct ftl_conn *api)
+static int config_patch(struct ftl_conn *api)
 {
 	// Is there a payload with valid JSON data?
 	const int ret = check_json_payload(api);
@@ -950,13 +950,21 @@ static int api_config_patch(struct ftl_conn *api)
 	return api_config_get(api);
 }
 
+static int api_config_patch(struct ftl_conn *api)
+{
+	lock_config();
+	const int ret = config_patch(api);
+	unlock_config();
+	return ret;
+}
+
 // Inspired by https://stackoverflow.com/a/32496721
 //static void replace_char(char* str, char find, char replace)
 //{
 //	for (char *current_pos = strchr(str, find); (current_pos = strchr(str+1, find)) != NULL; *current_pos = replace);
 //}
 
-static int api_config_put_delete(struct ftl_conn *api)
+static int config_put_delete(struct ftl_conn *api)
 {
 	if(api->item == NULL || strlen(api->item) == 0)
 		return 0;
@@ -1175,6 +1183,14 @@ static int api_config_put_delete(struct ftl_conn *api)
 		send_http_code(api, NULL, 204, "");
 		return 204;
 	}
+}
+
+static int api_config_put_delete(struct ftl_conn *api)
+{
+	lock_config();
+	const int ret = config_put_delete(api);
+	unlock_config();
+	return ret;
 }
 
 // Endpoint /api/config router
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -586,6 +586,8 @@ components:
                   type: integer
                 delay_startup:
                   type: integer
+                restart_delay:
+                  type: integer
                 addr2line:
                   type: boolean
                 etc_dnsmasq_d:
@@ -935,6 +937,7 @@ components:
           misc:
             nice: -10
             delay_startup: 10
+            restart_delay: 0
             addr2line: true
             privacylevel: 0
             etc_dnsmasq_d: false
--- a/src/api/info.c
+++ b/src/api/info.c
@@ -116,9 +116,14 @@ int api_info_database(struct ftl_conn *api)
 	JSON_ADD_NUMBER_TO_OBJECT(json, "ctime", st.st_ctime); // Time of last status change (owner or mode change, etc.)
 
 	// Get owner details
+	// The reentrant lookups are required as this runs on a webserver worker
+	// thread: getpwuid()/getgrgid() share one buffer between all threads
+	char pwbuf[PWBUF_SIZE];
+	struct passwd pwstore, *pw = NULL;
+	getpwuid_r(st.st_uid, &pwstore, pwbuf, sizeof(pwbuf), &pw);
+
 	cJSON *user = JSON_NEW_OBJECT();
 	JSON_ADD_NUMBER_TO_OBJECT(user, "uid", st.st_uid); // UID
-	const struct passwd *pw = getpwuid(st.st_uid);
 	if(pw != NULL)
 	{
 		JSON_COPY_STR_TO_OBJECT(user, "name", pw->pw_name); // User name
@@ -130,9 +135,12 @@ int api_info_database(struct ftl_conn *api)
 		JSON_ADD_NULL_TO_OBJECT(user, "info");
 	}
 
+	char grbuf[PWBUF_SIZE];
+	struct group grstore, *gr = NULL;
+	getgrgid_r(st.st_gid, &grstore, grbuf, sizeof(grbuf), &gr);
+
 	cJSON *group = JSON_NEW_OBJECT();
 	JSON_ADD_NUMBER_TO_OBJECT(group, "gid", st.st_gid); // GID
-	const struct group *gr = getgrgid(st.st_gid);
 	if(gr != NULL)
 	{
 		JSON_COPY_STR_TO_OBJECT(group, "name", gr->gr_name); // Group name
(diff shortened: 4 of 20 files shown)


DL6ER:
#3011 touches the same `test_final.bats` block: it changes the `custom.list` counts, this one the `pihole.toml` count. The numbers are independent, so whichever lands second just needs a rebase.

poindexter12:
Ran the branch against the reproducer from #3015 as requested. TL;DR: **the lost-update fix works** — zero resurrections in 8 runs — but restart frequency roughly doubles under the same workload, which surfaces as an availability regression for API clients.

## Setup
`pihole-FTL-arm64` CI artifact from this branch (`vDev-82caed4`) swapped into a `pihole/pihole:2026.07.2` container (aarch64/Docker). Reproducer: our Terraform provider acceptance stress suite (pre-mitigation client), bulk create → verified delete → immediate re-create cycles of 30 `dns.hosts` entries + 30 `dns.cnameRecords` per run, strictly serialized requests. Fresh container per batch to keep API session seats out of the picture.

## Results (4 runs per batch)
| Build | `Item already present` (lost-delete resurrection) | Failure mode observed |
|---|---|---|
| stock v6.7 | **3 / 4 runs** | resurrections |
| this branch (2 fresh-container batches, 8 runs) | **0 / 8 runs** | none — remaining failures were availability (below) |
| development `0af0d74` (control for the binary-swap procedure) | 0 / 4, but confounded — session seats filled before resurrections could surface | session/auth |

## Availability observation
Under identical 4-run load: **790** `Restarting FTL` events on this branch vs **415** on stock v6.7 (~2×). The restart churn produces API blackouts long enough that our client's re-auth (5 retries spanning ~15 s) gives up, so runs fail with auth/session errors instead of resurrections. Not a correctness issue, but mutation-heavy API workloads regress from "sometimes corrupts" to "often unavailable".

## Related per-op finding (also on stock v6.7)
Isolated probes show every `dns.cnameRecords` PUT triggers a full FTL restart (hosts PUT/DELETE do not). That restart is what converts a transiently lost TOML write into a *visible* resurrection — the reload faithfully restores the stale disk state. It's also why a hosts-only curl reproducer never fired. If restart requests were coalesced/debounced (or CNAME changes handled via reload), both the amplification and the blackout windows would shrink considerably.

Minor secondary note: sessions being stored/restored across every restart means seats never free under restart-churn, so long mutation-heavy runs eventually wedge auth at `max_sessions` — easy to hit in CI-style workloads.

Happy to re-run against an updated branch.

DL6ER:
Your run is a better confirmation than my own testing: I reproduced the API-vs-API case with 4 to 12 parallel writers, but your client sends one request at a time. That a strictly serialized client still loses writes says the other writer is in-process - `reread_config()` picking up the inotify event on `pihole.toml`, and the restart path re-reading the file while a write-back is still in flight. That is the case the new lock closes, and it is also why your CNAME finding matters: the restart is what makes a lost write visible, because it restores whatever is on disk.

The doubled restart count is that same fix showing up. Previously the losing writer's change was dropped silently, and a dropped change never got as far as requesting its restart. Now every write lands, so every write flagged `FLAG_RESTART_FTL` restarts too.

We do have `?restart=false` plus `POST /api/action/restartdns`, but that only helps a client that knows when its sequence ends, and a Terraform provider does not - it gets one Create or Delete per resource and never learns which one is the last of an apply. So you are right that this belongs on our side, and I have added it here rather than deferring it to another issue.

The branch now has a third commit adding `misc.restart_delay`: the number of seconds FTL waits after a restart-requiring change before actually restarting. Changes arriving in the meantime extend the wait and are applied together, so a burst costs one restart instead of one per request. A restart is postponed by no more than five times the delay, so a client writing continuously cannot defer it forever. Only config writes are collected - `/api/action/restartdns` and the NTP time-jump restart stay immediate.

The default is 0, which keeps today's behavior of restarting immediately. Changing restart timing for every installation is not something I want to slip into a bugfix PR, so it is opt-in for now and we can revisit the default once it has some mileage. For your suite that means one line in `pihole.toml`:

```toml
[misc]
  restart_delay = 2
```

With that, a burst of writes spaced less than two seconds apart restarts FTL once instead of per request, without your provider needing to know where the batch ends.

On the session seats: we keep sessions across restarts on purpose so a restart does not log everybody out, and expired seats are reclaimed when the next login looks for a free slot. What fills your table is that your client re-authenticates after each restart rather than reusing its SID, which is still valid, so every restart burns a fresh seat. That should largely disappear along with the restarts.

Re-running against the updated branch would be welcome, in particular whether `restart_delay = 2` is enough for the request spacing your suite produces.

poindexter12:
Tested this on arm64 (Docker [`pihole/pihole:nightly`](https://hub.docker.com/r/pihole/pihole) with the CI branch binary from [ftl.pi-hole.net](https://ftl.pi-hole.net/fix%2Fconfig-hosts-race/pihole-FTL-arm64), `vDev-`[`fee7a86`](https://github.com/pi-hole/FTL/commit/fee7a8671edf6a8adb10d20fe147d227725db7de)) and can confirm it fixes the issue for me.

Ran the reproduction from the PR description — 4 parallel workers each doing 40 serialized PUT/DELETE pairs against `/api/config/dns/hosts`:

- **This branch:** 160/160 request pairs succeeded, 0 leftover hosts entries afterward.
- **Stock nightly (`vDev-511e369`), same container, same script:** 152/160 pairs failed and FTL went down with `SEGV_MAPERR` partway through — matches the "usually by taking FTL down" behavior described above.

For extra context: I maintain a Terraform provider for Pi-hole ([poindexter12/terraform-provider-pihole](https://github.com/poindexter12/terraform-provider-pihole)), where I've had to serialize all record mutations client-side ([#68 on the upstream provider](https://github.com/ryanwholey/terraform-provider-pihole/issues/68)) and treat `Item already present` as success specifically because of this race (same symptoms as #3015). The provider's full acceptance suite, including its concurrency stress tests, passes cleanly against this branch. Thanks for the fix!

poindexter12:
Re-ran on `vDev-fee7a86`. `restart_delay = 2` is plenty; 1 gives the same result.

Same suite as before (232 config writes per run, 8 ms median between writes, 12 ms p90). Binary and setting verified through `/api/info/version` and `GET /api/config/misc/restart_delay`.

| `misc.restart_delay` | restart requests | actual FTL restarts | wall time |
|---|---|---|---|
| 0 | 116 | 7 | 33.9 s |
| 1 | 1 | 1 | 14.4 s |
| 2 | 2 | 2 | 16.5 s |

The two restarts at delay 2 are the five times cap firing once (writes run flat out for about 14 s) and the deferred one 1.4 s after the last write. Both match `check_pending_restart()`. Full acceptance suite at delay 2: 368 writes, 122 restart requests, 3 restarts, all green.

Worth noting the delay 0 row: 116 requests already collapse to 7 restarts because a request arriving mid-restart just re-sets the flag. So my "restarts doubled" number from the 9th was stock dropping writes vs. this branch applying them, not a regression in the lock.

Restart counts are `Restarting FTL:` and `internal restart` lines in FTL.log with a 12 s tail after the suite, otherwise the deferred restart isn't in the capture.

👍

DL6ER:
Thanks for re-running it. `1` looking the same as `0` was not the wait being too short: the window was measured with `time(NULL)`, so a change arriving just before the next full second already counted as a full second of quiet. It now runs on a monotonic millisecond clock and the default is 1 second.

poindexter12:
Thanks for fixing! Let me know if anything else comes up!

PR #3026, opened 2026-08-16
follow-up: bound IFLA_VF_INFO by its own header length

010011110:

`RTA_PAYLOAD()` is unsigned, so a nested `IFLA_VF_INFO` attribute claiming fewer bytes than its own header wraps to a huge length. Nothing is parsed out of bounds today, as `parse_rtattr_flags()` takes an `int`, the truncated value is negative and `RTA_OK()` rejects it - but that is a conversion, not a check. `vfinfo` is the only nested attribute we take straight from `RTA_DATA()` without validating it first, so it gets an explicit lower bound here.

Also fixes the wording of a comment added in #2986.

## Additional information

---

## Checklist:

- [ ] The code change is tested and works locally.

### DIFF
--- a/src/tools/netlink.c
+++ b/src/tools/netlink.c
@@ -1186,8 +1186,10 @@ static int nlparsemsg_link(struct ifinfomsg *ifi, void *buf, size_t len, cJSON *
 					break;
 
 				// The nested attribute carries its own length, so it
-				// must not claim more than the outer payload holds
-				if(RTA_PAYLOAD(rta) < vfinfo->rta_len)
+				// must cover its header and must not claim more than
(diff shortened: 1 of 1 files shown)


DL6ER:
Your analysis is right: `RTA_LENGTH(0)` is `size_t`, so `RTA_PAYLOAD()` wraps to a huge unsigned value for a short `rta_len`. It only worked because `parse_rtattr_flags()` takes an `int`, which narrows it back to something negative that `RTA_OK()` then rejects. That is not a property I want to keep relying on.

Merging once riscv64 is through.

DL6ER (approved):

PR #3028, opened 2026-08-17
lua: add `pihole.download()`

DL6ER:
> [!IMPORTANT]
> Blocked on pi-hole/docker-base-images#186 being merged and tagged first. Until `.github/Dockerfile` is bumped off `ftl-build:v2.25`, this builds with `HTTP client support: NO` and the new tests fail.

`pihole.download(url [, path])` for the Lua `pihole` module: streams a body into a string, or writes it to a file, returning `nil, "message"` on failure. We had no HTTP client - CivetWeb is built with `NO_SSL`, so its `mg_download()` silently fetches `https://` in plaintext.

`http`, `https`, `ftp`, `ftps` and `file` are accepted, gzip is decompressed on the way in. The 16 MiB cap sits in the write callback and so counts decompressed bytes; redirects may never reach `file`/`ftp` or leave TLS; files land via `mkostemp()` + `rename()`, as libcurl would otherwise truncate the destination up front.

libcurl is optional: without it the function says the build has no HTTP client.

`test/run.sh` covers it with 16 bats tests (http, https, cert verification, fail-closed, atomic write, size cap, chunked, redirects, gzip bomb, `file://`, redirect-to-`file://` refusal, bad schemes and paths). Locally: 207 bats, 151 pytest, 46 dotdoh, 0 failures on a static build.

```
./pihole-FTL lua -e 'print(pihole.download("https://ftl.pi-hole.net/libraries/"))'
```

Also worth compiling without libcurl to check the fallback path.

## Additional information

---

## Checklist:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -623,6 +623,35 @@ int main(void) { return 0; }
     else()
         message(STATUS "Building FTL with HTTP/3 support: NO (nghttp3 not found)")
     endif()
+
+    # libcurl backs pihole.download(). It sits inside this block because the
+    # bundled libcurl is built against OpenSSL, so it can only be linked where
+    # libssl/libcrypto are. Optional: without it pihole.download() reports that
+    # this build has no HTTP client.
+    find_library(LIBCURL NAMES libcurl${LIBRARY_SUFFIX} curl)
+    find_library(LIBZ NAMES libz${LIBRARY_SUFFIX} z)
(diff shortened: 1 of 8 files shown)


DL6ER:
It seems we decided to not add this to `lua` but instead use `go` in the core code. This branch does work (if we ever want to revisit it) and `libcurl` is now available, but there is no point in adding this when we don't plan to use it - even if fully written. Hence, closing!

PR #3031, opened 2026-08-18
Release the SHM lock and roll back on early returns

010011110:

Two macro families return from the calling function on error: `SQL_bool()`/`SQL_void()` when a query fails, and the `JSON_*` macros when a cJSON allocation fails. Where that return happens between `lock_shm()` and `unlock_shm()`, the SHM lock stays taken with nothing left to release it, and every thread that touches shared memory blocks from then on.
`queries_to_database()` uses `SQL_bool()` for four transaction statements. Two of them sit inside the locked region, and all four run after the snapshot array is allocated, so they leak it as well. A failed `END` also left the transaction open on `_memdb`, which lives for the whole process, so every later `BEGIN` on that handle failed too. They now go through `dbquery()` and share one cleanup path that rolls back, unlocks and frees as required.
The same applies to eleven `JSON_*` uses in `api_history()`, `api_history_clients()`, `api_padd()` and `api_stats_recentblocked()`. Those sites now test the allocation themselves and release everything they own on the way out: the lock, the partially built JSON tree and, in `api_history_clients()`, the temporary client array. The guard cannot live in the macros instead, as they return from whichever function uses them, which is not always the one holding the lock - `api_list_read()` runs with the lock held by `api_list()`, so unlocking there would unlock twice. `JSON_SEND_OBJECT_UNLOCK()` is removed, it had no users.

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/history.c
+++ b/src/api/history.c
@@ -27,15 +27,25 @@ int api_history(struct ftl_conn *api)
 
 	cJSON *history = JSON_NEW_ARRAY();
 	const unsigned int max_slot = get_max_overtime_slot();
-	// Loop over all overTime slots and add them to the array
+	// Loop over all overTime slots and add them to the array. The JSON_*
+	// macros cannot be used while we hold the lock, as their early return
+	// would leave it taken for good
 	for(unsigned int slot = 0; slot <= max_slot; slot++)
 	{
 		cJSON *item = JSON_NEW_OBJECT();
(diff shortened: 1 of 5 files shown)


DL6ER (changes requested):
One thing to keep from `SQL_bool()`: it logged `SQLITE_BUSY` as a warning and only everything else as an error. The four replacements always call `log_err()`, so a transient busy database now reports as an error.

010011110:
updated

DL6ER (approved):

PR #3032, opened 2026-08-19
let a write wait for the gravity database instead of failing

010011110:

the gravity connection has no busy handler so that a DNS lookup never waits for the database. API writes use that same connection, and once per second gravity_updated() opens a second, read-only connection to gravity.db from the database thread. gravity.db uses the normal rollback journal, so that reader blocks the writer and a COMMIT landing in the window fails right away:

```
ERROR: gravityDB_delFromTable(0): SQL error exec("COMMIT TRANSACTION;"): database is locked
```

the rollback works, so nothing is corrupted, but the edit is also lost. It depends on timing, the test suite here hit it once in six runs. the relevant now set the busy handler while the write runs and remove it again afterwards, so a write waits and retries instead of failing.

adding, editing and deleting list items works as before. a new test holds a read transaction on gravity.db and writes over the API while it is held, revert gravity-db.c and keep it and the suite says 192 tests, 1 failure. please infor me if you have different test needs. regarding comment and test style I tried to stay close to existing code

## Additional information

---

## Checklist:

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -1827,26 +1827,34 @@ bool gravityDB_get_regex_client_groups(clientsData *client, const unsigned int n
 	return true;
 }
 
-// The gravity connection deliberately runs without a busy handler so that a DNS
-// lookup never waits for the database (see gravityDB_open()). Writes need the
-// opposite: gravity_updated() opens its own read connection once per second, and
-// a COMMIT landing in that window fails outright without one. Arm it for the
-// duration of a write and take it off again afterwards
-static void gravity_write_wait(const bool wait)
+// Writes get their own connection. sqlite3_busy_handler() is a property of the
+// connection, and the shared one is deliberately left without one so a DNS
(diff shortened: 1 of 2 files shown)


DL6ER (changes requested):
`sqlite3_busy_handler()` applies to the whole connection, and `gravity_db` is the one the DNS path uses. Its missing busy handler is deliberate, which is what the comment you quote is about. While a write has it armed, a lookup that runs between the write's statements and meets the file lock from the once-per-second reader now waits up to `DATABASE_BUSY_TIMEOUT` (1 s) instead of failing straight away.

I would rather the write used its own connection, the way `gravity_updated()` already opens its own for reading. The DNS connection then stays untouched and there is no window at all.

The test also depends on `sleep 0.1` and `sleep 0.4` lining up. On a loaded runner the reader may not hold its transaction when the write lands, and the test then passes without exercising anything.

010011110:
updated as requested

DL6ER (approved):

PR #3036, opened 2026-08-20
Do not drop a database connection that refused to close

010011110:

sqlite3_close() answers SQLITE_BUSY while a statement of that connection is still alive, and then does nothing at all. the connection stays open, with its transaction and its file locks. _dbclose() logs the error and sets the pointer to NULL anyway, so from that moment nothing can close the database or release the lock any more, and the next dbopen() just adds another connection

it finalizes what the caller left behind now, and says which statement that was and where the close came from, as that is a bug in the caller either way. the connection is then handed over with sqlite3_close_v2(), which also takes over blob handles and backups that are still in flight

get_row_count() did reach that, its prepare-failure path closed unconditionally while the success path already guarded with if(!memory), so with memory set it handed _memdb to dbclose(). that was inert while sqlite3_close() refused, but finalizing the statements query-table.c caches on that connection would have left them dangling, so the call site is guarded now and dbclose() refuses the shared in-memory handle outright, the call-site discipline has failed once already

the finalize-and-hand-over part sits in dbclose_handle() and the other places that own a connection use it too: the gravity write connection, which can hold an exclusive lock and is the worst place to lose one, and macvendor_db. gravityDB_close() reports a refused close instead of dropping it. checkFTLDBrc() stays in dbclose(), a refused close on gravity.db or macvendor.db says nothing about the FTL database

nothing changes in normal operation, the suite passes and neither the "Statement not finalized" nor the in-memory guard line appears. there is no test for the failures themselves, provoking them means leaking a statement or closing someone else's connection on purpose

## Additional information

---

## Checklist:

### DIFF
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -68,6 +68,16 @@ void _dbclose(sqlite3 **db, const char *func, const int line, const char *file)
 	if(FTLDBerror())
 		return;
 
+	// The shared in-memory connection is owned by close_memory_database() and
+	// its prepared statements live as long as FTL does. Closing it here would
+	// finalize them behind the back of whoever cached them
+	if(db != NULL && is_memdb(*db))
+	{
+		log_err("dbclose() called on the in-memory database in %s() (%s:%i)",
+		        func, short_path(file), line);
+		return;
+	}
+
 	if(config.debug.database.v.b)
 		log_debug(DEBUG_DATABASE, "Closing FTL database in %s() (%s:%i)", func, short_path(file), line);
 
@@ -1208,7 +1218,8 @@ int64_t get_row_count(const char *table_name, const bool memory)
 		log_err("Failed to prepare statement to get size of in-memory table %s: %s",
 		        table_name, sqlite3_errmsg(db));
 		sqlite3_free(query);
-		dbclose(&db);
+		if(!memory)
+			dbclose(&db);
 		return -3;
 	}
 	sqlite3_free(query);
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -72,12 +72,30 @@ void _dbclose(sqlite3 **db, const char *func, const int line, const char *file)
 		log_debug(DEBUG_DATABASE, "Closing FTL database in %s() (%s:%i)", func, short_path(file), line);
 
 	// Only try to close an existing database connection
-	int rc = SQLITE_OK;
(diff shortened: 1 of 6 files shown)


DL6ER (changes requested):
The diagnosis is right: `sqlite3_close()` refuses with `SQLITE_BUSY` while a statement is alive and does nothing, and dropping the pointer afterwards leaves the connection unreachable. The finalize loop is what actually fixes that - `sqlite3_close_v2()` on its own would not, because a zombie connection is only freed once the last statement finalizes, and a leaked statement is unreachable by definition.

What needs sorting first is that the loop makes `_dbclose()` destructive when it is handed a connection it does not own.

`get_row_count()` in `common.c` takes `db = memory ? get_memdb() : dbopen(true, false)`. The success path closes it under `if(!memory)`, but the prepare-failure path calls `dbclose(&db)` unconditionally. So with `memory` set, `_dbclose()` runs on `_memdb`. Today that is inert: the statements `query-table.c` caches on that connection make `sqlite3_close()` return `SQLITE_BUSY` and nothing happens. With this change the loop finalizes all of them, leaves those file-scope pointers dangling and `sqlite3_close_v2()` then tears the connection down, so every later use of the in-memory query database is a use-after-free.

It takes some doing to get there: the only callers passing `memory` are arguments to a `log_debug(DEBUG_TIMING, ...)`, so `debug.timing` has to be on before they are evaluated at all, and the prepare has to fail on top of that. Narrow, but the guard is one line and the failure mode is not one to leave standing.

I went through the other places that hold a memdb handle and this is the only one that reaches `dbclose()` unguarded, so it is one missing guard rather than a pattern.

Concretely:

- `common.c`, `get_row_count()`: give the prepare-failure `dbclose(&db)` the same `if(!memory)` guard the success path already has. This is the instance.
- `common.c`, `_dbclose()`: return early, before the finalize loop, when `*db` is the shared in-memory handle. That fixes the class rather than the instance, and matters because call-site discipline has already failed once.
- `gravity-db.c`, `gravityDB_close()`: check the `sqlite3_close()` return value and log a refusal. It finalizes the per-client and all six shared statements first so it should not normally refuse, but today a refusal there would be silent and permanent in the same way.

DL6ER (approved):
Please fix the PR text according to you new changes, otherwise fine and can be merged as-in. Thank you again for your continued work on improving Pi-hole. I will try to keep up with timely reviewing.

PR #3041, opened 2026-08-25
Search gravity on a connection of its own

010011110:

api_search() walks a table cursor on the shared gravity connection and takes no lock, while FTL_reload_all_domainlists() holds the SHM lock and reopens that connection under it. a GET /api/search that runs into a list edit setting RELOAD_GRAVITY therefore has its cursor pulled away mid-walk, which is the case gravityDB_close() had to be made survivable for in #3036

taking the SHM lock like api_list_read() does would be wrong here. a partial search is a LIKE '%term%' over the whole gravity table joined against adlist, so it cannot use an index and scans everything, and DNS would block for as long as that takes. api_list_read() gets away with the lock because it is a single indexed lookup on domainlist

the search opens its own read-only connection instead, the way gravity_updated() and the write paths already do, and closes it when it is done. gravityDB_readTable() takes the connection to use and NULL keeps the shared one, so api_list_read() is unchanged. gravityDB_readTableGetRow() reports errors on the connection its cursor belongs to rather than on the global

searching in the web interface returns what it did before, on all of domains, gravity and the regex lists. the suite passes. overall the defect is hard to provoke but it happens. on retry it usually immediately works so users are likely not reporting it. the code change is small and review by code should make clear that this is a hardening mechanism more than a specific bugfix

## Additional information

---

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -339,6 +339,48 @@ static void gravity_check_list_presence(void)
 }
 
 // Open gravity database
+// Connection settings every gravity connection gets. They live in one place
+// because a pragma added to one opener and forgotten in the other does not fail
+// anywhere, it just makes that connection slower - which is how the search
+// connection ended up without them
+static const struct gravity_pragma {
+	const char *sql;
+	bool required;
+} gravity_pragmas[] = {
+	// Temporary tables, indices and views are read from memory rather than disk
+	{ "PRAGMA temp_store = MEMORY", true },
+	// Read B-tree pages straight from the kernel page cache by virtual address
+	// instead of pread() plus a copy. gravity.db is effectively read-only at
+	// runtime (journal_mode = OFF, "pihole -g" swaps in a new file), 256 MiB
+	// covers every real-world gravity, and SQLite falls back to regular I/O
+	// where mmap is unavailable. The in-process page cache is raised globally
+	// via -DSQLITE_DEFAULT_CACHE_SIZE, so no cache_size pragma here
+	{ "PRAGMA mmap_size = 268435456", false },
+};
+
+// Returns false only when a required pragma failed
+static bool gravity_apply_pragmas(sqlite3 *db, const char *func)
+{
+	for(unsigned int i = 0; i < ArraySize(gravity_pragmas); i++)
+	{
+		const struct gravity_pragma *p = &gravity_pragmas[i];
+		char *zErrMsg = NULL;
+		if(sqlite3_exec(db, p->sql, NULL, NULL, &zErrMsg) == SQLITE_OK)
+			continue;
+
+		if(p->required)
+			log_err("%s(%s) - SQL error: %s", func, p->sql, zErrMsg);
+		else
+			log_warn("%s(%s) - SQL error: %s", func, p->sql, zErrMsg);
+		sqlite3_free(zErrMsg);
+
+		if(p->required)
+			return false;
+	}
+
+	return true;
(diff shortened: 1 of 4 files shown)


DL6ER (changes requested):
Two things before this goes in.

**1. The comments name the one case we already handle.** `gravityDB_close()` uses `sqlite3_close_v2()` precisely so a cursor an API thread is walking survives a reload, and [says so](https://github.com/pi-hole/FTL/blob/3eeb0ae68f5119f85faec47e45fb5528965ff50b/src/database/gravity-db.c#L1189-L1195). "The shared connection can be closed by a gravity reload while we walk our cursor" contradicts that, in [search.c](https://github.com/pi-hole/FTL/blob/3eeb0ae68f5119f85faec47e45fb5528965ff50b/src/api/search.c#L233-L238) and again in [gravityDB_readTable()](https://github.com/pi-hole/FTL/blob/3eeb0ae68f5119f85faec47e45fb5528965ff50b/src/database/gravity-db.c#L2451-L2457).

The race you do close is a different one: `gravityDB_readTable()` reads the global into `db`, builds the query string, and only then prepares on it. A reload landing in that window, with no statement outstanding, has `close_v2()` free the handle right away and we prepare on freed memory. On top of that `api_search()` makes six of these calls without the lock, so a reload between two of them surfaces as `gravity_db == NULL`, i.e., a spurious 400, or as an answer split over two gravity generations. Pinning one connection fixes both - please say that in both comments.

**2. The new connection drops the pragmas.** `gravityDB_open()` sets [`temp_store = MEMORY` and `mmap_size = 268435456`](https://github.com/pi-hole/FTL/blob/3eeb0ae68f5119f85faec47e45fb5528965ff50b/src/database/gravity-db.c#L375-L410), [`gravityDB_open_RO()`](https://github.com/pi-hole/FTL/blob/3eeb0ae68f5119f85faec47e45fb5528965ff50b/src/database/gravity-db.c#L2419-L2435) sets neither, so a search is now slower than it was before this PR, on the one path that scans the largest table unindexed. `gravity_write_open()` and `gravity_updated()` are no precedent - one is a writer, the other reads a single row from `info`. Set both here too.

src/database/gravity-db.c
```diff
@@ -339,6 +339,48 @@ static void gravity_check_list_presence(void)
 }
 
 // Open gravity database
```
DL6ER:
```suggestion
```
You additional code left this as orphan, needs to be re-added further down

src/database/gravity-db.c
```diff
@@ -339,6 +339,48 @@ static void gravity_check_list_presence(void)
...
+			log_warn("%s(%s) - SQL error: %s", func, p->sql, zErrMsg);
+		sqlite3_free(zErrMsg);
+
+		if(p->required)
+			return false;
+	}
+
+	return true;
+}
+
```
DL6ER, line 382:
```suggestion

// Open gravity database (read-write mode)
```

DL6ER (dismissed):
Just one orphaned comment for cleanup, otherwise cleanly done

010011110:
branch rewritten, your two suggestion commits are gone from it and the same two lines are back as a fixup on top of c2c9fd57, so the tree is what you asked for

one thing came with it: the log_debug right below now says read-write mode, it said read-only while the open has always been SQLITE_OPEN_READWRITE. predates this PR, shout if you would rather have it separate

ready for another look

DL6ER (approved):

PR #3042, opened 2026-08-25
Close the raw socket when dropping privileges fails

010011110:

dhcpv6_discover_iface() creates a raw socket, drops privileges and only then hands the socket to do_discoverv6(). when the setuid() fails it returns without closing it. the process exits right after, so nothing is really lost, it is just untidy. this is not a fix but a cleanup to not leave anything behind future ai workflows may find and report about

found with gcc -fanalyzer over the tree

pihole-FTL dhcpv6-discover behaves as before, the path only differs when dropping privileges fails

## Additional information

---


darkexplosiveqwx (approved):
You're right that leaking the file descriptor has no real impact due to the short lifetime of the process when taking that path. Given that your fix is quite simple to please the analyzers, I think its fine to include it.

DL6ER (approved):

PR #3048, opened 2026-08-27
Check DHCP addresses against the configured netmask

DL6ER:

The DHCP validation rejected any address ending in `.0` or `.255`, which is only correct for a /24. With `dhcp.netmask = "255.255.0.0"`, network and broadcast are, e.g., 10.20.0.0 and 10.20.255.255, so a pool ending at 10.20.5.255 is a perfectly ordinary range and was refused nevertheless.

Both addresses are now derived from the netmask and only those two are rejected, for start, end and router alike. Without a configured netmask there is no way to know the one dnsmasq takes from the interface, so the fallback is the classful network, as dnsmasq does for relayed networks, and the usual /24 stays as protected as it was.

As `dhcp.netmask` was only checked to be a valid dotted quad, a mask that is not a contiguous block of one-bits is now rejected before it is used: such a mask has neither a network nor a broadcast address and would only lead to confusing rejections further down.

The second commit moves that question to the moment the value is set. Whether the one-bits are contiguous needs no other config item, so `validate_netmask()` answers it for the CLI, the API and `FTLCONF_` variables alike, instead of only when DHCP is switched on. The check in `write_dnsmasq_config()` stays as the backstop, because reading `pihole.toml` does not run validators at all.

```
pihole-FTL --config dhcp.netmask 255.255.0.0
pihole-FTL --config dhcp.start 10.20.5.1
pihole-FTL --config dhcp.end 10.20.5.255
pihole-FTL --config dhcp.router 10.20.3.1
pihole-FTL --config dhcp.active true
```

This is accepted now and failed with "DHCP end address is not valid" before. An end of 10.20.255.255 or a start of 10.20.0.0 are still rejected, now naming the broadcast resp. network address, and 192.168.1.255 stays rejected both with `dhcp.netmask` set to 255.255.255.0 and with it left empty.

```
pihole-FTL --config dhcp.netmask 255.254.255.0
```

is refused straight away, with or without DHCP enabled. `test/test_suite.bats` covers the netmask validator; the address checks still have no dedicated automated test.

## Additional information

**Related issue or feature (if applicable):** Fixes https://github.com/pi-hole/FTL/issues/3043

---

## Checklist:

### DIFF
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -832,7 +832,7 @@ void initConfig(struct config *conf)
 	conf->dhcp.netmask.t = CONF_STRUCT_IN_ADDR;
 	conf->dhcp.netmask.f = FLAG_RESTART_FTL;
 	memset(&conf->dhcp.netmask.d.in_addr, 0, sizeof(struct in_addr));
-	conf->dhcp.netmask.c = validate_stub; // Only type-based checking
+	conf->dhcp.netmask.c = validate_netmask;
 
 	conf->dhcp.leaseTime.k = "dhcp.leaseTime";
 	conf->dhcp.leaseTime.h = "If the lease time is given, then leases will be given for that length of time. If not given, the default lease time is one hour for IPv4 and one day for IPv6.";
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -263,6 +263,37 @@ static void write_config_header(FILE *fp, const char *description)
 	CONFIG_CENTER(fp, HEADER_WIDTH, "%s", "################################################################################");
 }
 
+// The netmask of the subnet the DHCP server is going to serve. If it is not
+// configured, we have to guess: dnsmasq takes it from the interface it serves
+// on, which we cannot know here, so we fall back to the classful network the
+// address belongs to (what dnsmasq itself does for relayed networks)
+static uint32_t dhcp_netmask(struct config *conf)
+{
+	const uint32_t netmask = ntohl(conf->dhcp.netmask.v.in_addr.s_addr);
+	if(netmask != 0)
+		return netmask;
+
+	const uint32_t addr = ntohl(conf->dhcp.start.v.in_addr.s_addr);
+	if((addr & 0x80000000u) == 0)
+		return 0xFF000000u; // class A
+	if((addr & 0xC0000000u) == 0x80000000u)
+		return 0xFFFF0000u; // class B
+	return 0xFFFFFF00u; // class C
+}
+
+// Neither the network nor the broadcast address of the subnet can be used by a
+// client. Which addresses these are depends on the netmask, e.g., x.x.x.255 is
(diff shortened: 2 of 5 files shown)


rdwebdesign (dismissed):

rdwebdesign:
**NOTE:**

If you configure DHCP options (`dhcp.netmask`, `dhcp.router`, `dhcp.end` and `dhcp.start`) while `dhcp.enable = false`, any value is accepted.

Then, when you try to enable the DHCP service, the validator won't allow and it will return error messages explaining what is wrong.

<details><summary>Example:</summary>

```sh
piholetest:/# pihole-FTL --config dhcp.end 192.168.0.255
192.168.0.255
piholetest:/# pihole-FTL --config dhcp.start 192.168.0.1
192.168.0.1
piholetest:/# pihole-FTL --config dhcp.router 192.168.0.1
192.168.0.1
piholetest:/# pihole-FTL --config dhcp.netmask 255.254.255.0
255.254.255.0
piholetest:/# pihole-FTL --config dhcp.active true
Unable to update dnsmasq configuration: DHCP netmask is not valid
piholetest:/# pihole-FTL --config dhcp.netmask 255.255.254.0
255.255.254.0
piholetest:/# pihole-FTL --config dhcp.active true
Unable to update dnsmasq configuration: DHCP router address should not be within DHCP range
piholetest:/# pihole-FTL --config dhcp.start 192.168.0.2
192.168.0.2
piholetest:/# pihole-FTL --config dhcp.active true
true
```

</details>

DL6ER:
Correct, and that is the `dhcp.active` guard rather than anything this PR changes: all early DHCP checks in `write_dnsmasq_config()` sit inside `if(conf->dhcp.active.v.b)`, so with DHCP off none of them run. On their own the four items only get type-based validation (`validate_stub`), which is why any dotted quad is accepted.

I think this is fine as it stands. While `dhcp.active = false` the values are inert: the `dhcp-range=` and `dhcp-option=` lines are written under the same condition, so nothing is handed to dnsmasq and no lease can come out of them. The first moment they could do harm is the moment the check runs.

Lifting the guard would also be worse than it looks. `dhcp.start` defaults to `0.0.0.0` and `readFTLconf()` calls `write_dnsmasq_config()` at startup, so every default install would fail its own config test at boot.

The netmask is the one check that needs no other field, so rejecting a non-contiguous mask at set time is a sensible follow-up. Network, broadcast and router-in-range need `dhcp.start`, `dhcp.end` and `dhcp.netmask` together, so those stay where they are.

DL6ER:
The netmask part turned out small enough to keep here rather than in a follow-up, so it is now the second commit: `validate_netmask()` rejects a non-contiguous mask at set time, on the CLI, the API and `FTLCONF_` alike. The rest is unchanged, and the check in `write_dnsmasq_config()` stays as the backstop for a hand-edited `pihole.toml`. Worth a second look if you want one.

rdwebdesign (approved):

PR #3049, opened 2026-08-28
cache API doc asset over builds

010011110:

two cmake rules that embed a generated file into the binary re-ran on every build

the api docs rule declares OUTPUT hex/<name>.h, which cmake resolves against the build tree, while the command writes into the source tree. cmake never sees that output appear, so it re-runs all 29 assets through gzip and xxd on every single build and recompiles docs.c behind them. it also has no DEPENDS, so a changed asset does not regenerate on its own. that goes unnoticed because the rule runs every time anyway

the lua scripts rule passes SOURCES where it means DEPENDS. add_custom_command has no SOURCES keyword, so the word and the path after it are absorbed as two further OUTPUT entries: a file named SOURCES that the command never writes, and inspect.lua itself, declared as an output of the rule that reads it. the first keeps the rule permanently out of date, the second makes every build run touch_nocreate on a tracked source file. inspect.lua.hex is rewritten each time and ftl_lua.c.o recompiles behind it

in both rules OUTPUT now names the file the command actually writes and DEPENDS tracks the input. in the docs rule the list(APPEND COMPILED_RESOURCES ...) below it appended to the list the foreach was iterating over, which cmake copies, so it never did anything and is gone

build twice in a row without changing anything, the second build compresses and compiles nothing where it used to redo all 29 doc assets plus inspect.lua and two objects. touch src/api/docs/content/index.css or src/lua/scripts/inspect.lua and only that one asset is regenerated. the suite passes, 192 bats and 151 pytest

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/docs/CMakeLists.txt
+++ b/src/api/docs/CMakeLists.txt
@@ -69,11 +69,11 @@ foreach(INPUT_FILE ${COMPILED_RESOURCES})
     string(REPLACE "/" "_" VARIABLE_NAME ${INPUT_FILE})
     set(OUTPUT_FILE ${CMAKE_CURRENT_SOURCE_DIR}/hex/${INPUT_FILE_MODIFIED}.h)
     add_custom_command(
-        OUTPUT hex/${INPUT_FILE_MODIFIED}.h
+        OUTPUT ${OUTPUT_FILE}
         COMMAND ${GZIP_COMPRESSOR} -9 -n -c ${IN} | ${RESOURCE_COMPILER} -iC -n ${VARIABLE_NAME} - | sed "s/unsigned /static const unsigned /" > ${OUTPUT_FILE}
+        DEPENDS ${IN}
         COMMENT "Compressing and compiling ${INPUT_FILE}"
         VERBATIM)
-    list(APPEND COMPILED_RESOURCES ${OUTPUT_FILE})
 endforeach()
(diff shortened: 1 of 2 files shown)


DL6ER (approved):

DL6ER:
I reproduced both rules standalone. Two confirmations and two notes.

1. The docs rule really did redo everything on every build. With the old `OUTPUT hex/x.h` against a command writing to `${OUTPUT_FILE}`, the second build recompresses the asset and recompiles `docs.c`. With the patch it is a no-op, and touching `content/index.css` regenerates that one asset only.

2. The `SOURCES` reading is confirmed in the generated `build.make`:

   ```
   src/SOURCES: src/inspect.lua.hex
   	@$(CMAKE_COMMAND) -E touch_nocreate src/SOURCES
   	@$(CMAKE_COMMAND) -E touch_nocreate .../src/inspect.lua
   ```

3. `hex/*.h` are declared source tree outputs now, so `make clean` deletes them out of `src/api/docs/hex/`. They are gitignored and come back on the next build, so nothing breaks - just worth knowing.

4. The `GLOB_RECURSE` still has no `CONFIGURE_DEPENDS`, so a newly added asset needs a re-configure before it is seen. `sources` is hand maintained anyway, so I would leave that as it is.

PR #3052, opened 2026-08-29
give the internal ptr resolver a connected socket per lookup

010011110:

create_socket() connected only in the tcp case, so the udp socket the internal ptr resolver uses never received icmp errors. with nothing listening on the dns port the kernel drops the port unreachable and recvfrom sits out the full two second SO_RCVTIMEO, once for every client in the sweep. the ECONNREFUSED string in strsockerr() was unreachable on this path

connecting the udp socket makes the kernel report that right away, and it drops datagrams from foreign sources for us. a connected socket queues icmp errors on itself and hands them out on the next call though, so one socket for a whole sweep would surface one client's refusal on the next client's lookup. resolveAndAddHostname() creates and closes its own socket per lookup instead, next to the tcp socket it already handles that way. one socket per lookup also means a fresh source port each time, so a late reply to the previous query is dropped by the kernel rather than having to be recognised as stale

rebased on top of #2979. its reply correlation is untouched and stays as the layer that holds if socket reuse ever comes back, but the manual source comparison in the receive loop goes, a connected socket has the kernel do exactly that. checked rather than assumed: a valid correlated reply sent to ftl's own source port from a second port never reaches the socket, ftl sits out its full deadline as if nothing was sent

one line in ptr_response_regression.c, the harness builds its client socket with a bare socket() and relied on the destination travelling in dest. ftl has no unconnected resolver sockets left, so the harness has to connect too or it hangs, and it would be pinning down a setup that no longer exists

pihole-FTL ptr with nothing listening on the configured port, two seconds and a timeout before, 0.002 s and "Connection refused by upstream DNS server" after. the new bats test pins that down, ptr_response_regression passes plain and with the sanitizers. 194 bats and 151 pytest

## Additional information

**Related issue or feature (if applicable):** https://github.com/pi-hole/FTL/issues/2978

---

## Checklist:


DL6ER:
@010011110 Please rebase this. You may need to redo it even from scratch as #2979 landed two hours ago. It touches the same file you are doing many changes in.

010011110:
rebased, one commit on top of #2979 now. resolve.c was redone from scratch against the new base rather than merged, the old hunks did not survive it

two things in there that touch your side. the manual source comparison in the udp receive loop is gone, connect() has the kernel do that for us. checked rather than assumed: a valid, correctly correlated reply sent to ftl's own source port from a second port never reaches the socket at all, ftl sits out its full deadline as if nothing was sent. the correlation from #2979 stays untouched, it is what catches a stale reply from the right source, which the kernel cannot

ptr_response_regression.c needed one connect() on its client socket. it builds a bare unconnected one, and with the destination no longer travelling in dest the query never left, so the harness hung in pthread_join instead of failing. ftl has no unconnected resolver sockets left, so a harness with one would be pinning down a setup that does not exist. happy to split that into its own commit if you would rather see it separately

DL6ER:
Four things before it goes in.

1. `dest` is dead now. `ngethostbyname()` no longer reads it, only `create_socket()` fills and uses it, yet it is still threaded through `resolveHostname()` at every call site plus the harness prototype. The harness sets `config.dns.port` itself, so the whole `struct sockaddr_in dest` plumbing can go and `create_socket()` becomes `create_socket(bool tcp)`.

2. The receive error is logged twice, and this PR is what makes that path reachable:

   ```
   Could not recvfrom() in ngethostbyname() (src/resolve.c:588): Connection refused
   Cannot receive UDP DNS reply: Connection refused by upstream DNS server
   ```

   Same thing I fixed for TCP in 3126faff. There is no `recvfrom_nowarn`, but the socket is connected now, so `recv_nowarn(sock, buf, sizeof(buf), MSG_DONTWAIT)` silences the generic warning and drops the two `NULL` arguments in one go. Likewise `send(sock, buf, len, 0)` for the query, matching the TCP branch below it.

3. After setting `*sock_error` the function still falls into the `getNameFromIP()` fallback, and on a hit `addstr()` leaves a string in shared memory that the caller discards by returning. Return right there instead.

4. The log line you moved into `resolveUpstreams()` still says "client host name resolution failed".

One for the harness: it now reuses a single connected socket across both lookups, a shape the resolver no longer has. That is fine as a unit test of the correlation from #2979, but put that in the file - otherwise someone aligns it with production later and takes the coverage with it.

010011110:
all five in, one thing pushed back on

1. gone. create_socket(bool tcp) now, dest is a local in there, and it is off resolveHostname(), ngethostbyname(), resolve.h, args.c and the harness prototype

2. done both halves. recv_nowarn drops the wrapper warning and the two NULLs in one go. for the send side there was no sendto_nowarn, so FTLsendto() now takes the same warn flag FTLrecv() already has, which keeps its EINTR retry that a plain send() would have lost. every caller goes through the macro so only the macro and the two definitions moved, plus the stub in the harness. one log line on a refused lookup now

3. fixed, and it was a real leak rather than just waste: on a getNameFromIP hit addstr() left a string in shared memory that the caller then threw away. returns oldnamepos right after setting sock_error now, before the fallback

4. says upstream now

5. called out in the file, that the reuse is deliberate and that a fresh port per lookup would have the kernel drop the stale reply, so the correlation under test would never be reached

DL6ER:
Two small things.

`refute_output --partial "Timeout - no response from upstream DNS server"` cannot fail. That string only comes from `strsockerr(EAGAIN)`, and the UDP loop `continue`s on `EAGAIN`, so a real timeout takes the `poll()` branch and logs `Cannot receive UDP DNS reply: Timed out after %d ms` instead. Refuting `Timed out after` makes the assertion do what it says.

The second is taste: `resolveAndAddHostname()` now returns a position plus two independent `bool` out-parameters. An enum return would carry that more directly, but it works as is.

The rest looks right to me, and `tcp_sock > 0` -> `>= 0` is a good catch.

010011110:
good catch, that assertion was decorative. the string only comes from strsockerr(EAGAIN) and the udp loop continues on EAGAIN, so nothing on that path can ever log it. it came from writing the test against the pre-#2979 base, where SO_RCVTIMEO did produce it, and carrying the line through the rebase without rechecking it

refutes "Timed out after" now, which is what the poll branch actually logs. checked that it is the right string rather than assuming again: an unanswered lookup on this build logs "Cannot receive UDP DNS reply: Timed out after 2000 ms", so the two cases are told apart properly

left the out-parameters as they are for now since you called it taste, say if you would rather have the enum and i will do it

010011110:
went with the enum after all, ignore the "left it as is" in my last comment, it crossed with the decision to do it

resolveAndAddHostname() returns enum resolve_result now with the name position as an out-parameter: RESOLVE_OK, RESOLVE_NO_NAME, RESOLVE_NO_SOCKET. the call sites read if(res == RESOLVE_NO_SOCKET) for the sweep abort and if(res != RESOLVE_OK) where if(!success) was, with no flag preamble

it also drops an implicit bit of the old shape: "resolution is switched off for this address" had to be encoded as success with an empty name, it is RESOLVE_OK with position 0 now

194 bats, 151 pytest, harness passes

DL6ER (approved):
Thanks for you contributions to FTL lately, I saw that you also started stepping into a few issues to help there. That's great 👍

PR #3053, opened 2026-08-30
generate the api documentation assets into the build tree

010011110:

the api documentation assets are generated into the source tree, one hex directory shared by every branch and every build directory. a file left there by a build of another revision is newer than its input, so DEPENDS never fires and the wrong assets stay embedded. a tree that has built master keeps 29 uncompressed assets, development then hands them out with content-encoding gzip and /api/docs fails with "incorrect header check"

they go to CMAKE_CURRENT_BINARY_DIR now, the way src/lua/scripts/CMakeLists.txt already does. every build directory gets its own copy, so nothing is shared and a branch switch costs only the assets whose input actually changed, no clean and no sweep needed

one wrinkle the move needs on top: docs.h includes hex/... by relative path, and a quoted include is searched next to the including file before the include path, so a leftover src/api/docs/hex would shadow the generated one for good and nothing would ever rewrite it. that is worse than what we have today, so the old directory is removed once at configure time. it triggers itself, changing CMakeLists.txt forces a reconfigure, so it is gone on the first build after the upgrade

the two sweeps in build.sh go with it. the api/docs one existed because the rule had no DEPENDS and a changed asset would otherwise never regenerate, it has one since #3049. the lua one looks for src/lua/scripts/*.lua.hex, which that rule has never written, its output has always gone to the build tree. neither of them caught the case above either, both only remove a hex file older than its input. .gitignore loses the two matching entries

put junk in src/api/docs/hex/index_html.h on development and build: it survives and gets embedded. with this change the directory is gone after configure, the 29 assets land in the build tree, a null rebuild regenerates nothing, touching content/index.css regenerates that one, and a second build directory gets its own copy without disturbing the first. 193 bats and 151 pytest

## Additional information

**Related issue or feature (if applicable):** follow-up to #3049

---

## Checklist:


DL6ER (changes requested):
The diagnosis is right and I checked each part of it: the lua loop scans for files `src/lua/scripts/CMakeLists.txt` never writes, since its `OUTPUT_FILE` goes to `CMAKE_CURRENT_BINARY_DIR`; the api/docs rule does carry `DEPENDS` since #3049; and I reproduced a planted asset surviving `./build.sh clean nobuild` on `development` and being removed with your change. A full `clean` build afterwards regenerates all 29 assets correctly.

Two things before this goes in.

`.gitignore` still has `/src/lua/scripts/*.hex`, which is the same leftover as the sweep you are deleting. Drop it here too.

The bigger one: I would rather not have `clean` be the mechanism we depend on. Cleaning after every branch switch costs a full recompile, which is exactly what I want to avoid. The dependency tracking we need is already in place - `DEPENDS` is set - what defeats it is that `OUTPUT_FILE` points at `${CMAKE_CURRENT_SOURCE_DIR}/hex`, so a single shared directory is written by every branch and every build directory, and a file from another revision is newer than its input. Point the output at `${CMAKE_CURRENT_BINARY_DIR}` the way `src/lua/scripts/CMakeLists.txt` already does and the problem goes away without any sweep and without a clean: every build directory gets its own assets, and cmake regenerates just the ones whose input actually changed.

Keeping the `rm -rf` in `clean` on top of that is fine, it just should not be what we rely on.

010011110:
done, the assets go to CMAKE_CURRENT_BINARY_DIR now and the clean based approach is gone

one thing the move needed on top: docs.h includes hex/... by relative path, and a quoted include is searched next to the including file before the include path, so a leftover src/api/docs/hex would shadow the generated one for good and nothing would ever rewrite it. that is worse than what we have today, so the old directory is removed once at configure time. it triggers itself, changing CMakeLists.txt forces a reconfigure, so it is gone on the first build after the upgrade

checked with a legacy directory planted first: configure removes it, 29 assets land in the build tree gzipped, null rebuild regenerates nothing, touching content/index.css regenerates exactly that one, and a second build directory gets its own 29 without touching the first

dropped both .gitignore entries, the lua one you pointed at and src/api/docs/hex which the move makes obsolete. also dropped the rm -rf from build.sh clean, the configure time removal covers it and leaving it in would be one more sweep that is not needed

193 bats, 151 pytest, and the source tree has no hex directory left after a full ci build

010011110:
body and title rewritten to match, both still described the clean based version

DL6ER (dismissed):
This is what I had in mind, and quickly done. I went over it: editing a content file regenerates the asset on a plain incremental build, a second build directory stays independent, and a stale `src/api/docs/hex` planted by hand is removed at configure time instead of shadowing the generated one. Good catch on the quoted include being searched next to `docs.h` first - that one would have been unpleasant to debug later.

Two small things. Take them here or leave them for follow-up work, neither blocks this:

`.gitignore` now carries two section comments with nothing under them, `# Documentation files generated by cmake` and `# Compiled LUA scripts`.

The `file(REMOVE_RECURSE ...)` runs on every configure and only earns its keep for as long as old checkouts still have the directory. Switching `docs.h` to angle includes would skip the including file's own directory outright, so that line could go away entirely at some point.

010011110:
took the first one here, both headings are gone. that was mine, i removed the entries and left their section comments behind

left the angle include for a follow-up. it touches the 30 include lines in docs.h and whether project generated headers are <> or "" is a call for you rather than something to slip into this one. once that is in, the file(REMOVE_RECURSE) has nothing left to protect and can go with it

no test run for the last push, it is two comment lines out of .gitignore and cannot reach the build

DL6ER (approved):
Re-approving, the force-push dismissed the previous one. The section headings are gone, and leaving the angle includes for later is fine by me - that is a separate question about how we include generated headers in general, and it does not belong in this change.

Nothing but `.gitignore` moved since my last pass, so the rest still stands as reviewed.

PR #3057, opened 2026-09-01
catch a pihole.toml written before the watcher exists

010011110:

nothing watches /etc/pihole for most of ftl's startup. the inotify instance is created by the gc thread, which starts long after readFTLconf() has loaded the file, with delay_startup() and all of dnsmasq's initialization in between. a pihole.toml written in that stretch is never announced, so ftl keeps serving what it loaded while the file on disk says something else, and it stays that way until the file is written a second time. for a setting put in place once while a box is provisioned that can be for good

creating the watcher earlier does not cover it either. writeFTLtoml() arms one whenever it rewrites the file, startup included, but create_inotify_watcher() overwrites inotify_fd without looking at what is already there, so the gc thread's instance replaces that one along with whatever it had collected

the window is closed from the other end instead. the watcher goes up first, which leaves every later change certain to be reported, and the file is then compared against the loaded configuration once. a normal start pays a single sha256 of pihole.toml for it, reread_config() gives up as soon as the checksum turns out to be the one it recorded

mid impact as i read it. the window is narrow and a hand edited config will almost never land in it, but a provisioning script that writes pihole.toml straight after starting ftl lands in it every time, and what follows is silent, indefinite and looks exactly like the setting having been ignored. the cost on every other start is one checksum

widen the window with a sleepms(10000) in front of watch_config() in GC_thread(), start ftl and write into pihole.toml while it sleeps. FTL.log then gets "Reloading config due to pihole.toml change" and ftl rewrites the file with the "### CHANGED" marker on that setting, on development neither happens and the running config keeps what it loaded at startup. 194 bats and 151 pytest

## Additional information

---

## Checklist:

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -627,6 +627,11 @@ void *GC_thread(void *val)
 	// Create inotify watcher for pihole.toml config file
 	watch_config(true);
 
+	// The watcher above reports only what follows it, so compare the file
+	// against the loaded configuration once to pick up anything written
+	// during startup
+	reread_config();
+
 	// Run as long as this thread is not canceled
 	while(!killed)
 	{
--- a/src/main.c
+++ b/src/main.c
@@ -14,6 +14,8 @@
 #include "config/setupVars.h"
 #include "args.h"
 #include "config/config.h"
+// watch_config()
+#include "config/inotify.h"
 #include "main.h"
 // exit_code
 #include "signals.h"
@@ -141,7 +143,15 @@ int main (int argc, char *argv[])
 	// Skip it here if we jump back to this point from die()
 	const int jmpret = setjmp(exit_jmp);
 	if(jmpret == 0)
+	{
+		// main_dnsmasq() opens by closing every descriptor it
+		// inherited from us, a config watcher armed while the config
+		// was read included, and the numbers are reissued afterwards.
+		// Close it here, while it is still ours to close
+		watch_config(false);
+
 		main_dnsmasq(argc_dnsmasq, (char**)argv_dnsmasq);
+	}
 	else
 	{
 		// We are jumping back to this point from dnsmasq's die()


DL6ER:
The mechanism is right, and arming the watcher before the comparison rather than after is the part that matters - anything written from that point raises an event, anything earlier the checksum finds. `last_checksum` is only ever set by `writeFTLtoml()`, and `readFTLconf(rewrite=true)` always ends there, so an untouched start really does stop at the `memcmp`.

Two things:

1. The comment tells the bug story rather than what the code does. "Nothing was watching /etc/pihole while the rest of the startup ran ... produced no event at all" describes a state that stops existing the moment this merges, and the commit message already carries that account in far more detail. Something like "the watcher above only reports what follows it, so compare the file against the loaded config once to pick up anything written during startup" ages better.

https://github.com/pi-hole/FTL/blob/9a7c5b3bcf1e83cb1fb1dae2f898140739a79659/src/gc.c#L627-L635

2. Not this PR's doing, but it follows directly from what the commit message describes: since `create_inotify_watcher()` assigns `inotify_fd` unconditionally, the `watch_config(true)` right above leaks the instance `writeFTLtoml()` armed on any start that rewrote `pihole.toml` - one fd and one watch, for the life of the process. Closing an existing instance first would settle it.

https://github.com/pi-hole/FTL/blob/9a7c5b3bcf1e83cb1fb1dae2f898140739a79659/src/config/inotify.c#L32-L40

010011110:
comment reworded along your lines, it says what the code does and the story stays in the commit message

second commit in here takes care of the descriptors, from a slightly different angle than closing them in create_inotify_watcher(). main_dnsmasq() opens with close_fds(max_fd, -1, -1, -1) at dnsmasq.c:171 and writeFTLtoml() arms that early watcher back in readFTLconf(), so by the time the gc thread runs the instance has been closed for a while and the number is in use elsewhere. probe printing readlink of /proc/self/fd/N right where the close would go:

```
PROBE: inotify_fd 3 (wd 1) currently is socket:[1695386337]
WARNING: Cannot remove watch from inotify instance: Invalid argument
```

which takes 16 of the 25 dotdoh tests with it. so the fd and the wd are dropped on the way into main_dnsmasq() instead, at the point where the sweep is what makes them meaningless. nothing is closed there, dnsmasq does that itself, and nothing is left for a stray watch_config(false) to act on:

```
DEBUG_INOTIFY: Created inotify watcher for /etc/pihole
DEBUG_INOTIFY: Dropping inotify watcher for /etc/pihole, dnsmasq closes it
DEBUG_INOTIFY: Created inotify watcher for /etc/pihole
```

194 bats, 151 pytest, 12, 25, 9, all green with both commits

010011110:
second commit reshaped: it closes the watcher on the way into main_dnsmasq() now, rather than dropping the two numbers and letting dnsmasq's sweep do the closing

read_opts() at dnsmasq.c:122 runs ahead of close_fds() at :171, and die() in there longjmps back to the setjmp in main() before the sweep has happened, so numbers merely forgotten would leave that watcher open for the life of the process with nothing able to reach it. closing covers both paths, the descriptor is still ours at that point and the sweep then finds nothing of ours

```
DEBUG_INOTIFY: Created inotify watcher for /etc/pihole   <- writeFTLtoml()
DEBUG_INOTIFY: Closed inotify watcher                    <- handover to dnsmasq
DEBUG_INOTIFY: Created inotify watcher for /etc/pihole   <- gc thread
```

no new function left over either, it is ten lines in main.c. 194 bats, 151 pytest, 12, 25, 9, still green

DL6ER (approved):
Clean nice fix. Thank you again!

PR #3062, opened 2026-09-04
count a database connection from the moment it exists

010011110:

_dbopen() raises dbopen_cnt at the very end, once everything has succeeded, but the two failure paths after the open give the connection back through dbclose(), which lowers it unconditionally. a connection refused for being read-only, or one whose busy handler could not be set, takes the counter down without ever having raised it, so dbopen_cnt walks towards negative numbers for as long as ftl runs

the increment moves up to just after sqlite3_open_v2() returns success. that is the moment there is something to count, and the point from which dbclose() is the way back out

sqlite3_open_v2() also hands back a handle when it fails, and the sqlite documentation is explicit that resources are associated with it whether or not an error occurs. that path returned NULL and dropped it, so every failed open leaked an sqlite3 object. it closes the handle now, with a plain sqlite3_close() rather than dbclose(), because the connection was never counted and dbclose() would take a decrement with it

low impact. the counter feeds one line, the DEBUG_DATABASE "Database busy - waiting %d ms (dbopen: %d)" message, and a failed open is not the common case. but that line exists to say how many connections are open while the database is contended, which is exactly when someone reads it

nothing user visible changes, it is by inspection: the counter is raised and lowered on the same paths now, and the failed-open path releases what sqlite handed it. full suite in the container is 194 bats, 151 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -210,9 +210,18 @@ sqlite3* _dbopen(const bool readonly, const bool create, const char *func, const
 	if( rc != SQLITE_OK )
 	{
 		log_err("Error while trying to open database: %s", sqlite3_errstr(rc));
+
+		// sqlite3_open_v2() associates resources with the handle even
+		// when it fails. dbclose_handle() releases it without the
+		// decrement dbclose() carries, as it is not counted yet
+		dbclose_handle(db);
 		checkFTLDBrc(rc);
 		return NULL;
 	}
+
+	// Count the connection as soon as it exists. The failure paths below
+	// release it through dbclose(), which is the matching decrement
+	atomic_fetch_add_explicit(&dbopen_cnt, 1, memory_order_relaxed);
 
 	// If the database is opened in read-write mode, actually check if it is
 	// writable. If it is not, close the database and return an error
@@ -233,9 +242,6 @@ sqlite3* _dbopen(const bool readonly, const bool create, const char *func, const
 		checkFTLDBrc(rc);
 		return NULL;
 	}
-
-	// Increment the number of open database connections
-	atomic_fetch_add_explicit(&dbopen_cnt, 1, memory_order_relaxed);
 
 	return db;
 }


darkexplosiveqwx (dismissed):

src/database/common.c
```diff
@@ -210,10 +210,19 @@ sqlite3* _dbopen(const bool readonly, const bool create, const char *func, const
 	if( rc != SQLITE_OK )
 	{
 		log_err("Error while trying to open database: %s", sqlite3_errstr(rc));
+
+		// sqlite3_open_v2() hands a handle back even when it fails and
+		// leaves it to us to release. It was never counted, so it is
+		// closed directly rather than through dbclose()
+		sqlite3_close(db);
```
DL6ER: `dbclose_handle()` exists for exactly this case, closing a connection the caller owns without touching the counter, and is what `network-table.c` and `gravity-db.c` use. It also goes through `sqlite3_close_v2()` and logs a close failure, which the raw call swallows. `dbclose_handle(NULL)` returns `SQLITE_OK` right away, so the allocation-failure case stays covered.

```suggestion
		// sqlite3_open_v2() associates resources with the handle even
		// when it fails. dbclose_handle() releases it without the
		// decrement dbclose() carries, as it is not counted yet
		dbclose_handle(db);
```

DL6ER (commented):
The counter fix is right, both failure paths took a decrement they never earned, and the leaked handle is real - SQLite hands one back whether or not the open succeeded.

Two things before this goes in, both inline.

Separately, and outside this diff: the counter drifts the other way once `DBerror` is set, because `_dbclose()` returns at the `FTLDBerror()` check before reaching the decrement and nothing ever clears `DBerror` again. Same class of bug, low impact for the same reason as the rest. Up to you whether it belongs in this commit or a follow-up.

src/database/common.c
```diff
@@ -210,10 +210,19 @@ sqlite3* _dbopen(const bool readonly, const bool create, const char *func, const
...
 		checkFTLDBrc(rc);
 		return NULL;
 	}
 
+	// Count the connection as soon as it exists. The failure paths below
+	// give it back through dbclose(), which decrements unconditionally
```
DL6ER: `dbclose()` does not decrement unconditionally - it returns early on `FTLDBerror()` and for the in-memory handle. Neither can be reached from here, but the comment should claim only what holds.

```suggestion
	// Count the connection as soon as it exists. The failure paths below
	// release it through dbclose(), which is the matching decrement
```

010011110:
both taken. dbclose_handle() is the right call there and i had missed that it exists for exactly that case, and the comment now claims only what holds: dbclose() returns early on FTLDBerror() and for the in-memory handle, neither reachable from here

on the drift you raise outside the diff: that is #3064, which drops the FTLDBerror() early return in _dbclose() so a broken database is closed like any other. the two overlap in common.c and will conflict. merge this one first and i will rebase #3064 on top, it is the smaller of the two rebases

rebased onto current development, 194 bats, 151 pytest, 12, 25, 9

DL6ER (approved):
Both points taken, and `dbclose_handle()` is the right call there.

PR #3063, opened 2026-09-04
include the generated api doc assets with angle brackets

010011110:

follow-up to #3053, where DL6ER suggested this

the assets are generated into the build tree since #3053, but docs.h includes them with quotes, and a quoted include is searched next to the including file before the include path is consulted. a hex/ directory left in src/api/docs/ by a build from before that change is found first and goes on being found, with nothing left that could ever overwrite it. removing that directory at configure time was the guard against exactly this

angle brackets do not search next to the including file, so a leftover cannot shadow anything: api_docs carries CMAKE_CURRENT_BINARY_DIR on its include path, and PROJECT_SOURCE_DIR/src next to it, where a stale src/api/docs/hex is not reachable as hex/... either. the guard goes with the quotes

low impact, it removes a workaround rather than fixing a live bug. the guard does its job today, this makes it unnecessary rather than makes something work that did not

one trade-off worth naming, since the line was doing double duty: #3053 also dropped src/api/docs/hex from .gitignore and from build.sh, so in a checkout that still has those headers they are now untracked and un-ignored, and this line is the only thing sweeping them. without it they sit in git status until someone deletes them by hand. i still think it should go - touching the source tree at configure time is what #3053 moved away from, and a line that needs a backwards-looking comment to justify itself is a line to remove - but it is a cost, not free

the mechanism on its own: two files in one directory including the same name, one quoted and one angled, with a second copy of the header on the -I path. gcc -H shows the quoted one reading the header beside it and the angled one reading the header from -I

in the tree: build, then plant a src/api/docs/hex/index_html.h containing an #error and rebuild without reconfiguring, with cmake --build cmake_ci rather than build.sh, since build.sh reruns cmake and on development the guard would sweep the file first. on development the build stops with that #error, on this branch it is ignored. full suite is 194 bats, 151 pytest, 12, 25, 9

## Additional information

**Related issue or feature (if applicable):** follow-up to #3053

---

## Checklist:


darkexplosiveqwx (approved):

src/api/docs/CMakeLists.txt
```diff
@@ -47,14 +47,6 @@ file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/hex/specs)
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/hex/images)
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/hex/external)
```
DL6ER, line 48: #3053 also dropped `src/api/docs/hex` from `.gitignore` and from `build.sh`, so those headers are now untracked *and* un-ignored in a checkout that has them, and this line is the only thing still sweeping them. Without it they sit in `git status` until someone deletes them by hand.

I lean towards dropping it as you have. Touching the source tree at configure time is what #3053 moved away from, and any comment that justifies keeping the line has to point backwards to explain itself, which is a decent sign it should go.

If you would rather keep it for one cycle, it needs to say what it does rather than what it guarded:

```suggestion
file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/hex/external)

# Nothing generates into the source tree and src/api/docs/hex is not ignored,
# so remove it to keep it out of the working tree
file(REMOVE_RECURSE ${CMAKE_CURRENT_SOURCE_DIR}/hex)
```

Either way, the description should name the trade-off rather than present the line as pure dead weight.

DL6ER (commented):
Checked the mechanism against `development`: `api_docs` carries `CMAKE_CURRENT_BINARY_DIR` on its include path, the sources list is already fully qualified, `src/hex` does not exist, and `docs.h` has exactly one includer. A leftover `src/api/docs/hex` is unreachable as `hex/...`, so the guard really is redundant. That part holds.

The one thing the description leaves out is that the line was doing double duty, which is the inline note.

010011110:
you are right that it was doing double duty and the description hid it. rewritten to name the trade-off: since #3053 dropped src/api/docs/hex from .gitignore and build.sh, a checkout that still has those headers now has them untracked and un-ignored, and this line is the only thing removing them

i have left the line out rather than keeping it for a cycle, for the reason you give: a line that needs a backwards-looking comment to explain itself is a line to remove, and touching the source tree at configure time is exactly what #3053 moved away from. no code change, only the description

DL6ER (approved):
Description names the trade-off now.

PR #3064, opened 2026-09-04
leave the database in a usable state when a write fails

010011110:

three failure paths in the database layer that hand back something worse than they found, grouped because they are the same mistake in three shapes

queries_to_database() clears flags.database.changed in phase 1 under the shm lock, on purpose, so a dns thread updating the query during phase 2 marks it again. every route out after that left the flag clear: the phase1_error return, the BEGIN, a step error that breaks out with succeeded < snap_count, and the END, which rolls everything back, and nothing else will, because dnsmasq_interface.c only sets it when the query itself changes and a query in its final state does not change again. those queries are skipped by every later run and reach neither database, silently. phase 3 walks the uncommitted tail now and says so once through log_err, the failure labels do the same for the whole snapshot, taking the lock where phase 2 gave it up, and phase1_error leaves through them instead of returning on its own

_dbclose() returned before doing anything once FTLDBerror() was latched. that latch is one way, so from the first corruption on, every close leaked its handle, left the caller holding a pointer it believes was closed, and skipped the open counter. sqlite3_close_v2() copes with a corrupt file, so the check goes

db_create() opened a connection and returned on five failure paths without closing it, keeping a half-created pihole-FTL.db open and locked for the life of the process while db_init() reported the failure and carried on. the table statements move into db_create_tables() so every return comes back through db_create(), which closes once

export_queries_to_disk() ended its transaction with a bare SQL_bool(), which returns without rolling back, on the shared in-memory connection that lives as long as ftl. a failing END left every later caller inside a transaction nobody opened

low to medium impact. each needs a failing write or a corrupt database to reach, but what they produce is silent data loss, a file locked until restart, and a connection stuck in a transaction

not reachable without forcing the failures: make a step or an END fail and watch the next run pick the queries up rather than skip them, with "Could not store N queries" in FTL.log. full suite in the container is 194 bats, 151 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/src/database/common.c
+++ b/src/database/common.c
@@ -87,11 +87,6 @@ int _dbclose_handle(sqlite3 *db, const char *func, const int line, const char *f
 
 void _dbclose(sqlite3 **db, const char *func, const int line, const char *file)
 {
-	// Silently return if the database is known to be broken. It may not be
-	// possible to close the connection properly.
-	if(FTLDBerror())
-		return;
-
 	// The shared in-memory connection is owned by close_memory_database() and
 	// its prepared statements live as long as FTL does. Closing it here would
 	// finalize them behind the back of whoever cached them, so return before
@@ -327,12 +322,11 @@ static bool create_counter_table(sqlite3* db)
 	return true;
 }
 
-static bool db_create(void)
+// Split from db_create() so that every failure below returns through it and
+// the connection is closed exactly once. SQL_bool() returns on failure, so the
+// close cannot live in here
+static bool db_create_tables(sqlite3 *db)
 {
-	sqlite3 *db = dbopen(false, true);
-	if(db == NULL)
-		return false;
-
 	// Create Queries table in the database
 	SQL_bool(db, CREATE_QUERIES_TABLE_V1);
 
@@ -350,10 +344,23 @@ static bool db_create(void)
 	if(!db_set_FTL_property(db, DB_LASTTIMESTAMP, 0))
 		return false;
 
-	// Close database handle
+	return true;
+}
+
+static bool db_create(void)
+{
+	sqlite3 *db = dbopen(false, true);
+	if(db == NULL)
+		return false;
+
+	const bool okay = db_create_tables(db);
+
+	// Close database handle whether or not the tables were created: a
+	// half-created database that stays open holds the file for the life of
+	// the process while db_init() carries on without it
 	dbclose(&db);
 
-	return true;
+	return okay;
 }
 
 void SQLite3LogCallback(void *pArg, int iErrCode, const char *zMsg)
--- a/src/database/query-table.c
+++ b/src/database/query-table.c
@@ -1009,8 +1009,16 @@ bool export_queries_to_disk(const bool final)
 		log_debug(DEBUG_DATABASE, "Exported %i rows to disk.%s", sqlite3_changes(memdb), subtable_names[i]);
 	}
 
-	// End transaction
-	SQL_bool(memdb, "END");
+	// End transaction. A bare SQL_bool() would return with the transaction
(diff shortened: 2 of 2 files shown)


010011110:
force-pushed. reviewing this again i found a fourth way out that i had missed: the phase1_error return just after the unlock. it gives up with the snapshot already taken and the changed flags already cleared, so it dropped exactly the queries this pr is about. it leaves through the same restore as the transaction failures now, and the body above says so

also re-split the two commits by file, query-table.c in one and common.c in the other. an amend had folded a hunk into the wrong commit and the messages no longer described their own diffs

194 bats, 151 pytest, 12, 25, 9 again

src/database/common.c
```diff
@@ -87,10 +87,10 @@ int _dbclose_handle(sqlite3 *db, const char *func, const int line, const char *f
...
 void _dbclose(sqlite3 **db, const char *func, const int line, const char *file)
 {
-	// Silently return if the database is known to be broken. It may not be
-	// possible to close the connection properly.
-	if(FTLDBerror())
-		return;
+	// A broken database is closed like any other. sqlite3_close_v2() copes
+	// with a connection whose file is corrupt, and skipping it here leaked
+	// the handle, left the caller holding a pointer it believes was closed,
+	// and skipped the counter below for the rest of the process
```
DL6ER: This is a comment with no code under it, and what it explains is the state of the branch before this commit. It belongs in the commit message.

```suggestion
```

src/database/query-table.c
```diff
@@ -2330,11 +2352,32 @@ bool queries_to_database(void)
 rollback_unlock_fail:
 	dbquery(memdb, "ROLLBACK");
 unlock_fail:
+	// Nothing was committed. Phase 1 cleared the changed flag under the
+	// lock, and only dnsmasq_interface.c ever sets it again, when the query
+	// itself changes, so without putting it back these queries are skipped
+	// by every later run and reach neither database. The lock is still ours
+	// here
```
DL6ER: Same again: most of this is why the old code was wrong rather than what this one does.

```suggestion
	// Nothing was committed, so give the snapshotted queries their changed
	// flag back and let the next run pick them up. The lock is still ours
```

DL6ER (commented):
I traced the label flow and it holds up. `snaps` is allocated at `query-table.c:1927` with its own early return and `snap_count` is zeroed at `:1935`, so all four jumps land after both and no restore loop can walk an unallocated array. The lock state per label is right too, including the one pre-existing `goto fail` at `:2216`, which is after `unlock_shm()`. And the `succeeded < snap_count` restore is correct because `END` at `:2268` still commits the rows that did step.

Two things, both inline. Plus one across PRs: this overlaps #3062, which also rewrites the `_dbopen()`/`_dbclose()` counter handling, and removing the `FTLDBerror()` early return here is the same drift I raised there. They will conflict in the file. Pick a merge order and note it on one of the two.

Last, the restore loop appears three times, identically. A small static helper taking `snaps` and `snap_count` would leave `fail:` as the only site that needs the `lock_shm()` wrapper around it.

010011110:
both comments cut - the one in common.c had no code under it at all and was describing the branch before the commit, which belongs in the message and now only lives there

the restore is requeue_snapshots() now, shared by the three sites, so fail: is the only one that needs the lock_shm() wrapper. that meant lifting struct query_snap to file scope, which is the only structural change

on the overlap: #3062 rewrites the same counter handling and removing the FTLDBerror() early return here is the drift you raised there. merge #3062 first and i will rebase this on top

rebased onto current development, 194 bats, 151 pytest, 12, 25, 9

DL6ER (approved):
Comments fixed and the restore loop deduplicated as suggested.

PR #3065, opened 2026-09-04
read shared state while it is still protected

010011110:

two places where shared state is read outside the protection that is supposed to cover it

_get_and_clear_event() learned an event's state by setting it, because an atomic_flag cannot be read any other way, and cleared it a few lines further down with the debug logging in between. a producer running in that window found the flag already set, took its own event for a duplicate and dropped it. the queue has one consumer and many producers, RELOAD_GRAVITY alone comes from the SIGRTMIN handler, from two places in the api, from the teleporter, from dnsmasq_interface.c and from the database thread. a lost one means a list edit or pihole reloaddns reports success and never takes effect. an _Atomic bool can be read and written in one atomic_exchange(), so the window closes, and a static_assert pins the lock-free property the signal handler depends on

_unlock_shm() released both mutexes and only then took the end timestamp and read time.begin. both live in the shared memory region, so by then another thread can hold the lock and have overwritten time.begin: the number reported is the time since somebody else's acquisition, and the read races their write. the timestamps are taken into locals before the unlocks now. while there, the scaling: seconds were divided by 1000 where they should be multiplied, so a lock held for two seconds was reported as 0.002 ms plus the nanosecond remainder

third commit, out of review here: SIGRT_handler() states in its own comment that nothing async-signal-unsafe may run in it and that its events are logged where they are processed, but set_event() reaches log_debug() and from there _FTL_log(), which formats with printf and can take the shm lock. a signal arriving while the interrupted code holds that lock, or sits inside malloc, deadlocks the process. set_event_from_signal() is the exchange and nothing else, and the five calls in the handler use it. the wider async-signal-safe logging api darkexplosiveqwx is working towards is a separate matter, this only removes the call that contradicts the comment already in the file

medium impact for the first, it is rare and unreproducible but produces a silent no-op on exactly the operation someone just asked for. the second only affects a debug line, which is nevertheless the line someone reads when they are trying to understand lock contention

by inspection: a set followed by a clear is one exchange now, and the timestamps are read while the lock is held. reproducing the lost event means widening the window on purpose, a sleep between the old set and clear plus set_event from a second thread. full suite is 194 bats, 151 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/src/events.c
+++ b/src/events.c
@@ -49,6 +49,17 @@ void _set_event(const enum events event, int line, const char *function, const c
 	}
 }
 
+// Raise an event from a signal handler
+// SIGRT_handler() already states that nothing async-signal-unsafe may run in
+// it, but set_event() reaches log_debug() and from there _FTL_log(), which
+// formats with printf and can take the SHM lock. Neither is allowed to happen
+// with a signal interrupting arbitrary code, so the signal path gets the
+// exchange on its own: the event is logged where it is processed anyway
+void set_event_from_signal(const enum events event)
+{
+	atomic_exchange(&eventqueue[event], true);
+}
+
 // Get and clear event
 // Read and clear in one exchange, so there is no window in which a producer
 // sees a value the consumer has already taken. Compiles to XCHG on x86_64 and
--- a/src/events.c
+++ b/src/events.c
@@ -12,7 +12,7 @@
 #include "FTL.h"
 // public prototypes
 #include "events.h"
-// atomic_flag_test_and_set()
+// atomic_exchange()
 #include <stdatomic.h>
 // struct config
 #include "config/config.h"
@@ -23,7 +23,12 @@
 static const char *eventtext(const enum events event);
 
 // Queue containing all possible events
-static volatile atomic_flag eventqueue[EVENTS_MAX] = { ATOMIC_FLAG_INIT };
+// An atomic_bool rather than an atomic_flag: a flag cannot be read without
+// setting it, which is what used to lose events (see below). RELOAD_GRAVITY is
+// raised from the SIGRTMIN handler, so the type has to be lock-free for the
+// exchange to be async-signal-safe
+static _Atomic bool eventqueue[EVENTS_MAX];
+static_assert(ATOMIC_BOOL_LOCK_FREE == 2, "Event queue must be lock-free, it is used from a signal handler");
 
 // Set/Request event
 // We set the events atomically to ensure no race collisions can happen. If an
@@ -31,10 +36,8 @@ static volatile atomic_flag eventqueue[EVENTS_MAX] = { ATOMIC_FLAG_INIT };
 // added multiple times
 void _set_event(const enum events event, int line, const char *function, const char *file)
 {
-	bool is_set = false;
-	// Set eventqueue bit
-	if(atomic_flag_test_and_set(&eventqueue[event]))
-		is_set = true;
+	// Set eventqueue bit, learning what it was in the same operation
+	const bool is_set = atomic_exchange(&eventqueue[event], true);
 
 	// Possible debug logging
 	if(config.debug.events.v.b)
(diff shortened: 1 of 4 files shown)


darkexplosiveqwx:
https://github.com/pi-hole/FTL/commit/03663dd1b9381ee22da5b00f48bc8e0f2e15412f looks good, but I noticed one thing:

@DL6ER why is log_debug() called from here?

set_event() can be called directly from the SIGRTMIN signal handler, so this makes the signal-handler path enter _FTL_log(). _FTL_log() is not async-signal-safe: among other things it does printf-style formatting and can take SHM locks.

I have been looking into an async-signal-safe logging path for this. It is a bit more involved than it initially sounds: timestamp formatting, structured/JSON formatting, and even writing to the journald socket all have to stay out of the signal handler. I arrived at this while working through the logging layers (log-dnsmasq → log-structured → log-journal → log-async).

Since I am already working on an asynchronous logger for FTL, this fits: Vyukov's MPSC allows async-signal-safe insertions into the rind and the consumer takes care of I/O and the SHM FIFO. The only unsafe thing done in `_FTL_log()` is printf formatting, but that can be worked around with a new, specifically async-signal-safe API taking a preformatted string. I can also implement a minimal async-signal-safe formatter for our particular needs (string concatenation plus integer formatting) to not make this too annoying to use. 
I’m mainly pinging you to make sure you’re aware of the issue and to get your preliminary opinion on that direction before I start changing the logging API.

In particular, I think the important invariant should be that **nothing below the signal-handler entry point may depend on locks, stdio/printf formatting, malloc, or other non-async-signal-safe facilities.**

Depending on if I remember, I will also open a new issue to track this.

010011110:
you are right, and it is fixable today without waiting for the bigger change: pushed as a third commit here

SIGRT_handler() already carries a comment saying nothing async-signal-unsafe may run in it and that its events are logged where they are processed. set_event() quietly broke that, since it reaches log_debug() and from there _FTL_log(). set_event_from_signal() is the exchange and nothing else, and the five calls in the handler use it now. nothing is lost, the consumer logs the event when it picks it up

your invariant is the right one to hold the code to, and this pr only satisfies half of it: the exchange itself is lock-free (there is a static_assert on ATOMIC_BOOL_LOCK_FREE now, precisely because the handler uses it), while every other path into the logger is still what you describe. the preformatted-string api plus a small async-signal-safe formatter sounds right to me, and an mpsc ring for the handoff means the handler never touches the shm fifo at all

not trying to pre-empt any of that, this is one call site that contradicted a comment already in the file. happy to be a reviewer on the logging work when you open it

src/events.c
```diff
@@ -46,26 +49,32 @@ void _set_event(const enum events event, int line, const char *function, const c
...
+// Reading and clearing happen in one exchange, so there is no window in which
+// the queue holds a value neither side owns. This used to be an atomic_flag,
+// which cannot be read without setting it: the consumer set the flag to learn
+// its value and cleared it a few lines later, and any producer that ran in
+// between saw a flag that was already set, treated its own event as a duplicate
+// and dropped it. A lost RELOAD_GRAVITY means a list edit or `pihole reloaddns`
+// reports success and never takes effect. Debug logging sits inside that window
+// and widens it considerably.
+// The exchange compiles to XCHG on x86_64 and i686, and to the LDAXRB/STLXRB
+// pair on ARM64, the same instructions the flag used.
```
DL6ER: Eight of these ten lines are the old bug. The instruction note is worth keeping, the rest reads as a changelog entry that will outlive anyone's interest in it.

```suggestion
// Get and clear event
// Read and clear in one exchange, so there is no window in which a producer
// sees a value the consumer has already taken. Compiles to XCHG on x86_64 and
// i686, and to the LDAXRB/STLXRB pair on ARM64
```

DL6ER (commented):
The race is real and the fix is the right shape: the consumer's `test_and_set` left the flag set, a producer landing in that window read it as already pending and folded its own request in, and the consumer then cleared both. `static_assert` resolves through `assert.h` at `FTL.h:67`, C11 7.14.1.1 allows a lock-free atomic in a handler so dropping `volatile` is fine, and all five `set_event` calls in `SIGRT_handler` are converted - `rtsig == 2` only reaches `terminate_error()`, which is `raise()` and an assignment. The `/1000.0` was genuinely wrong.

Two things beyond the inline note. `ShmLock.time.end` (`shmem.c:143`) is now neither written nor read, so the field is dead. Dropping it shifts the shm layout and would want a `SHARED_MEMORY_VERSION` bump (`shmem.c:39`), which is probably reason enough to leave it - but then say so, rather than leaving a member nothing touches.

Second, `end` is now sampled before both `pthread_mutex_unlock()` calls, so the number no longer includes the release. That is more honest for "held", but it is a change in what the line reports and the commit body should say it.

010011110:
comment cut to your version - the eight lines of old-bug archaeology are gone

both other points are in the commit body now rather than left implicit. time.end stays because dropping a member of ShmLock shifts the shared memory layout and would want a SHARED_MEMORY_VERSION bump for a field only that debug line ever read, and the message says that instead of leaving a dead member unexplained. and end being sampled before both unlocks does change what the line reports: it is now the time the lock was held, without the release, which is what "held for" claims

DL6ER (approved):
Comment trimmed, and the two other points are in the commit body.

PR #3066, opened 2026-09-04
give a query's bookkeeping the same treatment on every path

010011110:

three places in dnsmasq_interface.c where a query's bookkeeping is started and not finished

the analyzeOnlyAandAAAA skip returns before the query record is created, but findClientID() has already incremented the client's counter and gc only decrements per query record. every ptr, srv, txt or https query therefore inflates that client's total for the life of the process and top clients drifts upward with it. the rate-limit branch twelve lines above already undoes its own increment for exactly this reason, with a comment saying why. this one does the same now, and its "we already counted it" comment goes, since it was justifying the bug

detect_blocked_IP() assigned blockingreason and cacheStatus inside if(config.debug.queries) in all four branches while returning the status outside it. those two statics are what _FTL_make_answer() logs and derives the extended dns error from, so with debug off an upstream-blocked answer carried the previous query's reason and no ede, whatever dns.blocking.edns said. turning debug logging on changed what went out on the wire. only inet_ntop() and the log line stay inside the guard now

the regex cache-hit path restored force_next_DNS_reply and last_regex_idx from the cache entry but not cname_target, which the miss path sets from the same entry. a configured cname redirect therefore worked once and fell back to the default blocking mode from the second hit on, because REPLY_CNAME without a target is nothing. it is restored there and cleared where blockingreason is, so it cannot travel into the next query either

medium impact. none of them crashes anything, all three produce wrong output that looks like a configuration problem: a client total that only grows, a missing ede, a cname redirect that stops working after the first query

the counter: set dns.analyzeOnlyAandAAAA, send ptr queries, watch the client's total in /api/clients grow without any query appearing. the ede: set dns.blocking.edns and have an upstream answer 0.0.0.0, the ede is absent with debug off and present with it on. the cname: set a regex cname redirect and query the domain twice, the second answer is the default blocking mode. full suite is 194 bats, 151 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/src/dnsmasq_interface.c
+++ b/src/dnsmasq_interface.c
@@ -792,8 +792,10 @@ size_t _FTL_make_answer(struct dns_header *header, char *limit, const size_t len
 	if (trunc)
 		header->hb3 |= HB3_TC;
 
-	// Unset the blocking reason
+	// Unset the blocking reason and the CNAME target that went with it, so
+	// neither travels into the next query
 	blockingreason = "<not set>";
+	cname_target = NULL;
 
 	return p - (unsigned char *)header;
 }
@@ -1085,10 +1087,13 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
 	// but user wants to see only A and AAAA queries (pre-v4.1 behavior)
 	if(config.dns.analyzeOnlyAandAAAA.v.b && querytype != TYPE_A && querytype != TYPE_AAAA)
 	{
-		// Don't process this query further here, we already counted it
+		// Don't process this query further here
 		if(config.debug.queries.v.b)
 			log_debug(DEBUG_QUERIES, "Skipping new query (%i)", id);
 
+		// Undo the findClientID() increment, as for the rate-limited
+		// branch above: no query record is created here
+		change_clientcount(client, -1, 0, -1, 0);
 		unlock_shm();
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


src/dnsmasq_interface.c
```diff
@@ -1085,10 +1087,16 @@ bool _FTL_new_query(const unsigned int flags, const char *name,
...
-		// Don't process this query further here, we already counted it
+		// Don't process this query further here
 		if(config.debug.queries.v.b)
 			log_debug(DEBUG_QUERIES, "Skipping new query (%i)", id);
 
+		// Undo the client count increment from findClientID() above:
+		// we return before the query record is created, so GC will
+		// never decrement this counter - leaving it permanently
+		// inflated for the lifetime of the process, exactly as for
+		// rate-limited queries above
```
DL6ER: The rate-limit branch twelve lines up already carries this rationale in full. Repeating it here means two copies to keep true, and the last two lines describe what would happen without the call rather than what it does.

```suggestion
		// Undo the findClientID() increment, as for the rate-limited
		// branch above: no query record is created here
```

DL6ER (commented):
Verified all three. `change_clientcount(client, -1, 0, -1, 0)` matches the rate-limit branch at `:1051` exactly; `findClientID()` is called with `count=true, aliasclient=false` at `:1004`, so the increment always happened and the undo can be unconditional, and overTime is only touched later at `:1168`, so the `-1, 0` there is right. The `cname_target` restore mirrors the miss path at `:1717-1718` and sits in the regex branch alongside `last_regex_idx`. The `NULL` reset joins the existing `blockingreason` reset, which is what keeps it out of the gravity branches that never set it.

Good catch on `detect_blocked_IP()` - that one changed what went out on the wire depending on a debug flag.

One inline note, otherwise this is ready.

010011110:
trimmed - the rate-limit branch twelve lines up carries the rationale and one copy is enough. what is left says what the call does rather than what would happen without it

DL6ER (approved):
Trimmed - one copy of the rationale is enough.

PR #3067, opened 2026-09-04
answer database failures with a server error, and free the half-built row

010011110:

three api paths report a failure as something the client did wrong, or as nothing at all

networkTable_readDevices() and api_network_devices_DELETE() are static int handlers whose return value is the http status, and both said return false when dbopen() failed. zero is what the router reads as "no handler answered", so a database that could not be opened came back as 404 not found for an endpoint that plainly exists. both send 500 now, and the delete path no longer logs the other function's name

the two gravity readers in api/search.c and api/list.c reported a failed gravityDB_readTable() as 400 bad request. nothing about the request is wrong when the gravity database cannot be read, so both are 500

while in those two files: the row object built inside the read loop is only appended to the array at the end of the loop body, so the parse_groupIDs() failure path in the middle returned with it owned by nobody

low impact, but the 404 is the kind of answer that sends someone hunting for a typo in a path that is perfectly correct

make the database unreadable (chmod 000 on pihole-FTL.db as a non-root ftl) and call /api/network/devices: 500 with a database_error key instead of 404 not found. same for the gravity readers with gravity.db. full suite is 194 bats, 151 pytest, 12, 25, 9

not in this pr, deliberately: a wrong method still answers 404 rather than 405 with an Allow header. the router has the endpoint list to build that from, but send_json_error() cannot add a header, my_send_http_error_headers() emits the whole block, so it needs a header-capable error path first and that is a change of its own

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/network.c
+++ b/src/api/network.c
@@ -177,8 +177,11 @@ static int api_network_devices_GET(struct ftl_conn *api)
 	sqlite3 *db = dbopen(true, false);
 	if(db == NULL)
 	{
-		log_warn("Failed to open database in networkTable_readDevices()");
-		return false;
+		log_warn("Failed to open database in api_network_devices_GET()");
+		return send_json_error(api, 500,
+		                       "database_error",
+		                       "Could not open the long-term database",
+		                       NULL);
 	}
 
 	const char *sql_msg = NULL;
@@ -293,8 +296,11 @@ static int api_network_devices_DELETE(struct ftl_conn *api)
 	sqlite3 *db = dbopen(false, false);
 	if(db == NULL)
 	{
-		log_warn("Failed to open database in networkTable_readDevices()");
-		return false;
+		log_warn("Failed to open database in api_network_devices_DELETE()");
+		return send_json_error(api, 500,
+		                       "database_error",
+		                       "Could not open the long-term database",
+		                       NULL);
 	}
 
 	// Delete row from network table by ID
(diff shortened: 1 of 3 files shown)


src/api/network.c
```diff
@@ -178,7 +178,10 @@ static int api_network_devices_GET(struct ftl_conn *api)
 	if(db == NULL)
 	{
 		log_warn("Failed to open database in networkTable_readDevices()");
```
DL6ER: You fixed this exact message in the DELETE handler three lines from here, but the GET one still names a different function. `networkTable_readDevices()` does exist, which makes it worse than a typo - it sends the reader to `network-table.c`.

```suggestion
		log_warn("Failed to open database in api_network_devices_GET()");
```

DL6ER (commented):
Confirmed the mechanism: `api_handler()` turns `ret == 0` into a 404 at `api/api.c:264-270`, so `return false` really did answer "Not found" for an endpoint that exists. The `400` to `500` change is right for the same reason - nothing about the request is wrong when gravity cannot be read.

The ownership fix is correct in both files. In `search.c` the array belongs to the caller and `row` is only appended at the end, so the one `JSON_DELETE(row)` is all that is needed; `list.c` needs both and has both.

One thing inline.

010011110:
fixed - it now names api_network_devices_GET(). worse than a typo is right: networkTable_readDevices() exists in network-table.c, so the message sent the reader to the wrong file, and i had corrected the same thing three lines down without noticing this one

DL6ER (approved):
Both handlers name their own function now.

Issue #3003, opened 2026-08-05
misc.hide_connection_error has no effect on "TCP connection failed while receiving payload length" warnings (FTL v6.7)


tobschndr:
## Versions

- Pi-hole FTL: v6.7
- Docker image: pihole/pihole:latest
- Platform: Debian, Docker (bridge network), upstream is unbound in a
  separate container at 172.20.0.2#53
- unbound: 1.25.2

## Expected behavior

With `misc.hide_connection_error` set to `true`, FTL should not write
connection warnings to FTL.log. PR #2749 describes the setting as
hiding "all connection warnings" and names "Connection prematurely
closed by remote server" as the example case.

## Actual behavior

The setting is active, confirmed by FTL itself, but the warnings are
still written to FTL.log.

$ docker exec pihole pihole-FTL --config misc.hide_connection_error
true

$ docker exec pihole pihole-FTL --version
v6.7

$ docker exec pihole grep "receiving payload length" /var/log/pihole/FTL.log
2026-08-05 17:11:23.783 CEST [100233/F51] WARNING: Connection error (172.20.0.2#53): TCP connection failed while receiving payload length from upstream (Connection prematurely closed by remote server)
2026-08-05 17:16:29.652 CEST [100325/F51] WARNING: Connection error (172.20.0.2#53): TCP connection failed while receiving payload length from upstream (Connection prematurely closed by remote server)
2026-08-05 17:16:29.848 CEST [100326/F51] WARNING: Connection error (172.20.0.2#53): TCP connection failed while receiving payload length from upstream (Connection prematurely closed by remote server)

$ docker inspect pihole --format '{{.State.StartedAt}}'
2026-08-01T10:00:16.68238222Z

Ruling out the obvious:

- The entries are from today, the container has been running since
  2026-08-01, so they were produced by the running instance with the
  setting already active.
- FTL.log was rotated today at 00:00, so these are not stale lines
  from before the setting was enabled.
- The setting is applied via the environment variable
  `FTLCONF_misc_hide_connection_error=true` in docker compose, which
  makes it read-only, and FTL logs `[i] FTLCONF_misc_hide_connection_error
  is used` on startup.

## Steps to reproduce

1. Run Pi-hole v6.7 in Docker with unbound as the only upstream.
2. Set `FTLCONF_misc_hide_connection_error=true` and recreate the
   container.
3. Verify with `pihole-FTL --config misc.hide_connection_error`.
[...]

PR #3068, opened 2026-09-04
say something when the configuration does not do what was asked

010011110:

three places where ftl knows more about the configuration than it says

a pihole.toml value of the wrong type is dropped exactly like a missing one, with a DEBUG_CONFIG line nobody has enabled, and readFTLconf() then rewrites the file with the compiled-in default: the edit disappears from the file and nothing is said. the two cases are told apart now, absent stays a debug line, present-but-wrong-type is a warning naming the key and the type expected, which is what the environment and cli paths already do

closeFTLtoml() was void. a short write sets the error indicator on the stream and fprintf() carries on, so on a full filesystem the temporary file was truncated, files_different() reported it as different, the good config was rotated away and the truncated one renamed over it. it returns a status now, from fflush(), ferror(), fsync() and fclose(), and writeFTLtoml() leaves the live file alone when the new one could not be written

check_capabilities() ran only when debug.caps was set, which also hid the warnings about capabilities ftl needs and does not have. the call is unconditional now, but warning about all of them always would be worse than silence: a container started without --cap-add would report capabilities its owner deliberately withheld, every boot. each warning is gated on the configuration that uses it - CAP_NET_ADMIN and CAP_NET_RAW on dhcp.active, CAP_NET_BIND_SERVICE on a privileged dns.port, CAP_SYS_NICE on a negative misc.nice, CAP_SYS_TIME on ntp.sync.active - with CAP_CHOWN unconditional. the second call site in dnsmasq_interface.c no longer hides behind debug.caps either, so the two agree

one thing worth stating about the write check: nine of writeFTLtoml()'s ten callers ignore its return, so what a user sees is the log_err, not a caller reacting. the value of the change is that the live pihole.toml survives a failed write, not that anything downstream handles it

mixed impact. the config write is the serious one, it destroys a working pihole.toml on a full disk, though that takes a full disk to reach. the rest is diagnostics: things that are already known and simply not said

wrong type: put dns.port = "not a number" in pihole.toml and start ftl, the warning names the key and the type instead of nothing. full disk: fill the filesystem and set a config value, pihole.toml is left intact and the temporary file removed, where before it was replaced with a truncated one. capabilities: drop CAP_NET_RAW and start, the warning appears without debug.caps. full suite is 194 bats, 151 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/src/capabilities.c
+++ b/src/capabilities.c
@@ -15,6 +15,8 @@
 #include "FTL.h"
 #include "capabilities.h"
 #include "config/config.h"
+// DOT_PORT
+#include "dotdoh/server.h"
 #include "log.h"
 
 static const unsigned int capabilityIDs[]   = { CAP_CHOWN ,  CAP_DAC_OVERRIDE ,  CAP_DAC_READ_SEARCH ,  CAP_FOWNER ,  CAP_FSETID ,  CAP_KILL ,  CAP_SETGID ,  CAP_SETUID ,  CAP_SETPCAP ,  CAP_LINUX_IMMUTABLE ,  CAP_NET_BIND_SERVICE ,  CAP_NET_BROADCAST ,  CAP_NET_ADMIN ,  CAP_NET_RAW ,  CAP_IPC_LOCK ,  CAP_IPC_OWNER ,  CAP_SYS_MODULE ,  CAP_SYS_RAWIO ,  CAP_SYS_CHROOT ,  CAP_SYS_PTRACE ,  CAP_SYS_PACCT ,  CAP_SYS_ADMIN ,  CAP_SYS_BOOT ,  CAP_SYS_NICE ,  CAP_SYS_RESOURCE ,  CAP_SYS_TIME ,  CAP_SYS_TTY_CONFIG ,  CAP_MKNOD ,  CAP_LEASE ,  CAP_AUDIT_WRITE ,  CAP_AUDIT_CONTROL ,  CAP_SETFCAP };
@@ -110,6 +112,66 @@ bool check_capability(const unsigned int cap)
  *
  * @return true if all required capabilities are available, false otherwise.
  */
+// Does anything in this configuration want a port below 1024? 1024 itself is
+// the first unprivileged one (net.ipv4.ip_unprivileged_port_start). dns.port,
+// the DoT listener on DOT_PORT and every entry of webserver.port count, the
+// last being a list like "80o,443os,[::]:80o", so the port is what follows the
+// final colon, if any. DoH needs nothing extra, it rides on a webserver port
+static bool binds_privileged_port(void)
+{
+	if(config.dns.port.v.u16 != 0 && config.dns.port.v.u16 < 1024)
+		return true;
+
+	if(config.dns.dot.v.b && DOT_PORT < 1024)
+		return true;
+
+	const char *list = config.webserver.port.v.s;
+	if(list == NULL)
+		return false;
+
+	char *copy = strdup(list);
+	if(copy == NULL)
+		return false;
+
+	bool privileged = false;
+	char *save = NULL;
+	for(char *tok = strtok_r(copy, ",", &save); tok != NULL && !privileged; tok = strtok_r(NULL, ",", &save))
+	{
+		// Skipped through a separate pointer so the loop variable itself
+		// is not modified in the body
+		const char *ent = tok;
+		while(*ent == ' ')
+			ent++;
+
+		// An entry may carry an address, so the port follows the last colon
+		const char *port = strrchr(ent, ':');
+		port = port != NULL ? port + 1 : ent;
+
+		const long p = strtol(port, NULL, 10);
+		if(p > 0 && p < 1024)
+			privileged = true;
+	}
+
+	free(copy);
+	return privileged;
+}
+
+// Warn when a capability this configuration needs is missing. Returns false
+// only in that case: a capability the running configuration does not use is
+// not reported at all
+static bool warn_missing_cap(const cap_user_data_t data, const unsigned int capid,
+                             const char *name, const bool needed, const char *what)
+{
+	if(!needed)
+		return true;
+
+	if((data->permitted & (1u << capid)) && (data->effective & (1u << capid)))
+		return true;
+
+	log_warn("Linux capability %s is not available, needed for %s", name, what);
+	return false;
+}
+
 bool check_capabilities(void)
 {
 	cap_user_data_t data = NULL;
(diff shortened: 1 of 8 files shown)


rdwebdesign:
**NOTE:**

> misc.hide_connection_error filtered the message table but not the log line. its help text is about the log file and says it hides all connection warnings, so it skips the my_syslog() call and the message-table insert now. the read-time filters stay as they are, and the DEBUG_QUERIES line stays too, being opt-in already. 

There is another open PR (https://github.com/pi-hole/FTL/pull/3019) trying to fix this, but using a different approach.

@DL6ER 
What do you think?

darkexplosiveqwx:
> misc.hide_connection_error filtered the message table but not the log line. its help text is about the log file and says it hides all connection warnings, so it skips the my_syslog() call and the message-table insert now. the read-time filters stay as they are, and the DEBUG_QUERIES line stays too, being opt-in already. closes #300

I think that `misc.hide_connection_error` should only apply to FTL.log and the connected message in the web interface. I don't think we should filter dnsmasq's log directly.

rdwebdesign:
> I don't think we should filter dnsmasq's log directly.

I agree.

010011110:
dropped the hide_connection_error change, you are both right. it skipped my_syslog(), which is dnsmasq's log and not ours to filter, and it skipped the message-table insert, which loses the entry instead of quieting the log: turning the setting back off would have shown nothing

#3019 has the right shape and covers hide_dnsmasq_warn as well, so this pr keeps the other three and no longer claims to close #3003

what is left here: a pihole.toml value of the wrong type is dropped with a debug line and then overwritten with the default, so the edit disappears silently; closeFTLtoml() cannot tell its caller a write failed, so on a full disk the good config is rotated away and a truncated one renamed over it; and check_capabilities() only ran under debug.caps, which hid the warnings about capabilities ftl needs and does not have

src/main.c
```diff
@@ -117,9 +117,12 @@ int main (int argc, char *argv[])
...
 
-	// Check for availability of capabilities in debug mode
-	if(config.debug.caps.v.b)
-		check_capabilities();
+	// Check for availability of capabilities. The per-capability table this
+	// prints is behind DEBUG_CAPS, but the warnings about the ones FTL needs
+	// and does not have are not, and they are the reason to run this at all:
+	// hiding them behind a debug flag means nobody sees them until they
+	// already suspect the problem
+	check_capabilities();
```
DL6ER, line 125: These warnings are not unconditionally true. `CAP_SYS_TIME` only matters when the NTP client sets the clock, and `CAP_NET_ADMIN` is for ARP injection when we are the DHCP server. A container started without `--cap-add` now warns on every boot about capabilities its owner deliberately withheld and does not need, and that arrives back at us as bug reports.

I am with you that hiding a real problem behind `debug.caps` is wrong. But the fix is to warn about what the running configuration actually needs - gate the feature-specific ones on the config that uses them - rather than to warn about all of them always.

src/config/toml_helper.c
```diff
@@ -90,9 +92,29 @@ void closeFTLtoml(FILE *fp, const bool locked)
 	if (mode == -1)
 		log_err("Cannot get access mode for FTL's config file: %s", strerror(errno));
 
+	// A write error shows up here and nowhere else: fprintf() sets the
+	// error indicator and carries on, and the bytes are lost at the flush.
+	// The caller renames this file over the live config, so it has to know
+	if(mode != -1 && (mode & O_ACCMODE) != O_RDONLY)
```
DL6ER: If `fcntl` failed we get `mode == -1`, this whole block is skipped and `okay` stays true. The one case where we know least about the file is the one where nothing is checked. I would treat a failed `fcntl` as a reason to check anyway, or as a failure in its own right.

DL6ER (commented):
The config-write half is the valuable part and it is sound - `fn` is the `fileno(fp)` already fetched for `mode`, and the fsync lands before the fclose. Destroying a working `pihole.toml` on a full disk is worth fixing.

The capability change I would not take as it stands, see inline. The type-vs-absent split is a clear improvement.

Two smaller ones. `dnsmasq_interface.c:2493` still calls `check_capabilities()` only under `debug.caps`, so the two call sites now disagree and the comment there still says "if requested via debug flag" - square them up either way you go. And the new `false` from `writeFTLtoml()` is ignored by nine of its ten callers, so the `log_err` is what a user actually sees; worth saying in the description rather than implying the caller reacts.

010011110:
the capability warnings are gated on the configuration now, which is your point and it was the right one - warning about a capability a container owner deliberately withheld is a bug report waiting to happen

CAP_NET_ADMIN and CAP_NET_RAW on dhcp.active, CAP_NET_BIND_SERVICE on a privileged dns.port, CAP_SYS_NICE on a negative misc.nice, CAP_SYS_TIME on ntp.sync.active, and CAP_CHOWN stays unconditional since ftl always takes ownership of what it creates. the gates follow dnsmasq's own list at dnsmasq.c:513 rather than my reading of what each is for

checked in the container, where dhcp.active is false: no CAP_NET_ADMIN or CAP_NET_RAW warning at all, and the three the config does use still warn

dnsmasq_interface.c:2493 no longer hides behind debug.caps, so both call sites agree, and its comment goes with it

closeFTLtoml() now checks when fcntl() failed instead of skipping - you are right that the case where we know least about the file was the one checking nothing

and the description says plainly that nine of writeFTLtoml()'s ten callers ignore the return, so the log_err is what a user sees. the value is that the live pihole.toml survives, not that anything reacts

DL6ER (commented):
Gating per configuration is the right shape and `warn_missing_cap()` reads well. Two of the gates are wrong though, both inline. The `fcntl` change is right.

One to be aware of rather than fix: `check_capabilities()` now runs on every `FTL_dnsmasq_reload()`, so a missing capability warns again on every `pihole reloaddns`.

src/capabilities.c
```diff
@@ -129,49 +145,29 @@ bool check_capabilities(void)
...
-		// Necessary for setting the system time in the NTP client
-		log_warn("Required Linux capability CAP_SYS_TIME not available");
-		capabilities_okay = false;
-	}
+	capabilities_okay &= warn_missing_cap(data, CAP_NET_ADMIN, "CAP_NET_ADMIN",
+	                                      config.dhcp.active.v.b,
+	                                      "ARP injection while acting as the DHCP server");
+	capabilities_okay &= warn_missing_cap(data, CAP_NET_RAW, "CAP_NET_RAW",
+	                                      config.dhcp.active.v.b,
+	                                      "ICMP from the DHCP server");
```
DL6ER: `dhcp-discover.c:763` already says the opposite in the tree: "DHCP uses normal UDP datagrams, so we cdon't need CAP_NET_RAW". The actual user is `arp-scan.c:206`, `AF_PACKET`/`SOCK_RAW`, which is a CLI subcommand with no relation to `dhcp.active` - and it already checks for itself at `:625` and prints a usable error. So this gate warns where the capability is not needed and stays quiet where it is. I would drop `CAP_NET_RAW` from the daemon check and leave it to `arp-scan`.

src/capabilities.c
```diff
@@ -129,49 +145,29 @@ bool check_capabilities(void)
...
+	capabilities_okay &= warn_missing_cap(data, CAP_NET_BIND_SERVICE, "CAP_NET_BIND_SERVICE",
+	                                      config.dns.port.v.u16 != 0 && config.dns.port.v.u16 <= 1024,
+	                                      "binding a privileged port");
```
DL6ER, line 203: `webserver.port` is missing here, and its default includes 80 and 443. Set `dns.port` to 5353, keep the web UI where it is, and the warning disappears while the capability is still required. `dns.dot` and DoH on 443 widen it further. `split_terminator_ports()` in `webserver.c:736` already walks that list if you want to reuse the parsing rather than write it twice.

010011110:
both gates were wrong and both are fixed

CAP_NET_RAW is out of the daemon check entirely. dhcp-discover.c:763 says it plainly, dhcp is ordinary udp, and arp-scan is the only user with its own check at :625 and a better error than mine. i had reasoned from the old comment in check_capabilities() instead of from the code

CAP_NET_BIND_SERVICE now looks at webserver.port as well as dns.port. it walks the list and takes the port after the last colon, so [::]:80o counts. checked your case directly: dns.port 5353 with the web ui left on 80/443 still warns, where before the change it went quiet

on the reload: you are right that a missing capability now warns again on every pihole reloaddns. i have left it, since suppressing repeats means keeping state about what was already warned about, and a reload is not frequent enough for that to be worth it. say if you would rather have it once per process

src/capabilities.c
```diff
@@ -110,6 +110,58 @@ bool check_capability(const unsigned int cap)
  *
  * @return true if all required capabilities are available, false otherwise.
  */
+// Does anything in this configuration want a port below 1024? dns.port and
+// every entry of webserver.port count, the latter being a list of the form
+// "80o,443os,[::]:80o", so the port is what follows the last colon, if any
+static bool binds_privileged_port(void)
+{
+	if(config.dns.port.v.u16 != 0 && config.dns.port.v.u16 <= 1024)
```
DL6ER: `dns.dot` is missing. It defaults to true and the listener binds 853, hardcoded as `DOT_PORT` in `dotdoh/dot_server.c:57`. With the default config `dns.port` is 53 and this returns true anyway, so it does not show - but move DNS to 5353 and the web UI to a high port, leave DoT on, and the warning disappears while 853 still needs the capability. That is the same case I raised for `webserver.port`, one config item further along.

`dns.doh` needs nothing extra, it rides on the HTTPS port that is already in `webserver.port`.

src/capabilities.c
```diff
@@ -110,6 +110,58 @@ bool check_capability(const unsigned int cap)
...
+	{
+		while(*tok == ' ')
+			tok++;
+
+		// An entry may carry an address, so the port follows the last colon
+		const char *port = strrchr(tok, ':');
+		port = port != NULL ? port + 1 : tok;
+
+		const long p = strtol(port, NULL, 10);
+		if(p > 0 && p <= 1024)
```
DL6ER: Off by one: 1024 is the first *unprivileged* port (`net.ipv4.ip_unprivileged_port_start` defaults to 1024), so binding it needs no capability. Same on the `dns.port` line above.

```suggestion
		if(p > 0 && p < 1024)
```

DL6ER (commented):
Both gates are addressed. `CAP_NET_RAW` is gone from the daemon check, and `webserver.port` is parsed now - I checked the awkward forms and they come out right, `[::]:443os` as well as the `+80` notation. The `fcntl` case reads the right way round too: an unknown mode is a reason to check rather than to skip.

Two things left, both in `binds_privileged_port()`.

010011110:
all three fixed

dns.dot is in the gate now, and DOT_PORT moved from dot_server.c into dotdoh/server.h so the check names the constant rather than repeating 853. doh needs nothing extra as you say, it rides on a webserver port. checked your case: dns.port 5353, webserver on 8080, dot left on, and the warning still appears

off by one corrected in both places, 1024 is the first unprivileged port and binding it needs nothing

the codeql note was right too - i was advancing the strtok_r loop variable in the body. separate pointer now, the same shape split_terminator_ports() uses, and with the comment that says why

DL6ER (approved):
All of it is addressed. `dns.dot` is in the check now, both thresholds read `< 1024`, and the loop no longer modifies its own variable.

Moving `DOT_PORT` into `server.h` is the right place for it. I checked the two things that move could have broken: `dot_server.c` still includes that header, so it keeps the constant, and the header carries no TLS-conditional while `dns.dot` is unconditional too - so the new include is safe in a build without TLS, which is the configuration CI never exercises.

PR #3072, opened 2026-09-06
close two memory-safety holes on error and signal paths

010011110:

two memory-safety fixes on error and signal paths, unrelated to each other beyond that

the three gravity write wrappers call dbclose_handle(db) and then return the worker's *message, which every failure path sets to sqlite3_errmsg(db). that string belongs to the connection just closed, so a failing write on /api/domains, /api/lists, /api/groups or /api/clients copies freed heap into the json error body. it is a use-after-free, though the reader is an already authenticated admin, so nothing crosses a privilege boundary. keep_message() snapshots the string before the handle dies, thread-local because the api answers several requests at once, and the roughly thirty *message = sqlite3_errmsg(db) assignments in the workers are left alone

FTL_TCP_worker_terminating(false) is reached only from dnsmasq's sig_handler(), both SIGALRM arms, each of which _exit(0)s straight after. what ran there was log_debug(), lock_shm() on a process-shared mutex, gravityDB_close() and unlock_shm(), none of it async-signal-safe. the case that hurts is an alarm landing while the worker holds the shm lock mid-update: is_our_lock() is true, so the handler skips the lock, finalizes sqlite statements the interrupted code was stepping, unlocks and exits, publishing a half-written query record to the parent. none of that work is needed on the way out, the worker's connections go with the process and the shm mutex is robust, so the parent takes it back through EOWNERDEAD

low impact for the first, an authenticated admin sees a garbled error message or a crash on a path that is already failing. medium for the second, it needs a tcp client holding a connection past the timeout and an alarm landing inside a narrow window, but what it corrupts is shared memory the parent goes on using

the first by inspection, or under asan: make a gravity write fail and read the json error. the second the same way, the signal path is not reachable on demand. full suite in the container is 194 bats, 151 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -2134,6 +2134,25 @@ static bool addToTable(sqlite3 *db, const enum gravity_list_type listtype, table
 	return okay;
 }
 
+// sqlite3_errmsg() hands back a string owned by the connection, and the
+// wrappers below close that connection before their caller ever reads the
+// message. Copy it somewhere that outlives the handle. Thread-local because
+// the API answers several requests at once, and only called on failure: a
+// success leaves *message as an earlier iteration set it, which would make
+// this copy a string onto itself
+static const char *keep_message(const char *message)
+{
+	static _Thread_local char kept[256];
+
+	if(message == NULL)
+		return NULL;
+
+	strncpy(kept, message, sizeof(kept) - 1);
+	kept[sizeof(kept) - 1] = '\0';
+
+	return kept;
+}
+
 bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
                           const char **message, const enum http_method method)
 {
@@ -2142,6 +2161,8 @@ bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
 		return false;
 
 	const bool ret = addToTable(db, listtype, row, message, method);
+	if(!ret && message != NULL)
+		*message = keep_message(*message);
 	dbclose_handle(db);
 	return ret;
 }
@@ -2417,6 +2438,8 @@ bool gravityDB_delFromTable(const enum gravity_list_type listtype, const cJSON* 
 		return false;
 
 	const bool ret = delFromTable(db, listtype, array, deleted, message);
+	if(!ret && message != NULL)
+		*message = keep_message(*message);
 	dbclose_handle(db);
 	return ret;
 }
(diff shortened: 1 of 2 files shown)


src/database/gravity-db.c
```diff
@@ -2142,6 +2159,8 @@ bool gravityDB_addToTable(const enum gravity_list_type listtype, tablerow *row,
 		return false;
 
 	const bool ret = addToTable(db, listtype, row, message, method);
+	if(message != NULL)
+		*message = keep_message(*message);
```
DL6ER: `addToTable()` sets `*message` only on failure paths, so in the `cJSON_ArrayForEach` loop at `list.c:564` an item that succeeds after an earlier one failed arrives here with `*message` still pointing at `kept`. That makes this `strncpy(kept, kept, 255)`, which is undefined for overlapping objects. Snapshot only on failure, which is the only case the message is read anyway:

```suggestion
	if(!ret && message != NULL)
		*message = keep_message(*message);
```

Same in the other two wrappers.

DL6ER (dismissed):
Checked the claims rather than the description.

Both callers passing `false` are inside dnsmasq's `sig_handler()` and `_exit(0)` on the next line, so the early return skips nothing that is still needed. The SHM mutex is robust (`shmem.c:618`) and the parent takes it back through `EOWNERDEAD` at `:692`, so the recovery you rely on is really there.

`sql_msg` is initialized to NULL at all three call sites in `api/list.c`, so reading `*message` unconditionally in the wrappers is safe. The read wrappers hand the statement back with the handle still open, so they do not share the bug and the scope of the fix is right.

Splitting these out of the advisory was the right call - neither of them needs to wait for it.

DL6ER:
My approval was premature - I posted it without reading the open thread above, and the point it makes is right.

`addToTable()` sets `*message` only on failure paths, so in the loop at `list.c:564` an item that succeeds after an earlier one failed still carries the pointer from the previous iteration, and the snapshot then copies that buffer onto itself. Undefined for overlapping objects, and the message is only read when `okay` is false anyway. Same in the other two wrappers.

What I did check stands: both callers passing `false` are in `sig_handler()` and `_exit(0)` on the next line, and the SHM mutex is robust, so the parent recovers through `EOWNERDEAD`. It is the guard that is missing, not the approach. Re-approving once that lands.

010011110:
you are right and it is undefined behaviour, not just untidy: addToTable() and the other two set *message only on failure, so an item succeeding after an earlier failure in the list.c:564 loop arrives with *message already pointing at kept, and the copy becomes strncpy(kept, kept, 255) over itself

snapshotting only on failure now, in all three wrappers, which is the only case anything reads it anyway. the helper's comment says so, so the next person does not restore the unconditional call

rebased onto current development, 194 bats, 152 pytest, 12, 25, 9

DL6ER (approved):
The guard is in all three wrappers, and the helper's comment records why it is failure-only, which is the part that stops someone restoring the unconditional call later.

PR #3073, opened 2026-09-06
add --config -t to check a value without applying it

010011110:

the api tests fail intermittently on two assertions, `unique_domains == 77` and `top_blocked == "gravity.ftl"`, roughly one arm/v6 job in ten. #3071 tried to relax them and was rightly closed: this is the cause

`--config key value` is the only way to ask the cli whether a value is acceptable, and it answers by installing it. for an item carrying FLAG_RESTART_FTL the write that follows restarts a running ftl, so a caller who only wants an answer has to change the configuration and put it back, and lives with whatever the daemon did in between

that is what the netmask test does: it writes a valid netmask, then writes an empty one to restore it. between those two cli calls the config file holds a changed value, and ftl's watcher restarts the daemon if it reads the file in that window, which depends on nothing but how long the two invocations take. a restart rebuilds the domain table from the query database, which has no record of domains no query references, so seven cname targets and pi.hole do not come back and their blocking credits go with them: 70 domains instead of 77, and gravity.ftl level with denied.ftl on 4, which then wins top_blocked on table order

`--config -t key value` runs the parse, the validator and the dnsmasq config test, then stops. nothing is installed, pihole.toml is untouched, and the value that would have been stored is printed. exit codes are the setter's own, so an invalid value stays distinguishable from an unknown key or a failed dnsmasq test. the netmask cases use it, and a following read asserts nothing was applied

no assertion is weakened: unique_domains stays 77 and top_blocked stays gravity.ftl, exactly as before

```
pihole-FTL --config -t dhcp.netmask 255.255.254.0   # prints the value, exit 0
pihole-FTL --config dhcp.netmask                    # still unset
pihole-FTL --config -t dhcp.netmask 255.254.255.0   # "not a valid netmask", exit 3
```

the flake reproduces under emulated linux/arm/v6, two runs in twelve before the change: `Restarting FTL due to change of dhcp.netmask` mid-suite in FTL.log, then the two assertions failing in pytest. with the change, twelve iterations under the same emulation and no restart from that write. x86 suite: 194 bats, 151 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -432,17 +432,25 @@ void parse_args(int argc, char *argv[])
 				                "         The value shown below may not reflect the current configuration.\n",
 				        GLOBALTOMLPATH);
 		}
-		if(argc == 2)
+		if(argc > 2 && strcmp(argv[2], "-t") == 0)
+		{
+			if(argc == 5)
+				exit(set_config_from_CLI(argv[3], argv[4], true));
+		}
+		else if(argc == 2)
 			exit(get_config_from_CLI(NULL, false));
 		else if(argc == 3)
 			exit(get_config_from_CLI(argv[2], false));
 		else if(argc == 4 && strcmp(argv[2], "-q") == 0)
 			exit(get_config_from_CLI(argv[3], true));
 		else if(argc == 4)
-			exit(set_config_from_CLI(argv[2], argv[3]));
-		else
+			exit(set_config_from_CLI(argv[2], argv[3], false));
+
+		// Anything else, including `--config -t key` with the value
+		// forgotten, falls through to the usage text
 		{
 			printf("Usage: %s --config [<config item key>] [<value>]\n", argv[0]);
+			printf("       %s --config -t <config item key> <value>\n", argv[0]);
 			printf("Example: %s --config dns.CNAMEdeepInspect true\n", argv[0]);
 			exit(EXIT_FAILURE);
 		}
@@ -1347,7 +1355,11 @@ void parse_args(int argc, char *argv[])
 			printf("\t%s--config %skey%s        Get current value of config item %skey%s\n", green, blue, normal, blue, normal);
 			printf("\t                    Config items with non-default values may\n");
 			printf("\t                    be colored in %sred%s\n", red, normal);
-			printf("\t%s--config %skey %svalue%s  Set new %svalue%s of config item %skey%s\n\n", green, blue, cyan, normal, cyan, normal, blue, normal);
+			printf("\t%s--config %skey %svalue%s  Set new %svalue%s of config item %skey%s\n", green, blue, cyan, normal, cyan, normal, blue, normal);
+			printf("\t%s--config -t %skey %svalue%s\n", green, blue, cyan, normal);
+			printf("\t                    Check whether %svalue%s would be accepted for\n", cyan, normal);
+			printf("\t                    %skey%s, without applying it or writing the\n", blue, normal);
+			printf("\t                    config file\n\n");
 
 			printf("%sEmbedded GZIP un-/compressor:%s\n", yellow, normal);
 			printf("    A simple but fast in-memory gzip compressor\n\n");
--- a/src/config/cli.c
+++ b/src/config/cli.c
@@ -380,7 +380,7 @@ static bool readStringValue(struct conf_item *conf_item, const char *value, stru
 	return true;
 }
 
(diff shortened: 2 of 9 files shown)


src/config/cli.c
```diff
@@ -509,6 +509,15 @@ int set_config_from_CLI(const char *key, const char *value)
 			write_custom_list();
```
DL6ER, line 509: Same for `dns.hosts`: this rewrites `custom.list` from the running config, so the content is the same, but the file is replaced and dnsmasq re-reads it.

```suggestion
			if(!test_only)
				write_custom_list();
```

DL6ER (commented):
The diagnosis checks out. The GC thread polls inotify once per second and `reread_config()` restarts on a changed restart-flagged item, so the window is the wall time between the two CLI writes, which under emulation easily spans a tick. I also went through the other restart-flagged writes in the suite (`dns.port`, `dns.revServers`, `misc.nice`) - all of them are rejected before anything is installed, so the netmask case was indeed the only one.

`-t` is not write-free yet, though. For a restart-flagged item, `write_dnsmasq_config(&newconf, true, ...)` does not stop after the test: when the test passes and the content differs, it renames the temporary file into `/etc/pihole/dnsmasq.conf` (the `files_different()` branch at the end of `write_dnsmasq_config()`). So `-t` on, e.g., `dns.upstreams` leaves a `dnsmasq.conf` that no longer matches `pihole.toml`. The test mode needs a way to run `test_dnsmasq_config()` and always discard the temporary file, e.g., a third argument or a separate helper for the test path. Two smaller things inline.

test/test_suite.bats
```diff
@@ -1653,19 +1653,23 @@ setup() {
...
   assert_line --index 0 '255.255.254.0'
   assert_success
 
   # An empty netmask is valid, it is then taken from the interface
-  run bash -c './pihole-FTL --config dhcp.netmask ""'
+  run bash -c './pihole-FTL --config -t dhcp.netmask ""'
+  assert_success
+
+  # None of the above was applied
+  run bash -c './pihole-FTL --config dhcp.netmask'
```
DL6ER, line 1667: This read-back cannot tell the difference: the `""` case right above would have restored the empty value even if `-t` had applied everything. Moved directly after the `255.255.254.0` case it actually asserts nothing was written. Note that `-t ""` equals the current value and takes the unchanged branch, i.e., the validator does not run for it.

src/args.c
```diff
@@ -439,10 +439,13 @@ void parse_args(int argc, char *argv[])
 		else if(argc == 4 && strcmp(argv[2], "-q") == 0)
 			exit(get_config_from_CLI(argv[3], true));
 		else if(argc == 4)
-			exit(set_config_from_CLI(argv[2], argv[3]));
+			exit(set_config_from_CLI(argv[2], argv[3], false));
+		else if(argc == 5 && strcmp(argv[2], "-t") == 0)
```
DL6ER: `--config -t key` (value forgotten) still lands in the `argc == 4` branch above and reports "Unknown config option -t". Worth checking for `-t` before the generic branches so the new usage line is what the user sees.

010011110:
the write is real and i had missed it. write_dnsmasq_config(conf, true, ...) tests and then installs, so -t on a restart-flagged item left dnsmasq.conf out of step with pihole.toml, which is worse than what it was meant to avoid

the bool was the trap: test_config only ever meant "test it as well". it is an enum now, DNSMASQ_INSTALL, DNSMASQ_TEST_INSTALL and DNSMASQ_TEST_ONLY, and the last one removes the file it tested instead of renaming it into place. six call sites updated, all mechanical. write_custom_list() is skipped in test mode for the same reason you give, the content matches but the file is replaced and dnsmasq re-reads it

checked rather than assumed: -t on dns.upstreams now leaves no dnsmasq.conf and no leftover temp file

--config -t key without a value used to reach the argc == 4 branch and report "Unknown config option -t". -t is matched before the generic branches now, so it prints the usage instead

read-back moved to directly after the 255.255.254.0 case, where it can actually tell the difference, and the comment on the "" case notes it takes the unchanged branch so no validator runs

DL6ER (approved):
Checked the revision against the tree: all seven write_dnsmasq_config() callers are converted, the temp file is the only thing that function writes, and -t now leaves nothing behind. Thanks for turning the bool into an explicit mode, that was the actual trap here.

PR #3074, opened 2026-09-06
wait properly for the test resolver, and say so when it does not come

010011110:

test/pdns/setup.sh starts pdns_server and pdns_recursor and polls each for five seconds, then carries on whether or not they answered:

```sh
pdns_server --daemon
for _ in 1 2 3 4 5 6 7 8 9 10; do
  dig @127.0.0.1 -p 5554 ftl. SOA +tries=1 +time=1 +short > /dev/null 2>&1 && break
  sleep 0.5
done
```

on a slow or emulated runner that is not always enough, and if the resolver is not up the tests query something that is not there. what follows is a wall of dns failures with nothing saying why, which is a poor way to learn that a daemon did not start

this makes no claim on the intermittent unique_domains failure. that has a confirmed cause in #3073, the dhcp.netmask write tripping the config watcher, and a dead resolver does not fit it

both loops become one wait_for_pdns() that waits thirty seconds and returns failure naming the process and the port, and run.sh stops there rather than running a dns test suite against nothing

low impact and no behaviour change when the resolver comes up, which is the normal case. what it buys is that the other case says what happened

break it on purpose: point pdns at a port already in use, or chmod -x the binary, and the run stops with "ERROR: pdns_server did not answer on port 5554 within 30 s" instead of continuing into a wall of dns failures. setup.sh also runs under set -euo pipefail now, and the suite is unchanged at 194 bats, 151 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/test/pdns/setup.sh
+++ b/test/pdns/setup.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euo pipefail
 
 echo "************ Installing PowerDNS configuration ************"
 
@@ -156,7 +157,7 @@ pdnsutil zone secure dnssec
 #   dnssec. IN DS 42206 8 2 6d2007e292483fa061db37011676d9592649d1600e5b2ece1326f792ebedd412 ; ( SHA256 digest )
 # --->
 #   trust-anchor=dnssec.,42206,8,2,6d2007e292483fa061db37011676d9592649d1600e5b2ece1326f792ebedd412
-pdnsutil zone export-ds dnssec. | head -n1 | awk '{FS=" "; OFS=""; print "trust-anchor=",$1,",",$4,",",$5,",",$6,",",$7}' > /etc/dnsmasq.d/02-trust-anchor.conf
+pdnsutil zone export-ds dnssec. | awk 'NR==1{FS=" "; OFS=""; print "trust-anchor=",$1,",",$4,",",$5,",",$6,",",$7}' > /etc/dnsmasq.d/02-trust-anchor.conf
 
 # Create a locally-signed root zone so DNSSEC validation never leaves the test
 # environment. FTL ships the real ICANN root trust anchors, so without a local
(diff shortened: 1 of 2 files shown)


DL6ER:
The change is right, and with no `set -e` in either script the `run.sh` guard is doing real work.

Drop the attribution though. #3073 has a confirmed cause for the seven-domain shortfall - the `dhcp.netmask` write trips the inotify watcher, FTL restarts, and a restart cannot rebuild the seven domains no query references. A dead resolver gives a wall of DNS failures, as you say yourself, not a clean run with one count short. If both PRs claim it and it recurs, we will not know which mattered.

While in here, `set -euo pipefail` at the top of `setup.sh` would cover the rest of it too.

010011110:
attribution dropped. you are right that it cannot be both, and #3073 has the confirmed mechanism where this one only had a plausible story - a dead resolver gives a wall of dns failures, not a clean run seven domains short, as the description itself said. the body now claims only what this change does: a resolver that does not come up says so instead of being discovered later

set -euo pipefail added to setup.sh, suite unchanged

test/pdns/setup.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euo pipefail
```
DL6ER, line 2: `pipefail` reaches three pipelines further down that are shaped `pdnsutil zone export-ds ... | head -n1 | awk ...` (lines 159, 167 and 183). When `head` exits after the first line it closes the pipe, `pdnsutil` takes SIGPIPE, the pipeline reports 141, and `set -e` turns that into an abort before the resolver is ever started - with `run.sh` now exiting on a failed setup, that takes the whole run with it.

Whether it fires depends on whether `pdnsutil` finishes writing before `head` leaves, so it is a race, in a change meant to remove one. `awk 'NR==1{...; exit}'` instead of `head -n1` avoids it, or wrap the producer as `{ pdnsutil ... || true; }`.

test/pdns/setup.sh
```diff
@@ -230,17 +231,33 @@ done
 mkdir -p /var/run/pdns-recursor
 rm -f /var/run/pdns-recursor/*
 
+# Wait for one of the pdns processes to answer on its port. The old loop gave
+# up after five seconds and carried on regardless, which on a slow (emulated)
+# runner means the tests query a resolver that is not there yet: the CNAME
+# chain below never resolves, its intermediate domains are never created, and
+# the failure surfaces much later as an API test counting seven domains too
```
DL6ER: This attributes the seven-domain shortfall to a resolver that is not up yet. #3073 established a different cause, the `dhcp.netmask` write tripping the config watcher, and the PR body says as much - so the comment contradicts its own description.

I would drop the history entirely rather than correct it: say what the function waits for and why thirty seconds. The story of what used to go wrong belongs in the commit message, where it does not go stale.

DL6ER (commented):
The wait itself is right, and stopping the run rather than continuing into a wall of DNS failures is the correct call. Two things inline.

010011110:
both fixed

the comment no longer tells the story: it says what the function waits for and why thirty seconds. you are right that the history goes stale in a comment and belongs in the message

on the pipelines, i took the awk route but without the exit. awk 'NR==1{...; exit}' closes the pipe exactly as head does, so the producer can still take SIGPIPE and pipefail still turns it into 141 - the race would be narrower, not gone. NR==1 without exit lets awk drain the input, so there is nothing to signal. all three converted, and the two dnssec tests still pass, which is what those trust anchors feed

DL6ER (approved):
Both points are addressed. `awk 'NR==1{...}'` without an `exit` reads the producer to completion, so nothing takes SIGPIPE under `pipefail` - and those three were the only pipelines in the script, so the hazard is gone rather than narrowed. The comment now says what the function waits for and leaves the history out.

PR #3075, opened 2026-09-06
feat: support stdin for --config values via - marker

bcambl:

`pihole-FTL --config <key> <value>` puts the value in argv, which is visible via `/proc/<pid>/cmdline` and `ps -eo args=`. For secret values (notably the web password) this leaks to other unprivileged users for the lifetime of the child process.

Allow reading the value from stdin by passing `-` in place of the value: `echo secret | pihole-FTL --config webserver.api.password -`. fgets reads up to 4095 bytes into a stack buffer, trailing CR/LF is stripped, and the existing set_config_from_CLI validation pipeline is reused unchanged.

Buffer is bounded at 4096 bytes; values that fill the buffer without a terminating newline are rejected with a clear error rather than silently truncated.

TBD

---

---

### DIFF
--- a/test/test_suite.bats
+++ b/test/test_suite.bats
@@ -1740,6 +1740,82 @@ setup() {
   assert_line --partial --index 0 '"no password set"'
 }
 
+@test "CLI: Setting password via stdin (--config <key> -) leaves no net change" {
+  # Set password via stdin (value is piped, not in argv)
+  logsize_before=$(stat -c%s /var/log/pihole/FTL.log)
+  run bash -c "echo 'STDIN_PW' | ./pihole-FTL --config webserver.api.password -"
+  assert_success
+
+  # Wait for the running FTL instance to pick up the config file change
+  run bash -c "./pihole-FTL wait-for 'pihole.toml unchanged' /var/log/pihole/FTL.log 5 $logsize_before"
+  assert_success
+
+  # Verify login is required
+  run bash -c 'curl -s 127.0.0.1/api/auth'
+  assert_line --partial --index 0 '"valid":false'
+
+  # Verify the stdin-supplied password works
+  run bash -c 'curl -s -X POST 127.0.0.1/api/auth -d "{\"password\":\"STDIN_PW\"}" | jq .session.valid'
+  assert_line --index 0 "true"
+
+  # Verify the password was never present in the long-running FTL
+  # daemon's argv. The daemon was started without --config so its
+  # cmdline should not contain the password.
+  for pid in $(pidof pihole-FTL 2>/dev/null); do
+    if [ -r "/proc/$pid/cmdline" ]; then
+      ! tr '\0' ' ' < "/proc/$pid/cmdline" | grep -q "STDIN_PW"
+    fi
+  done
+
+  # Remove password via stdin (empty value)
+  logsize_before=$(stat -c%s /var/log/pihole/FTL.log)
+  run bash -c "echo '' | ./pihole-FTL --config webserver.api.password -"
+  assert_success
+
+  # Wait for the running FTL instance to pick up the config file change
+  run bash -c "./pihole-FTL wait-for 'pihole.toml unchanged' /var/log/pihole/FTL.log 5 $logsize_before"
+  assert_success
+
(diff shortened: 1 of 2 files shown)


DL6ER:
Worth more than the description claims: `pihole setpassword` does this itself. `setFTLConfigValue()` (`utils.sh:137`) runs `pihole-FTL --config "$1" "$2"` under sudo, so the plaintext sits in a root-owned `/proc/<pid>/cmdline`, which is world-readable - any unprivileged local user can read it, no `hidepid` in the default mount. It also means this needs a companion Core PR piping into `-`, or nothing changes for the normal path.

This has to wait for #2960 first: `-` is still a valid value for `files.log.dnsmasq`, and #2960 removes it - the last place `-` is a value rather than a marker.

Three things meanwhile:

1. `log_err()` prints to stdout here (`log.c:276`), and `args.c:411` explains why this block keeps stdout clean. Use `fprintf(stderr, ...)`.
2. The `pidof` loop passes with or without this change - the daemon never had the password in its argv.
3. #3073 landed today and gave `set_config_from_CLI()` a third parameter, so the rebase is not just textual. Docs PR is still TBD.

test/test_suite.bats
```diff
@@ -1740,6 +1740,135 @@ setup() {
...
+  # argv holds only the "-" marker, never the password itself
+  run bash -c "tr '\\0' ' ' < /proc/${cli_pid}/cmdline"
+  assert_success
+  assert_output --partial "pihole-FTL --config webserver.api.password -"
+  refute_output --partial "FIFO_PW"
+
+  # Release the blocked reader and let the CLI finish setting the password
+  echo 'FIFO_PW' >&9
+  exec 9>&-
+  wait "${cli_pid}"
```
DL6ER, line 1820: `curl` runs before the running FTL has reloaded `pihole.toml`. Wait for it like the other password tests do:

```suggestion
  logsize_before=$(stat -c%s /var/log/pihole/FTL.log)
  echo 'FIFO_PW' >&9
  exec 9>&-
  wait "${cli_pid}"
  run bash -c "./pihole-FTL wait-for 'pihole.toml unchanged' /var/log/pihole/FTL.log 5 $logsize_before"
  assert_success
```

test/test_suite.bats
```diff
@@ -1740,6 +1740,135 @@ setup() {
...
+  run bash -c 'curl -s -X POST 127.0.0.1/api/auth -d "{\"password\":\"MULTI\"}" | jq .session.valid'
+  assert_line --index 0 "true"
+
+  # The second line was NOT used
+  run bash -c 'curl -s -X POST 127.0.0.1/api/auth -d "{\"password\":\"SECOND_LINE\"}" | jq .session.valid'
+  refute_line --index 0 "true"
+
+  # Cleanup: remove the password
+  run bash -c "./pihole-FTL --config webserver.api.password \"\""
+  assert_success
```
DL6ER: Same here, otherwise the next test can still see the password set:

```suggestion
  logsize_before=$(stat -c%s /var/log/pihole/FTL.log)
  run bash -c "./pihole-FTL --config webserver.api.password \"\""
  assert_success
  run bash -c "./pihole-FTL wait-for 'pihole.toml unchanged' /var/log/pihole/FTL.log 5 $logsize_before"
  assert_success
```

DL6ER (dismissed):
All three points are addressed and #2960 is merged, so this is no longer blocked.

Two tests race the config reload, see the suggestions below.

This also only helps once `pihole setpassword` uses it. I found no companion Core PR (`setFTLConfigValue()` piping the value into `-`) and no docs PR yet. Please open both and link them here.

DL6ER (approved):
Thanks!

darkexplosiveqwx:
@bcambl 
image


I have also created https://github.com/pi-hole/pi-hole/pull/6705 for `setFTLConfigValue()`.

rdwebdesign:
@bcambl 

Your commits need [GPG](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) or [SSH](https://docs.pi-hole.net/guides/github/sign/) signing.

PR #3079, opened 2026-09-07
Fix timestamp losing ms accuracy for rolling-over seconds

darkexplosiveqwx:

Make `get_timestr()` take a `double` instead of `time_t` per **DL6ER**'s request in https://github.com/pi-hole/FTL/pull/3079#issuecomment-5580756721.
This avoids re-sampling the time via `gettimeofday()` if millies are requested, which may result in a different second losing millisecond accuracy.

We use `double` because it is an already established pattern across the codebase via functions like `double_time()`. Precision is not an issue, as it will remain accurate to at least a microsecond for over 100a

~~Make `get_timestr()` take a `struct timespec` instead of `time_t`.
This avoids re-sampling the time via `gettimeofday()` if millies are requested, which may result in a different second losing millisecond accuracy.~~

~~Why `struct timespec` and not `struct timeval`?
`timeval` may seem more fitting because we never use the nanosecond representation provided by `timespec`. I still prefer it because `timespec` is explicitly preferred by POSIX and recommended for new APIs. In fact, gettimeofday() got marked as obsolete in The Open Group Specification Issue 7 (POSIX.1-2008) and officially removed in The Open Group Specification Issue 8 (POSIX.1-2024). This shows that `timeval` is effectively obsolete and shouldn't be relied upon in new code.~~

This PR is not simply about fixing the bug, but also about modernization of the API, as it will be beneficial to my further logging changes and also for https://github.com/pi-hole/FTL/pull/3005.

~~Change the parameter of `get_timestr()`.
Convert all existing calls to use `clock_gettime()` if ms are wanted or the `(struct timespec){ .tv_sec = now, .tv_nsec = 0 }` pattern if ms are not required to easily convert an existing `time_t` to `struct timespec`.~~

---

---

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -80,10 +80,13 @@ double double_time(void)
 }
 
 // Get a human-readable time string
-void get_timestr(char timestring[TIMESTR_SIZE], const time_t timein, const bool millis, const bool uri_compatible)
+// timein is a double epoch-seconds timestamp as produced by double_time() so
+// that the millisecond fraction comes from the timestamp itself
+void get_timestr(char timestring[TIMESTR_SIZE], const double timein, const bool millis, const bool uri_compatible)
 {
+	const time_t seconds = (time_t)timein;
 	struct tm tm;
-	localtime_r(&timein, &tm);
+	localtime_r(&seconds, &tm);
 	char space = ' ';
 	char colon = ':';
 	if(uri_compatible)
@@ -94,9 +97,7 @@ void get_timestr(char timestring[TIMESTR_SIZE], const time_t timein, const bool 
 
 	if(millis)
 	{
-		struct timeval tv;
-		gettimeofday(&tv, NULL);
-		const int millisec = tv.tv_usec/1000;
+		const int millisec = (int)((timein - seconds) * 1000.0);
 
 		snprintf(timestring, TIMESTR_SIZE, "%d-%02d-%02d%c%02d%c%02d%c%02d.%03i%c%s",
 		        tm.tm_year + 1900, tm.tm_mon + 1, tm.tm_mday, space,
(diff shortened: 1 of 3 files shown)


darkexplosiveqwx:
From a quick local test, this merges without conflicts with https://github.com/pi-hole/FTL/pull/2960, so any can go first.
And I'd actually like for this PR to go before https://github.com/pi-hole/FTL/pull/2968.

DL6ER:
On the type, I would take a `double` instead of a `struct timespec`. `double_time()` sits right above `get_timestr()` in `log.c` and already carries your POSIX argument, and `lastQuery` is a `double` as well, so we have effectively standardized on double epoch seconds. It leaves all second-granular call sites untouched (`time(NULL)` converts implicitly), drops the eight `(struct timespec){ .tv_sec = ..., .tv_nsec = 0 }` literals and makes `timespec_from_double()` unnecessary - the diff shrinks to `log.c`, `log.h` and three call sites.

Precision is not the reason to prefer `timespec` here. A binary64 has a 53-bit significand, so for a value in `[2^e, 2^(e+1))` the spacing is `2^(e-52)`. Unix time currently sits in `[2^30, 2^31)`, which gives `2^-22` = 238 ns; that doubles to 477 ns past `2^31` in 2038 and first exceeds a microsecond at `2^33`, in 2242. `nextafter(t, INFINITY) - t` confirms it. So #3005 keeps all six of its digits, not just the three we print today.

darkexplosiveqwx:
I like `struct timespec` because it is explicit for storing time and the standardized way, both in POSIX (via `clock_gettime()`) and in ISO C11 (via `timespec_get()`), but if the Project already standardized on using `double` for storing time, then I will honor that.

DL6ER (approved):
Thanks!

PR #3080, opened 2026-09-07
Change CSP for img-src to allow inline Base64 encoded data URIs

rdwebdesign:

Allow the web interface to load Base64 encoded images from Bootstrap 5 CSS files.

Adding `data:` to the `img-src` Content Security Policy.

**NOTE:**
This will change the default value for new installations, but it won't change the current `webserver.headers` value from previous installations.

--- 

**CSP evaluation:**
image

### Link documentation PRs if any are needed to support this PR:

None

---

---

### DIFF
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -881,7 +881,7 @@ components:
             threads: 0
             headers:
               - "X-DNS-Prefetch-Control: off"
-              - "Content-Security-Policy: default-src 'none'; connect-src 'self'; font-src 'self'; frame-ancestors 'none'; img-src 'self'; manifest-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; form-action 'self'"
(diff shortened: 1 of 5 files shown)


darkexplosiveqwx:
You forgot to update `test/pihole.toml`

DL6ER:
@rdwebdesign 

> This will change the default value for new installations, but it won't change the current webserver.headers value from previous installations.

Do we want this? If so, we can write a small migration that replaces the existing string if it is 1:1 the old one

rdwebdesign:
If possible, we need a migration.

All new themes will use Bootstrap 5. 
The images from the CSS files will be needed for dropdown arrows, close buttons, checkboxes and a few other things. 

Without the new CSP header, the images won't be visible and the themes will look broken.

DL6ER:
I wrote the migration. It walks the stored `webserver.headers` and swaps the Content-Security-Policy only when it is byte-identical to the old default, so anyone who wrote their own policy keeps what they wrote.

That is the conservative reading and it leaves one gap: someone who customized their CSP for an unrelated reason still ends up without `img-src data:`, and still gets the broken themes.

DL6ER (dismissed):

DL6ER (approved):

rdwebdesign:
> That is the conservative reading and it leaves one gap: someone who customized their CSP for an unrelated reason still ends up without `img-src data:`, and still gets the broken themes.

I think the migration is enough. 

If users changed the headers to something else, we shouldn't touch the value.

PR #3081, opened 2026-09-08
latch a corrupt database when the api messages paths hit one

010011110:

delete_message(), count_messages() and format_messages() all return early on FTLDBerror(), so they expect the flag to be set for them. none of them sets it: they run raw sqlite3_prepare_v2() and sqlite3_step() and only log what comes back. a corruption first met on /api/info/messages is reported upward and retried by every later caller instead of being latched once and answered from the flag

each of the three hands its return code to checkFTLDBrc() now, which is what the rest of the database layer does - 38 call sites across network-table.c, common.c and aliasclients.c, and two in this file already. count_messages() and format_messages() had an rc and a cleanup label to hang it on, delete_message() gains the variable and reports where its loop breaks converge

same class as #3062 and #3064, in the file those two left out

low impact: it needs a corrupt or unwritable pihole-FTL.db to reach at all. what it changes there is that the failure is recorded once rather than rediscovered on every request

by inspection, or by corrupting pihole-FTL.db (truncate it mid-file) and calling /api/info/messages: FTL.log gets the "Database is not available" latch once, and later requests answer from the flag instead of reopening and failing again. full suite in the container is 194 bats, 152 pytest, 12, 25, 9

## Additional information

---

## Checklist:

### DIFF
--- a/src/database/message-table.c
+++ b/src/database/message-table.c
@@ -482,6 +482,12 @@ static int _add_message(const enum message_type type,
 
 end_of_add_message: // Close database connection
 
+	// Every path that jumps here carries a failing rc, and the duplicate
+	// removing DELETE above is on a hotter path than anything else in this
+	// file - a corrupt database met there has to latch as well
+	if(rc != SQLITE_OK && rc != SQLITE_DONE)
+		checkFTLDBrc(rc);
+
 	// Final database handling
 	if(stmt != NULL)
 		sqlite3_finalize(stmt);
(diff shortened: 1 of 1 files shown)


DL6ER (commented):
The additions look right - `rc` is set on every path that reaches them, and the placement matches the convention at lines 463/476. Three things around them:

`format_messages()` (1380) latches `DBerror` but still returns `true`, so `info.c:933` never takes its 500 branch. The request that finds the corruption answers 200 with a truncated array, and only the next one errors.

`_add_message()` (393) still gotos `end_of_add_message` from the duplicate-removing DELETE with no latch. Same class as what you are fixing, on a hotter path.

The new guard at 1373 excludes `SQLITE_OK`/`SQLITE_DONE` but not `SQLITE_ROW`, unlike the sibling at 1079. A no-op today, but asymmetric.

DL6ER (approved):
All three are addressed: `format_messages()` returning `rc == SQLITE_DONE` is what makes the 500 in `info.c` reachable at all, `_add_message()` now latches where its paths converge, and the `SQLITE_ROW` exclusion is documented rather than just symmetric.

PR #3083, opened 2026-09-09
four performance findings, no behaviour change except one new error path

010011110:

the payload buffer. every endpoint with API_PARSE_JSON allocated 64 KiB and
calloc zeroes all of it, before looking at the request. that flag is on almost
the whole table, GET endpoints included, so a dashboard polling
/api/stats/summary paid a 64 KiB allocation, a 64 KiB memset and a free per
request with no body to read. now allocated inside read_and_parse_payload() and
only when the request announces one

the clients list. GET /api/clients resolves a name per row, and both lookups
opened pihole-FTL.db, ran one SELECT and closed it again, once per client.
getNameFromMAC() gets the optional connection parameter getNameFromIP() already
had, and the list opens it once

gravity batch adds. POST /api/domains with an array called gravityDB_addToTable()
and gravityDB_edit_groups() per item, each opening a write connection,
committing and closing. N items cost 2N connections and 2N commits, and every
commit is an fsync. one connection and one transaction for the whole loop now.
modelled on the same schema and access pattern, 100 items go from about 1.2 s to
6 ms and 1000 from 12.5 s to 10 ms. the gain is the commits. nothing covered
that path, so there is a pytest case for it now, five domains in one request

query_storage(id). id is INTEGER PRIMARY KEY, so it is the rowid and the table
b-tree already is that index. idx_query_storage_id maintained a second one with
the same keys, costing 990 pages and 4 percent of the insert time per 300000
rows. only the in-memory database ever created it, no on-disk schema or
migration is touched

the one behaviour change is in the gravity batch: nothing reaches the disk until
the commit, so a failing commit now answers 500 instead of reporting items as
successes that were lost. a failing BEGIN is not fatal, it falls back to
committing per item

post an array of domains to /api/domains/deny/exact and compare the wall time
against development. for the index, create it by hand on a pihole-FTL.db and
diff EXPLAIN QUERY PLAN before and after: id lookups and id ranges do not move,
only MAX(id) picks the index up, and that is a seek to the last rowid either way

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -167,9 +167,10 @@ int api_handler(struct mg_connection *conn, void *ignored)
 
 			if(api_request[i].opts.flags & API_PARSE_JSON)
 			{
-				// Allocate memory for the payload
-				api.payload.raw = calloc(MAX_PAYLOAD_BYTES, sizeof(char));
-				if(!api.payload.raw)
+				// Read and try to parse payload. The buffer is
+				// allocated in there, and only for a request that
+				// actually carries a body
+				if(!read_and_parse_payload(&api))
 				{
 					log_crit("Cannot handle API request %s %s: %s",
 							api.request->request_method,
@@ -182,9 +183,6 @@ int api_handler(struct mg_connection *conn, void *ignored)
 					                      NULL);
 					break;
 				}
-
-				// Read and try to parse payload
-				read_and_parse_payload(&api);
 			}
 
 			// Verify requesting client is allowed to see this resource
--- a/src/api/list.c
+++ b/src/api/list.c
@@ -728,7 +728,10 @@ static int api_list_write(struct ftl_conn *api,
 		// own while the ones before it were undone, so neither the
 		// per-item results above nor a plain commit failure describe
 		// what is in the database
-		const char *commit_msg = sqlite3_get_autocommit(db) != 0
+		// SQLite has ended the transaction itself if the connection is
+		// back in autocommit, which means part of the batch is on disk
+		const bool partially_applied = sqlite3_get_autocommit(db) != 0;
+		const char *commit_msg = partially_applied
 		                       ? "Gravity database batch was only partially applied"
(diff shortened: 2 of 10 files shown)


DL6ER (commented):
The payload and index parts look right to me - `payload_expected()` is equivalent to civetweb's own framing, `payload.raw` is only read behind `payload.avail`, and `id` is a rowid alias so the index really is redundant. The batch is where I have concerns.

`BEGIN TRANSACTION` (622) is deferred, so the connection holds only SHARED when the first INSERT has to promote to RESERVED - and SQLite deliberately skips the busy handler on that promotion. During a `pihole -g` the whole batch now fails instantly with "database is locked" where each item previously got the full `DATABASE_BUSY_TIMEOUT`. `BEGIN IMMEDIATE` restores that.

The `JSON_*` macros expand to `return 500`, and both calls in the loop sit between `gravityDB_write_open()` and the close at 695. A cJSON allocation failure then leaks a handle with an open write transaction, which locks gravity.db for the rest of the process. Same shape at 103 for `namedb`, where `rows` and `stmt` already leak - one `goto` label would cover all of it.

The partial-apply 500 (667) returns before `set_event(RELOAD_GRAVITY)`, so rows that did reach disk stay unblocked until an unrelated reload.

DL6ER:
Set a bool where `sqlite3_get_autocommit()` is checked instead of inferring the partially-applied case from `commit_err[0] == '\0'`.

The two leaks the row loop picked up on the way - `rows`/`stmt` on a failing `JSON_*`, `processed` on the error returns - are worth having on their own.

DL6ER (approved):
The `partially_applied` bool reads far better than inferring it from `commit_err`, and dropping `idx_query_storage_id` is correct - `id` is an `INTEGER PRIMARY KEY` so the rowid B-tree already is that index, and `index_creation[]` only ever feeds the in-memory database.

PR #3084, opened 2026-09-09
405 with an Allow header instead of 404 for a wrong method

010011110:

a request to an endpoint that exists with a method it does not accept falls
through the whole table and comes back as 404 Not found. DELETE
/api/stats/summary is the easy example. RFC 9110 15.5.6 wants a 405 that names
the methods that would have worked, and the handler was already collecting
exactly that set to answer OPTIONS, it just threw it away

so the change is small: check the URI first and the method second, let every
URI match contribute its methods to allowed_methods, and answer 405 with an
Allow header when nothing matched by method but something did by URI. a URI that
does not exist at all is still a 404 with no Allow header

## on the header route, since it came up

i had this filed as blocked on there being no way to attach a header to an error
response, since my_send_http_error_headers() emits the whole block. that turns
out to be wrong. it calls send_additional_header(), which appends
pi_hole_extra_headers, so an error response has always been able to carry a
header, and that is the route the Location header of a list add and the session
cookie already take. i used the same one here

asked whether a real parameter on send_json_error() would be preferred to the
thread-local and the answer was no: send_additional_header() emits and clears
pi_hole_extra_headers in the same place, so it cannot reach the next response,
and a parameter would mean touching civetweb.c and mirroring that in
patch/civetweb/. so the thread-local stays

two smaller ones while you are looking

OPTIONS is in the Allow list. an earlier version of this description said it was
left out - that was wrong, allowed_methods ors in HTTP_OPTIONS and the formatter
runs j <= HTTP_OPTIONS, so it is advertised by both the new 405 and the existing
204. that also means the OPTIONS answer itself changed: it used to leave itself
out of the list it returned

and the openapi spec is untouched. a 405 belongs to the path, not to an
operation, and the method being refused has no operation object to hang a
response on, so i did not see a clean place for it. tell me if you want it
somewhere anyway

one behaviour note: this tells an unauthenticated caller which methods a URI
takes. /api/docs is public and documents that already, so i do not think
anything new leaks, but it is a change worth naming

curl -i -X DELETE /api/stats/summary and look at the status line and the Allow
header, then the same against a URI that does not exist

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -109,6 +109,58 @@ static struct {
 	{ "/api/docs",                              "",                           api_docs,                              { API_PARSE_JSON, 0                         }, false, HTTP_GET },
 };
 
+// Does the URI carry as many path components as this table row expects? Several
+// rows share a URI and are told apart only by their parameters, which
+// startsWith() does not look at - /api/domains has four of them. Answering a
+// 405 with the union of all four advertises a DELETE that /api/domains without
+// arguments would refuse with a 400, so count the components and keep the rows
+// that would really have taken this URI
+static bool __attribute__((pure)) parameters_match(const char *parameters, const char *item)
+{
+	unsigned int expected = 0;
+	for(const char *p = parameters; *p != '\0'; p++)
+		if(*p == '{')
+			expected++;
+
+	// A trailing slash does not open another component: /api/domains/deny/
+	// addresses the same one-parameter row as /api/domains/deny
+	size_t len = strlen(item);
+	if(len > 0 && item[len - 1] == '/')
+		len--;
+
+	unsigned int found = 0;
+	if(len > 0)
+	{
+		found = 1;
+		for(size_t i = 0; i < len; i++)
+			if(item[i] == '/')
+				found++;
+	}
+
+	return expected == found;
+}
+
+// Format the methods an endpoint accepts as an Allow header value, e.g.
+// "GET, POST, OPTIONS". OPTIONS is answered for every endpoint by the handler
+// itself, so it belongs in a header a client is meant to act on
+static void format_allowed_methods(char *buffer, const size_t size, const enum http_method allowed)
+{
+	size_t len = 0;
+	buffer[0] = '\0';
+	for(enum http_method j = HTTP_GET; j <= HTTP_OPTIONS; j <<= 1)
+	{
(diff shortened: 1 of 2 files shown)


DL6ER (commented):
The loop restructure is dispatch-equivalent and the extra-header thread-local is safe - `send_additional_header()` clears it on both paths. But `ret == 0` is overloaded: it also means "the handler ran and wants a 404". `api_docs()` returns 0 for a file it cannot find, so `GET /api/docs/specs/nope.yaml` now answers 405 with `Allow: GET`, denying a method the header says is allowed. That path is unauthenticated, and `test_openapi.py:46` skips it. A separate flag set on the method-check `continue` keeps the two cases apart.

`allowed_methods` also unions every row sharing a URI prefix, since `startsWith()` ignores `parameters`. `PATCH /api/domains` advertises DELETE, which then returns 400 `uri_error`. That was fine when it only fed OPTIONS; on a 405 it is what clients act on.

Smaller: HEAD maps to `HTTP_UNKNOWN`, so it gets 405 with `Allow: GET`, and `j < HTTP_OPTIONS` omits OPTIONS - which matters more now that the header is normative.

DL6ER:
On your question: keep the thread-local. `send_additional_header()` emits `pi_hole_extra_headers` and clears it in the same place (`civetweb.c:4232-4235`), so it cannot reach the next response. A parameter on `send_json_error()` would mean touching `civetweb.c`, and that change would then have to be mirrored in `patch/civetweb/` as well.

The description needs a correction rather than the code: it says the Allow list leaves OPTIONS out, but `allowed_methods |= api_request[i].methods | HTTP_OPTIONS` and the loop now runs `j <= HTTP_OPTIONS`. OPTIONS is advertised in both the new 405 and the existing 204. That is the right call, it just means the OPTIONS answer changed too and the PR should say so.

Leaving the OpenAPI spec alone is fine by me, your reasoning holds. Add a `parameters_match()` case with a trailing slash.

DL6ER (approved):
All three are addressed: `handler_ran` keeps a handler asking for its own 404 apart from a method mismatch, `parameters_match()` stops the union advertising a DELETE that `/api/domains` would refuse with a 400, and OPTIONS is in the header now.

PR #3086, opened 2026-09-11
two defects in the gravity database reader, one of them a double free

010011110:

two defects in the gravity database reader, one of them a double free

gravityDB_readTable() builds a LIKE pattern only when the item is non-empty, but
all four free() calls guard on !exact alone. so an item that is present but
empty makes the function free a string it never allocated

api_search() reaches that: it passes exact = !partial, and idn2_to_ascii_lz()
answers IDN2_OK with an empty heap string for input that IDNA maps away
entirely - a soft hyphen does it. GET /api/search/%C2%AD?partial=true frees
punycode, uses it in the five searches that follow and frees it again at the
end. whether the conversion gives an empty string rather than an error depends
on LC_CTYPE, which daemon.c inherits from the service environment, so this is
real wherever LANG is a UTF-8 one

second one, unrelated failure but the same file: the regex group statement is
shared between all clients and sqlite3_reset() does not clear bindings.
gravityDB_get_regex_client_groups() skips the carray bind when a client has no
groups and steps anyway, so that client is given the previously processed
client's deny or allow regexes. the gravity, allowlist and denylist lookups all
return early in that case, this one was the outlier

GET /api/search/%C2%AD?partial=true under a UTF-8 locale, before and after. for
the second one, put a client in no group at all and reload gravity with another
client that has regex rules

## Additional information

---

## Checklist:

### DIFF
--- a/src/database/gravity-db.c
+++ b/src/database/gravity-db.c
@@ -2554,7 +2554,12 @@ bool gravityDB_readTable(sqlite3 *db, const enum gravity_list_type listtype,
 		*message = "Failed to allocate memory for query string";
 		return false;
 	}
+	// like_name is the caller's item until we build a LIKE pattern of our own.
+	// The free() calls below have to follow the allocation, not just !exact:
+	// an empty non-exact item skips the allocation and would otherwise make
+	// this function free a string it never owned
 	char *like_name = (char*)item;
+	bool like_name_allocated = false;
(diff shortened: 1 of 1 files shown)


DL6ER (approved):
Both hold: an empty non-exact item made this free the caller's `punycode` rather than a pattern of its own, and the shared regex statement handed a group-less client the previously processed client's regexes.

PR #3087, opened 2026-09-11
upstream accounting, a stale query index, and three smaller things in the bridge

010011110:

upstream accounting, a stale query index, and three smaller things in the bridge

FTL_forwarded() counts a query against the upstream it first picks and returns
early for every further server of the same forward round, so a forwardall probe
counts only the first. update_upstream() then repoints query->upstreamID at
whichever server answered without moving the count, so the counted server keeps
a +1 nobody takes back and the answering one is decremented by the GC although
it was never counted - upstream->count is a signed int and goes negative.
query_blocked() has the opposite problem, handing the count back and leaving
upstreamID set so the GC hands it back again

there was no record of whether a query holds a count, which is the root of both.
flags.upstream_counted is that record. it fits in the existing bitfield and
queriesData stays 64 bytes, so SHARED_MEMORY_VERSION is unchanged - and since
nothing was checking that, the last commit pins the size of all ten structs that
live in shared memory. those sizes differ per target, clientsData is 688 bytes
on 64-bit, 672 on armhf and 664 on i386, so all three cases are held rather than
just the one this is built on

_FTL_new_query() drops the SHM lock around find_mac() and re-fetches its
pointers, which handles a remap but not the GC. a query index is logical,
_getQuery() resolves it as queryID + queries_offset, and runGC() advances that
offset - so after a GC run in that window the re-fetch returns a different query
that is very much alive, and the blocking decision lands on someone else's.
there is no way to rebase the index from there, so it gives up on the query
instead

three smaller ones: a DNS cache entry took a CNAME domain reference every time
the path ran but gives it back once when recycled, so a gravity reload leaks one
and the domain never leaves the table. _FTL_iface() cached its addresses against
the irec pointer alone, which does not move when another address on the same
interface changes, so the cached addr4/addr6 go stale with nothing to signal it
- the comment claiming the pointer is invalidated on SIGHUP had the wrong
invariant, an address change needs no SIGHUP. that cache is deleted rather than
taught to notice: any check covering the addresses has to visit all of them,
while the collection loop it guarded stops as soon as it has an IPv4 and a ULA
IPv6, and the expensive part of that loop is behind debug.networking anyway, so
the check costs at least what it saves. recomputing every time is correct and
smaller, three hunks of deletions. and the RFC 9462 test asked for len > 13 against a
13-character zone, so resolver.arpa itself was forwarded instead of answered,
with no label boundary either, so notresolver.arpa matched

two upstreams with dnsmasq probing both, then watch /api/stats/upstreams over a
few GC cycles. dig resolver.arpa and dig notresolver.arpa

## Additional information

---

## Checklist:

### DIFF
--- a/src/database/query-table.c
+++ b/src/database/query-table.c
@@ -1717,6 +1717,7 @@ void DB_read_queries(void)
 					{
 						upstream->lastQuery = queryTimeStamp;
 						upstream->count++;
+						query->flags.upstream_counted = true;
 					}
 				}
 				break;
--- a/src/datastructure.h
+++ b/src/datastructure.h
@@ -64,10 +64,6 @@ typedef struct {
 	} flags;
 } queriesData;
 
-// The layout above is deliberately packed to 64 bytes and lives in shared
-// memory, so a change that grows it needs a SHARED_MEMORY_VERSION bump
-_Static_assert(sizeof(queriesData) == 64, "queriesData must stay 64 bytes");
-
 typedef struct {
 	// Contains size_t and double fields -> size differs by architecture
 	// (64-bit: 64 bytes with 4 bytes internal padding before ippos; 32-bit:
--- a/src/datastructure.h
+++ b/src/datastructure.h
@@ -53,12 +53,20 @@ typedef struct {
 		bool complete :1;
 		bool blocked :1;
 		bool response_calculated :1;
+		// Set while this query holds a count on upstreamID, so the count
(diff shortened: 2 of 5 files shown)


DL6ER (approved):
Checked the accounting, every `count++` sets the flag and every `count--` is guarded by it, and dropping the interface cache is the better call over fingerprinting it since any check covering every address cannot stop early the way the collection loop does.

PR #3088, opened 2026-09-11
generate dnsmasq.conf from the candidate, notice failed writes, and do not race

010011110:

generate dnsmasq.conf from the candidate, notice failed writes, and do not race

write_dnsmasq_config() takes the config to render as conf, but six reads of
dns.domain.local and dns.domain.name went to the global config instead - the one
that is still live and has not had the change applied. changing either setting
generated the file from the old values and the next SIGHUP applied them

nothing noticed a failed write either. every directive is fputs/fprintf with the
result discarded, the fflush return was thrown away, and fclose reports success
after a short write, so a disk filling up part-way produced a truncated file
that was renamed over /etc/pihole/dnsmasq.conf and handed to dnsmasq.
write_custom_list() had the same hole

openFTLtoml("w") opened the temp file with fopen(..., "w"), which truncates
before the lock on the next line, so a second writer empties the file the first
is still filling. and the install was unserialised: closeFTLtoml() drops the
lock and only then does writeFTLtoml() compare, rotate and rename. the lock for
that has to sit on a path nothing renames, otherwise the next writer locks a
different inode and it serialises nothing, so it is a lock file of its own

plus two small ones: CONF_INT assigned a TOML int64 into a 32-bit int with no
range check, unlike CONF_UINT and CONF_UINT16 right below it, and
test_dnsmasq_config() never checked its fork so a failure fell through to
waitpid(-1, ...) and reported an unrelated child's exit status as the verdict on
the config. the identical fork in run_and_stream_command() is fixed in the
process handling PR, that file belongs with the action endpoints

change dns.domain.local and look at the generated dnsmasq.conf, then fill the
disk part-way through a write and check nothing is installed

## Additional information

---

## Checklist:

### DIFF
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -53,6 +53,19 @@ static bool test_dnsmasq_config(char errbuf[ERRBUF_SIZE])
 	pid_t cpid = fork();
 	int code = -1;
 	bool crashed = false;
+	if(cpid == -1)
+	{
+		// Without this the parent branch below would waitpid(-1, ...)
+		// and reap some unrelated child of ours - a dnsmasq TCP helper -
+		// then report its exit status as the result of the config test
+		log_err("Cannot fork to test new dnsmasq config: %s", strerror(errno));
+		close(pipefd[0]);
+		close(pipefd[1]);
+		strncpy(errbuf, strerror(errno), ERRBUF_SIZE - 1);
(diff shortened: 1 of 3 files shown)


DL6ER (approved):
The `config.` to `conf->` changes are the substance here, the candidate config is what gets written now instead of the live one, and the `fsync` before the rename stops a short write being installed over the running config.

PR #3089, opened 2026-09-11
the config endpoints apply changes they then throw away

010011110:

the config endpoints apply changes they then throw away

four defects in api/config.c, all the same shape: something escapes a request
that is afterwards rejected, or a guard PATCH has and the others do not

api_config_put_delete() never checked misc.readOnly. PATCH answers 403 for it,
PUT and DELETE went through: they changed the live config, rewrote
01-pihole.conf and the HOSTS file and restarted FTL, while writeFTLtoml()
deliberately wrote none of it to disk, so the next restart silently reverted
everything

delete_all_sessions() ran inside the item loop of api_config_patch(), and
api->ftl.restart was set there too. a request rejected further down - a later
item, or the dnsmasq test - logged every session out and left a restart pending
for a candidate config that free_config() had already discarded. both are noted
while walking the items now and applied after replace_config()

the failed dnsmasq test branch returned without free_config(&newconf, false),
unlike every other error exit in that function

and gen_config_path() splits the request path on every slash while the value was
taken to be the last component alone, so no string-array value containing a
slash could be added or removed - civetweb has already decoded %2F by then.
matching against the config item's own depth and rejoining everything past it is
what the old arithmetic already worked out to for a value without a slash

set misc.readOnly and try a PUT, then PATCH two items where the second is
invalid and check you are still logged in

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/config.c
+++ b/src/api/config.c
@@ -777,6 +777,8 @@ static int api_config_patch(struct ftl_conn *api)
 
 	// Read all known config items
 	bool config_changed = false;
+	bool privacy_level_decreased = false;
+	bool invalidate_sessions = false;
 	bool dnsmasq_changed = false;
 	bool rewrite_hosts = false;
 	struct config newconf;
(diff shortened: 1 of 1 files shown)


DL6ER (approved):
Noting the side effects and applying them only after `replace_config()` is right, a request rejected by a later item or by the dnsmasq test no longer logs every session out for a change that never happened.

PR #3090, opened 2026-09-11
endpoints that answer twice, or answer with the wrong numbers

010011110:

endpoints that answer twice, or answer with the wrong numbers

get_version_obj() returns a status code when it cannot read the versions file
and both callers discarded it, so after the 500 had gone out they carried on and
appended a whole second body to it. the same function also allocates seven cJSON
objects before opening that file and returned without freeing any of them

get_hwmon_sensors() broke out of the entire readdir loop when a hwmonX had no
readable name file, throwing away the directory-name fallback it had just
prepared and every sensor after that one, and left the stream open when the read
itself failed

compile_filter_regex() answered the request on a bad regex but reported it as a
plain bool, which every call site read as "filtering is on" before sending its
own response. it reports the code through an out-parameter now.
get_top_domains() and get_top_clients() return a cJSON object and have no status
to carry, so they pass NULL and a bad regex is logged rather than answered
behind their caller's back

get_string_var() ran mg_url_decode() over a value mg_get_var() had already
decoded, so any value with a percent sign, or a space that arrived as +, came
back as -1 from the second pass and the filter was silently dropped

api_history_clients() folded the lowest-ranked client into the unlabelled others
series whenever arrayID >= Nc - 1, while the clients map only gained an others
entry when there were more clients than fit. at the usual num_clients == Nc a
client's data vanished into a series with no name

GET /api/queries?domain=a b with a space in it, and compare /api/history/clients
against the number of clients you actually have

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/api.h
+++ b/src/api/api.h
@@ -52,7 +52,8 @@ int api_history_database_clients(struct ftl_conn *api);
 // Query methods
 int api_queries(struct ftl_conn *api);
 int api_queries_suggestions(struct ftl_conn *api);
-bool compile_filter_regex(struct ftl_conn *api, const char *path, cJSON *json, regex_t **regex, unsigned int *N_regex);
+bool compile_filter_regex(struct ftl_conn *api, const char *path, cJSON *json,
+                          regex_t **regex, unsigned int *N_regex, int *ret);
 void free_filter_regex(regex_t *regex, const unsigned int N_regex);
 
 // Statistics methods (database)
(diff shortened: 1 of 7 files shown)


DL6ER (approved):
The double-response fixes are the important half of this, and `mg_get_var()` already decoding means the second pass was silently dropping any filter value containing a percent sign or an encoded space.

PR #3091, opened 2026-09-11
free the teleporter archive, and end the reader on a rejected import

010011110:

free the teleporter archive, and end the reader on a rejected import

mz_zip_writer_finalize_heap_archive() hands the archive buffer to its caller and
clears m_pMem out of the writer state, so the mz_zip_writer_end() inside
free_teleporter_zip() releases the writer and nothing else. both callers treated
it as if it freed everything, so every authenticated GET /api/teleporter leaked
an entire ZIP of the config, the gravity database and the leases, and
pihole-FTL --teleporter leaked the same on its way out

on the import side the three error returns for pihole.toml, the leases and
gravity.db return straight out of the loop, past the only mz_zip_reader_end() at
the end, so a rejected import leaked the reader state and its copy of the
central directory

field_get() overwrote data->sid and data->import instead of keeping what was
already there. civetweb calls it again for each chunk of a field larger than its
buffer, and a client can just send the same field twice, so both were one POST
away from dropping an allocation

watch RSS across a few GET /api/teleporter, then POST an import that gets
rejected

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/teleporter.c
+++ b/src/api/teleporter.c
@@ -64,8 +64,12 @@ static int api_teleporter_GET(struct ftl_conn *api)
 	// Send raw (binary) ZIP content
 	mg_write(api->conn, ptr, size);
 
-	// Free allocated ZIP memory
+	// Free allocated ZIP memory. mz_zip_writer_finalize_heap_archive() has
+	// handed the buffer over to us and cleared it out of the writer's state,
+	// so mz_zip_writer_end() inside free_teleporter_zip() releases the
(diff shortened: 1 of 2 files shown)


DL6ER (approved):
Complete now that `write_teleporter_zip_to_disk()` frees the buffer on all three of its exits too - the only return in `read_teleporter_zip()` still without an `mz_zip_reader_end()` is the one before the reader is initialized, where there is nothing to end.

PR #3092, opened 2026-09-11
stop database failure sentinels reaching arithmetic

010011110:

stop database failure sentinels reaching arithmetic

db_query_int_from_until() and friends report a failure as a negative value and
three places used the result without asking

count_queries_on_disk() stored it straight into counters->queries, which is
unsigned, and init_queries_shm_sz() then sized the shared queries object from it
- an allocation of very nearly the whole address space, during startup

api_stats_database_upstreams() added two unchecked results into an unsigned
sum_queries, so a failed count wrapped the total into an enormous number and
served it as fact. api_stats_database_summary() already checks the same values
the same way

_dbclose() decremented dbopen_cnt outside the db != NULL && *db != NULL guard
while dbopen() only counts up once it has a connection, so closing an
already-NULL handle drifted the counter down. and api_client_suggestions()
skipped the dbopen() NULL check both its siblings in that file perform

two arithmetic ones in the same area. maxDBdays * 86400 was evaluated in 32-bit
unsigned arithmetic and only widened to double afterwards, so a large retention
wrapped into a cutoff that deletes nearly everything. and
find_recent_device_by_mock_hwaddr() wrote strftime('%%s','now') in a string that
never passes through a formatter, so SQLite got a literal %%s, strftime()
answered NULL, the cast gave 0 and the one-hour window never applied

set database.maxDBdays to 50000 and check what the cleanup would delete

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/network.c
+++ b/src/api/network.c
@@ -363,6 +363,18 @@ int api_client_suggestions(struct ftl_conn *api)
 
 	// Open pihole-FTL.db database file connection
 	sqlite3 *db = dbopen(true, false);
+	if(db == NULL)
+	{
+		// The two sibling handlers in this file check this. Without it
+		// the attach below fails on a NULL handle and answers with that
+		// instead of naming the real problem, and dbclose() takes a
+		// decrement for a connection that was never opened
+		log_err("Failed to open database in api_client_suggestions()");
+		return send_json_error(api, 500,
+		                       "database_error",
+		                       "Could not open long-term database",
+		                       NULL);
+	}
 
 	// Attach gravity database
 	const char *message = "";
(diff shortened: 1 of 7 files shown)


DL6ER (changes requested):
The `%%s` fix is correct, `db_query_int_str()` prepares the string directly so SQLite really did get two percent signs, the cast yielded 0 and the one-hour window never applied. `_dbclose()` and the `count_queries_on_disk()` sentinel check are right too.

`maxDBdays` needs more than the arithmetic, though. It is declared `CONF_INT` with `validate_stub`, yet every reader takes `.v.ui` - `database-thread.c`, `query-table.c` and `common.c`. Set it to `-1` and all three see 4294967295. Before this patch that wrapped in 32-bit and deleted nearly everything; after it, `mintime` goes hugely negative and nothing is ever deleted. Both are wrong, and the patch trades data loss for unbounded growth. Make the declared type and the readers agree and reject negatives, then the widening is not needed at all.

DL6ER (approved):
`CONF_UINT` is the right fix: the declared type now agrees with every `.v.ui` reader, so a negative cannot be stored at all and the widening is no longer load-bearing.

PR #3093, opened 2026-09-11
copy client data under the lock, and stop the server before freeing the api

010011110:

copy client data under the lock, and stop the server before freeing the api

resolveClients() reads client->ippos while holding the shared memory lock, keeps
both the resulting string pointer and the clientsData pointer, then releases the
lock and goes on using them - the IPv6 postponement check, the firstSeen
comparison and two debug lines. a concurrent mremap(MREMAP_MAYMOVE) of the
strings region or of the clients array while the resolver thread is unlocked
leaves both aimed at memory that is no longer mapped. resolveAndAddHostname() in
the same file already copies for exactly this reason

cleanup() called free_api() before http_terminate(), and mg_stop() runs inside
the latter, so the session table was backed up, zeroed and freed while civetweb
workers could still be serving requests against it

restart FTL under load with a few API sessions open

## Additional information

---

## Checklist:

### DIFF
--- a/src/daemon.c
+++ b/src/daemon.c
@@ -442,13 +442,16 @@ void cleanup(const int ret)
 	log_debug(DEBUG_ANY, "Terminating: Freeing regex filter memory");
 	free_regex();
 
+	// Terminate HTTP server (if running) before the API it serves. The other
+	// way round, free_api() backed up, zeroed and freed the session table
+	// while civetweb worker threads were still handling requests against it,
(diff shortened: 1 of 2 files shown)


DL6ER (approved):
Both right, `resolveClients()` kept shared-memory pointers across `unlock_shm()` where an `mremap()` invalidates them, and `mg_stop()` only runs inside `http_terminate()` so the session table was freed while workers could still serve against it.

PR #3094, opened 2026-09-11
check fork, terminate the sender name, and do not count a thread that failed

010011110:

check fork, terminate the sender name, and do not count a thread that failed

run_and_stream_command() tested only cpid == 0, so a failed fork fell into the
parent branch and reached waitpid(-1, &status, 0). that reaps any child of ours
- a dnsmasq TCP helper, whose tcp_pids slot then leaks - and reports its exit
status as the result of a command that was never run. the identical fork in
test_dnsmasq_config() is fixed in the config write PR, that file belongs there

log_sigterm_info() read up to all 256 bytes of kill_name and then replaced every
NUL separator in the cmdline with a space. a read that filled the buffer left no
terminator anywhere in it, and the trailing-space trim only writes one when the
last byte happens to be a space, so log_info() and the snprintf after it ran off
the end

run_arp_scan() increased tid even when pthread_create() had just failed, leaving
that slot at STATUS_INITIALIZING. the progress loop waits for every counted slot
to reach a finished state, so one failed thread hung pihole-FTL arp-scan for good

send SIGTERM from a process with a long cmdline and look at the log line

## Additional information

---

## Checklist:

### DIFF
--- a/src/api/action.c
+++ b/src/api/action.c
@@ -39,6 +39,20 @@ static int run_and_stream_command(struct ftl_conn *api, const char *path, const 
 	pid_t cpid = fork();
 	int code = -1;
 	bool crashed = false;
+	if(cpid == -1)
+	{
+		// Neither branch below tests for this, and the parent one ends in
+		// waitpid(-1, ...): it would reap an unrelated child of ours - a
+		// dnsmasq TCP helper, say, whose tcp_pids slot then leaks - and
+		// report that child's exit status as the result of this command.
+		// test_dnsmasq_config() has the same fork and is fixed alongside
+		// the config write path
+		log_err("Cannot fork to run command: %s", strerror(errno));
+		close(pipefd[0]);
+		close(pipefd[1]);
+		return false;
+	}
+
 	if (cpid == 0)
 	{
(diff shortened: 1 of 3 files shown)


DL6ER (changes requested):
The `continue` hangs `pihole-FTL arp-scan` harder than the bug it fixes. That loop advances `tmp` at the end of the body, so skipping to the next iteration without advancing retries the same interface forever and prints without bound - the `IFF_UP` skip twenty lines up does `tmp = tmp->ifa_next;` first for exactly that reason. CI being green says nothing here, `pthread_create` does not fail on a runner so the branch never executes.

No `continue` needed at all:

```c
if(pthread_create(&scanthread[tid], &attr, arp_scan_iface, &thread_data[tid]) != 0)
    printf("Unable to launch thread for interface %s, skipping...\n", tmp->ifa_name);
else
    tid++;
```

The `fork()` check and the `kill_name` termination are both right.

DL6ER (approved):
The `else tid++;` reads better than the `continue` did and leaves the loop to do its own advancing, which is what went wrong the first time round.

PR #3099, opened 2026-09-12
Add dns.fastRetry to retransmit unanswered upstream queries

nazar-pc:
Despite LLM wrote most of it under my guidance, I reviewed, edited and tested the changes to the best of my ability before opening the PR.

This is a partial solution to https://github.com/pi-hole/pi-hole/issues/6667, while the more fundamental fix probably belongs to the dnsmasq itself.

This change exposes dnsmasq's `fast-dns-retry` as a new config option `dns.fastRetry`, so that Pi-hole retransmits an upstream query that has not been answered instead of waiting for the client to notice.

Today it never does. `fast-dns-retry` defaults to off, and the only thing that turns it on is DNSSEC validation:

```c
/* Default fast retry on when doing DNSSEC */
if (option_bool(OPT_DNSSEC_VALID) && daemon->fast_retry_time == 0)
```

With the default `dns.dnssec = false`, `daemon->fast_retry_time` stays 0 and `fast_retry()` is a no-op, so a single lost packet on the way to or from an upstream server leaves the query hanging until the frec times out.

A repeated query does not necessarily rescue it either. `forward_query()` matches a new query against the in-flight one on the question alone - name, class and type (`lookup_frec(..., id = -1, ...)`) - but only treats it as a retry of that query when source address, source port *and* query ID all match as well. Anything else is attached to the stuck query and nothing is sent upstream for the first two seconds:

```c
/* closely spaced identical queries cannot be a try and a retry, so
   it's safe to wait for the reply from the first without
   forwarding the second. */
if (difftime(now, forward->time) < 2)
```

That assumption holds for stub resolvers that retry at multi-second intervals, but not for a caching resolver forwarding to Pi-hole - which is a common way to run it, e.g. pfSense/OPNsense with unbound in front. unbound retries an unanswered server after [376 ms](https://github.com/NLnetLabs/unbound/blob/5ccee67af22022f93dfc2b6723146b399955a4a4/util/config_file.c#L322) by default and derives later timeouts from measured RTT with a [50 ms](https://github.com/NLnetLabs/unbound/blob/5ccee67af22022f93dfc2b6723146b399955a4a4/util/rtt.c#L47) floor, so its retransmissions arrive comfortably inside the two-second window. They also cannot take the retry path: every unbound UDP send goes through `randomize_and_send_udp()`, which picks a fresh query ID via `select_id()` and a fresh source port via `select_ifport()`, so no retransmission ever matches the source-address/port/ID triple dnsmasq requires.

The practical result is that during the window where a lost packet needs covering, Pi-hole neither retries itself nor passes on the retries of the resolver in front of it. The query eventually fails upwards, the client falls back to its search domain, and the user sees NXDOMAIN for a domain that resolves perfectly well a second later.

## Measured effect

I extracted stats from two 24 h windows on the same installation, ~128k queries each, before and after adding `fast-dns-retry=250` via `/etc/dnsmasq.d`:

| metric | before | after |
|---|---|---|
| queries > 1 s | 1433 | 206 |
| queries > 5 s | 119 | 21 |
| `RETRIED` (status 12) | 274 | 54 |
| queries 0.5–1 s | 125 | 567 |

The growth in the 0.5–1 s bucket is the mechanism working: queries that used to hang for seconds now get an answer after a retransmission. Above 1 s, where clients start failing over, everything dropped by roughly 7x, and subjectively I stopped seeing them.

New `dns.fastRetry` is added, emitted as `fast-dns-retry=<ms>` in `dnsmasq.conf` when non-zero.
The default value is 1000 ms - the same `DEFAULT_FAST_RETRY` dnsmasq itself applies when it enables the feature for DNSSEC.

I realize a default that retransmits has an upstream-load cost, and that dnsmasq's conservative default made sense for its original role: a LAN resolver one hop from the ISP's resolver, in front of clients that retried at 5 s.

At least my Pi-hole's deployment is a different shape: it forwards across the internet to public resolvers, in front of unbound that has little patience.
Let me know if you prefer the default value to remain `0`/disabled so that previous behavior it retained (though in that case there is not as much benefit from this change since `fast-dns-retry=250` can already be customized with config file).

---

---

### DIFF
--- a/src/api/docs/content/specs/config.yaml
+++ b/src/api/docs/content/specs/config.yaml
@@ -227,6 +227,8 @@ components:
                   type: array
                   items:
                     type: string
+                fastRetry:
+                  type: integer
                 CNAMEdeepInspect:
                   type: boolean
                 EDNS0ECS:
@@ -770,6 +772,7 @@ components:
         config:
           dns:
             upstreams: [ "127.0.0.1#5353", "8.8.8.8" ]
+            fastRetry: 1000
             upstreamCA: ""
             doh: true
             dot: true
--- a/src/config/config.c
+++ b/src/config/config.c
@@ -420,6 +420,14 @@ void initConfig(struct config *conf)
 	conf->dns.upstreams.f = FLAG_RESTART_FTL;
 	conf->dns.upstreams.c = validate_upstreams;
 
+	conf->dns.fastRetry.k = "dns.fastRetry";
+	conf->dns.fastRetry.h = "Retransmission interval (in milliseconds) for unanswered upstream queries.\n\n Pi-hole forwards a query and then waits for the reply. If either the query or the reply is lost on the way, nothing happens until the client gives up and asks again. Repetitions of a query that is still in flight are attached to the running query instead of being forwarded again, so a client cannot recover such a loss on its own either - only Pi-hole can, by asking upstream once more.\n\n The first retransmission is sent after the configured number of milliseconds, the interval doubles for every further attempt, and retransmissions stop as soon as the query is answered or has timed out. Setting this to 0 disables retransmission. Note that DNSSEC validation enables it independently of this setting because an unanswered query would otherwise stall the whole validation chain.";
+	conf->dns.fastRetry.a = cJSON_CreateStringReference("An integer value between 50 and 10000 (milliseconds), or 0 to disable retransmission");
+	conf->dns.fastRetry.t = CONF_UINT;
+	conf->dns.fastRetry.f = FLAG_RESTART_FTL;
+	conf->dns.fastRetry.d.ui = 1000u;
+	conf->dns.fastRetry.c = validate_dns_fastRetry;
+
 	conf->dns.CNAMEdeepInspect.k = "dns.CNAMEdeepInspect";
 	conf->dns.CNAMEdeepInspect.h = "Use this option to control deep CNAME inspection. Disabling it might be beneficial for very low-end devices";
 	conf->dns.CNAMEdeepInspect.t = CONF_BOOL;
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -135,6 +135,7 @@ struct enum_options {
 struct config {
 	struct {
 		struct conf_item upstreams;
+		struct conf_item fastRetry;
 		struct conf_item CNAMEdeepInspect;
 		struct conf_item EDNS0ECS;
 		struct conf_item ignoreLocalhost;
--- a/src/config/dnsmasq_config.c
+++ b/src/config/dnsmasq_config.c
@@ -549,6 +549,14 @@ bool __attribute__((nonnull(1,3))) write_dnsmasq_config(struct config *conf, enu
 		fputs("\n", pihole_conf);
 	}
 
+	if(conf->dns.fastRetry.v.ui > 0)
+	{
+		fputs("# Retransmit an unanswered upstream query after this many milliseconds,\n", pihole_conf);
+		fputs("# doubling the interval for every further attempt\n", pihole_conf);
+		fprintf(pihole_conf, "fast-dns-retry=%u\n", conf->dns.fastRetry.v.ui);
+		fputs("\n", pihole_conf);
+	}
+
 	// Check if an explicit interface is configured
 	char interface[MAXIFACESTRLEN];
 	strncpy(interface, conf->dns.interface.v.s, sizeof(interface) - 1);
--- a/src/config/validator.c
+++ b/src/config/validator.c
@@ -637,6 +637,19 @@ bool validate_dns_revServers(union conf_value *val, const char *key, char err[VA
 	return true;
 }
 
+bool validate_dns_fastRetry(union conf_value *val, const char *key, char err[VALIDATOR_ERRBUF_LEN])
+{
+	// dnsmasq rejects intervals below 50 ms outright, and an interval
+	// beyond the forwarding timeout would never fire a retransmission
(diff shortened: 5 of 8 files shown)


darkexplosiveqwx:
I'm not sure if `fast-dns-retry` deserves its own option in Pi-hole's configuration. Why not use `misc.dnsmasq_lines` or `misc.etc_dnsmasq_d`? This is not a particularly strong opinion of mine, 
I'm open to arguments for it.

nazar-pc:
The biggest benefit is that the default value for new option is not 0, so it has (arguably desirable) change in "out of the box" behavior.

rdwebdesign:
> The biggest benefit is that the default value for new option is not 0

Is this really a benefit?

Do we want (or need) to change the default value?

DL6ER (commented):
How did you manage to get such an unreliable upstream? Is this a real existing bad-upstream-connectivity setup? Reason I'm asking is that I was not able to reproduce this with three setups over the past three days:

1. My usual setup: a local `unbound` docker container
2. Using an upstream destination via DoT (this is a v7.0 feature, so not yet released) over my FTTH-connection
3. Using an upstream destination plain via a mobile hotspot from my phone

In principle, this seems fine, but I concur to @rdwebdesign that setting a default `> 0` will change behavior but probably not in a bad way. I am fine trying this out in `development` and, if it proves to be useful there, release it into v7.0.

One thing on the numbers: the table measures `fast-dns-retry=250`, while this PR defaults to 1000. Since the default is the part we are actually debating, could you say what the same two windows look like at 1000? If the effect is mostly carried by the first retransmission arriving early, 250 and 1000 will not behave the same.

nazar-pc:
My upstream servers are Cloudflare and Google, not exactly sure why this is happening since my Internet setup with multi-WAN (3 FTTH ISPs) is not exactly very common.

Restarted with `fast-dns-retry=1000`, will let you know how it goes in a day or so.

DL6ER:
> my Internet setup with multi-WAN (3 FTTH ISPs) is not exactly very common.

are they up at the same time? If so, you should carefully start investigating what is actually happening. I am pretty sure the silent drops you are seeing are happening also for all your other traffic. TCP will notice it and retry, but it makes everything slower. Think "slower" rather as temporary, even if short, hangs, rather than actually noticeable throughput reduction. Pages will load a bit slower, downloads will need certain parts being resubmitted to complete.

I guess much of the solution lies in *how* you achieve the parallelism *exactly*, but I am almost sure that the replies are coming back to you, you just somehow don't route them back into your Pi-hole.

No need to discuss your particular situation here in details, take it just as a friendly suggestion.

nazar-pc:
I compared three non-overlapping 24 h windows with `fast-dns-retry` unset, then at 250 ms, then at 1000 ms:

| metric | before | after 250 ms | after 1000 ms |
|---|---|---|---|
| queries in window | 124062 | 127984 | 142813 |
| queries > 1 s | 1433 | 206 | 1938 |
| queries > 5 s | 119 | 21 | 173 |
| `RETRIED` (status 12) | 274 | 54 | 440 |
| queries 0.5–1 s | 125 | 567 | 54 |

At 250 ms the growth in the 0.5–1 s bucket is the mechanism working: queries that used to hang for seconds are now rescued by a retransmission and answered inside a second. Above 1 s (the range where clients start failing over to the search domain) everything dropped by roughly 7x, and observable failures mostly disappeared.

At 1000 ms the same queries are rescued but answered just after the retry fires, so they land in the 1–2 s bucket instead, and the count above 1 s is not better than the baseline. The `>= 5 s` tail is worse as well, because the 10 s frec lifetime fits only three retries at 1000 ms against five at 250 ms. The interval matters, not just the presence of retries. I added the total number of queries to the table for better comparison.

Note that fast retries do not fan out, every retransmission goes back to the upstream that already failed (or looked like failed), and additional upstreams do not help there.

---

I can try 500 ms value in between if you want, but 1000 ms as is doesn't seem to help meaningfully in practice.

nazar-pc:
Claude suggested trying to address it from a different angle:

> Use the duplicate as a **signal** rather than tuning how long to ignore it: when a duplicate arrives for an in-flight query whose last send is older than the retry interval, retransmit immediately instead of only attaching to the frec.
>
> That keeps dedup completely intact (genuinely simultaneous duplicates still collapse into one upstream query, because the last send is milliseconds old) while removing the blind period entirely. And it needs **no new configuration**: `fast_retry_time` is already the "minimum spacing between retransmissions" value, and `forward_timestamp` already tracks the last send. Downstream re-asking is strictly better evidence that the answer hasn't arrived than a timer firing blind.
>
> A smaller variant if that's too invasive for upstream taste: cap the window at the retry interval (`difftime(...) < min(2, fast_retry_time/1000)`) so the two policies can't contradict each other. Still no new config.

But these are dnsmasq changes.

DL6ER:
Thanks for the three windows, they settle the default. 1000 ms is no better than off and makes the tail worse, and 250 ms helps you but is not a default I want on every install: any upstream with more than 250 ms round trip would get every query twice, doubling from there.

So: keep `dns.fastRetry`, default 0. It stays discoverable and validated, and there is no out-of-the-box change, which is what @rdwebdesign and @darkexplosiveqwx were after. That needs the default, the help text (say it is off by default), `test/pihole.toml` and the bats test, which currently asserts `fast-dns-retry=1000` in `dnsmasq.conf`; with 0 the line is not written at all.

On the dnsmasq-side idea (a repeated downstream query as the retransmit trigger): that changes forwarding semantics inside dnsmasq, so it belongs on dnsmasq-discuss, not in a patch we carry here.

PR #3104, opened 2026-09-17
read the lists off the shared memory lock, and let a lookup reopen gravity

010011110:

two things about the gravity handle, found while looking at why the list
endpoints hold the global shared memory lock

api_list() wraps the GET handler in lock_shm(), with a comment saying the lock
is not actually needed and is taken for simplicity, to stop anyone editing the
lists during the read. the comment is wrong on both counts. api_list_read()
passes NULL to gravityDB_readTable(), which means the shared gravity_db handle,
and FTL_reload_all_domainlists() closes and rebuilds that handle in
gravityDB_reopen() from the database thread - under the same lock. drop the lock
without dealing with that and a list request racing a gravity reload answers
"Database not available"

that lock is the one every DNS query takes, and this handler held it across a
gravity database read, a name lookup per returned row, JSON serialisation and a
blocking write to a client that may be slow. gravityDB_readTable() documents the
way out in its own comment: a caller passing a connection pins it for the whole
read, which is what api_search() already does with gravityDB_open_RO(). doing
the same here removes the need for the lock on the read

only the read. PUT, POST and DELETE keep it: taking those off it as well made
the resolver log "Gravity database not available" 27 times in one suite run,
because a reload in the database thread then overlaps an API write still holding
gravity.db. that is worth untangling separately and is not this change

the second half is what that experiment turned up. a reload whose
gravityDB_reopen() fails leaves the database shut, and in_allowlist(),
in_denylist() and in_gravity() test the shared statements and answer
LIST_NOT_AVAILABLE before they reach gravityDB_prepare_client_statements(),
which holds the only retry on that path. so the retry is unreachable and every
lookup reports the lists as unavailable from then on. that state is meant to be
momentary - dns.replyWhenBusy exists for a database that is busy - and here it
is permanent: on the default ALLOW nothing is blocked any more. nothing retries
on a timer either, the database thread's gravity_updated() returns early when
the file is unreadable and asks for no reload when the file comes back unchanged

so the three lookups may reopen before they give up, backing off from a second
to a minute - a failed gravityDB_open() stats the file and warns, and gravity.db
can be away for minutes while a gravity run rebuilds it, so a fixed retry would
fill the log from a path that used to say nothing. a successful reopen
then asks for a reload: opening the database back up restores the statements and
the gravity_has_exact_* flags but not the rest of what the failed reload was
doing - read_regex_from_database() frees the regexes before it reads the table,
so there were none left, the counters->database.* fields kept DB_FAILED and the
per-client cache was never reset. exact lists and gravity would block again
while regex quietly did not. and FTL_reload_all_domainlists() now looks at what
gravityDB_reopen() returned - it discarded it, so the one event that puts FTL
into this state went unremarked, as a warning since before the first pihole -g
the file is simply not there yet

the write path needed one more thing. api_list_write() ends with a GET-style
reply of its own and api_list() runs it under the lock, so rendering there would
open the read-only connection - which does carry sqliteBusyCallback - while
holding it, and a gravity.db held by a pihole -g would park the request and the
resolver with it. the write is committed and its connection closed by then, so
api_list_write() hands back the status, the item and the processed object and
the caller renders after unlocking

on the write side: PUT, POST and DELETE keep the lock. taking them off it too
gets close - the "Database not available" flood goes away and the suite is down
to two warnings from a reload racing an API write - but closing those means
letting the reload wait on a busy database, and the reload holds the shared
memory lock while it does. that trades briefly unavailable lists for DNS
stalling up to DATABASE_BUSY_TIMEOUT, which is the thing this PR is trying to
avoid. it needs the reload and the API writes serialised by something that is
not the shared memory lock, and that is a design change rather than a fix

a third commit giving gravityDB_open() a busy handler during its setup was here
and has been dropped. it was the same trade in another place: with the second
commit the lookups call that function themselves while holding the shared memory
lock, so each failed open could wait a second, and every further open attempt in
the reload could too. the reasoning behind it was wrong as well - the list writes
run under the shared memory lock and cannot overlap the reload, and
gravity_updated() is a reader, which does not block the prepares in the setup. a
reopen that fails fast and is retried by the second commit is the better answer

with a reload forced while gravity.db is moved away: a domain blocked before it
stops being blocked, and the first query after the file returns blocks again,
with the reopen logged in between

## Additional information

---

## Checklist:


DL6ER (changes requested):
The first two commits look right to me. Reading through a connection of its own is what `api_search()` already does, the RO handle is released on every exit of `api_list_read()`, and `gravity_ensure_open()` is the missing retry: after a failed reopen the shared statements are NULL, so the lookups answered `LIST_NOT_AVAILABLE` before `gravityDB_prepare_client_statements()` could ever try again.

The third commit I would drop. The busy handler makes `gravityDB_open()` wait up to `DATABASE_BUSY_TIMEOUT` (1 s), and the callers that matter hold the SHM lock: `FTL_reload_all_domainlists()` on the database thread and, new with the second commit, the lookups themselves. A reopen that still fails after that second is followed by more than a dozen further open attempts in the same function (`gravityDB_count()` x13, `gravityDB_getTable()`, `check_inaccessible_adlists()`, `check_restored_gravity()` all retry on their own), each of which can now wait as well. Before this PR all of them failed instantly, there was no handler to unset. For as long as a writer holds `gravity.db`, DNS now stalls: through the reload while it runs, and afterwards through `gravity_ensure_open()`, up to a second at a time. That is the trade the PR description argues against. The rationale is also off: the list writes run under the SHM lock and cannot overlap the reload, and `gravity_updated()` is a reader, which does not block the prepares and selects in the setup. With the retry in place, a reopen that fails *fast* heals within a second of the file becoming free, which is what `dns.replyWhenBusy` is for.

Small one: the old comment above the `GET` branch in `api_list()` still says the lock is taken for simplicity, directly above the line saying there is no lock. Drop those three lines.

DL6ER (changes requested):
One thing before this goes in: `api_list_write()` ends with a GET-style `api_list_read()`, which now opens the read-only connection while `api_list()` still holds the shared memory lock. That connection carries `sqliteBusyCallback`, so a `gravity.db` locked by someone else - `pihole -g`, say - parks the write handler for up to `DATABASE_BUSY_TIMEOUT` with the lock held, and DNS waits with it. The shared connection has its busy handler cleared for exactly that reason. The write is committed and its connection closed by then, so the reply can be rendered outside the lock: move the read there, or hand that path a connection without the busy handler.

Smaller: the new `log_err()` in `FTL_reload_all_domainlists()` also fires when `gravity.db` is simply not there yet, where `gravityDB_open()` has already warned. Before the first `pihole -g` that turns a normal state into an error. Please make it a warning.

DL6ER (changes requested):
Both points from my last round are addressed: the write handler now hands the reply back and `api_list()` renders it after `unlock_shm()`, and the failed reopen is a warning.

One gap remains in the second commit. `gravity_ensure_open()` brings back the shared statements and the `gravity_has_exact_*` flags, but the reload that failed did more than open the database: `read_regex_from_database()` runs `free_regex()` first and then gets nothing from the shut table, so every regex allow and deny entry is gone; the `counters->database.*` fields are left at `DB_FAILED`; the per-client domain cache is not reset. After the lookup-side reopen, exact lists and gravity block again, regex silently does not, until the next `RELOAD_GRAVITY`. The state this commit sets out to cure is only half cured, and the half that stays broken is the one nothing logs. Have a successful open in `gravity_ensure_open()` end with `set_event(RELOAD_GRAVITY)`, so the database thread runs `FTL_reload_all_domainlists()` once more and finishes the job.

Small one: in the `PUT` and `POST` branches of `api_list()` the old comment about the lock being taken "for simplicity" is still there, directly above the new one saying the write needs it. Drop the old three lines, as the `GET` branch does.

DL6ER (changes requested):
Both points from round 3 are in: the successful reopen now asks for a `RELOAD_GRAVITY`, and the stale lock comments are gone.

One more, from the retry rate. `gravityDB_open()` logs a warning when the file does not exist, and `gravity_ensure_open()` tries once per second from the query path. With gravity.db gone - `rm gravity.db && pihole -g` is a common troubleshooting step and a gravity run takes minutes - that is one warning per second for as long as it lasts, where these lookups used to be silent. Back off instead: double the wait after each failed attempt, cap it at something like 60 s, and reset it on success.

Also trim the two comment blocks that recount the old behavior (the eight lines above `set_event()` and the "used to pass unremarked" lines in `FTL_reload_all_domainlists()`). Say what the code does now; the history is in this PR.

DL6ER (approved):
Backoff and the trimmed comments are in, and every point from the four rounds is addressed. Thanks for staying with it.

PR #3108, opened 2026-09-18
Enable `ntp.sync.rtc.set` by default

DL6ER:

A device without a real-time clock has nothing to update, and one that has a clock benefits without having to find the setting first: after a reboot the system comes up with a good time before it has reached an NTP server. Suggested by @yubiuser in pi-hole/docs#1418.

Only new installations are affected. An existing `pihole.toml` carries the value it was written with, so a Pi-hole that is already running keeps its clock untouched until the setting is changed by hand.

Remove `/etc/pihole/pihole.toml`, start FTL, and `ntp.sync.rtc.set` comes up as `true`. On a machine with an RTC, `hwclock -r` follows the system time after the next successful synchronization; on one without, nothing happens at all - every failure in `src/ntp/rtc.c` is logged at `DEBUG_NTP` only. `test/pihole.toml` carries the new default, so the BATS config comparison covers what FTL writes.

## Additional information

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** N/A - the NTP page keeps saying the setting is off by default until v7.0 ships, see pi-hole/docs#1418

---

## Checklist:


darkexplosiveqwx (approved):
Since this only applies if there is an RTC detected/configured and FTL already backs down if another NTP client is active I see no harm in this.

PR #3115, opened 2026-09-20
remove unneeded `FLAG_FTL_LOG`

darkexplosiveqwx:
`log.c` takes a private copy of the path and no longer uses the global conf item since https://github.com/pi-hole/FTL/pull/2960. 
The asynchronous logger also takes a copy, so `FLAG_FTL_LOG` is also unneded for it.

If desired, I can move `FLAG_API_READ_ONLY` to `(1 << 7)`

---

---


DL6ER (approved):
`set_log_path()` strdup's the path into `ftl_log.path` and both `write_log_line()` and `reopen_log_fd()` use only that copy, so the config item is read once at startup and the `FTL terminated` lines at the end of `cleanup()` are still fine.

It also removes a dangling pointer: `free_config()` NULLs `v.s`, so a second `cleanup()` on the crash path no longer double-frees.

No need to renumber `FLAG_API_READ_ONLY` - nothing reads the flag word numerically, so I'd rather leave `(1 << 7)` free and reuse it for the next flag.

PR #3116, opened 2026-09-20
key injected PTR records on the address, not on the spelling of the query

010011110:

check_pihole_PTR() decides whether it has already injected a PTR record by
comparing the query name against every ptr_record in daemon->ptr. that does not
bound anything, because many names decode to the same address.
in_arpa_name_2_addr() (src/dnsmasq/rfc1035.c:238) checks each label is digits
and then rotates: addr[3] = addr[2]; addr[2] = addr[1]; addr[1] = addr[0];
addr[0] = atoi(cp1). atoi() drops leading zeros and the rotation shifts extra
leading labels out, so 5.1.168.192, 005.01.0168.0192 and 9.9.5.1.168.192 are
all 192.168.1.5

so every spelling that matches one of our own addresses appends another record,
and nothing frees, caps or counts that list. it gets walked linearly on every
PTR query - twice in this function, twice more in dnsmasq's answer paths, once
in check_for_local_domain - and make_non_terminals() expands all of it into
cache entries on every cache_reload(). the per-query cost grows with whatever
has been sent, in the single-threaded resolver loop, so everyone's DNS slows
down and stays slow after the sender stops. only a restart clears it

no authentication, UDP/53, anywhere on the LAN. the call sits before lock_shm()
and before the per-client rate limiter, so dns.rateLimit suppresses the reply
but the record is already linked. dns.piholePTR defaults to PI.HOLE so this is
the default configuration - NONE disables the path for anyone who wants a
mitigation before this ships

c1a69ff8 closed the letter-case axis of the same hole and left a comment
describing the bound it was aiming for. this keys on the decoded address
instead, which is what get_ptrname() already documents as its precondition:
"check_pihole_PTR() ensures we are called at most once per address"

the calloc() three lines down goes from unchecked to checked while here, it was
dereferenced immediately

send 16 PTR queries for one of the Pi-hole's own addresses, spelled with
varying leading zeros and extra leading labels, and count "Generating PTR
record" in FTL.log with debug.queries on. in the ftl-build container against
the container's own eth0 address: 16 records before this change, 1 after, with
the other 15 reported as already covered by that address. it has to be a real
PTR query - an A query for the same name never reaches this function

## Additional information

---

## Checklist:


DL6ER (changes requested):
`strdup()` for the record name is still unchecked, while the allocation right above it is now guarded. dnsmasq passes that name into its own name comparisons without a NULL check, so an allocation failure takes the resolver down on the next PTR query rather than quietly skipping a record. Close that in the same pass.

The fixed bound of 64 is the part I would change. Keying on the address already limits the set to the number of our own addresses, so the constant adds no protection of its own - but on a host with more addresses than that, the remaining ones never get a record and every query for one writes a warning from a path that takes unauthenticated traffic. My development machine has 79. Take the bound from the interface list rather than a constant.

PR #3117, opened 2026-09-20
raise the open-file soft limit to the hard limit at startup

010011110:

FTL is one process holding, at once, the resolver's UDP and TCP sockets, the
connections dnsmasq forwards, the SQLite databases and every webserver
connection. it never touched RLIMIT_NOFILE, so it runs on whatever soft limit
the init system gives it - commonly 1024 while the hard limit is far higher
(1048576 under systemd on the build container)

that soft limit is a shared ceiling with no partition between the parts, and the
webserver can reach it on its own. the TLS terminator opens a loopback backend
socket per HTTP/2 and HTTP/3 stream, so one client multiplexing streams over a
few connections holds hundreds open at once - measured in the ftl-build
container, four connections of 100 held streams each took the process from 17
open files to about 390. once the table is full the resolver cannot open a
socket either: allocate_rfd() fails, TCP DNS stops accepting, the DoT listener
gets EMFILE, SQLite cannot open its files. the web side taking DNS down with it
is the part worth closing

raise the soft limit to the hard limit at startup - the headroom the platform
already allows. it never lowers a limit and never exceeds the hard cap, so where
the two are already equal it does nothing

this is headroom, not a cap on any one consumer. the terminator opening a
backend per stream with no in-flight limit is a separate matter and needs a
decision about its own design (a stream cap trades off legitimate multiplexing,
an in-flight budget needs a counter across its teardown paths) - i will raise
that separately with the reproduction. this change keeps that from reaching the
resolver through a low default the whole process shared

launch pihole-FTL under a lowered soft limit (ulimit -Sn 1024) and read
/proc/<pid>/limits: before, Max open files stays 1024; after, it is raised to
the hard limit, with an INFO line "Raised the open-file limit from 1024 to ...".
where soft already equals hard the function returns without doing or logging
anything

## Additional information

---

## Checklist:


DL6ER (commented):
Raising the ceiling is only safe as long as nothing in the process builds an `fd_set` from a real descriptor, since a descriptor above the set size would write past it. The two places that still do are a CLI subcommand and a standalone tool, both short-lived processes holding a handful of descriptors, and the webserver polls rather than selects. Nothing here is exposed by the change.

There is a stray blank line after the argument parsing.

On the terminator: that is the one I actually want. A backend socket per stream with no in-flight budget is the defect, and this only keeps it from reaching the resolver. Please raise it with the reproduction.

PR #3118, opened 2026-09-20
return after the /api/docs redirect so no second response follows it

010011110:

api_docs() sends a 301 to /api/docs/ when the request has no trailing slash and
then falls through. serve_index stays false, no docs_files entry matches the
empty item, and the function returns 0 - on which api_handler() sends a 404. so
one GET /api/docs produces a 301 with Content-Length: 0 immediately followed by
a not_found JSON body on the same connection

raw response before the change:

    HTTP/1.1 301 Moved Permanently
    Location: /api/docs/
    ...
    Content-Length: 0

    {"error":{"key":"not_found","message":"Not found","hint":"/api/docs"},...}

most clients follow the redirect and never read the trailing bytes, but it is a
malformed response: on a kept-alive connection those bytes are read as the head
of the next response and desync it. return 301 after the redirect, which
api_handler() passes through without adding a body

GET /api/docs with no trailing slash and read the raw bytes - before, a JSON
body trails the Content-Length: 0 of the 301; after, the 301 ends clean.
/api/docs/ still serves the index (200) and /api/docs/missing is still 404

## Additional information

---

## Checklist:


darkexplosiveqwx (dismissed):
Maybe trim the comment, otherwise looks good.

darkexplosiveqwx (approved):

PR #3119, opened 2026-09-20
give the route parser a local address buffer, not a shared static

010011110:

nlparsemsg_route() renders each route attribute address into a
static char ip[INET6_ADDRSTRLEN] and then copies it into the JSON object.
GET /api/network/routes reaches it, and API handlers run on civetweb worker
threads, so two concurrent requests share that single buffer - one can fill it
and the other overwrite it between the fill and the copy, so a response carries
the wrong address. it is a fixed buffer inet_ntop() always terminates, so this
is a wrong value under concurrency, not a memory fault

make it a local, which is exactly what the sibling nlparsemsg_address() already
does with its own char ip[INET6_ADDRSTRLEN]. the static bought nothing, the
value is used on the next line

honest scope: the window is narrow and i did not force an observably wrong
address. this is a zero-risk move of one buffer from static to automatic
storage, matching the parser beside it

GET /api/network/routes still returns the same routes with their addresses (8
routes with valid destination/gateway strings in the build container). the
change is only the storage class of the ip buffer

## Additional information

---

## Checklist:


darkexplosiveqwx (commented):
I don't think this deserves a that long explanation comment.
You already provide detailed commit messages, so blame provides enough information.

darkexplosiveqwx (approved):

